PR #14, opened 2021-05-25
Add proper CI for both the JavaScript and Python libs

tasn:
This depends on #13


svix-liam:
btw, if you change the branch origin to `tom/event-type-update` then github will automatically set it back to `main` when #13 gets merged pretty sure. It worked for me the other day.

tasn:
That's good to know!

tasn:
Fixed, needs another review now.

python/svix/receiver.py
```diff
@@ -29,7 +27,7 @@ def verify(self, data: t.Union[bytes, str], headers: t.Dict[str, str]) -> t.Dict
         msg_id = headers.get("svix-id")
         msg_signature = headers.get("svix-signature")
         msg_timestamp = headers.get("svix-timestamp")
-        if not (msg_id and msg_timestamp and msg_timestamp):
+        if not (msg_id and msg_timestamp and msg_signature):
```
svix-liam, line 30: 👀

python/svix/api.py
```diff
@@ -61,7 +51,7 @@ def __init__(self, configuration: Configuration) -> None:
     @contextmanager
     def _api(self) -> t.Generator[ApiClass, None, None]:
         with ApiClient(self._configuration) as api_client:
-            yield self._ApiClass(api_client)
+            yield t.cast(t.Any, self._ApiClass)(api_client)
```
svix-liam, line 54: Why do we need to cast this to Any?

svix-liam (approved):
LGTM, left one question though

python/svix/receiver.py
tasn: ?

python/svix/api.py
tasn: Because the OpenAPI generator APIs are not all inheriting from the same type (but all share the same structure). Don't ask me why.
This cast is a way to force it to work. Let me try and see if I can fix it in a less ugly way.

python/svix/receiver.py
svix-liam: I see you sneaking a bugfix in :P

python/svix/api.py
tasn: I made it less casty, but still with a cast that I couldn't figure out.

python/svix/receiver.py
tasn: Ah yeah, because it was actually discovered by the type checks so I let it slide. :)

PR #16, opened 2021-05-26
Go Support

### LINKED ISSUE #3: Create Go library
See the description in #1 for how it should be done.

svix-frank:
closes #3


tasn (changes requested):
See the other libraries, we don't want to commit automatically generated files.

svix-frank:
> See the other libraries, we don't want to commit automatically generated files.

Go does not have a build step for dependencies so if we want to provide the library as a module via this repo we will have to check in the generated code (a user would then be able to use the dependency via `github.com/svixhq/svix-libs/go`.

svix-frank:
> See the other libraries, we don't want to commit automatically generated files.

Unfortunately, go doesn't have a build step for modules (dependencies) so you have to have all the code checked in if you want a user to be able to start using our library via a simple `go get github.com/svixhq/svix-lib/go`.

svix-frank:
Btw moving the module to the root directory allows this lib to be used as such:
```sh
go get github.com/svixhq/svix-libs/go
```

```go
package main

import (
	svix "github.com/svix-frank/svix-libs/go"
)

func main() {
	sx := svix.New("API_TOKEN", nil)
	_, _ = sx.Application.Create(&svix.ApplicationIn{Name: "test"})
}
```

Which I think is a good enough workflow and lets us keep it in one repo

tasn:
We can't not have the generated files checked in in go?

svix-frank:
> We can't not have the generated files checked in in go?

Not if we want the library to be useable with a single `go get` go's module system just downloads the library at the given URL; there is no concept of a `build step` for dependencies in go, this is a common issue with go bindings for things like c libraries that require a per environment build step.  We will need to provide a git repo with generated openapi code either somewhere (either here or a separate repo).  Otherwise we have to ask our users to do the build step manually.

svix-frank:
Also for review, verifying works a little different than javascript & python since you can't just arbitrarily parse json into an object, so I opted for the verify to just return an error:

```go
secret := "MfKQ9r8GKYqrTwjUPD8ILPZIo2LaLaSw"

// These were all sent from the server
headers := http.Header{}
headers.Set("svix-id", "msg_p5jXN8AQM9LWM0D4loKWxJek")
headers.Set("svix-timestamp", "1614265330")
headers.Set("svix-signature", "v1,g0hM9SsE+OTPJTGt/tmIKtSyZlE3uFJELVlNIOLJ1OE=")

payload := `{"test": 2432232314}`

wh, _ := svix.NewWebhook(secret)
// Throws on error, returns nil
err := wh.Verify(payload, headers)
fmt.Println(err)
```

Alternatively, we can make it take an interface and expect it to json unmarshall to that
```go
type Example struct {
	Name  string
	Order string
}
var examples []Example
 // here jsonBlob would be a byte array and on successful verification it would unmarshal it into examples
err := webhook.Verify(jsonBlob, headers, &examples)
if err != nil {
	fmt.Println("error:", err)
}
```

tasn:
Generated files: OK, it is what it is then...
Verification without creating an object: that's fine.

Is the PR ready for review?

svix-frank:
> Generated files: OK, it is what it is then...
> Verification without creating an object: that's fine.
> 
> Is the PR ready for review?

Yep!

tasn (approved):
1. This PR LGTM
2. Go looks so weird! 😂

PR #18, opened 2021-05-26
py: make type checking and linting more strict.

tasn:


python/.flake8
```diff
@@ -1,3 +1,4 @@
 [flake8]
 max-line-length = 120
+extend-ignore = E203, W503, E501, E711, E712, N803
```
svix-frank, line 3: It doesn't look like we break any of these rules, why not enforce them? except for maybe `W503` which is deprecated

svix-frank (changes requested):

python/.flake8
tasn: This is the recommended list from `black`. It's the list of rules where flake8 is not compliant and is incompatible with black. That's why we want to ignore them.

svix-frank (approved):

PR #20, opened 2021-05-28
go: export event-type list command

svix-frank:
I accidentally made the `List` command for the event-type api private 🙈


tasn:
Don't blame yourself, blame weird-ass Golang. :P

PR #21, opened 2021-05-28
all: add function to get endpoint stats

svix-frank:
depends on #20


svix-yair:
This is a Development only endpoint. I think we should abandon this PR

tasn:
It's not a development only endpoint, it's used in the frontend. Though it's not exposed on purpose, because it's kinda hacky.

PR #25, opened 2021-05-29
Rename 'data' to 'payload' in the Message model

tasn:
Passing data is still accepted (and will work), though return will always be
called payload.

We could potentially make this change fully API compatible, though
really, it's such a minor change, and it's not worth the trouble.


tasn:
@svix-frank: please make sure I haven't messed up the Go lib. :)

svix-frank (approved):
The order inputs in one of the openapi functions changed (just fixed), other than that LGTM.

Note this is technically a breaking change as go is strongly typed (and the generated model no longer contains the Data member), but since we are at v0 this is fine IMO and not worth polluting the go api to handle.  Maybe just add a note in the changelog just incase =)

tasn:
Yeah, it's also a breaking change in TypeScript and Python. Though honestly, it's so minor and the libs are so early, it's not worth carrying this baggage yet.
It's backwards compatible for existing library users (as in, old versions of the libraries will continue to work), and people upgrading will just need a tiny fix.

PR #28, opened 2021-05-30
go: allow api url override

svix-frank:


tasn (changes requested):
1. We are just releasing 0.14.0, so the changelog will actually be for the next version.
2. Is there a way to make it more "debuggy-looking"? Something like "DebugURL", or actually maybe just read an env var directly? This shouldn't be used, and shouldn't be there to confuse people, though it is useful for testing. So it needs to be non-obstrusive.

svix-frank:
>     1. We are just releasing 0.14.0, so the changelog will actually be for the next version.
> 
>     2. Is there a way to make it more "debuggy-looking"? Something like "DebugURL", or actually maybe just read an env var directly? This shouldn't be used, and shouldn't be there to confuse people, though it is useful for testing. So it needs to be non-obstrusive.

What about `OverrideBaseURL` with the comment `The base URL (protocol + hostname) used for all requests sent by this Svix client. (Useful for testing)`

tasn:
> What about `OverrideBaseURL` with the comment `Overrides the base URL (protocol + hostname) used for all requests sent by this Svix client. (Useful for testing)`?

It's useful for testing if you have a local svix server, but only we do. So it's really just confusing noise for people. As you may see in TypeScript and Python, it's not exposed at all. That's why an env var is nice, as it doesn't confuse people who are unaware of it.

svix-frank:
> > What about `OverrideBaseURL` with the comment `Overrides the base URL (protocol + hostname) used for all requests sent by this Svix client. (Useful for testing)`?
> 
> It's useful for testing if you have a local svix server, but only we do. So it's really just confusing noise for people. As you may see in TypeScript and Python, it's not exposed at all. That's why an env var is nice, as it doesn't confuse people who are unaware of it.

I get the desire to not pollute the API, but I _really_ don't like the idea of having an undocumented environment variable that can change the functionality of our library and potentially break it.  I personally think it's much better to just have a well documented (albeit mostly unused by our users) configuration option like I've done since there is no way to have a secret but public option in Go. 

For argument sake this is exactly what the official stripe-go library does:
https://github.com/stripe/stripe-go/blob/270a88ff0477d75717b76c24a7ea5f590c8f3532/stripe.go#L213

tasn:
I was convinced even before the stripe-go link.

Could you maybe at least prefix the parameter with "Debug", so "DebugUrl"?

tasn (approved):
Let me know we are OK to merge this.

svix-frank:
> Let me know we are OK to merge this.

Please do!

tasn:
Need a new release I presume?

fcjr:
Yeah, not a blocker for me though, I'm just using the commit hash for now

tasn:
Ah, OK. So it'll wait until the next release.

PR #32, opened 2021-06-03
php: initial verification lib

### LINKED ISSUE #30: Create Initial PHP Library
Create initial php project structure & webhook verification functionality

svix-frank:
Again not very familiar with PHP but added phpunit tests to try and cover our bases.

To test you can run:
```sh
composer install
./vendor/bin/phpunit tests
```
We will probably want to add this to https://packagist.org

Example Usage:
```php
$secret = 'MfKQ9r8GKYqrTwjUPD8ILPZIo2LaLaSw';
$payload = '{"test": 2432232314}';
$header = array(
        'svix-id'  => 'msg_p5jXN8AQM9LWM0D4loKWxJek',
        'svix-timestamp' => '1614265330',
        'svix-signature' => 'v1,g0hM9SsE+OTPJTGt/tmIKtSyZlE3uFJELVlNIOLJ1OE=',
    );

$wh = new \Svix\Webhook($secret);
$json = $wh->verify($payload, $header);
```

Closes #30

### DIFF
--- a/php/src/Webhook.php
+++ b/php/src/Webhook.php
@@ -32,7 +32,7 @@ class Webhook
                 continue;
             }
 
-            if (strcmp($signature, $expectedSignature) == 0) {
+            if (hash_equals($signature, $expectedSignature)) {
                 return json_decode($payload, true);
             }
         }
--- a/php/src/Webhook.php
+++ b/php/src/Webhook.php
@@ -2,20 +2,17 @@
 
 namespace Svix;
 
-class WebhookVerificationException extends \Exception {
-    public function __construct($message, $code = 0, Throwable $previous = null) {
-            parent::__construct($message, $code, $previous);
-    }
-}
-
-class Webhook {
-    public function __construct(string $secret) {
+class Webhook
+{
+    public function __construct(string $secret)
+    {
         $this->secret = base64_decode($secret);
     }
 
-    public function verify($payload, $headers) {
+    public function verify($payload, $headers)
+    {
         if (!isset($headers['svix-id']) || !isset($headers['svix-timestamp']) || !isset($headers['svix-signature'])) {
-            throw new WebhookVerificationException("Missing required headers");
+            throw new Exception\WebhookVerificationException("Missing required headers");
         }
 
         $msgId = $headers['svix-id'];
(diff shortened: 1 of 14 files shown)


tasn:
```
$wh = new \Svix\Webhook($secret);
```
What's up with the backslash? Is this actual PHP syntax?!

php/src/Webhook.php
```diff
@@ -0,0 +1,47 @@
...
+        foreach ($passedSignatures as $versionedSignature) {
+            $sigParts = explode(',', $versionedSignature, 2);
+            $version = $sigParts[0];
+            $expectedSignature = $sigParts[1];
+
+            if (strcmp($version, "v1") != 0) {
+                continue;
+            }
+
+            if (strcmp($signature, $expectedSignature) == 0) {
```
tasn, line 35: Not the end of the world in this case, but to prevent timing attacks you want to use a constant time comparison function.

I tried to look up an example online, and I found this: https://github.com/stripe/stripe-php/blob/f6724447481f6fb8c2e714165e092adad9ca470a/lib/WebhookSignature.php#L51

tasn (changes requested):
I can't really evaluate PHP properly, but found one issue.

php/src/Webhook.php
svix-frank: looks like `hash_equals` is available in php 5.6+ (which I currently have this lib locked to) so we should be good using that (just updated) unless you think we need to widen our support window.

tasn: Support Windows? The OS? Why, is newer PHP not available there?

svix-frank: Sorry by `widen our support window` I meant support PHP lower than 5.6 =)

tasn: I didn't see the `widen our`. Sorry, I'm a skimmer, not a reader. :)

No need, it's god the way it is.

tasn (approved):
I'm not a PHPer, but it looks good to me.

svix-frank:
This should now be ready to merge with the following caveats:

1) To publish to packagist the `composer.json` has to be at the top level of the repo.
2) Looks like when composer pulls a vendored repo it just pulls the all files (unlike go which just pulls .go files) not sure if that;s a big deal or not but wanted to point it out
3) we cannot publish until this is merged into main.

tasn:
Merged. Annoying that PHP has these caveats, but this is a good start.

PR #35, opened 2021-06-04
go: use secure hash compare & add webhook tests

svix-frank:
based on @tasn comments on the PHP & Ruby PRs.


tasn (approved):
Looks good. One question: is `sign` exposed? We should return bytes if yes, not base64.

svix-frank:
> Looks good. One question: is `sign` exposed? We should return bytes if yes, not base64.

nope, sign is not exposed

PR #38, opened 2021-06-06
Create Java Library (#1)

### LINKED ISSUE #1: Create Java Library
Take a look at `regen_openapi.sh` and all of the files in `javascript` and `python` to see how to generate from the openapi spec, configure the generator to not generate some of the files, and make it into a package.
You can also look at https://github.com/etesync/etebase-java for how to create a java package. Though it's an Android package, which is a bit different to normal Java packages.

Keep the same API structure as there is in Python and JS (e.g. `dh.message.create()`).

talhacohen:
Closes #1.


java/src/main/java/svix/SvixOptions.java
```diff
@@ -0,0 +1,19 @@
+package svix;
+
+public class SvixOptions {
+	private final boolean debug;
+	private final String testUrl;
+
+	public SvixOptions(boolean debug, String testUrl) {
```
svix-frank, line 7: Ideally we'd be able to set only one of these options and have the others still be the default value.
Not sure the best way to do this in java but in Go we'd use pointers and only change the value if explicitly set
i.e. `new SvixOptions(true, null)` => `debug: true, testUrl: https://api.svix.com`

java/src/main/java/svix/SvixOptions.java
```diff
@@ -0,0 +1,19 @@
+package svix;
+
+public class SvixOptions {
+	private final boolean debug;
+	private final String testUrl;
```
svix-frank, line 5: can we rename this to `debugUrl` to match our go library

regen_openapi.sh
```diff
@@ -10,4 +10,6 @@ yarn openapi-generator-cli generate -i openapi.json -g python -o python/ -c pyth
 
 yarn openapi-generator-cli generate -i openapi.json -g typescript -o javascript/src/openapi -c javascript/openapi-generator-config.json
 
-yarn openapi-generator-cli generate -i openapi.json -g go -o go/internal/openapi -c go/openapi-generator-config.json
\ No newline at end of file
+yarn openapi-generator-cli generate -i openapi.json -g go -o go/internal/openapi -c go/openapi-generator-config.json
+
+yarn openapi-generator-cli generate -i openapi.json -g java -o java -c java/openapi-generator-config.json
```
svix-frank, line 15: Is it possible to generate the openapi code into it's own subfolder like `java/openapi` and then gitignore it?

svix-frank (changes requested):

java/src/main/java/svix/SvixOptions.java
talhacohen: Done. Note that this is the name in the python implementation.

java/src/main/java/svix/Webhook.java
```diff
@@ -0,0 +1,62 @@
...
+			if (sigParts.length < 2) {
+				continue;
+			}
+			String version = sigParts[0];
+			String signature = sigParts[1];
+			if (!version.equals("v1")) {
+				continue;
+			}
+			if (signature.equals(expectedSignature)) {
+				return new JSON().serialize(payload);
```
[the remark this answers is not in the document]
talhacohen: That's a good question. I'm not using JSON nowadays so I'm not sure what most people are used to. What I did is actually wrong.

I can think of 3 options here, not sure what is the best approach that will be consistent with the other languages.
1. Just return the string as is (a bit weird, as we accept it as an argument and don't use it)
2. Change the method to return boolean
3. If the payload can be represented with an object then return that object

what do you think?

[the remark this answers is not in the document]
svix-frank: I personally think its best to just return a bool (or really nothing and just throw an error) and punt on handling the JSON marshaling and punt that to the user, This is how I did it in the go library and will probably do going forward.  Parsing made sense as a convince in the javascript library but since the payload is arbitrary it really doesn't make sense for us to do it here.

@tasn do you agree?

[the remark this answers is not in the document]
tasn: Yeah, I'm 100% with Frank on this one. Just throw an error on failure and return nothing on success.

P.s, @talhacohen, you are definitely using the string though, for the verification of the signature (or at least should, haha).

java/src/main/java/svix/Webhook.java
```diff
@@ -0,0 +1,62 @@
...
+	public String verify(String payload, HttpHeaders headers) throws WebhookVerificationError {
+		Optional<String> msgId = headers.firstValue("svix-id");
+		Optional<String> msgSignature = headers.firstValue("svix-signature");
+		Optional<String> msgTimestamp = headers.firstValue("svix-timestamp");
+
+		if (msgId.isEmpty() || msgSignature.isEmpty() || msgTimestamp.isEmpty()) {
+			throw new WebhookVerificationError("Missing required headers");
+		}
+
+		String toSign = String.format("%s.%s.%s", msgId, msgSignature, msgTimestamp);
```
svix-frank, line 30: This should be
```java
String toSign = String.format("%s.%s.%s", msgId, msgTimestamp, payload);
```

svix-frank:
It would also be great to setup a testing framework in this PR if possible as well, no need to add API mocks and go crazy just yet but, but at least a framework that covers basic webhook verification cases (just the equivalent cases from [this file](https://github.com/svixhq/svix-libs/blob/main/ruby/spec/webhook_spec.rb) would be a great start)

svix-frank:
> It would also be great to setup a testing framework in this PR if possible as well, no need to add API mocks and go crazy just yet but, but at least a framework that covers basic webhook verification cases (just the equivalent cases from [this file](https://github.com/svixhq/svix-libs/blob/main/ruby/spec/webhook_spec.rb) would be a great start)

java/src/main/java/svix/Webhook.java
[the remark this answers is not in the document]
talhacohen: True, I'm using it, changed to boolean then.
I can also remove all the exceptions here and just return false when it's not valid. for now, keeping the exception for consistency with go.

This code isn't working yet (at least the tests are failing, still looking into it)

java/src/main/java/svix/SvixOptions.java
talhacohen: Done, added overload for SvixOptions and handled null (just in case)

java/src/main/java/svix/SvixOptions.java
talhacohen: Done. Note that this is the name in the python implementation.

regen_openapi.sh
talhacohen: Done.

java/src/main/java/svix/Webhook.java
talhacohen: Yeap, missed that before, fixed. thanks!

talhacohen:
Missed your comments about testing, already added testing which suggested that the Webhook.java isn't working.
I'm digging into it, not sure whether the code or the tests are broken tbh

java/bin/.openapi-generator-ignore
```diff
@@ -0,0 +1,23 @@
...
+
+# You can recursively match patterns against a directory, file or extension with a double asterisk (**):
+#foo/**/qux
+# This matches foo/bar/qux, foo/baz/qux, and foo/bar/baz/qux
+
+# You can also negate patterns with an exclamation (!).
+# For example, you can ignore all files in a docs folder with the file extension .md:
+#docs/*.md
+# Then explicitly reverse the ignore rule for a single file:
+#!docs/README.md
```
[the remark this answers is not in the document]
talhacohen: Done

java/src/main/java/svix/Webhook.java
```diff
@@ -0,0 +1,62 @@
...
+			String[] sigParts = versionedSignature.split(",");
+			if (sigParts.length < 2) {
+				continue;
+			}
+			String version = sigParts[0];
+			String signature = sigParts[1];
+			if (!version.equals("v1")) {
+				continue;
+			}
+			if (signature.equals(expectedSignature)) {
```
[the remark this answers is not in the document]
talhacohen: Done, I hope this is the correct way to use it (never used MessageDigest before)
Oops, sorry about that. I copied from the JS implementation.

java/src/main/java/svix/SvixOptions.java
tasn: I think the right way of doing it is by doing a builder-like structure. So init things to a certain value by default, and make it possible to do things like:
```
options = new SvixOptions().setDebug(true).setTestUrl("http...");
```
See what I mean?

talhacohen: Done. I didn't create another class for the Builder though. (followed your suggestion above)

tasn: Yeah, no need for a real builder. That's why I said builder-like. :P

java/src/main/java/svix/Webhook.java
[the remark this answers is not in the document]
talhacohen: Done

regen_openapi.sh
svix-frank: I'm still seeing the code generated to java/ i think that force push may have removed some of your changes

talhacohen: I changed it in the openapi-generator-config.json (the package). 
Do you mean that we should have all generated files (including the non-Java code) in an internal file?

tasn: Why are the generated files even committed to the repo? They shouldn't be... Add the directory to .gitignore, and remove all of the files...

talhacohen: ah got it
I think I can just add all the bin to the gitignore, another option is just to create everything inside an inner folder. I see that this is what you did for golang. I can do either

svix-frank: Inner folder would probably be the the best, and then as @tasn mentioned add that folder to the `java/.gitignore` file.

talhacohen: sg, done

svix-frank: looks like the most recent changes broke some things, the internal code is now being generated to `java/internal/src/main/java/openapi...` but your bindings still expect them in `java/src/main/java/openapi`

talhacohen: hmm that's weird, I changed build.gradle to internal/openapi/src/main/java and it works here. are you sure you are synced to the latest?

java/build.gradle
```diff
@@ -0,0 +1,58 @@
...
+        classpath 'com.android.tools.build:gradle:2.3.+'
+        classpath 'com.github.dcendents:android-maven-gradle-plugin:1.5'
+    }
+}
+
+repositories {
+    jcenter()
+}
+sourceSets {
+    main.java.srcDirs = ['src/main/java', 'internal/openapi/src/main/java']
```
svix-frank, line 23: should be 
```suggestion
    main.java.srcDirs = ['src/main/java', 'internal/src/main/java']
```
right?

svix-frank (changes requested):

java/build.gradle
talhacohen: Sorry about that, I accidentally didn't modify the .sh . I changed it to be internal/openapi to be similar to golang

java/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,3 @@
+<manifest package="openapi.svix" xmlns:android="http://schemas.android.com/apk/res/android">
```
tasn, line 1: It should be a java package. Not android specific. Just making sure.

java/lib/src/main/java/svix/Application.java
```diff
@@ -13,23 +13,23 @@
 		api = new ApplicationApi();
 	}
 
-	public ListResponseApplicationOut list(String iterator, Integer limit) throws ApiException {
+	public ListResponseApplicationOut list(final String iterator, final Integer limit) throws ApiException {
```
talhacohen, line 16: As far as I know it's not common to set 'final' on method params
It's fine to keep it and doesn't really change a lot though

svix-frank: Ah, I added Github's Superlinter who's default setup yelled about this, I can disable the rule and undo this change if you think this is just added noise.

svix-frank (approved):
This looks great to me as the first implementation, I'll make a separate PR setting up maven central distribution.

java/lib/src/main/java/svix/Application.java
tasn: I think having `final` is probably good. It's maybe just not Google-style, but I don't think it detracts anything. Following what the linter says sounds good.

.github/workflows/other-lint.yml
```diff
@@ -0,0 +1,36 @@
+name: Lint Code Base
+
+on:
+  pull_request:
+    paths-ignore:
```
tasn, line 5: Why paths-ignore and not just white list the correct path?

tasn (commented):
Made just one comment, but looks good to me otherwise.

.github/workflows/other-lint.yml
svix-frank: The `SuperLinter` lints a ton of different languages (including ruby, php, etc) so i just ignored the paths that our other linters already cover

tasn:
Great job, and thanks a lot!

PR #40, opened 2021-06-08
Bump version and update changelog.

tasn:


tasn:
I assume everything is ready for a new release?

svix-frank (approved):
I bumped the ruby gem as well (although there isn't anything new for ruby in 0.16.0), now looks good to me.

Also assigned myself https://github.com/svixhq/svix-libs/issues/41 to add a "version bump" command as there are already too many files that need to be changed

PR #47, opened 2021-06-09
Java: improved imports

svix-frank:
Improves the usage a bit and pulls in the new openapi changes.

Models are now at `com.svix.models`, api is at `com.svix.internal.api`.

It still exposes the internal openapi code but I don't think its possible to avoid this.  Its now all under `com.svix.internal` though so at least we can be explicit about our compatibility promise (i.e. dont use com.svix.internal).

Creating an application now looks like this
```java
import com.svix.Svix;
import com.svix.models.ApplicationIn;
import com.svix.exceptions.ApiException;

Svix svix = new Svix("AUTH_TOKEN");
try {
    svix.getApplication().create(new ApplicationIn().name("hello");
} catch(ApiException e) {
    System.out.println(e);
}
```
And a Webhook Verification looks like this:
```java
import com.svix.Webhook;
import com.svix.exceptions.WebhookVerificationException;

Webhook webhook = new Webhook("SECRET");
try {
    webhook.verify("payload", headers);
} catch(WebhookVerificationException e) {
    System.out.println(e);
}
```

### DIFF
--- a/java/build.gradle
+++ b/java/build.gradle
@@ -37,7 +37,7 @@ jar {
                    'Implementation-Version': VERSION_NAME,
                    'Implementation-Vendor': VENDOR_NAME,
                    'Bundle-SymbolicName': POM_ARTIFACT_ID,
-                   'Export-Package': 'com.svix.*')
+                   'Export-Package': 'com.svix,com.svix.models.*')
 
         archiveVersion = VERSION_NAME
     }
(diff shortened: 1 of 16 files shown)


tasn (commented):
Looks good! Actually though, one suggestion: maybe change it to `com.svix.exceptions.api`?

Also, do we really raise a bare Exception from signing? We haven't created a special exception type?

tasn (approved):
I hate how redundant it is to manually set the exception transformations. Though let's not spend any more time on this. It's at a good stage now, I think it's good enough for now.

PR #49, opened 2021-06-09
Java: Change package structure to improve imports & remove generated code from main package 

svix-frank:
Improves the usage a bit and pulls in the new openapi changes.

Models are now at `com.svix.models`, api is at `com.svix.internal.api` (since theres a bunch of extra things that openapi generates one above `api`.

It still exposes the internal openapi code but I don't think its possible to avoid this. Its now all under `com.svix.internal` though so at least we can be explicit about our compatibility promise (i.e. dont use `com.svix.internal`).

Creating an application now looks like this
```java
import com.svix.Svix;
import com.svix.models.ApplicationIn;
import com.svix.exceptions.ApiException;

Svix svix = new Svix("AUTH_TOKEN");
try {
    svix.getApplication().create(new ApplicationIn().name("hello");
} catch(ApiException e) {
    System.out.println(e);
}
```
And a Webhook Verification looks like this:
```java
import com.svix.Webhook;
import com.svix.exceptions.WebhookVerificationException;

Webhook webhook = new Webhook("SECRET");
try {
    webhook.verify("payload", headers);
} catch(WebhookVerificationException e) {
    System.out.println(e);
}
```


tasn:
Why a new PR?

svix-frank:
Don't love this but I'm not sure there's anything better we can do without modifying our own templates.

@talhacohen does this seem alright to you?

svix-frank:
> Why a new PR?

For some reason it said the branch was `protected` and wouldn't let me push to it anymore so I had to close and reopen.

java/lib/src/main/java/com/svix/MessageAttempt.java
```diff
@@ -16,28 +16,52 @@
 	}
 
 	public ListResponseMessageAttemptOut list(final String appId, final String msgId, final FetchOptionsMessageAttempt options) throws ApiException {
-		return api.listAttemptsApiV1AppAppIdMsgMsgIdAttemptGet(msgId, appId, options.getIterator(), options.getLimit(), options.getMessageStatus());
+		try {
+			return api.listAttemptsApiV1AppAppIdMsgMsgIdAttemptGet(msgId, appId, options.getIterator(), options.getLimit(), options.getMessageStatus());
+		} catch (com.svix.internal.ApiException e) {
```
talhacohen, line 21: Can this just be ApiException?

java/lib/src/main/java/com/svix/exceptions/ApiException.java
```diff
@@ -0,0 +1,44 @@
...
+
+import java.util.Map;
+import java.util.List;
+
+public class ApiException extends Exception {
+    private int code = 0;
+    private Map<String, List<String>> responseHeaders = null;
+    private String responseBody = null;
+
+    public ApiException(final String message, final Throwable throwable, final int code, final Map<String, List<String>> responseHeaders, final String responseBody) {
```
talhacohen, line 11: Maybe we should accept com.svix.internal.ApiException instead?
Then we can change below to call getCode of that member

Another option is to extend com.svix.internal.ApiException and not create the public method

svix-frank: I originally had it accept a `com.svix.internal.ApiException` but IMO it felt like it broke the point of having `com.svix.internal` if its publicly exposed in a constructor in `com.svix.exceptions`

Extending sounds great though, done!

talhacohen:
> Don't love this but I'm not sure there's anything better we can do without modifying our own templates.
> 
> @talhacohen does this seem alright to you?

I don't have a better suggestion, I think this is fine 

btw, thanks for fixing WebhookVerificationException

java/lib/src/main/java/com/svix/exceptions/ApiException.java
talhacohen: Got it, though now we expose it in the Utils class, not sure if that's better. Should me move Utils to an internal package?

java/openapi-generator-config.json
```diff
@@ -1,4 +1,4 @@
 {
-  "apiPackage": "com.svix.generated.api",
-  "modelPackage": "com.svix.generated.model"
+  "apiPackage": "com.svix.internal.api",
+  "modelPackage": "com.svix.models"
```
talhacohen, line 3: Should we have any indication that it's openapi (or generated)? and only add internal to the api

It's fine not keep as is

java/lib/src/main/java/com/svix/exceptions/ApiException.java
svix-frank: Isn't `Utils` only accessible form within `com.svix`?

talhacohen: Right, sorry. 
Anyway, it's fine to keep as is. You can change the constructor to be private package if you prefer (and accept com.svix.internal.ApiException).

svix-frank: ah right 🤦. great idea!

java/openapi-generator-config.json
fcjr: I personally think its cleaner like this but i can be convinced if you think this is weird 🤷

talhacohen: I don't feel strongly about it, let's keep as is

tasn (approved):
Overall looks good to me. I trust you guys to conclude the open discussions correctly.

java/lib/src/main/java/com/svix/exceptions/ApiException.java
talhacohen: We can move it to `com.svix`, though dunno if that's better. Up to you, both sounds ok to me

tasn:
Can this go in? What's missing?

svix-frank:
> Can this go in? What's missing?

Let's merge this in so we can apply #57 on top of it

tasn:
Can you please change the title to be more descriptive so I can just squash and merge?

svix-frank:
> Can you please change the title to be more descriptive so I can just squash and merge?

Done

PR #56, opened 2021-06-14
Fix java build

talhacohen:


svix-frank:
Hey Tal, this is fixed in #49 which is about to be merged, so I am going to close this in favor of that PR.

PR #59, opened 2021-06-14
php: comply with psr-4 autoloading

svix-frank:
Follows naming convention for psr-4 autoloading


tasn (approved):
Good on my end, should I merge it in?

svix-frank:
> Good on my end, should I merge it in?

Yes please!

PR #62, opened 2021-06-16
ruby: Add openapi bindings

svix-frank:
Builds, and is functional but still exposes all of the openapi code in an unacceptable way.  Still trying to figure out how to encapsulate the openapi code in a good way.

- [ ] Encapuslate/Hide Openapi's API & root classes
- [ ] finish implementeding FetchOptionsMessageAttempt
- [ ] test iterator
- [ ] look into whether or not its possible to rename `Client` to `Svix` so that it matches the other libraries.


tasn:
How are we on this one?

tasn:
Why was this closed?

PR #65, opened 2021-06-20
tests: Add Multi Signature Tests

### LINKED ISSUE #55: [0.2] Tests: make sure the verification libraries can handle multiple signatures
We may get signatures such as:
```
v1,BADSIG v1,GOODSIG v2,BADSIG
```
And we need to make sure that all the libraries know how to handle this and verify correctly

svix-frank:
Assuming that multi-sig payloads will always have all of the signature set as a single `svix-signature` header and not multiple ones.

Closes #55

### DIFF
--- a/go/webhook_test.go
+++ b/go/webhook_test.go
@@ -4,6 +4,7 @@ import (
 	"encoding/base64"
 	"fmt"
 	"net/http"
+	"strings"
 	"testing"
 	"time"
 )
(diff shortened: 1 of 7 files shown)


go/webhook_test.go
```diff
@@ -96,6 +97,19 @@ func TestWebhook(t *testing.T) {
 			testPayload: newTestPayload(time.Now().Add(tolerance + time.Second)),
 			expectedErr: true,
 		},
+		{
+			name:        "valid multi sig is valid",
+			testPayload: newTestPayload(time.Now()),
+			modifyPayload: func(tp *testPayload) {
+				sigs := []string{
+					"v1,Ceo5qEr07ixe2NLpvHk3FH9bwy/WavXrAFQ/9tdO6mc=",
```
tasn, line 105: Could you maybe make the correct signature *not* be the first but rather sandwiched between two failing ones?
This will make sure we don't make mistakes like: only working when the first passes, or when the last passes.

javascript/src/webhook.test.ts
```diff
@@ -36,64 +36,67 @@ class TestPayload {
   }
 }
 
-test('missing id raises error', () => {
```
tasn, line 39: Could you maybe move the prettier run to its own PR? Same with the Python changes.
Not the end of the world if too annoying this time, but at least for next time then...

go/webhook_test.go
svix-frank: It was always the _last_ one but I added an invalid one to the end of all the tests to cover future edge cases

tasn (approved):

PR #66, opened 2021-06-20
go: expose api errors

svix-frank:


tasn (approved):
Not proficient in Go, but it looks alright.

tasn:
Actually

tasn:
What's the "error" field? Is it `code` from the API? Or what?

The problem is that we have two kinds of errors, we have validation errors (422) and we have the HTTP errors (the rest) which are structured differently.

svix-frank:
> What's the "error" field? Is it `code` from the API? Or what?

It's the display string for the error `422 Validation Error` or `409 Conflict`, not ideal.  The only way I can get the actual status code is to grab it from the resp object. =/

tasn:
> > What's the "error" field? Is it `code` from the API? Or what?
> 
> It's the display string for the error `422 Validation Error` or `409 Conflict`, not ideal. The only way I can get the actual status code is to grab it from the resp object. =/

Ahhhh, cool. :)

PR #82, opened 2021-06-28
 Python: Cleanup exports, export exceptions as `svix.exceptions` and webhooks as `svix.webhooks`

svix-frank:
Export exceptions as `svix.exceptions`.  @tasn please let me know if I did this correctly 🙈


python/svix/api/exceptions.py
```diff
@@ -0,0 +1,11 @@
+from ..openapi_client.exceptions import (  # noqa
```
tasn, line 1: Why not expose all of the exceptions from `svix/exceptions.py`? Both API and receiver?

python/svix/api/api.py
```diff
@@ -2,32 +2,32 @@
 from contextlib import contextmanager
```
tasn, line 2: This is a very confusing name for this file. :)

I think you don't need to change it though. Just keep as it was.

python/svix/api/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+    DashboardAccessOut,
+    FetchOptions,
+    ListResponseApplicationOut,
+    ListResponseMessageOut,
+    MessageIn,
+    MessageOut,
+    Svix,
+    SvixOptions,
+)
+from .exceptions import (  # noqa
```
tasn, line 13: This exports everything like this: `svix.api.ApiException`, which I don't think is what you want...

python/svix/__init__.py
```diff
@@ -1,13 +1 @@
-from .api import (  # noqa
```
tasn, line 1: This means these are no longer expected from the top level...

tasn (changes requested):
We exposed everything (well, other than the things we forgot) in the top-level, now we expose the webhooks stuff in the top level, but not Svix and etc.

This is the correct way to expose things btw: 
https://github.com/svix/python-ksuid/blob/main/ksuid/__init__.py#L3
Rather than the `# noqa` hack we did (which also works)>

In summary:
If you want to fix the organization, I'd probably do something like this:
1. Split to two submodules (svix.api and svix.webhooks - instead of receiver)
2. Keep the existing imports in `__init__.py` for backwards compat (though update the docs to the new location).

tasn (approved):
A beauty. 😍

PR #86, opened 2021-07-02
Python: Disable return type checking

svix-frank:
It looks like the openapi-generator code is checking optional fields as required on return types, I haven't had a chance to dive deep into why it is actually happening but I figure we can just disable it for now which "fixes" the problem, I'll open an issue with openapi about it.

Context:
When making any calls that return classes with optional fields it checks them as required and errors out.
for example:
`svix.message.create("app_ID", MessageIn(event_type="test", payload={}))`
throws:
`svix.openapi_client.exceptions.ApiTypeError: Invalid type for variable 'event_id'. Required value type is str and passed type was NoneType at ['received_data']['event_id']`

### DIFF
--- a/python/svix/api.py
+++ b/python/svix/api.py
@@ -74,11 +74,13 @@ class Authentication(ApiBase[AuthenticationApi]):
 
     def dashboard_access(self, app_id: str) -> DashboardAccessOut:
         with self._api() as api:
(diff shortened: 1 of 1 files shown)


tasn (approved):
I guess it doesn't really matter as we know the data from the server is correct, though still... 😭 

Great catch, and let's get it in!

PR #87, opened 2021-07-03
All: Regen openapi spec, fix message resend

svix-frank:
Fixes a bug where the javascript library will throw on message resend


.github/workflows/other-lint.yml
```diff
@@ -38,4 +38,5 @@ jobs:
           VALIDATE_JSCPD: false
           VALIDATE_PHP_PHPSTAN: false
           VALIDATE_PHP_PSALM: false
+          VALIDATE_OPENAPI: false
```
svix-frank, line 41: Made a ticket to remove the not so super super linter.

tasn (approved):
Haha, yeah, very not super. :)

PR #89, opened 2021-07-03
Bump version & update changelog

svix-frank:
Will merge and make a release after #87

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## Version 0.20.0
+* All: Update OpenAPI spec
+* Javascript: Fix bug with message attempt resend (would throw on return)
+* Python: Fix optional bug where optional return types would throw
+
 ## Version 0.19.1
 * Python: Fix issue with pip package not generating the `openapi_client` properly on deploy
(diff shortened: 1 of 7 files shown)


ChangeLog.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## Version 0.20.0
```
svix-frank, line 3: Though this kind of a bug fix, it changes the API so I bumped the minor version

tasn: Does it really change the API?

tasn: Ahh yes, it does for Go. :(

tasn (approved):

ChangeLog.md
svix-frank: Actually all languages, unfortunately.   javascript` Promise<any>` -> `Promise<void>`, java `Object` -> `void`, python `MessageAttemptOut` -> `None` (which i think was actually always wrong)

tasn: Yeah, but it's barely an API break. Will only break in weird scenarios, not under normal usage.

PR #92, opened 2021-07-09
Update Openapi Schema

svix-frank:
Update to dev spec.


go/internal/openapi/model_endpoint_stats.go
```diff
@@ -16,20 +16,20 @@ import (
...
+	Pending int32 `json:"pending"`
+	Success int32 `json:"success"`
 }
 
 // NewEndpointStats instantiates a new EndpointStats object
 // This constructor will assign default values to properties that have it defined,
 // and makes sure properties required by API are set, but the set of arguments
 // will change when the set of required properties is changed
-func NewEndpointStats(success int32, pending int32, fail int32, ) *EndpointStats {
+func NewEndpointStats(fail int32, pending int32, success int32, ) *EndpointStats {
```
tasn, line 28: The generated file seems to have changed the order of things. Is this normal, is this OK?

tasn (approved):
Looks fine, just made one comment.

Please double check yourself though, as Go code looks foreign to me.

go/internal/openapi/model_endpoint_stats.go
svix-frank: It's definitely not ideal, luckily we don't expose any of these constructors that take positional arguments, looked through all the other libraries and it looks like we use named args everywhere so we're good on all fronts.

tasn: Yeah, I used named parameters in JS and Python exactly because of this.

My main concern is Java, but I guess you checked? Because it's just Strings we won't even get type errors. :(

PR #94, opened 2021-07-10
Csharp: Initial Lib & Webhook Verification

### LINKED ISSUE #83: [0.5] Add C# webhook verification & signing
Including deployment

svix-frank:
Closes #83

Assuming we should get @svix-yair to sign off on this one.

### DIFF
--- a/csharp/Svix/Webhook.cs
+++ b/csharp/Svix/Webhook.cs
@@ -4,16 +4,20 @@ using System.Net;
 using System.Security.Cryptography;
 
 using Svix.Exceptions;
-namespace Svix {
-    public class Webhook {
+namespace Svix
+{
+    public class Webhook
+    {
 
         internal static readonly UTF8Encoding SafeUTF8Encoding = new UTF8Encoding(false, true);
 
         private static readonly int TOLERANCE_IN_SECONDS = 60 * 5;
         private static string prefix = "whsec_";
         private string key;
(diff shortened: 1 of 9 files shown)


tasn (approved):
I don't know csharp, but if the tests pass, and the code looks alright, I think this can go in. :)

tasn:
Yeah, probably wait on Yair

csharp/Svix/Webhook.cs
```diff
@@ -0,0 +1,104 @@
...
+            {
+                key = key.Substring(prefix.Length);
+            }
+
+            this.key = Convert.FromBase64String(key); ;
+        }
+
+        public void Verify(string payload, WebHeaderCollection headers)
+        {
+            string msgId = headers.Get("svix-id");
```
svix-yair, line 29: nit: Move keys to const and also use in the tests

svix-yair (approved):
LGTM one nit comment

csharp/Svix/Webhook.cs
svix-frank: I'll make them consts, the only question is do we want to expose these as part of the public API? The tests are in their own project so there's no way to access them there without publicly exposing afaik.

svix-frank: I made them consts but kept them private, if you think we should expose this we discuss and add them in a separate PR.

PR #96, opened 2021-07-11
Add signing functions

### LINKED ISSUE #12: Add functions to sign webhooks.

svix-frank:
Closes #12

### DIFF
--- a/go/webhook.go
+++ b/go/webhook.go
@@ -49,15 +49,18 @@ func (wh *Webhook) Verify(payload []byte, headers http.Header) error {
 	}
 
 	// enforce timestamp tolerance
-	err := verifyTimestamp(msgTimestamp)
+	timestamp, err := verifyTimestamp(msgTimestamp)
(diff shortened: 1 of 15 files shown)


go/webhook.go
```diff
@@ -77,12 +80,15 @@ func (wh *Webhook) Verify(payload []byte, headers http.Header) error {
 	return errNoMatchingSignature
 }
 
-func sign(key []byte, toSign string) []byte {
-	h := hmac.New(sha256.New, key)
+func (wh *Webhook) Sign(msgId, timestamp string, payload []byte) (string, error) {
```
svix-frank, line 83: My only question is should we accept the time objects for the timestamp in each language or just a string?

tasn: I'd change the timestamp parameter to int/long (rather than string). We care about it being seconds since epoch, so let's at least enforce that.

tasn (approved):
Even better. :)

PR #121, opened 2021-08-24
Add User Agents

svix-frank:


go/internal/version/version.go
```diff
@@ -0,0 +1,3 @@
+package version
+
+const Version = "0.25.0"
```
svix-frank, line 3: there's no other way to inject this in go

java/lib/src/main/java/com/svix/Svix.java
```diff
@@ -5,6 +5,7 @@
 import com.svix.internal.auth.HttpBearerAuth;
 
 public final class Svix {
+	public static final String VERSION = "0.25.0";
```
svix-frank, line 8: or java

javascript/src/index.ts
```diff
@@ -38,11 +41,22 @@ import * as base64 from "@stablelib/base64";
 import * as sha256 from "fast-sha256";
 
 const WEBHOOK_TOLERANCE_IN_SECONDS = 5 * 60; // 5 minutes
+const VERSION = "0.25.0";
```
svix-frank, line 44: We don't use a packager so this I don't know of another option here but in the future we could use rollup/webpack and inject this via a plugin

python/svix/api.py
```diff
@@ -68,6 +68,10 @@ def __init__(self, configuration: Configuration) -> None:
     @contextmanager
     def _api(self) -> t.Generator[ApiClass, None, None]:
         with ApiClient(self._configuration) as api_client:  # type: ignore
+            
+            import svix
+            api_client.user_agent = f"svix-libs/{svix.__version__}/python"
+
```
svix-frank, line 74: Does this seem reasonable?

tasn (approved):
Can you just add a point on the readme/some process to make sure that we bump these too?

python/svix/api.py
```diff
@@ -68,6 +68,11 @@ def __init__(self, configuration: Configuration) -> None:
     @contextmanager
     def _api(self) -> t.Generator[ApiClass, None, None]:
         with ApiClient(self._configuration) as api_client:  # type: ignore
+
+            import svix
```
tasn, line 72:
```suggestion
            from . import __version__
```
Maybe? (make sure it works)

PR #130, opened 2021-09-09
Java: APIException set error message to error detail if possible

### LINKED ISSUE #129: Java: ApiException stacktrace doesn't contain the actual error message
I noticed that when a request to Svix fails, the stacktrace doesn't contain a description of what went wrong exactly.

E.g. this is the stack trace you get when updating an endpoint fails:

```
...
2021-09-08T12:19:36.711851+00:00 app[web.1]: at java.base/java.lang.Thread.run(Thread.java:829)
2021-09-08T12:19:36.711851+00:00 app[web.1]: Caused by: com.svix.exceptions.ApiException:
2021-09-08T12:19:36.711851+00:00 app[web.1]: at com.svix.Utils.wrapInternalApiException(Utils.java:10)
2021-09-08T12:19:36.711851+00:00 app[web.1]: at com.svix.Endpoint.update(Endpoint.java:45)
2021-09-08T12:19:36.711852+00:00 app[web.1]: at util.SvixClient.updateEndpoint(SvixClient.java:57)
2021-09-08T12:19:36.711853+00:00 app[web.1]: ... 35 common frames omitted
2021-09-08T12:19:36.711853+00:00 app[web.1]: Caused by: com.svix.internal.ApiException:
2021-09-08T12:19:36.711853+00:00 app[web.1]: at com.svix.internal.ApiClient.handleResponse(ApiClient.java:1023)
2021-09-08T12:19:36.711854+00:00 app[web.1]: at com.svix.internal.ApiClient.execute(ApiClient.java:936)
2021-09-08T12:19:36.711854+00:00 app[web.1]: at com.svix.internal.api.EndpointApi.updateEndpointApiV1AppAppIdEndpointEndpointIdPutWithHttpInfo(EndpointApi.java:894)
2021-09-08T12:19:36.711854+00:00 app[web.1]: at com.svix.internal.api.EndpointApi.updateEndpointApiV1AppAppIdEndpointEndpointIdPut(EndpointApi.java:868)
2021-09-08T12:19:36.711854+00:00 app[web.1]: at com.svix.Endpoint.update(Endpoint.java:43)
2021-09-08T12:19:36.711855+00:00 app[web.1]: ... 36 common frames omitted
```

I assume there's something wrong with the request body, but I don't know what exactly.

Now I could write code to print out the response body inside the exception, but it would be nice if this was done out of the box.

svix-frank:
Not all openapi internal exceptions set the message, which when printing, gives you minimal context unless the user calls `getCode()` or `getResponseBody()`.  One solution is to override the getMessage() function to return the body if the message is missing.

A more 'complete' solution is probably to try and parse the body and set the message to the `error detail`, both are trivial, here I parse the body and try to extract the detail, otherwise we default back to the responseBody.

Before:
```
com.svix.exceptions.ApiException:
```

After
```
com.svix.exceptions.ApiException: Invalid token. (authentication_failed)
```

Fixes #129.

### DIFF
--- a/java/lib/src/main/java/com/svix/exceptions/ApiException.java
+++ b/java/lib/src/main/java/com/svix/exceptions/ApiException.java
@@ -28,7 +28,7 @@ public class ApiException extends com.svix.internal.ApiException {
             try {
                 Gson gson = new Gson();
                 APIError error = gson.fromJson(responseBody, APIError.class);
-                this.message = error.getDetail();
+                this.message = String.format("%s (%s)", error.getDetail(), error.getCode());
             } catch (Exception e) {
                 this.message = responseBody;
             }
--- a/java/lib/src/main/java/com/svix/exceptions/ApiException.java
+++ b/java/lib/src/main/java/com/svix/exceptions/ApiException.java
@@ -7,4 +7,12 @@ public class ApiException extends com.svix.internal.ApiException {
     public ApiException(final String message, final Throwable throwable, final int code, final Map<String, List<String>> responseHeaders, final String responseBody) {
         super(message, throwable, code, responseHeaders, responseBody);
     }
+
+    public String getMessage() {
+        String msg = super.getMessage();
+        if (msg == "") {
+            msg = this.getResponseBody();
+        }
+        return msg;
+    }
 }
--- a/java/lib/src/main/java/com/svix/exceptions/ApiException.java
+++ b/java/lib/src/main/java/com/svix/exceptions/ApiException.java
@@ -2,17 +2,44 @@ package com.svix.exceptions;
 
 import java.util.Map;
 import java.util.List;
+import com.google.gson.Gson;
+
+class APIError {
+    private String code;
+    private String detail;
+
+    public String getCode() {
+        return this.code;
+    }
+
+    public String getDetail() {
+        return this.detail;
+    }
+}
 
 public class ApiException extends com.svix.internal.ApiException {
+
+    private String message;
+
     public ApiException(final String message, final Throwable throwable, final int code, final Map<String, List<String>> responseHeaders, final String responseBody) {
         super(message, throwable, code, responseHeaders, responseBody);
+
+        if ((message == null || message.isEmpty()) && responseBody != null) {
+            try {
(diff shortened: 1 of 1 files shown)


java/lib/src/main/java/com/svix/exceptions/ApiException.java
```diff
@@ -2,9 +2,44 @@
...
+    private String message;
+
     public ApiException(final String message, final Throwable throwable, final int code, final Map<String, List<String>> responseHeaders, final String responseBody) {
         super(message, throwable, code, responseHeaders, responseBody);
+
+        if ((message == null || message.isEmpty()) && responseBody != null) {
+            try {
+                Gson gson = new Gson();
+                APIError error = gson.fromJson(responseBody, APIError.class);
+                this.message = error.getDetail();
```
tasn, line 31: Maybe also show the code?

tasn (approved):

java/lib/src/main/java/com/svix/exceptions/ApiException.java
svix-frank: Yeah we can do that, although at that point it might not make sense to parse the body at all and just print the whole thing as `message`?

`com.svix.exceptions.ApiException: Invalid token. (Code: "authentication_failed")`
vs
`com.svix.exceptions.ApiException: {"code":"authentication_failed","detail":"Invalid token."}`

What do you think?

tasn:
```
com.svix.exceptions.ApiException: authentication_failed: Invalid token.
```
How about?

tasn: Actually, maybe your first suggestion or:
```
com.svix.exceptions.ApiException: Invalid token. (authentication_failed)
```
Or
```
com.svix.exceptions.ApiException: Invalid token. (code: authentication_failed)
```

svix-frank: i like `com.svix.exceptions.ApiException: Invalid token. (authentication_failed)`

PR #134, opened 2021-09-10
Python: improve README.

tasn:


python/README.md
```diff
@@ -1,7 +1,21 @@
...
+The Svix client API
+
+## Usage
+
+```sh
+pip install Svix
+```
+
+```python
+from svix import Svix, ApplicationIn
```
svix-frank, line 12:
```suggestion
from svix.api import Svix, ApplicationIn
```

svix-frank (approved):

python/README.md
tasn: This is where I started, but I changed it to this following the code in docs.svix.com
Which one is it? Should we update the docs then?

svix-frank: We should update the docs then to suggest this, we haven't been exporting new types from `__init__`

PR #137, opened 2021-09-16
bump openapi, and add new fields

svix-frank:


tasn (approved):
Approved, but wait for prod deployment.

PR #138, opened 2021-09-17
Add initial wrappers

talhacohen:


kotlin/lib/src/main/kotlin/Application.kt
```diff
@@ -0,0 +1,56 @@
+package com.svix.kotlin
+
+import com.svix.kotlin.internal.apis.ApplicationApi
+import com.svix.kotlin.internal.infrastructure.ServerException
+import com.svix.kotlin.internal.models.ApplicationIn
+import com.svix.kotlin.internal.models.ApplicationOut
+import com.svix.kotlin.internal.models.ListResponseApplicationOut
```
svix-frank, line 7: looking great so far, one concern i have is that the models are under `com.svix.kotlin.internal` in java we had the ability to move the models to a separate package like `com.svix.kotlin.models` via an openapi-generator config option but looking at the [kotlin generator](https://github.com/OpenAPITools/openapi-generator/blob/master/docs/generators/kotlin.md) I'm not seeing a similar option.  @talhacohen is there a way in kotlin to re-export the models package a level up? (otherwise we might need to get our hands dirty with the templates.

svix-frank: Nope, the bindings that @talhacohen is working on look good so far, but this will be a blocker for release

talhacohen: Done

kotlin/lib/src/main/kotlin/Message.kt
```diff
@@ -0,0 +1,35 @@
+package com.svix.kotlin
+
+import com.svix.kotlin.exceptions.ApiException
+import com.svix.kotlin.internal.apis.MessageApi
+import com.svix.kotlin.models.ListResponseMessageOut
+import com.svix.kotlin.models.MessageIn
+import com.svix.kotlin.models.MessageOut
+
+class Message(debugUrl: String = SvixOptions.DEFAULT_URL) {
```
tasn, line 9: Can we make the constructor private so that others can't instantiate it?  (same for all of the rest of the models)

kotlin/lib/src/main/kotlin/EventTypeListOptions.kt
```diff
@@ -0,0 +1,3 @@
+package com.svix.kotlin
+
+class EventTypeListOptions(var withContent: Boolean = false) : ListOptions()
```
tasn, line 3: Put this in the EventType file? (same for the rest of them)

tasn, line 3: This is wrong, it means you can't set iterator/limit in EventTypeListOption, you need to put them as params and pass to ListOptions iirc.

kotlin/lib/src/main/kotlin/SvixOptions.kt
```diff
@@ -0,0 +1,7 @@
+package com.svix.kotlin
```
tasn, line 1: This should be in Svix.kt

tasn (commented):
Made a few comments, but looks good overall.

kotlin/lib/src/main/kotlin/Application.kt
```diff
@@ -0,0 +1,51 @@
+package com.svix.kotlin
+
+import com.svix.kotlin.exceptions.ApiException
+import com.svix.kotlin.internal.apis.ApplicationApi
+import com.svix.kotlin.models.ApplicationIn
+import com.svix.kotlin.models.ApplicationOut
+import com.svix.kotlin.models.ListResponseApplicationOut
+
+class Application(debugUrl: String = SvixOptions.DEFAULT_URL) {
```
tasn, line 9: This is an internal helper that we use, maybe call it (and others) as such? Like `ApplicationApiWrapper` so it's kind of not really accessible just by the means of having a shit name. :)

talhacohen: I'm not sure how each client will use it, but if someone wants to do
val application : ** = svix.application 
they will need to have access to it

tasn: Yeah, they shouldn't. Only access it through svix.application. Func

kotlin/lib/src/main/kotlin/Message.kt
talhacohen: Done

kotlin/lib/src/main/kotlin/Application.kt
talhacohen: The class name is accessible as it's returned from Svix.kt, I think it's fine to keep this name, though I changed the constructor to be internal

kotlin/lib/src/main/kotlin/EventTypeListOptions.kt
talhacohen: Done.

kotlin/lib/src/main/kotlin/Application.kt
tasn: It's confusing with the model itself (as in, you'd think it may be the Application model).

kotlin/lib/src/main/kotlin/EventTypeListOptions.kt
talhacohen: Yeah it was WIP, changed to builder

kotlin/lib/src/main/kotlin/Application.kt
talhacohen: Not saying it's a good thing, but this is what we have in Java

I can change their package to model if you think it's better

in the model we currently only have ApplicationApi

tasn: Cool!

talhacohen:
@svix-frank ptal
I'll try to get to tests before Monday, not sure whether I'll do it

talhacohen:
@svix-frank @tasn I converted the java tests (most of it atleast), it's currently not working with payload containing { or } due to Illegal base64 character 7b. 
I'm not sure whether this is a real use case, will be able to take a look only tomorrow

kotlin/lib/src/main/kotlin/Application.kt
svix-frank: 🙌 nice! The documentation I was looking at didn't have this option listed for jolting

kotlin/lib/src/main/kotlin/MessageAttempt.kt
```diff
@@ -0,0 +1,70 @@
...
+import com.svix.kotlin.models.MessageAttemptOut
+import com.svix.kotlin.models.MessageStatus
+import com.svix.kotlin.models.ListResponseEndpointMessageOut
+import com.svix.kotlin.models.ListResponseMessageAttemptEndpointOut
+import com.svix.kotlin.models.ListResponseMessageAttemptOut
+import com.svix.kotlin.models.ListResponseMessageEndpointOut
+
+class MessageAttemptListOptions(var messageStatus: MessageStatus? = null) : ListOptions() {
+	fun messageStatus(messageStatus: MessageStatus) = apply { this.messageStatus = messageStatus }
+
```
svix-frank, line 14: Just to be sure, apply returns `this` so these can be chained? (sorry kotlin noob 😅)

kotlin/lib/src/main/kotlin/Application.kt
```diff
@@ -0,0 +1,57 @@
...
+}
+
+class Application internal constructor(debugUrl: String) {
+    val api = ApplicationApi(debugUrl)
+
+    suspend fun list(options: ApplicationListOptions): ListResponseApplicationOut {
+        try {
+            return api.listApplicationsApiV1AppGet(options.iterator, options.limit)
+        } catch (ex: Exception) {
+            throw ApiException(ex)
```
svix-frank, line 22: I think we should only need to catch internal.ApiExceptions, can these catches be narrowed down? I'd prefer the build to fail if for some reason new exceptions are added in the openapi generated code

kotlin/lib/src/main/kotlin/Application.kt
talhacohen: Yeah I couldn't find that either, I saw in https://openapi-generator.tech/docs/usage/#generate that it's supported as CLI flag, so just tried it out in the config and it works :)

kotlin/lib/src/main/kotlin/MessageAttempt.kt
talhacohen: I'm also a noob :) yes, apply returns `this`

kotlin/lib/src/main/kotlin/Application.kt
talhacohen: We have 3 different errors, `ServerException, UnsupportedOperationException, ClientException`, I can explicitly catch them if you prefer.

lmk what you think

tasn: Do they all inherit from the same basic one? Or are they all completely unrelated? If inherit from the same one, just use that, otherwise please catch all three.

talhacohen: They are all RuntimeException, though I don't think it's specific enough

As for catching all of them, I think that kotlin doesn't support multi-catch in the same line, there are 2 options (see https://discuss.kotlinlang.org/t/does-kotlin-have-multi-catch/486/20), I don't like either but leaning towards:
`
try {
    // do some work
} catch (ex: Exception) {
    when(ex) {
        is IllegalAccessException, is IndexOutOfBoundsException -> {
            // handle those above
        }
        else -> throw ex
    }
}
`

tasn: Bahh, yeah, the when trick looks fine.

talhacohen: done. 

We can probably create a util class to do that instead

talhacohen: Wrapped it

PR #139, opened 2021-09-18
Add override for builder methods

talhacohen:


java/lib/src/main/java/com/svix/ApplicationListOptions.java
```diff
@@ -1,4 +1,13 @@
 package com.svix;
 
 public class ApplicationListOptions extends ListOptions {
+    @Override
```
tasn, line 4: oh god, no way Java is so crazy. 😱

tasn (commented):
Let's see what @svix-frank says, but looks good to me.

svix-frank (approved):

PR #166, opened 2021-11-06
Add recover functions

svix-frank:


tasn (dismissed):

svix-frank:
@tasn made a few fixes to the ruby library can you take another look?

tasn (approved):
Sorry I missed that you've updated it.

PR #175, opened 2021-12-03
go: add function to verify ignoring timestamp enforcement

svix-frank:
Will enable a `--no-enforce-timestamp` flag for the cli's `verify` command.

If there is a compelling use case for this outside of the command line tool we can consider porting this to the the other libraries.

### DIFF
--- a/go/webhook.go
+++ b/go/webhook.go
@@ -74,10 +74,15 @@ func (wh *Webhook) verify(payload []byte, headers http.Header, enforceTolerance 
 		}
 	}
 
-	// parse timestamp and enforce tolerance
-	timestamp, err := verifyTimestamp(msgTimestamp, enforceTolerance)
+	timestamp, err := parseTimestampHeader(msgTimestamp)
 	if err != nil {
 		return err
+	}
+
+	if enforceTolerance {
+		if err := verifyTimestamp(timestamp); err != nil {
+			return err
+		}
 	}
 
 	computedSignature, err := wh.Sign(msgId, timestamp, payload)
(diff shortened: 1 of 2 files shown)


tasn:
I think that for the CLI we don't want to have a flag at all. We want to always ignore the timestamp, but *also* check whether the timestamp is valid. So if the timestamp is valid return what we do now, and if not (but the rest is), print a nice message saying that and close the CLI with a non-zero error code.

go/webhook.go
```diff
@@ -39,7 +39,28 @@ func NewWebhook(secret string) (*Webhook, error) {
...
+
+// VerifyIgnoringTolerance validates the payload against the svix signature headers
+// using the webhooks signing secret.
+//
+// Returns an error if the body or headers are missing/unreadable
+// or if the signature doesn't match.
+//
+// WARNING: This function does not check the signature's timestamp.
+// We recommend using the `Verify` function instead.
+func (wh *Webhook) VerifyIgnoringTolerance(payload []byte, headers http.Header) error {
```
tasn, line 59:
```suggestion
func (wh *Webhook) VerifyIgnoringTimestamp(payload []byte, headers http.Header) error {
```

I think this better explains what's up.

go/webhook.go
```diff
@@ -96,19 +117,21 @@ func (wh *Webhook) Sign(msgId string, timestamp time.Time, payload []byte) (stri
 
 }
 
-func verifyTimestamp(timestampHeader string) (time.Time, error) {
+func verifyTimestamp(timestampHeader string, enforceTolerance bool) (time.Time, error) {
```
tasn, line 120: Isn't this whole function used to verify the tolerance? I think it shouldn't be a parameter to this, just an if statement whether to call this one or not, no?

svix-frank: it also was parsing the timestamp to a time.Time but pulled this out to a separate function.

tasn (approved):

PR #177, opened 2021-12-06
ruby: Use proper case when requiring 'base64' library

mikelarkin:
Using the latest version of the gem in Ruby 2.7 raises `! Unable to load application: LoadError: cannot load such file -- Base64`

This is because the library needs to be required as 'base64' (no caps)

### DIFF
--- a/ruby/lib/svix.rb
+++ b/ruby/lib/svix.rb
@@ -2,7 +2,7 @@
 
 require "json"
 require "openssl"
-require "Base64"
+require "base64"
 
 # API
 require "svix/api/application_api"


svix-frank:
Thanks for the PR @mikelarkin! Looks like the issue stems from our test suite being run on a case-insensitive file system, sorry about that!  I'll get this in for 0.38.0

svix-frank (approved):

svix-frank:
@mikelarkin this has been released in version 0.38.0 please let me know if you run into more issues.  Thanks again for the fix!

mikelarkin:
Wow that was fast, thank you!

PR #192, opened 2021-12-20
Add a way to set a custom server url for all of the libs.

tasn:
This is currently used for testing, but will be also used for
multi-region support.

### DIFF
--- a/kotlin/lib/src/main/kotlin/SvixOptions.kt
+++ b/kotlin/lib/src/main/kotlin/SvixOptions.kt
@@ -1,6 +1,6 @@
 package com.svix.kotlin
 
-data class SvixOptions(val debugUrl: String = DEFAULT_URL) {
+data class SvixOptions(val serverUrl: String = DEFAULT_URL) {
     private val version = "0.40.0"
 
     companion object {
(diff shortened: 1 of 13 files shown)


kotlin/lib/src/main/kotlin/SvixOptions.kt
```diff
@@ -1,6 +1,6 @@
 package com.svix.kotlin
 
-data class SvixOptions(val debugUrl: String = DEFAULT_URL) {
+data class SvixOptions(val serverUrl: String = DEFAULT_URL) {
```
svix-frank, line 3: you need to use the new serverUrl in all the clients now

svix-frank (changes requested):
overall looks good, but koltlin is still reffering to `debugUrl` in all of the sub clients.

kotlin/lib/src/main/kotlin/SvixOptions.kt
tasn: Oops, fixed.

tasn:
Feel free to merge and release a new version if really fixed.

svix-frank (approved):

PR #194, opened 2021-12-27
Kotlin close tries to avoid memory leak

svix-frank:
Though I'm not sure where the message in https://github.com/svix/svix-libs/issues/191 was seen, this looks to  be the root cause of the leak.


tasn (approved):

tasn:
Maybe also open a patch against upstream later this week?

PR #196, opened 2021-12-28
JS: Enable keepalives if using node-fetch

svix-frank:
Uses a fork of isomporphic-fetch which enables keepalives if using node-fetch.  Should significantly improve performance for subsequent requests in node, with no performance change when running in a browser context.

For changes see: https://github.com/svix-frank/isomorphic-fetch/commit/fa230b6d27d2d6de1861d4edef58ca116eff38bf

Cursory review greatly improves performance:

Before:
```
calling create message...
Call took 404.8625639677048 milliseconds
calling create message...
Call took 300.6274639368057 milliseconds
```

After:
```
calling create message...
Call took 412.2045749425888 milliseconds
calling create message...
Call took 119.86126601696014 milliseconds
```

### DIFF
--- a/javascript/package.json
+++ b/javascript/package.json
@@ -32,7 +32,7 @@
     "@stablelib/utf8": "^1.0.0",
     "es6-promise": "^4.2.4",
     "fast-sha256": "^1.3.0",
-    "isomorphic-fetch": "^3.0.0",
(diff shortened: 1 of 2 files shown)


svix-liam (approved):
🏃 💨

tasn (approved):

PR #199, opened 2022-01-11
Bump openapi spec

svix-liam:


tasn:
Is this post-recent-rename? Because that's maybe why it's breaking, need to fix the recover endpoints.

svix-liam (commented):
LGTM, ( can't approve since I opened it :P )

tasn (approved):

PR #201, opened 2022-01-14
Add Environment Import & Export API

svix-frank:


svix-liam (approved):
LGTM

svix-frank:
> LGTM

What do you think about the naming for python and java?

svix-liam:
I don't think there's a better option. Only other question: do we want to rename `organization` in libs to `environment` to be consistent with the UI?

tasn (changes requested):
Need to be renamed to environment (per the API).

PR #207, opened 2022-01-21
Python: Allow explicit `None` to be passed for optional fields

svix-frank:


tasn (approved):
Goes without saying: don't squash it. :)

PR #218, opened 2022-01-27
Add integration

svix-nolan:
Update schema and add integrations apis to each lib


java/lib/src/main/java/com/svix/Application.java
```diff
@@ -23,7 +23,7 @@ public ListResponseApplicationOut list(final ApplicationListOptions options) thr
 
 	public ApplicationOut create(final ApplicationIn applicationIn) throws ApiException {
 		try {
-			return api.createApplicationApiV1AppPost(applicationIn);
+			return api.createApplicationApiV1AppPost(applicationIn, false);
```
svix-frank, line 26:
```suggestion
			return api.createApplicationApiV1AppPost(applicationIn, null);
```

kotlin/lib/src/main/kotlin/Application.kt
```diff
@@ -24,7 +24,7 @@ class Application internal constructor(token: String, options: SvixOptions) {
 
     suspend fun create(applicationIn: ApplicationIn): ApplicationOut {
         try {
-            return api.createApplicationApiV1AppPost(applicationIn)
+            return api.createApplicationApiV1AppPost(applicationIn, false)
```
svix-frank, line 27:
```suggestion
            return api.createApplicationApiV1AppPost(applicationIn, null)
```

svix-frank (changes requested):

java/lib/src/main/java/com/svix/Application.java
svix-nolan: Fixed

kotlin/lib/src/main/kotlin/Application.kt
svix-nolan: Fixed

svix-frank (approved):

PR #223, opened 2022-02-01
Python: Add get-or-create for application

svix-stephanie:


python/svix/api.py
```diff
@@ -138,6 +138,12 @@ def get(self, app_id: str) -> ApplicationOut:
         with self._api() as api:
             return api.get_application_api_v1_app_app_id_get(app_id=app_id, _check_return_type=False)
 
+    def get_or_create(self, application_in: ApplicationIn, get_if_exists: bool) -> ApplicationOut:
+        with self._api() as api:
+            return api.create_application_api_v1_app_post(
+                application_in=application_in, get_if_exists=get_if_exists, _check_return_type=False
+            )
```
svix-frank, line 145:
```suggestion
    def get_or_create(self, application_in: ApplicationIn) -> ApplicationOut:
        with self._api() as api:
            return api.create_application_api_v1_app_post(
                application_in=application_in, get_if_exists=True, _check_return_type=False
            )
```

svix-frank (changes requested):
Since it's its own function we don't need to pass this as an optional, it should just be set to True internally.

tasn (approved):

svix-frank (approved):

PR #228, opened 2022-02-01
Python: Add filter by channel for message list APIs

svix-stephanie:
Listing of messages and listing of attempts per endpoint can 
have channel passed in through the options parameter, in the same way
these can already be filtered by limit or event_type


tasn:
It's not actually used? Also, can you please update the description?

svix-frank (approved):

PR #230, opened 2022-02-02
Python: Add support for idempotency_key

svix-stephanie:
Functionality for client to be able to pass their own idempotency_key
to POST requests through svix-libs in Python, support for
other languages to come

### DIFF
--- a/python/svix/api.py
+++ b/python/svix/api.py
@@ -125,15 +125,15 @@ class ApiBase(t.Generic[ApiClass]):
 class Authentication(ApiBase[AuthenticationApi]):
     _ApiClass = AuthenticationApi
 
-    def dashboard_access(self, app_id: str) -> DashboardAccessOut:
+    def dashboard_access(self, app_id: str, idempotency_key: str = "") -> DashboardAccessOut:
         with self._api() as api:
             return api.get_dashboard_access_api_v1_auth_dashboard_access_app_id_post(
-                app_id=app_id, _check_return_type=False
+                app_id=app_id, _check_return_type=False, idempotency_key=idempotency_key
             )
 
-    def logout(self) -> None:
+    def logout(self, idempotency_key: str = "") -> None:
         with self._api() as api:
-            return api.logout_api_v1_auth_logout_post(_check_return_type=False)
+            return api.logout_api_v1_auth_logout_post(_check_return_type=False, idempotency_key=idempotency_key)
 
 
 class Application(ApiBase[ApplicationApi]):
@@ -143,18 +143,23 @@ class Application(ApiBase[ApplicationApi]):
         with self._api() as api:
             return api.list_applications_api_v1_app_get(**options.to_dict(), _check_return_type=False)
 
-    def create(self, application_in: ApplicationIn) -> ApplicationOut:
+    def create(self, application_in: ApplicationIn, idempotency_key: str = "") -> ApplicationOut:
         with self._api() as api:
-            return api.create_application_api_v1_app_post(application_in=application_in, _check_return_type=False)
+            return api.create_application_api_v1_app_post(
+                application_in=application_in, _check_return_type=False, idempotency_key=idempotency_key
+            )
 
     def get(self, app_id: str) -> ApplicationOut:
         with self._api() as api:
             return api.get_application_api_v1_app_app_id_get(app_id=app_id, _check_return_type=False)
 
-    def get_or_create(self, application_in: ApplicationIn) -> ApplicationOut:
+    def get_or_create(self, application_in: ApplicationIn, idempotency_key: str = "") -> ApplicationOut:
         with self._api() as api:
             return api.create_application_api_v1_app_post(
-                application_in=application_in, get_if_exists=True, _check_return_type=False
+                application_in=application_in,
+                get_if_exists=True,
+                _check_return_type=False,
+                idempotency_key=idempotency_key,
             )
 
     def update(self, app_id: str, application_in: ApplicationIn) -> ApplicationOut:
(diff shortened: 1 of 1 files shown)


python/svix/api.py
```diff
@@ -125,15 +125,15 @@ def _api(self) -> t.Generator[ApiClass, None, None]:
 class Authentication(ApiBase[AuthenticationApi]):
     _ApiClass = AuthenticationApi
 
-    def dashboard_access(self, app_id: str) -> DashboardAccessOut:
+    def dashboard_access(self, app_id: str, idempotency_key: str = "") -> DashboardAccessOut:
```
svix-frank, line 128:
```suggestion
    def dashboard_access(self, app_id: str, idempotency_key: t.Optional[str] = None) -> DashboardAccessOut:
```
Would setting default to None work here? If not that's fine but I don't love explicitly setting to "", not sure what issues this could cause down the line, but if you can't set it to `None` by default can you check if the libs now send an empty `Idempotency-Key` header with your change?

python/svix/api.py
```diff
@@ -143,18 +143,23 @@ def list(self, options: ApplicationListOptions = ApplicationListOptions()) -> Li
         with self._api() as api:
             return api.list_applications_api_v1_app_get(**options.to_dict(), _check_return_type=False)
 
-    def create(self, application_in: ApplicationIn) -> ApplicationOut:
+    def create(self, application_in: ApplicationIn, idempotency_key: str = "") -> ApplicationOut:
```
tasn, line 146: @svix-frank, should it be a key or a "fetch options" like structure? I think maybe in Python a key makes more sense (option stucts aren't really that common).

python/svix/api.py
```diff
@@ -125,15 +125,15 @@ def _api(self) -> t.Generator[ApiClass, None, None]:
...
-    def dashboard_access(self, app_id: str) -> DashboardAccessOut:
+    def dashboard_access(self, app_id: str, idempotency_key: str = "") -> DashboardAccessOut:
         with self._api() as api:
             return api.get_dashboard_access_api_v1_auth_dashboard_access_app_id_post(
-                app_id=app_id, _check_return_type=False
+                app_id=app_id, _check_return_type=False, idempotency_key=idempotency_key
             )
 
-    def logout(self) -> None:
+    def logout(self, idempotency_key: str = "") -> None:
```
tasn, line 134:
```suggestion
    def logout(self, *, idempotency_key: str = "") -> None:
```
For all of them.

What it essentially does: forces people to only set it by key, and never as an argument.

So:
```python
# This works (and shouldn't):
obj.logout("some_idempotency_key")
# With my change it'll have to be called like this:
obj.logout(idempotency_key="some_idempotency_key")
```

python/svix/api.py
svix-frank: I think that's reasonable, its how we'll have to do it in other languages so might as well be consistent here.

tasn: I was actually saying the opposite, that maybe option structures are not that common in Python. Though I'm happy to do an options structure for consistency.

python/svix/api.py
```diff
@@ -125,15 +133,15 @@ def _api(self) -> t.Generator[ApiClass, None, None]:
 class Authentication(ApiBase[AuthenticationApi]):
     _ApiClass = AuthenticationApi
 
-    def dashboard_access(self, app_id: str) -> DashboardAccessOut:
+    def dashboard_access(self, app_id: str, options: CreateOptions = CreateOptions()) -> DashboardAccessOut:
         with self._api() as api:
             return api.get_dashboard_access_api_v1_auth_dashboard_access_app_id_post(
-                app_id=app_id, _check_return_type=False
+                app_id=app_id, _check_return_type=False, **options.to_dict()
```
svix-frank, line 139: any reason we aren't doing this instead of making a special `to_dict` function?
```suggestion
                app_id=app_id, _check_return_type=False, **asdict(options)
```

tasn: I think it's to skip `None` values. I also had the same question in my head. :P

svix-stephanie: Right, asdict does cause an issue when no options are passed

svix-frank: Nones are fine since we pass _check_return_type=False no?

svix-stephanie: svix.application.create(ApplicationIn(name="Application name"))


ApiTypeError: Invalid type for variable 'idempotency_key'. Required value type is str and passed type was NoneType at ['idempotency_key']

tasn: @svix-frank, you created the original code I think (she just copied what's done elsewhere).

Anyhow, what she did is correct, `None` are not a valid value for these.

python/svix/api.py
```diff
@@ -62,6 +62,14 @@ def to_dict(self) -> t.Dict[str, t.Any]:
         return {k: v for k, v in asdict(self).items() if v is not None}
 
 
+@dataclass
+class CreateOptions:
```
svix-frank, line 66: Overall LGTM though my only concern is that we are using the same `CreateOptions` class in every method,  We used to have this with the `ListOptions` in the past but when we wanted to add an additional option to only one type of API this became an issue, more so it's a bit confusing to have a function like `logout` take `CreateOptions` as a parameter.

I know @tasn will disagree with me on having a `<FunctionName>Options` class per function that inherits from some BaseOptions class, but IMO that would be the best path forward long term here.  At the very least though we should name this something more generic like `PostOptions`.

@tasn @svix-stephanie thoughts?

tasn: What I'd do tbh:

`PostOptions`, we can always later have it support: `t.Union[PostOptions, LogoutPostOptions]`, so it's not a biggie.

svix-stephanie: Agreed CreateOptions could be confusing, I've updated it to PostOptions.

svix-frank (approved):

PR #239, opened 2022-02-10
Fix input values for Java & Kotlin

fcjr:
#231, broke the inputs for java & kotlin, unfortunately I didn't realize this until running tests while prepping the latest release, this fixes the expected inputs for idempotency key & channel filtering.


tasn (approved):
Approving in case they are needed, but it seems like that there are unrelated changes in Ruby?

tasn (approved):

PR #248, opened 2022-02-15
Add Svix Api client for csharp 

IAmArchaic:
## Motivation

Add Svix Api client support for csharp developers.

## Solution

Implemented OpenApi generator for csharp to add generated client to pre-existing csharp Project/Solution. Added a wrapper around generated client to create a facade for an Api consumers.


regen_openapi.sh
```diff
@@ -17,3 +17,5 @@ yarn openapi-generator-cli generate -i openapi.json -g java -o java/lib/generate
 yarn openapi-generator-cli generate -i openapi.json -g kotlin -o kotlin/lib/generated/openapi -c kotlin/openapi-generator-config.json -t kotlin/templates
 
 yarn openapi-generator-cli generate -i openapi.json -g ruby -o ruby -c ruby/openapi-generator-config.json -t ruby/templates
+
```
svix-frank, line 20: I like moving each one of these to its own yarn command, honestly can just remove this shell script in favor of your new approach to avoid duplication

package.json
```diff
@@ -1,7 +1,15 @@
...
+    "gen:python": "openapi-generator-cli generate -i openapi.json -g python -o python/ -c python/openapi-generator-config.json -t python/templates",
+    "gen:javascript": "openapi-generator-cli generate -i openapi.json -g typescript -o javascript/src/openapi -c javascript/openapi-generator-config.json",
+    "gen:go": "openapi-generator-cli generate -i openapi.json -g go -o go/internal/openapi -c go/openapi-generator-config.json -t go/templates",
+    "gen:java": "openapi-generator-cli generate -i openapi.json -g java -o java/lib/generated/openapi -c java/openapi-generator-config.json -t java/templates",
+    "gen:kotlin": "openapi-generator-cli generate -i openapi.json -g kotlin -o kotlin/lib/generated/openapi -c kotlin/openapi-generator-config.json -t kotlin/templates",
+    "gen:ruby": "openapi-generator-cli generate -i openapi.json -g ruby -o ruby -c ruby/openapi-generator-config.json -t ruby/templates",
+    "gen:csharp": "openapi-generator-cli generate -i openapi.json -g csharp-netcore -o csharp/ -c csharp/openapi-generator-config.json --global-property apis,models,supportingFiles,apiTests=false,apiDocs=false,modelTests=false,modelDocs=false"
```
svix-frank, line 12: since you are ignoring these files in the `.openapi-generator-ignore` file is it still necessary to set these global properties to false here?

csharp/Svix/Models/Webhook.cs
```diff
@@ -1,116 +1,117 @@
...
-        }
-    }
+﻿using System;
+using System.Text;
+using System.Net;
+using System.Security.Cryptography;
+
+using Svix.Exceptions;
+
+namespace Svix.Models
```
svix-frank, line 8: Unfortunately moving the Webhook class to the `Models` namespace is a breaking change, I think its best we leave this in the `Svix` namespace for now.

csharp/Svix/Models/SvixClientOptions.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace Svix.Models
+{
+    public sealed class SvixClientOptions
+    {
+        public string AccessToken { get; set; }
+        
+        public string ServiceUrl { get; set; }
```
svix-frank, line 7:
```suggestion
        public string ServerUrl { get; set; }
```

svix-frank: We name this field ServerUrl in our other libraries so we should be consistent here

csharp/Svix/SvixClient.cs
```diff
@@ -0,0 +1,89 @@
...
+
+namespace Svix
+{
+    public sealed class SvixClient : ISvixClient
+    {
+        private readonly ILogger _logger;
+        
+        private readonly SvixClientOptions _options;
+
+        public SvixClient(SvixClientOptions options, ILogger<SvixClient> logger = null)
```
svix-frank, line 20: In our other languages we use `options` as our optional variables and pass the token in as its own parameter to the client, I'm wondering if we should do the same here for consistency
 ```suggestion
        public SvixClient(string accessToken, SvixClientOptions options, ILogger<SvixClient> logger = null)
```

File: csharp/Svix/SvixClient.cs [hunk 6]
```diff
@@ -0,0 +1,89 @@
...
+            catch (ApiException e)
+            {
+                _logger?.LogError(e, $"{nameof(IsHealthy)} failed");
+
+                if (_options.Throw)
+                    throw;
+                
+                return false;
+            }
+        }
```
svix-frank, line 58: The implementation itself here looks great though to keep consistent with our other libraries we should define a class called `Health` and have it hold all the functions for the health api (and do the same for all other api tags `Message`, `Application`, etc.

That way the caller would call:
```
svixClient.Health.IsHealthy()
```
```
svixClient.Message.Create(...)
```
etc.

Doing this would also allow us to construct the internal generated api classes once when we construct the SvixClient as well to avoid additional overhead when making the calls

svix-frank (commented):
@IAmArchaic, thanks so much for the contribution, so far looking great, just left a few comments/questions before you move forward with other API tags

regen_openapi.sh
IAmArchaic: Sounds great!

package.json
IAmArchaic: I'm not sure but I'll try messing around with the generator again to verify.

csharp/Svix/Models/Webhook.cs
IAmArchaic: I agree. I think moving this back to the Svix namespace makes more sense given your later comment about the Api client.

csharp/Svix/SvixClient.cs
IAmArchaic: I'll double check the other clients and to see what you mean. I've tried to create multiple constructors here to support one-off client use, and .Net Core Dependency Injection.

svix-frank:
> I think I got everything you were looking for @svix-frank. If you agree, I'll get started implementing the remaining Api resources.

💯 @IAmArchaic that was quick! Looks great!

regen_openapi.sh
```diff
@@ -1,19 +0,0 @@
-#!/bin/sh
-
-set -ex
-
-if [ -n "$1" ]; then
-    curl "$1" | python -m json.tool > openapi.json
-fi
```
tasn, line 7: @svix-frank, was this being used?

svix-frank: It shouldn't be, it has only caused confusion it leads new users to try and build from the latest openapi spec on the docs page which is no longer what should be done

svix-frank:
> @svix-frank, I've added Mock and refactored the unit tests to correctly test the Api wrappers without requiring a valid access token/serverUrl.
> 
> I also started implementing the Application Api resource.
> 
> Just wanted to open it up to any further comments or recommendations before I continue.

everything looks great so far!

tasn:
Hey @IAmArchaic, any updates? Very much looking forward to having this! :)

IAmArchaic:
Hey @tasn, I'm expecting to make progress this weekend. Shouldn't have to wait too much longer. lol

tasn:
No rush, but also yay! :)

Thanks a lot!

tasn:
Hey @IAmArchaic, again no rush, but really looking forward to having it. :)

IAmArchaic:
Hey @tasn, I believe all Api resources have now been implemented. All that should remain are Unit Tests.

tasn:
Yay! 😍 

Ready to take a look, or do are you still changing things?

IAmArchaic:
You can certainly take a look/try it out. I'm planning to do a second-pass through the API client classes just in case before moving to tests.

IAmArchaic:
> So I tried taking a look, and I don't understand C# at all. :)
> 
> I can take a look at the tests, but do you have may simple code samples so I can check out the API? Or should I just rely on the tests.

Right now there are only the tests. They exist to validate that the wrapper created around the generated client is passing the parameters correctly. I can certainly create additional code samples to help!

tasn:
I think I can probably figure it out from the tests, but yeah examples always help!

tasn:
From what I was able to tell it looks good! Can you please verify and add tests for the verification based on existing Svix signatures (or e.g. ones taken from other examples).

Also, can you please rebase on top of main?

Excited to get it in!

tasn:
\o/

Btw, please rebase when you have a moment (it looks like there aren't any real conflicts).

IAmArchaic:
Hey @tasn , only a few more test classes to complete. Planning to finish them today! I'll rebase once I'm code complete.

tasn:
Thanks a lot! Looking forward to having this in!! 🔥 🔥 🔥

IAmArchaic:
Hey @tasn and @svix-frank, I've completed all Api client wrappers and unit tests for each one. There are currently 103 unit tests that validate the Api Wrappers are calling the generated open-api client methods correctly. These changes have also been rebased on main.

I haven't had the time to add additional examples, but the unit tests to demonstrate how to use the Csharp lib in its entirety. let me know what you think.

tasn:
Looks amazing, I'll take a better look either later today or tomorrow morning! \o/

tasn:
There are some lint errors it seems.

tasn:
Want me to fix them for you?

IAmArchaic:
Hey @tasn , I'm so sorry for missing your messages. I've been busy with work and newborn classes that they slipped through.

I can look at fixing them today.

tasn:
No rush, just wanted to check on you!

IAmArchaic:
@tasn , I ran the formatter used in the linter across the project. I think we should be good now.

tasn:
Cool, thanks! Letting CI run, let's see!

tasn (approved):

tasn:
It's in! Yay, thanks a lot!

IAmArchaic:
You're most welcome! Reach out to me if I can be of further assistance.

tasn:
Will do, thanks!

tasn:
Yay, thanks a lot! :)

IAmArchaic:
You're welcome. Ping me again if you need something else.

tasn:
I'll probably ping you once there's a PR out. :)

tasn:
Creating a simple message:
```csharp
C#

using Svix;
using Svix.Model;
using Svix.Models;

var svix = new SvixClient("sk_IrlFPEh3VYctuyHhKTCxamGV", new SvixOptions("https://api.svix.com"));

// Send an event to Rock Inc's webhook endpoints
await svix.Message.CreateAsync(
    "app_Xzx8bQeOB1D1XEYmAJaRGoj0",
    new MessageIn(
      eventType: "user.created",
      payload: new {
        username = "new_user",
        email = "new_user@example.com",
      }
    ),
    new MessageCreateOptions {});
```

tasn:
I just realized something: the functions require passing Options, is it possible to make them have a default value? Can you show me one example of how to change them and I'll change them everywhere? 🙏🏻

tasn:
I also added https://github.com/svix/svix-docs/pull/82 I hope it's correct. :)

tasn:
Also added C# examples to https://api.svix.com/docs I hope those are correct too. :P

Please let me know if you have any comments.

IAmArchaic:
@tasn the interface and classes can be updated to assign a default value for the options parameter. Here is an example for the Application Create.

Application.cs
```
public ApplicationOut Create(ApplicationIn application, ApplicationCreateOptions options = null, string idempotencyKey = default)
```

IApplication.cs
```
public ApplicationOut Create(ApplicationIn application, ApplicationCreateOptions options = null, string idempotencyKey = default);
```

tasn:
Would `null` work, or would it break the code that uses options inside? That was my main thing. Should we have a default ApplicationCreateOptions created? If so, how do we do it?

tasn:
OK, I'll open a PR to add nulls there, thanks!

IAmArchaic:
If the given ```options``` is ```null``` the field off the options will not be evaluated and ```null``` will be passed into the Api method where the param will be omitted by auto-gen client.

tasn:
No trouble at all, I missed it at review!

PR #256, opened 2022-02-19
Implement GET/PUT for Endpoint Headers

jaymell:
## Motivation
Add GET and PUT routes for retrieving and adding Endpoint Headers

## Solution
Represent headers as jsonb column in database; support basic validation of header names and values upon creation using functionality in 'http' crate; mask sensitive headers on retrieval.

### DIFF
--- a/server/svix-server/migrations/20220218061228_endpoint_headers.up.sql
+++ b/server/svix-server/migrations/20220218061228_endpoint_headers.up.sql
@@ -1,0 +1,1 @@
+ALTER TABLE endpoint ADD COLUMN headers jsonb;
--- a/server/svix-server/src/core/types.rs
+++ b/server/svix-server/src/core/types.rs
@@ -1,7 +1,7 @@
 // SPDX-FileCopyrightText: © 2022 Svix Authors
 // SPDX-License-Identifier: MIT
 
-use std::collections::HashSet;
+use std::collections::{HashMap, HashSet};
 
 use chrono::{DateTime, Utc};
 use lazy_static::lazy_static;
@@ -490,6 +490,42 @@ pub struct ExpiringSigningKey {
     pub expiration: DateTime<Utc>,
 }
 
+#[derive(Clone, Debug, PartialEq, Serialize)]
+pub struct EndpointHeaders(pub HashMap<String, String>);
+json_wrapper!(EndpointHeaders);
+
+impl<'de> Deserialize<'de> for EndpointHeaders {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        HashMap::deserialize(deserializer)
+            .map(|x: HashMap<String, String>| {
+                x.into_iter().map(|(k, v)| (k.to_lowercase(), v)).collect()
+            })
+            .map(EndpointHeaders)
+    }
+}
+
+impl Validate for EndpointHeaders {
+    fn validate(&self) -> std::result::Result<(), ValidationErrors> {
+        let mut errors = ValidationErrors::new();
+        self.0.iter().for_each(|(k, v)| {
+            if let Err(_e) = http::header::HeaderName::try_from(k) {
+                errors.add(ALL_ERROR, ValidationError::new("Invalid Header Name."));
+            }
+            if let Err(_e) = http::header::HeaderValue::try_from(v) {
+                errors.add(ALL_ERROR, ValidationError::new("Invalid Header Value."));
+            }
+        });
+        if errors.is_empty() {
+            Ok(())
+        } else {
+            Err(errors)
+        }
+    }
+}
+
 #[repr(i16)]
 #[derive(Clone, Debug, Copy, PartialEq, IntoPrimitive, TryFromPrimitive)]
 pub enum MessageAttemptTriggerType {
@@ -513,7 +549,8 @@ enum_wrapper!(MessageStatus);
 mod tests {
     use crate::core::types::{EventChannel, EventTypeName};
 
-    use super::{ApplicationId, ApplicationUid};
+    use super::{ApplicationId, ApplicationUid, EndpointHeaders};
+    use std::collections::HashMap;
     use validator::Validate;
 
     #[test]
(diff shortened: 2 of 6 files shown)


server/svix-server/src/core/types.rs
```diff
@@ -490,6 +490,42 @@ pub struct ExpiringSigningKey {
...
+            })
+            .map(EndpointHeaders)
+    }
+}
+
+impl Validate for EndpointHeaders {
+    fn validate(&self) -> std::result::Result<(), ValidationErrors> {
+        let mut errors = ValidationErrors::new();
+        self.0.iter().for_each(|(k, v)| {
+            if let Err(_e) = http::header::HeaderName::try_from(k) {
```
tasn, line 514: Not saying one way or the other, but why use `HeaderName` only in the validation here, and not also in the type of the `HashMap` used in `EndpointHeaders`?

As in, why not change the above to:
```
 pub struct EndpointHeaders(pub HashMap<http::header::HeaderName, http::header::HeaderValue>);
```

Not saying one way is better than the other, I was just wondering.

server/svix-server/src/core/types.rs
```diff
@@ -490,6 +490,42 @@ pub struct ExpiringSigningKey {
...
+json_wrapper!(EndpointHeaders);
+
+impl<'de> Deserialize<'de> for EndpointHeaders {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        HashMap::deserialize(deserializer)
+            .map(|x: HashMap<String, String>| {
+                x.into_iter().map(|(k, v)| (k.to_lowercase(), v)).collect()
```
tasn, line 504: Why is moving all the keys to lowercase necessary?

server/svix-server/src/v1/endpoints/endpoint.rs
```diff
@@ -170,6 +170,60 @@ struct RecoverIn {
...
+struct EndpointHeadersIn {
+    #[validate]
+    headers: EndpointHeaders,
+}
+
+#[derive(Clone, Debug, PartialEq, Serialize)]
+#[serde(rename_all = "camelCase")]
+struct EndpointHeadersOut {
+    headers: HashMap<String, String>,
+    sensitive: Vec<String>,
```
tasn, line 184:
```suggestion
    sensitive: HashSet<String>,
```

server/svix-server/src/v1/endpoints/endpoint.rs
```diff
@@ -170,6 +170,60 @@ struct RecoverIn {
...
+#[serde(rename_all = "camelCase")]
+struct EndpointHeadersIn {
+    #[validate]
+    headers: EndpointHeaders,
+}
+
+#[derive(Clone, Debug, PartialEq, Serialize)]
+#[serde(rename_all = "camelCase")]
+struct EndpointHeadersOut {
+    headers: HashMap<String, String>,
```
tasn, line 183: Same comment as elsewhere, why not use the `http::header` types?

server/svix-server/src/v1/endpoints/endpoint.rs
```diff
@@ -170,6 +170,60 @@ struct RecoverIn {
...
+        }
+    }
+}
+
+impl From<EndpointHeaders> for EndpointHeadersOut {
+    fn from(hdr: EndpointHeaders) -> Self {
+        let (sens, remaining) = hdr.0.into_iter().partition(|(k, _)| {
+            Self::SENSITIVE_HEADERS
+                .iter()
+                .any(|&x| x == k.to_lowercase())
```
tasn, line 209: See my other comment about lowercasing, that's exactly my point. You don't need it there because you have it here.

tasn, line 209: Doesn't this lowercase k on every iteration? Isn't it better to do it once?

server/svix-server/src/v1/endpoints/endpoint.rs
```diff
@@ -483,6 +537,46 @@ async fn recover_failed_webhooks(
...
+    }: AuthenticatedApplication,
+) -> Result<Json<EndpointHeadersOut>> {
+    let endp = endpoint::Entity::secure_find_by_id_or_uid(app.id, endp_id)
+        .one(db)
+        .await?
+        .ok_or_else(|| HttpError::not_found(None, None))?;
+
+    match endp.headers {
+        Some(h) => Ok(Json(h.into())),
+        None => Ok(Json(EndpointHeadersOut::empty())),
```
tasn, line 555: Just a thought: maybe it's better to implement `Default` rather than have our own custom function (`empty()`)?

server/svix-server/src/core/types.rs
```diff
@@ -577,4 +614,46 @@ mod tests {
...
     }
+
+    #[test]
+    fn test_endpoint_headers_validation() {
+        let hdr_map = HashMap::from([
+            ("valid".to_owned(), "true".to_owned()),
+            ("also-valid".to_owned(), "true".to_owned()),
+        ]);
+        let endpoint_headers = EndpointHeaders(hdr_map);
+        assert!(endpoint_headers.validate().is_ok());
```
tasn, line 625: Better to just unwrap instead of `is_ok()` because then you'll get the actual error printed if it fails.

server/svix-server/migrations/20220218061228_endpoint_headers.up.sql
```diff
@@ -0,0 +1,3 @@
+-- Add up migration script here
```
tasn, line 1: Could you please fix the linting issue?

tasn (commented):
Thanks for the PR, it looks great!

I made a bunch of small nitpicks all around, but the PR is solid!

server/svix-server/src/core/types.rs
jaymell: This was an attempt to ensure compatibility with existing API, as in my testing it seems to always return lower-case returned results. though in theory it shouldn't matter since header names are case-insensitive per HTTP spec. 

I would suggest leaving it in here for now for the sake of the other code where header names are checked against SENSITIVE header list.

server/svix-server/src/v1/endpoints/endpoint.rs
jaymell: Fixed.

jaymell: Removed the redundant call.

server/svix-server/src/v1/endpoints/endpoint.rs
jaymell: Fixed.

server/svix-server/src/core/types.rs
jaymell: Fixed.

server/svix-server/src/core/types.rs
tasn: Cool, just wanted to make sure it's intentional.

server/svix-server/src/core/types.rs
tasn: OK, we chatted off-band, let's leave it like this for now.

tasn:
Only thing left is the SQL linter, which we'll disable.

@svix-frank, can you please disable the SQL super-linter? I'm not sure why it's complaining, and having a linter you can't run locally is terribly annoying.

server/svix-server/src/core/types.rs
jaymell: Sorry meant to post a comment here earlier -- I agree that either using the `http::header` types here are a better long-term solution. However, I've run into some scenarios using them where I can cause thread panics. I think it makes sense to leave current implementation in place until I can find a way to cleanly avoid unwanted panics.

server/svix-server/src/v1/endpoints/endpoint.rs
jaymell: Leaving this for now, see my previous comment.

svix-frank:
> @svix-frank, can you please disable the SQL super-linter?

Just disabled it.  @jaymell can you do a quick rebase on main when you have a second so it pulls in the fixed CI?

tasn (approved):

PR #259, opened 2022-02-25
Libs: Add support for reverse iteration (`prev_iterator`) & after param in list commands

tasn:
Match the libs to the spec changes.


svix-frank (approved):

tasn (commented):
Looks good to me.

PR #260, opened 2022-02-25
Python: Don't regen requirements.txt

svix-frank:
We have a new requirement that doesn't come from the generator
so we shouldn't regenerate this file anymore.

### DIFF
--- a/python/.openapi-generator-ignore
+++ b/python/.openapi-generator-ignore
@@ -35,3 +35,4 @@ docs/
 test/
 test-requirements.txt
 tox.ini
+requirements.txt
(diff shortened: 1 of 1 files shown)


python/.openapi-generator-ignore
```diff
@@ -34,4 +34,5 @@
 docs/
 test/
 test-requirements.txt
-tox.ini
\ No newline at end of file
+tox.ini
+requirements.txt
```
svix-frank, line 38:
```suggestion
requirements.txt
tox.ini
```

tasn (approved):

PR #264, opened 2022-02-28
Idempotency Key Option

svix-stephanie:
## Motivation

Previously implemented in Python, now supported 
in the rest of the languages

## Solution

- Ruby: options hash
- Kotlin: under PostOptions
- Java: under PostOptions
- Javascript: under PostOptions
- Go: under PostOptions


java/lib/src/main/java/com/svix/MessageAttempt.java
```diff
@@ -70,9 +70,9 @@ public MessageAttemptOut get(final String msgId, final String appId, final Strin
 		}
 	}
 
-	public void resend(final String msgId, final String appId, final String endpointId) throws ApiException {
+	public void resend(final String msgId, final String appId, final String endpointId, final PostOptions options) throws ApiException {
```
tasn, line 73: This breaks API. Every language that doesn't have a default will either need a new API added that's named differently, or in Java's case, the same API is fine, just need to create another one that's named the same.

svix-frank: In java this is called method overloading: https://www.programiz.com/java-programming/method-overloading and is fine, python and go don't have this so it will have to be a function with a different name, perhaps `<X>WithOptions`

tasn: Thanks, I was blanking. :)

java/lib/src/main/java/com/svix/Endpoint.java
```diff
@@ -13,97 +13,121 @@
...
+        }
+    }
+
+    public EndpointOut create(final String appId, final EndpointIn endpointIn, final PostOptions options) throws ApiException {
+        try {
+            return api.createEndpointApiV1AppAppIdEndpointPost(appId, endpointIn, options.getIdempotencyKey());
+        } catch (com.svix.internal.ApiException e) {
+            throw Utils.wrapInternalApiException(e);
+        }
+    }
```
svix-frank, line 44: Rather than duplicating all the logic in these functions it would be much cleaner to make your 

`create(final String appId, final EndpointIn endpointIn)` function call your `create(final String appId, final EndpointIn endpointIn, final PostOptions options)` function (just without an idempotency key.

you should do this for all of the overloaded functions

svix-frank, line 44: Rather than duplicating all the logic in these functions it would be much cleaner to make your

`create(final String appId, final EndpointIn endpointIn)` function call your `create(final String appId, final EndpointIn endpointIn, final PostOptions options)` function (just without an idempotency key.

you should do this for all of the overloaded functions

svix-frank (changes requested):

PR #267, opened 2022-03-01
Support custom headers for webhooks in worker threads

svix-daniel:
## Motivation
While much of the CRUD of setting customs headers per endpoint has already been done, they have not actually been applied to any of the messages thus far.

## Solution
Simply utilizes the header values already in memory and adds them to relevant webhook events.

### DIFF
--- a/server/svix-server/src/worker.rs
+++ b/server/svix-server/src/worker.rs
@@ -24,7 +24,7 @@ fn generate_msg_headers(
     cfg: &Configuration,
     body: &str,
     msg_id: &MessageId,
-    configured_headers: Option<EndpointHeaders>,
+    configured_headers: Option<&EndpointHeaders>,
     endpoint_signing_keys: Vec<&EndpointSecret>,
     _endpoint_url: &str,
 ) -> HeaderMap {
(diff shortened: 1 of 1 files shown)


server/svix-server/src/worker.rs
```diff
@@ -17,15 +17,17 @@ use std::iter;
 
 const USER_AGENT: &str = concat!("Svix-Webhooks/", env!("CARGO_PKG_VERSION"));
 
-fn generate_msg_headers(
+/// Generates a set of headers for any one webhook event
+async fn generate_msg_headers(
```
tasn, line 21: As said below, shouldn't be async.

server/svix-server/src/worker.rs
```diff
@@ -47,6 +49,27 @@ fn generate_msg_headers(
...
     }
 
+    let configured_headers =
+    	// FIXME: Don't clone if possible
+        endpoint::Entity::secure_find_by_id(msg_task.app_id.clone(), msg_task.endpoint_id.clone())
+            .one(db)
+            .await
+            .unwrap_or_default()
+            .map(|model| model.headers)
+            .flatten();
```
tasn, line 59: You already have the endpoint outside of this function, so no need to fetch it again, just pass it. :)

server/svix-server/src/worker.rs
```diff
@@ -17,15 +17,17 @@ use std::iter;
 
 const USER_AGENT: &str = concat!("Svix-Webhooks/", env!("CARGO_PKG_VERSION"));
 
-fn generate_msg_headers(
+/// Generates a set of headers for any one webhook event
+async fn generate_msg_headers(
     cfg: &Configuration,
     body: &str,
-    msg_id: &MessageId,
+    db: &DatabaseConnection,
```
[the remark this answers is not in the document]
svix-frank: why even pass the headers to the `generate` function? Seems simpler to just add them after the generate and before we append the `User Agent`.  Related, if you take this approach, this function should probably be named `generate_signature_headers`

tasn (approved):

PR #271, opened 2022-03-01
Bump version & update changelog

svix-frank:


tasn:
Wait for the idempotency and others since we are already doing it.

tasn (approved):
Approved, but we can probably wait for a few more things.

PR #278, opened 2022-03-04
server: PATCH Endpoint Headers

jaymell:
## Motivation

Implement PATCH method for updating endpoint headers; update axum extension setup to remove deprecated APIs

## Solution

Merge new/modified headers with existing

### DIFF
--- a/server/svix-server/src/v1/endpoints/endpoint.rs
+++ b/server/svix-server/src/v1/endpoints/endpoint.rs
@@ -177,6 +177,14 @@ struct EndpointHeadersIn {
     headers: EndpointHeaders,
 }
 
+impl ModelIn for EndpointHeadersIn {
+    type ActiveModel = endpoint::ActiveModel;
+
+    fn update_model(self, model: &mut Self::ActiveModel) {
+        model.headers = Set(Some(self.headers));
+    }
+}
+
 #[derive(Clone, Debug, PartialEq, Serialize, Default)]
 #[serde(rename_all = "camelCase")]
 struct EndpointHeadersOut {
@@ -208,11 +216,23 @@ impl From<EndpointHeaders> for EndpointHeadersOut {
     }
 }
 
-impl ModelIn for EndpointHeadersIn {
+#[derive(Clone, Debug, PartialEq, Validate, Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct EndpointHeadersPatchIn {
+    #[validate]
+    headers: EndpointHeaders,
+}
+
+impl ModelIn for EndpointHeadersPatchIn {
     type ActiveModel = endpoint::ActiveModel;
 
     fn update_model(self, model: &mut Self::ActiveModel) {
-        model.headers = Set(Some(self.headers));
+        model.headers = if let Some(Some(mut hdrs)) = model.headers.take() {
+            hdrs.0.extend(self.headers.0);
+            Set(Some(hdrs))
+        } else {
+            Set(Some(self.headers))
+        };
     }
 }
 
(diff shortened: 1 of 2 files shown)


server/svix-server/src/main.rs
```diff
@@ -102,12 +102,12 @@ async fn main() {
     let mut app = Router::new()
         .nest("/api/v1", v1::router())
         .layer(TraceLayer::new_for_http().on_request(()))
-        .layer(AddExtensionLayer::new(pool.clone()))
-        .layer(AddExtensionLayer::new(queue_tx.clone()))
-        .layer(AddExtensionLayer::new(cfg.clone()));
+        .layer(Extension(pool.clone()))
+        .layer(Extension(queue_tx.clone()))
+        .layer(Extension(cfg.clone()));
```
tasn, line 107: Are you sure this is correct? (really deprecated)
If so, can you please:
1. Split it to its own PR.
2. Reference in the commit where you got this information from (e.g. a link to the docs).

server/svix-server/src/v1/endpoints/endpoint.rs
```diff
@@ -208,11 +216,25 @@ impl From<EndpointHeaders> for EndpointHeadersOut {
...
+    #[validate]
+    headers: EndpointHeaders,
+}
+
+impl ModelIn for EndpointHeadersPatchIn {
     type ActiveModel = endpoint::ActiveModel;
 
     fn update_model(self, model: &mut Self::ActiveModel) {
-        model.headers = Set(Some(self.headers));
+        let existing = mem::replace(&mut model.headers, Set(None)).unwrap();
```
tasn, line 230: Not gonna lie, I don't remember what `mem::replace` does, but it feels very wrong to use `mem`. What are you trying to do here?

server/svix-server/src/main.rs
tasn: I did something with Axum now (testing a fix for a bug I reported) and this deprecation seems to only be done in the to-be-released version, and not in the currently released one. No?

jaymell: No problem, I will remove from PR for now. `AddExtensionLayer` is deprecated since 0.4.7, which was released a few days ago: https://github.com/tokio-rs/axum/releases/tag/axum-v0.4.7

tasn: Yeah, you're right, I dunno why it's not in the axum changelog. :|

https://github.com/tokio-rs/axum/blob/main/axum/CHANGELOG.md

tasn: Please open a PR for it, happy to insta-merge it, we don't want the warnings. :)

jaymell: In any event, I should have updated Cargo.toml as well to explicitly reference the latest version :|

tasn: Yeah, probably. :P 

Cargo "best practices" is not to commit the lock file for binaries, only libs (dunno whether we'll continue following this, but this is the recommendation I've seen). So I didn't get the update that showed it for you as deprecated. :) 

But yeah, just bump to 0.4.8 (I think that's the latest) and open a PR with this small change. Looking forward to having it in!

server/svix-server/src/v1/endpoints/endpoint.rs
jaymell: I agree it seems strange, but `mem::replace` is the lowest-overhead way I know of to modify a non-`Copy` field from a mutably borrowed struct: https://doc.rust-lang.org/error-index.html#E0507. 

The simplest alternative I know is to clone the field but was trying to avoid it.

tasn: Yeah, definitely shouldn't copy the structure, but why wouldn't something like this work?
```rust
        model.headers = match model.headers {
            Some(&mut existing) => {
                existing.0.extend(self.headers.0);
                Set(Some(existing))
            }
            None => Set(Some(self.headers)),
        };
```
As in without that line? Why do you even need to touch model.headers before?

jaymell: If I use something like above:
```rust
        model.headers = match model.headers {
            Some(existing) => {
                existing.0.extend(self.headers.0);
                Set(Some(existing))
            }
            None => Set(Some(self.headers)),
        };
```

I get the move error. Let me go back to the drawing board for a bit and see if there is a cleaner way.

tasn: What are you unwrapping? Also, I think this is actually something that also changed in axum (though maybe in the unreleased version).

jaymell: `model.headers` is a `sea_orm::ActiveValue` -- unwrapping if I want direct access to the underlying EndpointHeaders struct. I'm trying to see if I can do the pattern matching on the ActiveValue itself rather than the unwrapped value.

tasn: Yeah, just do pattern matching there too, or maybe you can use `take()`.
```rust
        model.headers = match model.headers.take() {
            Some(existing) => {
                existing.0.extend(self.headers.0);
                Set(Some(existing))
            }
            None => Set(Some(self.headers)),
        };
```

server/svix-server/src/main.rs
tasn: I can't unsee it now because of you. Let me know if you prefer me to open a PR and I will. I want it fixed, haha. :)

jaymell: Raised #283

tasn: Thanks!

server/svix-server/src/v1/endpoints/endpoint.rs
jaymell: Done.

tasn (approved):

PR #282, opened 2022-03-04
Application endpoint cache

svix-daniel:
Implements a generic Redis cache which as of now is used to store the endpoint information for a given application on message creation. This is read on repeated message  creations for the same application and when fetching endpoint information in the worker.

### DIFF
--- a/server/svix-server/src/core/cache/create_message_app.rs
+++ b/server/svix-server/src/core/cache/create_message_app.rs
@@ -1,0 +1,100 @@
+use std::convert::TryInto;
+
+use chrono::{DateTime, FixedOffset};
+use sea_orm::{ConnectionTrait, DatabaseConnection};
+use serde::{Deserialize, Serialize};
+
+use super::{kv_def, CacheKey, CacheValue};
+use crate::{
+    core::types::{
+        ApplicationId, ApplicationUid, EndpointHeaders, EndpointSecret, EventChannelSet,
+        EventTypeNameSet, ExpiringSigningKeys, OrganizationId,
+    },
+    db::models::{application, endpoint},
+    error::{Error, Result},
+};
+
+/// The information cached during the creation of a message. Includes a [`Vec`] of all endpoints
+/// associated with the given application and organization ID.
+#[derive(Deserialize, Serialize, Debug, Clone)]
+pub struct CreateMessageApp {
+    id: ApplicationId,
+    uid: Option<ApplicationUid>,
+    org_id: OrganizationId,
+    // TODO: org_group_id
+    rate_limit: Option<u16>,
+    endpoints: Vec<CreateMessageEndpoint>,
+    deleted: bool,
+}
+
+/// The information for each individual endpoint cached with the creation of a message.
+#[derive(Deserialize, Serialize, Debug, Clone)]
+pub struct CreateMessageEndpoint {
+    url: String,
+    key: EndpointSecret,
+    old_signing_keys: Option<ExpiringSigningKeys>,
+    event_type_ids: Option<EventTypeNameSet>,
+    channels: Option<EventChannelSet>,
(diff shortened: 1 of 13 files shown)


server/svix-server/src/v1/endpoints/message.rs
```diff
@@ -154,12 +158,35 @@ async fn list_messages(
...
+            let cma = CreateMessageApp::fetch(db, app.clone()).await?;
+            if let Err(e) = redis_cache.set(&cache_key, &cma, Some(30)).await {
+                tracing::error!("Redis cache error on set: {}", e);
+            }
+            cma
+        }
+    } else {
+        // If Redis is disabled, there will be no RedisCache so fallback to postgres
+        CreateMessageApp::fetch(db, app.clone()).await?
+    };
```
tasn, line 188: This looks like it should just be a function that does all of this, and is shared between the worker and the message.

I realize that in `create_message` we already have the `app` fetched, so we can just pass an optional `app` parameter to not fetch the app if it exists, but other than that, they should both be identical.

tasn (commented):
Looks great overall, I just made a few comments that I think would be useful.

tasn:
I just merged #283 which means you'll need to rebase over main and fix warnings. :)

server/svix-server/src/core/cache/create_message_app.rs
```diff
@@ -0,0 +1,109 @@
...
+            url: m.url,
+            key: m.key,
+            old_signing_keys: m.old_keys,
+            event_types_ids: m.event_types_ids,
+            channels: m.channels,
+            rate_limit: m
+                .rate_limit
+                .map(|v| v.try_into())
+                .transpose()
+                .map_err(|_| Error::Validation("Endpoint rate limit out of bounds".to_owned()))?,
```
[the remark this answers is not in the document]
svix-daniel: So I just saw rate limits being defined as `u16`s in code minus the PostgreSQL models which use `i32`s, although I think the conversion only went one way. I'm fine changing it to an `i32`and  removing the `map`entirely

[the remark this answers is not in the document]
tasn: Ah, it's because PG doesn't support unsigned, and seaorm for whatever reason, doesn't just automatically use the correct type for it (i32).

server/svix-server/src/main.rs
```diff
@@ -98,13 +102,18 @@ async fn main() {
         }
     };
 
+    let redis_cache = redis_pool
+        .as_ref()
+        .map(|pool| RedisCache::new(pool.clone()));
```
tasn, line 107: It's fine for this PR, but we probably need to improve it in a subsequent PR:
We should make the cache object always exist, and just have different backends like Queue.

The backends can be:
1. Redis
2. Memory
3. No-op (though tbh, probably memory with size 0 is a no-op).

This also makes the code using it much cleaner because it means the cache object always exists, it just happens to always cache miss with it's a no-op, or etc.

server/svix-server/src/core/message_app.rs
```diff
@@ -0,0 +1,157 @@
...
+        } else {
+            return Ok(None);
+        };
+
+        // Fetch the actual [`CreateMessageApp`]
+        let out = Self::fetch_from_pg_by_model(&db, app).await?;
+
+        // Insert it into Redis
+        if let Some(redis) = redis {
+            let _ = redis.set(&cache_key, &out, Duration::from_secs(30)).await;
```
tasn, line 102: Can you move this duration into a constant, or even just a variable? We'll want to make it configurable in the near future, and anyway we want to have it be set somewhere explicit.

tasn (commented):
Made a couple of small comments, but once you fix the clippy issues it can go in.

tasn (approved):

PR #286, opened 2022-03-08
Launch dependencies before server unit tests

svix-daniel:
Launches PostgreSQL and Redis via the  `docker-compose.yml` for the server before running unit tests.

### DIFF
--- a/.github/workflows/server-ci.yml
+++ b/.github/workflows/server-ci.yml
@@ -54,7 +54,7 @@ jobs:
     - uses: Swatinem/rust-cache@v1
 
     - name: Start dependencies
-      run: docker-compose -f "server/docker-compose.yml" up --no-build -d
+      run: docker-compose -f "server/testing-docker-compose.yml" up -d
 
     - name: Run tests
       uses: actions-rs/cargo@v1
--- a/.github/workflows/server-ci.yml
+++ b/.github/workflows/server-ci.yml
@@ -25,11 +25,13 @@ jobs:
         profile: minimal
         components: clippy, rustfmt
     - uses: Swatinem/rust-cache@v1
+
     - name: Check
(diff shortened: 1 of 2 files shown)


.github/workflows/server-ci.yml
```diff
@@ -50,12 +52,24 @@ jobs:
         override: true
         profile: minimal
     - uses: Swatinem/rust-cache@v1
+
+    - name: Start dependencies
+      run: docker-compose -f "server/docker-compose.yml" up -d --build
+
```
svix-frank, line 58: Does the github actions linux have docker & docker-compose by default? I don't see an install step

tasn: I think you need to maybe say "uses docker", though given that the tests pass, I'm inclined to say the answer is yes? :)

server/testing-docker-compose.yml
```diff
@@ -0,0 +1,31 @@
...
+    healthcheck:
+      test: "pg_isready -h localhost"
+      interval: 30s
+      timeout: 10s
+      retries: 3
+    environment:
+      DB_HOST: "postgres"
+      DB_USER: "postgres"
+      DB_PASSWORD: "postgres"
+      MAX_CLIENT_CONN: 500
```
tasn, line 21:
```suggestion
```
No need for pgbouncer in the tests

.github/workflows/server-ci.yml
svix-daniel: `docker-compose` is installed in the actions image as shown here: https://github.com/actions/virtual-environments/blob/c8b63063441fb63048458f7907292a814c25189f/images/linux/ubuntu2004.json#L181

tasn (approved):

PR #287, opened 2022-03-08
Update OpenAPI spec to add msg_id to attempt output.

tasn:


svix-frank (approved):
LGTM but can't release to prod till we release

tasn:
Yup, thanks!

PR #288, opened 2022-03-08
Draft: Enable clippy::pedantic lints and correct them

svix-daniel:
Simply enables most clippy pedantic lints (with some exclusions for false positives or necessary deviations).

### DIFF
--- a/server/svix-server/src/core/message_app.rs
+++ b/server/svix-server/src/core/message_app.rs
@@ -1,3 +1,5 @@
+// Lots of false positives in this file
+
 use std::{
     convert::{TryFrom, TryInto},
     time::Duration,
@@ -32,7 +34,7 @@ pub struct CreateMessageApp {
 }
 
 impl CreateMessageApp {
-    /// Fetch all requisite information for creating a [`CreateMessageApp`] from the PostgreSQL
+    /// Fetch all requisite information for creating a [`CreateMessageApp`] from the postgres
     /// database
     async fn fetch_from_pg_by_model(
         db: &DatabaseTransaction,
@@ -51,7 +53,7 @@ impl CreateMessageApp {
             org_id: app.org_id,
             rate_limit: app
                 .rate_limit
-                .map(|v| v.try_into())
+                .map(TryInto::try_into)
                 .transpose()
                 .map_err(|_| {
                     Error::Validation("Application rate limit out of bounds".to_owned())
@@ -62,8 +64,8 @@ impl CreateMessageApp {
     }
 
     /// Fetches all information for creating a [`CreateMessageApp`] from the Redis cache if it
-    /// exists or from PostgreSQL otherwise. If the RedisCache is Some, but does not contain the
-    /// requisite information, fetch it from PostgreSQL and insert the data into the cache.
+    /// exists or from postgres otherwise. If the [`RedisCache`] is Some, but does not contain the
+    /// requisite information, fetch it from postgres and insert the data into the cache.
     pub async fn layered_fetch(
         redis: Option<&RedisCache>,
         pg: &DatabaseConnection,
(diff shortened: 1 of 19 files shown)


server/svix-server/src/v1/endpoints/attempt.rs
```diff
@@ -105,24 +107,24 @@ async fn list_messageattempts(
...
 
     if let Some(status) = status {
-        query = query.filter(messageattempt::Column::Status.eq(status))
+        query = query.filter(messageattempt::Column::Status.eq(status));
     }
 
     Ok(Json(MessageAttemptOut::list_response(
-        query.all(db).await?.into_iter().map(|x| x.into()).collect(),
-        limit as usize,
+        query.all(db).await?.into_iter().map(Into::into).collect(),
```
tasn, line 126: I can see why this is preferred, but god, this looks worse. :)

server/svix-server/src/v1/endpoints/endpoint.rs
```diff
@@ -51,16 +51,14 @@ pub fn validate_event_types_ids(
     }
 }
 
-pub fn validate_channels_endpoint(
-    channels: &EventChannelSet,
-) -> std::result::Result<(), ValidationError> {
+pub fn validate_channels(channels: &EventChannelSet) -> std::result::Result<(), ValidationError> {
```
tasn, line 54: Why did formatting change all of a sudden?

server/svix-server/src/v1/endpoints/endpoint.rs
```diff
@@ -449,11 +452,7 @@ async fn rotate_endpoint_secret(
 
         old_keys: Set(Some(ExpiringSigningKeys(
             iter::once(last_key)
-                .chain(
-                    old_keys
-                        .map(|x| x.0.into_iter())
-                        .unwrap_or_else(|| vec![].into_iter()),
-                )
+                .chain(old_keys.map_or_else(|| vec![].into_iter(), |x| x.0.into_iter()))
```
tasn, line 455: Side comment: silly decision on Rust to have the map the SECOND parameter and the else the first. It's literally called "map or else", in this order...

server/svix-server/src/worker.rs
```diff
@@ -179,7 +179,11 @@ async fn dispatch(
     };
     let attempt = match res {
         Ok(res) => {
-            let status_code = res.status().as_u16() as i16;
+            let status_code = res
+                .status()
+                .as_u16()
+                .try_into()
+                .expect("HTTP status does not fit into u16");
```
tasn, line 186:
```suggestion
                .expect("HTTP status does not fit into i16");
```

tasn (commented):
Made some comments. I'm OK with most of the changes, not with all of them. Some look very funky and I wonder whether we should turn them off (docstring, I'm looking at you). Also not a huge fan of all of the `try_into().expect("bla bla bla")`, it feels just tedious to write, especially for something we know is fine.

Also, if possible/easy, I'd much rather if we first got in all of the "easy ones" so that this PR is more manageable, as it's scary as it is. E.g. all of the missing `;`, the Display for error, the `to_owned()` -> `.clone()`, the removing of iter on arrays. Etc.

Everything contentious can go on its own PR.

server/svix-server/src/core/message_app.rs
```diff
@@ -1,3 +1,5 @@
+// Lots of false positives in this file
```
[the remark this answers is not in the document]
svix-daniel: Leftover comment I forgot to delete.  I was ignoring the lint that was interpreting "PostgreSQL" as a struct name in doc comments, but figured it was just easier to rename to "postgres" so I could catch the real occurrences.

PR #298, opened 2022-03-10
Add integration testing of application endpoints

svix-daniel:
Sets up tests which launch  the entire server and send requests to the application endpoints.


server/Cargo.toml
```diff
@@ -1,2 +1,2 @@
 [workspace]
-members = ["svix-server", "svix-server_derive"]
+members = ["svix-server", "svix-server_derive", "svix_test_suite"]
```
tasn, line 2:
```suggestion
members = ["svix-server", "svix-server_derive", "svix-test-suite"]
```
I know the _derive is inconsistent and confusing, but dashes are much prettier. :P

server/svix_test_suite/src/lib.rs
```diff
@@ -0,0 +1,93 @@
...
+pub struct SvixToken(pub String);
+
+/// Fetches the environmental variables needed to connect to and authenticate with a running Svix
+/// server
+fn fetch_env_vars() -> (SvixUri, SvixToken) {
+    let _ = dotenv::dotenv();
+
+    (
+        SvixUri(env::var("SVIX_SERVER_URL").unwrap()),
+        SvixToken(env::var("SVIX_AUTH_TOKEN").unwrap()),
```
tasn, line 35: This you don't want to fetch, but rather generate one with a custom org_id on every test.

server/svix-server/src/main.rs
```diff
@@ -53,35 +61,7 @@ enum JwtCommands {
...
-    if let Some(Commands::Jwt {
-        command: JwtCommands::Generate,
-    }) = &args.command
-    {
-        let token = generate_token(&cfg.jwt_secret).unwrap();
-        println!("Token (Bearer): {}", token);
-        exit(0);
-    }
-
+pub(crate) async fn run(cfg: Configuration, listener: Option<TcpListener>) {
```
tasn, line 64: No particular objection, but why do we need to potentially pass listener? Why can't we just pass the custom URL in the config in the tests?

server/svix-server/src/test_util.rs
```diff
@@ -0,0 +1,232 @@
...
+impl TestClient {
+    pub fn new(base_uri: String, auth_token: &str) -> TestClient {
+        TestClient {
+            base_uri,
+            auth_header: format!("Bearer {}", auth_token),
+            client: Client::new(),
+        }
+    }
+
+    pub async fn request<I: Serialize, O: DeserializeOwned>(
```
tasn, line 32: As I said on Slack, I think this can be much simpler without all of the optionals and the base request class. It'll provide better type safety and much shorter code. Here's what I had in mind (should work, but haven't tested):
```rust
pub struct TestClient {
    base_uri: String,
    auth_header: String,
    client: Client,
}

impl TestClient {
    pub fn new(base_uri: String, auth_token: &str) -> TestClient {
        TestClient {
            base_uri,
            auth_header: format!("Bearer {}", auth_token),
            client: Client::new(),
        }
    }

    fn build_url(&self, path: &str) -> String {
        format!("{}/{}", self.base_uri, path)
    }

    fn add_headers(&self, request: RequestBuilder) -> RequestBuilder {
        request.header("Authorization", &self.auth_header)
    }

    pub async fn post<I: Serialize, O: DeserializeOwned>(
        &self,
        path: &str,
        input: I,
        status_code: StatusCode,
    ) -> std::result::Result<O, String> {
        let req = self.client.post(self.build_url(path));
        let req = self.add_headers(req).json(&input);
        let resp = req.send().await.unwrap();
        assert!(resp.status() == status_code);
        Ok(resp.json().await.unwrap())
    }

    pub async fn put<I: Serialize, O: DeserializeOwned>(
        &self,
        path: &str,
        input: I,
        status_code: StatusCode,
    ) -> std::result::Result<O, String> {
        let req = self.client.put(self.build_url(path));
        let req = self.add_headers(req).json(&input);
        let resp = req.send().await.unwrap();
        assert!(resp.status() == status_code);
        Ok(resp.json().await.unwrap())
    }

    pub async fn get<O: DeserializeOwned + Clone>(
        &self,
        path: &str,
        status_code: StatusCode,
    ) -> std::result::Result<O, String> {
        let req = self.client.get(self.build_url(path));
        let req = self.add_headers(req);
        let resp = req.send().await.unwrap();
        assert!(resp.status() == status_code);
        Ok(resp.json().await.unwrap())
    }

    pub async fn delete<O: DeserializeOwned + Clone>(
        &self,
        path: &str,
        status_code: StatusCode,
    ) -> std::result::Result<O, String> {
        let req = self.client.delete(self.build_url(path));
        let req = self.add_headers(req);
        let resp = req.send().await.unwrap();
        assert!(resp.status() == status_code);
        Ok(resp.json().await.unwrap())
    }
}
```
Some points:
- We don't need to return the status_code because we always assert that it's what we expect (what we want in the context of a test suite).
- Body is not optional in post/put - if you sent to send an empty Body send an empty serialize struct (we have one in utils called `EmptyResponse`).
- Response is also not optional, if you want to parse nothing try to parse into  `EmptyResponse`.
- We don't want to have the test client have logic about comparing the body because that's very specific, and anyway we want the body out because we sometimes do different kinds of comparisons (you'll see as you add more tests).
- No need for callbacks/closures, we just return the body so we can do whatever comparisons we want.
- I made these functions return a result and was too lazy to add an error type/`anyhow`. Though to be honest, they should probably just return `O` and always unwrap inside. We just assume everything works, and if it doesn't we want to throw anyway. No need to propagate the errors.

I removed the support for custom headers for now, but the way to do it is probably by having *state* on the client and just letting people change the headers associated with the client before making a request. Yeah, it's a bit ugly, but it doesn't really matter and it's much simpler.

server/svix-server/src/test_util.rs
```diff
@@ -0,0 +1,232 @@
...
+            expected_code,
+        )
+        .await
+    }
+}
+
+pub fn start_svix_server() -> (TestClient, tokio::task::JoinHandle<()>) {
+    let _ = dotenv::dotenv();
+    let cfg = crate::cfg::load().unwrap();
+    let token = crate::generate_token(&cfg.jwt_secret).unwrap();
```
tasn, line 225: You probably want to generate a different token for each test each with a random OrganizationId. This way tests will be completely isolated from one another rather than reusing the same account (so we can potentially run them in parallel!).

server/svix-server/src/main.rs
svix-daniel: I wanted to pass a `TcpListener` so I could bind to port `0` and get the one that was actually assigned via the `local_addr` method. I figured this was easier than requiring a port block be free for all the test instances.

server/svix-server/src/test_util.rs
svix-daniel: The advantage to leaving the `unwrap` in the test body instead of in a helper function is the panic message includes a file/line number in the test. Otherwise I see what you mean now :).

server/svix-server/src/main.rs
tasn: Yeah, cool, you're 100% right. :)

server/svix-server/src/test_util.rs
tasn: Good point. I guess you can make it return a result and unwrap in the test!

server/svix-server/src/core/security.rs
```diff
@@ -19,6 +19,10 @@ use crate::{
     error::{Error, HttpError, Result},
 };
 
+// False positive as the Trait is used with the [`BaseId::generate_`] method
+#[allow(unused_imports)]
```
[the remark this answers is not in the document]
tasn: Ha, you already have such a module! Just put it there :)

server/svix-server/src/test_util.rs
```diff
@@ -0,0 +1,152 @@
...
+            .await
+            .context("error receiving/parsing response")?)
+    }
+}
+
+pub fn start_svix_server() -> (TestClient, tokio::task::JoinHandle<()>) {
+    let _ = dotenv::dotenv();
+    let cfg = crate::cfg::load().unwrap();
+    let token = generate_token_random_org(&cfg.jwt_secret).unwrap();
+    println!("{}", token);
```
tasn, line 145: 🤔

server/docker-compose.yml
```diff
@@ -20,11 +20,6 @@ services:
 
   pgbouncer:
     image: edoburu/pgbouncer:1.15.0
-    healthcheck:
-      test: "pg_isready -h localhost"
-      interval: 30s
-      timeout: 10s
-      retries: 3
```
tasn, line 27: Unrelated, please revert this change.

server/svix-server/src/db/mod.rs
```diff
@@ -8,19 +8,24 @@ use crate::cfg::Configuration;
...
     } else {
         panic!("db_dsn format not recognized. {}", &cfg.db_dsn)
     }
 }
+
+pub async fn init_db(cfg: &Configuration) -> DatabaseConnection {
+    SqlxPostgresConnector::from_sqlx_postgres_pool(connect(cfg).await)
+}
+
+pub async fn run_migrations(cfg: &Configuration) {
```
tasn, line 26: We still want to automatically run it in release mode (for now), so please call it explicitly from `main()`.

The way it is now, it changes behavior.

server/svix-server/src/db/mod.rs
```diff
@@ -8,19 +8,24 @@ use crate::cfg::Configuration;
...
+    let db = connect(cfg).await;
+    MIGRATIONS.run(&db).await.unwrap();
+
+    println!("Migrations run");
```
tasn, line 30: You can print it in the place that handles the command (so when called explicitly), but it should probably be silent when called implicitly.

server/docker-compose.yml
svix-daniel: Oops -- removed it from the wrong  `docker-compose`

tasn (approved):

PR #304, opened 2022-03-11
Libs: Remove ListOptionsDouble

svix-frank:
prev_iterator is not used, prev iterators should be passed as `iterator`


tasn (approved):
Missing go, no?

tasn (approved):

PR #309, opened 2022-03-11
Python: Fix headers typing

svix-frank:
Same as the fix for payload in #308.


svix-stephanie (approved):

svix-frank:
@svix-stephanie is working on a long-term fix to this hack

Issue #316, opened 2022-03-15
Vendor data added since version 0.53.0


galaktipus:

tasn:
@fcjr, any idea?

PR #319, opened 2022-03-15
CI: Decrease ruby bundle size

fcjr:
The bundle cache directory is in `${pwd}` so we were erroneously including
in our CI builds.  This disables the cache for now, decreasing our build size.

Fixes https://github.com/svix/svix-webhooks/issues/316.

### DIFF
--- a/.github/workflows/ruby-release.yml
+++ b/.github/workflows/ruby-release.yml
@@ -3,7 +3,7 @@ name: Ruby Release
 on:
   push:
     tags:
-      - 'v*'
+      - "v*"
 
 jobs:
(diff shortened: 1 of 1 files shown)


tasn (dismissed):

.github/workflows/ruby-release.yml
```diff
@@ -3,7 +3,7 @@ name: Ruby Release
 on:
   push:
     tags:
-      - 'v*'
+      - "v*"
```
fcjr, line 6:
```suggestion
      - 'v*'
```

PR #327, opened 2022-03-18
Correct release path in release CI

svix-daniel:


.github/workflows/server-release.yml
```diff
@@ -1,6 +1,9 @@
 name: Server Release
 
 on:
+  pull_request:
```
tasn, line 4: Need to remove this

tasn: I did it.

tasn (approved):

PR #321, opened 2022-03-16
Add basic message dispatch tests

svix-daniel:
A simple create/read/list test for the message endpoints and utility functions for testing the dispatch of messages. To be expanded on.

PR #328, opened 2022-03-20
Add more message tests.

tasn:
This is just the code I removed from #321


tasn:
This has already been merged back I believe.

PR #332, opened 2022-03-21
Adds List Attempted Messages Endpoint

svix-daniel:
Adds an endpoint which lists all attempted messages associated with a given endpoint object.


server/svix-server/src/v1/endpoints/attempt.rs
```diff
@@ -65,6 +65,101 @@ impl From<messageattempt::Model> for MessageAttemptOut {
     }
 }
 
+// TODO: Mostly similar to `v1::endpoints::message::MessageOut, is there some way to reduce
+// duplication cleanly?
```
tasn, line 69: Did the `flatten` idea not work?

server/svix-server/src/v1/endpoints/attempt.rs
```diff
@@ -65,6 +65,101 @@ impl From<messageattempt::Model> for MessageAttemptOut {
...
+/// Additional parameters (besides pagination) in the query string for the "List Attempted Messages"
+/// enpoint.
+#[derive(Debug, Deserialize, Validate)]
+pub struct ListAttemptedMessagesQueryParameters {
+    #[validate]
+    channel: Option<EventChannel>,
+    status: Option<MessageStatus>,
+}
+
+/// GET /api/v1/app/:app_id/enpoint/:endp_id/msg/
```
tasn, line 116: I'm not a fan of this, we don't add this comment elsewhere, and it's nice that the routing is handled in one place, this makes it disjoin.

server/svix-server/src/v1/endpoints/attempt.rs
```diff
@@ -88,7 +183,7 @@ async fn list_messageattempts(
         permissions: _,
         app,
     }: AuthenticatedApplication,
-) -> Result<Json<ListResponse<MessageAttemptOut>>> {
+) -> Result<Json<ListResponse<MessageAttemptOut>>, Error> {
```
tasn, line 186: Why not use the `Result` that is already defined this way like it was?

server/svix-server/src/v1/endpoints/attempt.rs
```diff
@@ -154,21 +249,17 @@ async fn get_messageattempt(
 async fn resend_webhook(
     Extension(ref db): Extension<DatabaseConnection>,
     Extension(queue_tx): Extension<TaskQueueProducer>,
-    Path((_app_id, msg_id, endpoint_id)): Path<(
-        ApplicationIdOrUid,
-        MessageIdOrUid,
-        EndpointIdOrUid,
-    )>,
+    Path((_app_id, msg_id, endp_id)): Path<(ApplicationIdOrUid, MessageIdOrUid, EndpointIdOrUid)>,
```
tasn, line 252: These changes are unrelated. It's not the end of the world this time, but please try to avoid renaming variables in a PR without a good reason.

server/svix-server/src/v1/endpoints/attempt.rs
```diff
@@ -65,6 +65,101 @@ impl From<messageattempt::Model> for MessageAttemptOut {
...
+    let limit = pagination.limit;
+    let iterator = pagination.iterator.clone();
+
+    let mut msg_and_dest = messagedestination::Entity::secure_find_by_endpoint(endp_id)
+        .order_by_desc(messagedestination::Column::MsgId)
+        .limit(limit + 1)
+        .join(
+            sea_orm::JoinType::Join,
+            messagedestination::Entity::has_one(message::Entity).into(),
+        );
```
tasn, line 139: I know you said it's weird, but I think this actually looks fine. :)

You can also just define the relation on the message destination model and then just use that instead of doing it custom here.

server/svix-server/src/v1/endpoints/attempt.rs
```diff
@@ -65,6 +65,101 @@ impl From<messageattempt::Model> for MessageAttemptOut {
...
+
+    if let Some(iterator) = iterator {
+        msg_and_dest = msg_and_dest.filter(messagedestination::Column::MsgId.gt(iterator));
```
tasn, line 142:
```suggestion
        msg_and_dest = msg_and_dest.filter(messagedestination::Column::MsgId.lt(iterator));
```
Watch out for these things, we order by descending order, so it should be `lt`, not `gt`.

server/svix-server/src/v1/endpoints/attempt.rs
svix-daniel: Deriving `FromQueryResult` when the custom `Result` alias errors  because `this  type alias takes 1 generic argument, but 2 generic arguments  were supplied`.  I realizes this has been fixed in the latest release of SeaORM (by you nonetheless :smile: ), so we can fix this by bumping the version.

tasn: Fix: 😎 

Though why is it needed all of a sudden in all of these places? Happy to also upgrade if needed.

server/svix-server/src/v1/endpoints/attempt.rs
```diff
@@ -65,6 +65,101 @@ impl From<messageattempt::Model> for MessageAttemptOut {
...
+        ListAttemptedMessagesQueryParameters,
+    >,
+    Path((_app_id, endp_id)): Path<(ApplicationId, EndpointId)>,
+    AuthenticatedApplication {
+        permissions: _,
+        app: _,
+    }: AuthenticatedApplication,
+) -> Result<Json<ListResponse<AttemptedMessageOut>>, Error> {
+    let limit = pagination.limit;
+    let iterator = pagination.iterator.clone();
```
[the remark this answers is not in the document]
svix-daniel: It requires an owned value to be used in the filter. I tried taking it instead, but that'd require a `DerefMut` implementation on  `ValidatedQuery`.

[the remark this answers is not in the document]
tasn: OKie.

server/svix-server/src/v1/endpoints/attempt.rs
```diff
@@ -65,6 +65,95 @@ impl From<messageattempt::Model> for MessageAttemptOut {
...
+
+    let mut msg_and_dest = messagedestination::Entity::secure_find_by_endpoint(endp_id)
+        .order_by_desc(messagedestination::Column::MsgId)
+        .limit(limit + 1)
+        .column_as(message::Column::Id, "id")
+        .column_as(message::Column::EventType, "event_type")
+        .column_as(message::Column::Uid, "event_id")
+        .column_as(message::Column::Channels, "channels")
+        .column_as(message::Column::Payload, "payload")
+        .column_as(message::Column::CreatedAt, "created_at")
```
tasn, line 125: This is painfully manual, why does it have to be like this? What happened to how it was before?

svix-daniel: So, with the previous version it didn't actually select  any of the `message` columns and panicked unable to fill the struct. This is the only way I found with a quick glance at the SeaORM documentation, but I'll look for a better solution before we merge.

server/svix-server/src/v1/endpoints/attempt.rs
```diff
@@ -65,6 +65,112 @@ impl From<messageattempt::Model> for MessageAttemptOut {
...
 }
 
+// FIXME: Contains all members of a [`v1::endpoints::message::MessageOut`], so find a way to
+// #[serde(flatten)]` a [`super::message::MessageOut`] in cleanly. An attempt was made, but it would
+// require a custom [`ModelOut`] impl and another impl of [`sea_orm::TryGettable`] on the
+// [`super::message::MessageOut`] type in order for [`FromQueryResult`] to work.
+
+/// A model containing information on a given message plus additional fields on the last attempt for
+/// that message.
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize, ModelOut, FromQueryResult)]
```
tasn, line 75:
```suggestion
#[derive(Clone, Debug, PartialEq, Serialize, Deserialize, ModelOut)]
```
No longer needed, no?

server/svix-server/src/v1/endpoints/attempt.rs
```diff
@@ -65,6 +65,112 @@ impl From<messageattempt::Model> for MessageAttemptOut {
...
+            .map(
+                |(dest, msg): (messagedestination::Model, Option<message::Model>)| {
+                    let msg = msg.ok_or_else(|| {
+                        tracing::error!(
+                            "messagedestination::Enitity has no associated message::Entity"
+                        );
+                        Error::Database("Malformed data".to_owned())
+                    })?;
+
+                    Ok(AttemptedMessageOut::from_dest_and_msg(dest, msg))
```
tasn, line 166: Does this actually work? I don't understand how the error is propagated outside of the closure.

Also, I think doing:
```suggestion
                    AttemptedMessageOut::from_dest_and_msg(dest, msg.expect("Message missing for message endpoint exist"))
```
Is sufficient, as Axum automatically catches unwraps and returns a 500 (I think, check me!).

Though it's good either way, just wanted to keep our code concise.

PR #333, opened 2022-03-22
Update SeaORM to 0.6.0

svix-daniel:
Simply updates the version and fixes compiler errors associated with the update's breaking changes. Motivated by the changes @tasn made for increased sanitation in macros.


tasn (approved):
Glad to see my patch is released and useful. :)

PR #338, opened 2022-03-24
Add implementation for "List Attempts by Endpoint" endpoint

svix-daniel:
Adds the endpoint, associated models, a basic test, and associated test utilities. Additionally changes the queue type for CI tests to in-memory to avoid aborted threads mid-database-transaction.

### DIFF
--- a/server/svix-server/src/test_util.rs
+++ b/server/svix-server/src/test_util.rs
@@ -201,11 +201,8 @@ async fn test_receiver_route(
     code
 }
 
-pub async fn run_with_retries<F: Future<Output = Result<()>>, C: Fn() -> F>(
-    f: C,
-    max_attempts: usize,
-) -> Result<()> {
-    for attempt in 0..max_attempts {
+pub async fn run_with_retries<F: Future<Output = Result<()>>, C: Fn() -> F>(f: C) -> Result<()> {
+    for attempt in 0..20 {
         let out = f().await;
         if out.is_ok() {
             return Ok(());
--- a/server/svix-server/src/test_util.rs
+++ b/server/svix-server/src/test_util.rs
@@ -1,9 +1,10 @@
-use std::net::TcpListener;
+use std::{future::Future, net::TcpListener};
 
 use anyhow::{Context, Result};
 
 use reqwest::{Client, RequestBuilder, StatusCode};
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
+use tokio::sync::mpsc;
 
 use crate::core::security::test_util::generate_token_random_org;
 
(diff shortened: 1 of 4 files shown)


server/svix-server/src/test_util.rs
```diff
@@ -152,3 +153,68 @@ pub fn start_svix_server() -> (TestClient, tokio::task::JoinHandle<()>) {
...
+    axum::extract::Extension(ref tx): axum::extract::Extension<mpsc::Sender<serde_json::Value>>,
+    axum::extract::Extension(code): axum::extract::Extension<axum::http::StatusCode>,
+) -> axum::http::StatusCode {
+    tx.send(json).await.unwrap();
+    code
+}
+
+pub async fn run_with_retries<F: Future<Output = Result<()>>, C: Fn() -> F>(
+    f: C,
+    max_attempts: usize,
```
tasn, line 206: Hardcode the number of attempts here. Don't pass as param. Should be 20 or something.

server/svix-server/src/v1/endpoints/attempt.rs
```diff
@@ -158,6 +162,108 @@ async fn list_attempted_messages(
...
+        query = query.filter(messageattempt::Column::Status.eq(status));
+    }
+
+    query = match status_code_class {
+        Some(StatusCodeClass::Code1xx) => {
+            query.filter(messageattempt::Column::ResponseStatusCode.lt(100i16))
+        }
+
+        Some(StatusCodeClass::Code2xx) => {
+            query.filter(messageattempt::Column::ResponseStatusCode.between(199i16, 300i16))
```
tasn, line 223: Do we really need to explicitly mark it as i16? Doesn't rust do it automatically?

server/svix-server/src/v1/endpoints/attempt.rs
```diff
@@ -373,4 +482,117 @@ mod tests {
...
+                    .unwrap();
+                let list_2: ListResponse<MessageAttemptOut> = client
+                    .get(
+                        &format!("api/v1/app/{}/attempt/endpoint/{}/", app_id, endp_id_2),
+                        StatusCode::OK,
+                    )
+                    .await
+                    .unwrap();
+
+                for (num, list) in [list_1, list_2].into_iter().enumerate() {
```
tasn, line 571: This shouldn't be inside run with retries
This function is just a "sync gate" to make sure we got everything

server/svix-server/src/v1/endpoints/attempt.rs
```diff
@@ -158,6 +162,108 @@ async fn list_attempted_messages(
...
+
+        Some(StatusCodeClass::Code4xx) => {
+            query.filter(messageattempt::Column::ResponseStatusCode.between(399i16, 500i16))
+        }
+
+        Some(StatusCodeClass::Code5xx) => {
+            query.filter(messageattempt::Column::ResponseStatusCode.between(599i16, 600i16))
+        }
+
+        Some(StatusCodeClass::CodeNone) => query,
```
tasn, line 238: This is wrong, it should be 0-99

tasn (approved):

tasn:
I just realized something.

You made the tests pass in CI but not locally. They are flaky here. You need to override the config in `start_svix_server`, not in the github actions file.

PR #342, opened 2022-03-26
Reorganize integration tests based on Rust best practices.

tasn:
Following the recommendations from the rust book:
https://doc.rust-lang.org/book/ch11-03-test-organization.html

This is in order to make integration tests simpler.

We can potentially make the things we share more strict, but it doesn't
really matter as this shouldn't be used as a lib.


tasn:
@svix-daniel, I think that since it's now actual separate tests, we can remove the feature and instead just tell people to run `cargo test --lib` to only test the unit tests, otherwise they can just run the full suite with `cargo test`.

What do you think?

svix-daniel (approved):

PR #350, opened 2022-03-29
Adds `prev` iterators as query filters for `list_attempts_by_endpoint` and `list_attempts_by_msg`

svix-daniel:
Adds `prev` iterators for the `list_attempts_by_endpoint` and `list_attempts_by_msg` endpoints. To be followed by expanding endpoints covered by this structure and to add  `before` and `after` parameters.

### DIFF
--- a/server/svix-server/src/v1/endpoints/attempt.rs
+++ b/server/svix-server/src/v1/endpoints/attempt.rs
@@ -182,7 +182,7 @@ pub struct ListAttemptsByEndpointQueryParameters {
 
 // Applies filters common to [`list_attempts_by_endpoint`] and [`list_attempts_by_msg`]
 fn list_attempts_by_endpoint_or_message_filters(
-    mut query: Select<messageattempt::Entity>,
+    query: Select<messageattempt::Entity>,
     limit: u64,
     iterator: Option<ReversibleIterator<MessageAttemptId>>,
     status: Option<MessageStatus>,
@@ -190,7 +190,7 @@ fn list_attempts_by_endpoint_or_message_filters(
     event_types: Option<EventTypeNameSet>,
     channel: Option<EventChannel>,
 ) -> Select<messageattempt::Entity> {
-    query = match iterator {
+    let query = match iterator {
         Some(ReversibleIterator::Prev(id)) => query.filter(
             Condition::any().add(
                 messageattempt::Column::Id.in_subquery(
@@ -207,7 +207,7 @@ fn list_attempts_by_endpoint_or_message_filters(
         None => query,
     };
 
-    query = query
+    let mut query = query
         .limit(limit + 1)
         .order_by_desc(messageattempt::Column::Id);
 
--- a/server/svix-server/src/v1/endpoints/attempt.rs
+++ b/server/svix-server/src/v1/endpoints/attempt.rs
@@ -190,28 +190,25 @@ fn list_attempts_by_endpoint_or_message_filters(
     channel: Option<EventChannel>,
 ) -> Select<messageattempt::Entity> {
     query = match iterator {
-        Some(PaginationIterator::Prev(id)) => query
-            .filter(
-                Condition::any().add(
-                    messageattempt::Column::Id.in_subquery(
-                        Query::select()
-                            .column(messageattempt::Column::Id)
-                            .and_where(messageattempt::Column::Id.gt(id))
-                            .order_by_columns(vec![(messageattempt::Column::Id, Order::Asc)])
-                            .limit(limit + 1)
-                            .to_owned(),
-                    ),
+        Some(PaginationIterator::Prev(id)) => query.filter(
+            Condition::any().add(
+                messageattempt::Column::Id.in_subquery(
+                    Query::select()
+                        .column(messageattempt::Column::Id)
+                        .and_where(messageattempt::Column::Id.gt(id))
+                        .order_by_columns(vec![(messageattempt::Column::Id, Order::Asc)])
+                        .limit(limit + 1)
+                        .to_owned(),
                 ),
-            )
-            .order_by_desc(messageattempt::Column::Id),
-        Some(PaginationIterator::Normal(id)) => query
-            .limit(limit + 1)
-            .order_by_desc(messageattempt::Column::Id)
-            .filter(messageattempt::Column::Id.lt(id)),
-        None => query
-            .limit(limit + 1)
-            .order_by_desc(messageattempt::Column::Id),
+            ),
+        ),
+        Some(PaginationIterator::Normal(id)) => query.filter(messageattempt::Column::Id.lt(id)),
+        None => query,
     };
+
+    query = query
+        .limit(limit + 1)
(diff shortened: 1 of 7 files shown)


server/svix-server/src/v1/utils.rs
```diff
@@ -24,13 +24,85 @@ const fn default_limit() -> u64 {
 }
 
 #[derive(Debug, Deserialize, Validate)]
-pub struct Pagination<T: Validate> {
+pub struct Pagination<T: Validate + Clone> {
```
tasn, line 27: Adding clone makes me sad. Why is it needed?

server/svix-server/src/v1/utils.rs
```diff
@@ -24,13 +24,85 @@ const fn default_limit() -> u64 {
...
     #[serde(default = "default_limit")]
     pub limit: u64,
     #[validate]
     pub iterator: Option<T>,
 }
 
+#[derive(Debug, Deserialize, Validate)]
+pub struct ReversablePagination<T: 'static + Validate + Clone + From<String>> {
```
tasn, line 35: Reversible. :)

server/svix-server/src/v1/utils.rs
```diff
@@ -24,13 +24,85 @@ const fn default_limit() -> u64 {
...
+    #[validate]
+    pub iterator: Option<PaginationIterator<T>>,
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub enum PaginationIterator<T: Validate + Clone> {
+    Normal(T),
+    Prev(T),
+}
+
```
tasn, line 47: Deserialize alternative (so instead of all of the below) would be something like in https://github.com/svix/svix-webhooks/blob/main/server/svix-server/src/core/types.rs#L439

So:
1. Deserialize to string.
2. Check if it starts with `-`. If yes: serialize as `MessageId` for everything after the first char. If no: serialize as `MessageId`. You can just take the string you serialized in other case.
3. Validate should remain like you did it.

server/svix-server/src/v1/endpoints/attempt.rs
```diff
@@ -174,21 +181,37 @@ pub struct ListAttemptsByEndpointQueryParameters {
...
+                        Query::select()
+                            .column(messageattempt::Column::Id)
+                            .and_where(messageattempt::Column::Id.gt(id))
+                            .order_by_columns(vec![(messageattempt::Column::Id, Order::Asc)])
+                            .limit(limit + 1)
+                            .to_owned(),
+                    ),
+                ),
+            )
+            .order_by_desc(messageattempt::Column::Id),
```
tasn, line 206: You can move the `order_by_desc` outside of the match, it's used in all three options. Heck, you can just do it before the match even like where it was before and even keep the limit (because it doesn't matter).

server/svix-server/src/v1/utils.rs
```diff
@@ -31,6 +31,47 @@ pub struct Pagination<T: Validate> {
...
+    for PaginationIterator<T>
+{
+    fn deserialize<D>(deserializer: D) -> std::result::Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        String::deserialize(deserializer).map(|s| {
+            if s.starts_with('-') {
+                let s = s.trim_start_matches('-');
+                PaginationIterator::Prev(T::from(s.to_owned()))
```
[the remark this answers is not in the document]
svix-daniel: While it works in this case, I'm always hesitant to use `&str` slices -- they cause unexpected behavior with characters that aren't a single byte in width.

server/svix-server/src/v1/endpoints/attempt.rs
```diff
@@ -174,21 +181,37 @@ pub struct ListAttemptsByEndpointQueryParameters {
...
-    }
+    query = match iterator {
+        Some(PaginationIterator::Prev(id)) => query
+            .filter(
+                Condition::any().add(
+                    messageattempt::Column::Id.in_subquery(
+                        Query::select()
+                            .column(messageattempt::Column::Id)
+                            .and_where(messageattempt::Column::Id.gt(id))
+                            .order_by_columns(vec![(messageattempt::Column::Id, Order::Asc)])
```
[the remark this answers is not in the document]
svix-daniel: The API is just different for `SelectStatement` versus `Select`.

server/svix-server/src/v1/utils.rs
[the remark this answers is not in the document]
tasn: But you know that `-` is a single byte...

server/svix-server/src/v1/endpoints/attempt.rs
```diff
@@ -176,19 +184,32 @@ pub struct ListAttemptsByEndpointQueryParameters {
...
+        Some(ReversibleIterator::Normal(id)) => query.filter(messageattempt::Column::Id.lt(id)),
+        None => query,
+    };
 
-    if let Some(iterator) = iterator {
-        query = query.filter(messageattempt::Column::Id.lt(iterator));
-    }
+    let mut query = query
+        .limit(limit + 1)
+        .order_by_desc(messageattempt::Column::Id);
```
tasn, line 212: Put it where it was (above) at line 185 to minimize the diff.

server/svix-server/src/v1/utils.rs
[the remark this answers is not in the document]
svix-daniel: This caused a clippy warning which I fixed leading to a better solution in the latest commit

server/svix-server/src/v1/utils.rs
```diff
@@ -51,13 +84,28 @@ pub trait ModelIn {
...
         let done = data.len() <= limit;
-        data.truncate(limit);
+
+        if is_prev_iter {
+            if data.len() > limit {
+                data = data.drain(data.len() - limit..).collect();
+            }
+        } else {
+            data.truncate(limit);
+        }
```
tasn, line 100:
```suggestion
        if data.len() > limit {
             if is_prev_iter {
                data = data.drain(data.len() - limit..).collect();
             } else {
                data.truncate(limit);
            }
        }
```
Makes the behavior more clear.

server/svix-server/src/v1/endpoints/attempt.rs
```diff
@@ -353,9 +378,16 @@ async fn list_attempts_by_msg(
...
 
+    let mut out: Vec<_> = query.all(db).await?.into_iter().map(Into::into).collect();
+
+    if is_prev {
+        out = out.into_iter().rev().collect();
+    }
+
     Ok(Json(MessageAttemptOut::list_response(
-        query.all(db).await?.into_iter().map(Into::into).collect(),
+        out,
```
tasn, line 388:
```rust
    let mut out: Vec<_> = query.all(db).await?.into_iter();

    if is_prev {
        out = out.rev();
    }

    Ok(Json(MessageAttemptOut::list_response(
        out.map(Into::into).collect(),
```

How about this? Saves the double vector creation.

tasn (approved):

PR #354, opened 2022-03-31
Extend `prev_iterator` implementation

svix-daniel:
Extends the `prev_iterator` implementation to include `before`/`after` query parameters and to apply this functionality to all list operations in the attempt API.

### DIFF
--- a/server/svix-server/src/v1/endpoints/attempt.rs
+++ b/server/svix-server/src/v1/endpoints/attempt.rs
@@ -5,7 +5,7 @@ use crate::{
     core::{
         security::AuthenticatedApplication,
         types::{
-            ApplicationId, ApplicationIdOrUid, BaseId, EndpointId, EndpointIdOrUid, EventChannel,
+            ApplicationId, ApplicationIdOrUid, EndpointId, EndpointIdOrUid, EventChannel,
             EventTypeNameSet, MessageAttemptId, MessageAttemptTriggerType, MessageEndpointId,
             MessageId, MessageIdOrUid, MessageStatus, StatusCodeClass,
         },
@@ -491,14 +491,19 @@ pub struct ListAttemptsForEndpointQueryParameters {
     #[validate]
     pub channel: Option<EventChannel>,
     pub status: Option<MessageStatus>,
+    pub before: Option<DateTime<Utc>>,
+    pub after: Option<DateTime<Utc>>,
 }
 
 async fn list_attempts_for_endpoint(
     extension: Extension<DatabaseConnection>,
-    pagination: ValidatedQuery<Pagination<MessageAttemptId>>,
-    ValidatedQuery(ListAttemptsForEndpointQueryParameters { channel, status }): ValidatedQuery<
-        ListAttemptsForEndpointQueryParameters,
-    >,
+    pagination: ValidatedQuery<Pagination<ReversibleIterator<MessageAttemptId>>>,
+    ValidatedQuery(ListAttemptsForEndpointQueryParameters {
+        channel,
+        status,
+        before,
+        after,
+    }): ValidatedQuery<ListAttemptsForEndpointQueryParameters>,
     list_filter: MessageListFetchOptions,
     Path((app_id, msg_id, endp_id)): Path<(ApplicationIdOrUid, MessageIdOrUid, EndpointIdOrUid)>,
     auth_app: AuthenticatedApplication,
@@ -510,6 +515,8 @@ async fn list_attempts_for_endpoint(
             endpoint_id: Some(endp_id),
             channel,
             status,
+            before,
+            after,
         }),
         list_filter,
         Path((app_id, msg_id)),
@@ -525,15 +532,19 @@ pub struct AttemptListFetchOptions {
     #[validate]
     pub channel: Option<EventChannel>,
     pub status: Option<MessageStatus>,
+    pub before: Option<DateTime<Utc>>,
+    pub after: Option<DateTime<Utc>>,
 }
 
 async fn list_messageattempts(
     Extension(ref db): Extension<DatabaseConnection>,
-    pagination: ValidatedQuery<Pagination<MessageAttemptId>>,
+    ValidatedQuery(pagination): ValidatedQuery<Pagination<ReversibleIterator<MessageAttemptId>>>,
     ValidatedQuery(AttemptListFetchOptions {
         endpoint_id,
         channel,
         status,
+        before,
+        after,
     }): ValidatedQuery<AttemptListFetchOptions>,
     list_filter: MessageListFetchOptions,
     Path((_app_id, msg_id)): Path<(ApplicationIdOrUid, MessageIdOrUid)>,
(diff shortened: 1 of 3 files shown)


server/svix-server/src/v1/endpoints/attempt.rs
```diff
@@ -174,30 +174,52 @@ pub struct ListAttemptsByEndpointQueryParameters {
...
+    after: Option<DateTime<Utc>>,
+    iterator: Option<ReversibleIterator<MessageAttemptId>>,
 }
 
 // Applies filters common to [`list_attempts_by_endpoint`] and [`list_attempts_by_msg`]
 fn list_attempts_by_endpoint_or_message_filters(
     query: Select<messageattempt::Entity>,
-    limit: u64,
-    iterator: Option<ReversibleIterator<MessageAttemptId>>,
+    pagination: PaginationInformation,
```
tasn, line 191: Change it back (see other comment).

server/svix-server/src/v1/endpoints/attempt.rs
```diff
@@ -174,30 +174,52 @@ pub struct ListAttemptsByEndpointQueryParameters {
...
     status: Option<MessageStatus>,
     status_code_class: Option<StatusCodeClass>,
     event_types: Option<EventTypeNameSet>,
     channel: Option<EventChannel>,
 ) -> Select<messageattempt::Entity> {
-    let mut query = match iterator {
+    let mut query = match pagination.iterator {
```
tasn, line 197: Change it back (see other comment).

server/svix-server/src/v1/endpoints/attempt.rs
```diff
@@ -275,7 +299,8 @@ async fn list_attempts_by_endpoint(
     let limit = pagination.limit;
     let iterator = pagination.iterator.take();
```
tasn, line 300: Just add:
```rust
if iterator.is_none() {
    if before {
        iterator = ReversibleIterator::Prev(MessageAttemptId::start_id(before));
    } else if after {
        iterator = ReversibleIterator::Prev(MessageAttemptId::end_id(after));
    }
}
```

server/svix-server/src/v1/endpoints/attempt.rs
```diff
@@ -344,7 +358,15 @@ async fn list_attempts_by_msg(
     }: AuthenticatedApplication,
 ) -> Result<Json<ListResponse<MessageAttemptOut>>> {
     let limit = pagination.limit;
-    let iterator = pagination.iterator.take();
+    let iterator = pagination.iterator.take().or_else(|| {
+        if let Some(before) = before {
+            Some(ReversibleIterator::Normal(MessageAttemptId::start_id(
```
tasn, line 363: The indentation is different here compared to the other place.

server/svix-server/tests/e2e_attempt.rs
```diff
@@ -145,3 +145,191 @@ async fn test_list_attempts_by_endpoint() {
...
+        assert_eq!(all_attempts.data[0], first_three_by_time.data[0]);
+        assert_eq!(all_attempts.data[1], first_three_by_time.data[1]);
+        assert_eq!(all_attempts.data[2], first_three_by_time.data[2]);
+
+        // `before field`
+        let last_three_by_time: ListResponse<MessageAttemptOut> = client
+            .get(
+                &format!(
+                    "api/v1/app/{}/attempt/endpoint/{}/?before={}",
+                    app.id, ep.id, all_attempts.data[2].created_at,
```
tasn, line 322: Does this actually work? I've had issues with doing something like this because `created_at` doesn't get url-escaped (it has a colon in it).

server/svix-server/tests/e2e_attempt.rs
```diff
@@ -145,3 +145,191 @@ async fn test_list_attempts_by_endpoint() {
     receiver_1.jh.abort();
     receiver_2.jh.abort();
 }
+
+#[tokio::test]
+async fn test_pagination() {
```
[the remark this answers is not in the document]
svix-daniel: The test includes all iterators -- standard forward/reverse and before/after, so I figured this was a more appropriate name even if the `prev_iterator` was the tricky part.

server/svix-server/tests/e2e_attempt.rs
```diff
@@ -145,3 +145,191 @@ async fn test_list_attempts_by_endpoint() {
...
+    let (client, _jh) = start_svix_server();
+
+    // Setup three endpoints and six messages so there's a sufficient number to test pagination
+    let app = create_test_app(&client, "app1").await.unwrap();
+
+    let mut receivers = Vec::new();
+    for _ in 0..3 {
+        receivers.push(TestReceiver::start(StatusCode::OK));
```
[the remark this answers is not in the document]
svix-daniel: This was just to prevent extra attempts from showing up as they fail going to a dead URL

server/svix-server/tests/e2e_attempt.rs
```diff
@@ -145,3 +145,191 @@ async fn test_list_attempts_by_endpoint() {
...
+            )
+            .await
+            .unwrap();
+
+        assert_eq!(all_attempts.data.len(), 6);
+        assert_eq!(first_three.data.len(), 3);
+
+        assert_eq!(all_attempts.data[0], first_three.data[0]);
+        assert_eq!(all_attempts.data[1], first_three.data[1]);
+        assert_eq!(all_attempts.data[2], first_three.data[2]);
```
tasn, line 230:
```suggestion
        assert_eq!(&all_attempts.data[0..=2], &first_three.data);
```
Shouldn't this do the trick?

server/svix-server/tests/e2e_attempt.rs
tasn: It could be that the tests are wrong. :)

For example, it could be that you wrote `before` as `befffore` and it's not actually parsing the correct thing and that's why it works. Need to make sure it's correct by e.g. adding a test that uses a timestamp from before that expects 0 results or something of that sort.

server/svix-server/tests/e2e_attempt.rs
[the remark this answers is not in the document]
tasn: 👍🏻

server/svix-server/src/v1/utils.rs
```diff
@@ -63,6 +64,44 @@ impl<T: Validate> Validate for ReversibleIterator<T> {
     }
 }
 
+pub enum ReversibleSelect<T: QuerySelect> {
+    Forward(T),
+    Reversed(T),
+}
+
+pub fn apply_pagination<
```
tasn, line 72: I'm not sure this actually makes things simpler. Between `ReversibleSelect` and and everything I think it just adds complexity rather than reduces it.

I can *maybe* see value in a function that accepts an iterator, after, before and returns ReversibleIterator that can be shared everywhere. Heck, I can even see value in the other part being a function (even called apply_pagination). Though I think passing before/after which are *not* pagination related into a pagination function is confusing. And no, changing the function name won't help. :)

What do you think?

server/svix-server/tests/e2e_attempt.rs
```diff
@@ -234,3 +234,512 @@ async fn test_message_attempts() {
...
+            .await
+            .unwrap();
+
+        forward_msgs.append(&mut out.data);
+        done = out.done;
+        prev_iterator = out.prev_iterator;
+        iterator = out.iterator;
+    }
+
+    assert_eq!(forward_msgs.len(), 28);
```
tasn, line 713:
```suggestion
    assert_eq!(forward_msgs.len(), messages.len());
```

tasn (dismissed):

tasn (approved):

PR #359, opened 2022-04-01
Tests: make a common test for testing list endpoints.

tasn:
While doing it, also improved run_with_retries to return the successful value.


svix-james:
Very nice!

svix-james (approved):

PR #360, opened 2022-04-01
Server: Worker HTTP error handling fixes

jaymell:
## Motivation

Rust request-worker not currently handling HTTP non-200 errors successfully. These failed requests should be getting retried as with other types of failures.

## Solution
Properly wrap non-200s in Result::Err for retries/further error handling.


server/svix-server/src/worker.rs
```diff
@@ -193,20 +193,24 @@ async fn dispatch(
...
         Err(err) => {
             let attempt = messageattempt::ActiveModel {
                 response_status_code: Set(0),
                 response: Set("".to_owned()),
                 status: Set(MessageStatus::Fail),
-
                 ..attempt
             };
-            Err((attempt, err))
+            Err((attempt, crate::error::Error::Generic(format!("{}", err))))
```
tasn, line 213: What was wrong with what was there before?

server/svix-server/src/worker.rs
```diff
@@ -193,20 +193,24 @@ async fn dispatch(
...
+                    crate::error::Error::Generic(format!("HTTP STATUS {}", status_code)),
+                )),
+            }
         }
         Err(err) => {
             let attempt = messageattempt::ActiveModel {
                 response_status_code: Set(0),
                 response: Set("".to_owned()),
                 status: Set(MessageStatus::Fail),
-
```
tasn, line 206: Unrelated change.

server/svix-server/src/worker.rs
```diff
@@ -193,20 +193,24 @@ async fn dispatch(
                 response_status_code: Set(status_code),
                 response: Set(body),
                 status: Set(status),
-
```
tasn, line 196: Unrelated change.

server/svix-server/src/worker.rs
```diff
@@ -193,20 +193,24 @@ async fn dispatch(
...
                 ..attempt
             };
-            Ok(attempt)
+            match status {
+                MessageStatus::Success => Ok(attempt),
+                _ => Err((
+                    attempt,
+                    crate::error::Error::Generic(format!("HTTP STATUS {}", status_code)),
```
tasn, line 202: Generic feels wrong. Does it actually need to be a `crate::error::Error`? Why not just a string error?

server/svix-server/src/worker.rs
svix-james: This is fixed -- was using as a placeholder b/c I was struggling w/ getting a reqwest::Error in the http-non-200 scenario.

.gitignore
```diff
@@ -3,4 +3,6 @@
 
 #PHP
 /vendor/
-composer.lock
\ No newline at end of file
+composer.lock
+
+.vscode
```
tasn, line 8: Separate commit please.

server/svix-server/tests/e2e_attempt.rs
```diff
@@ -145,3 +150,130 @@ async fn test_list_attempts_by_endpoint() {
...
+
+#[tokio::test]
+async fn test_successful_attempt() {
+    let _ = dotenv::dotenv();
+    let mut cfg = svix_server::cfg::load().unwrap().as_ref().clone();
+    cfg.queue_type = svix_server::cfg::QueueType::Memory;
+    cfg.retry_schedule = (0..2)
+        .into_iter()
+        .map(|_| Duration::from_millis(1))
+        .collect();
```
tasn, line 232: As you can see, this is a lot of boilerplate. Can you maybe just add a `get_default_test_config` function that gives you a default config that you can use? (it'll also set the queue_type to memory).

server/svix-server/tests/e2e_attempt.rs
```diff
@@ -145,3 +150,130 @@ async fn test_list_attempts_by_endpoint() {
...
+            .get(
+                &format!("api/v1/app/{}/attempt/msg/{}", app_id, &msg.id),
+                StatusCode::OK,
+            )
+            .await
+            .unwrap();
+
+        if list.data.len() != cfg.retry_schedule.len() + 1 {
+            anyhow::bail!("Attempt count does not match retry_schedule length");
+        }
```
tasn, line 191:
```suggestion
        assert_eq!(list.data.len(), cfg.retry_schedule.len() + 1);
```
Use asserts in tests.

server/svix-server/tests/e2e_attempt.rs
```diff
@@ -145,3 +150,130 @@ async fn test_list_attempts_by_endpoint() {
...
+            if i.response_status_code != expected_status {
+                anyhow::bail!(
+                    "HTTP status should be 200, found {}",
+                    i.response_status_code
+                );
+            }
+            if i.endpoint_id != endp_id {
+                anyhow::bail!("Unexpected endpoint id");
+            }
+        }
```
tasn, line 205: This should be outside of the `run_with_retries`. The idea behind that function is to just have a sync point so that we know that we got everything we need at that point and then we can commence normal testing.

server/svix-server/tests/e2e_attempt.rs
```diff
@@ -145,3 +150,130 @@ async fn test_list_attempts_by_endpoint() {
...
+            }
+            if i.response_status_code != expected_status {
+                anyhow::bail!(
+                    "HTTP status should be 200, found {}",
+                    i.response_status_code
+                );
+            }
+            if i.endpoint_id != endp_id {
+                anyhow::bail!("Unexpected endpoint id");
+            }
```
tasn, line 204:
```suggestion
            assert_eq!(i.status, MessageStatus::Fail);
            assert_eq!(i.response_status_code, expected_status);
            assert_eq!(i.endpoint_id, endp_id);
```

Again, use asserts. Using ifs and bails just makes thing opaque. Using asserts will have useful prints when things fail but also much more concise.

server/svix-server/tests/e2e_attempt.rs
```diff
@@ -145,3 +150,130 @@ async fn test_list_attempts_by_endpoint() {
     receiver_1.jh.abort();
     receiver_2.jh.abort();
 }
+
+async fn test_attempt_retries(endpoint: &str, expected_status: i16) {
```
tasn, line 154: Oh, I thought this was a test (obviously it isn't) because of the `test_` prefix. Maybe change that?

server/svix-server/tests/e2e_attempt.rs
```diff
@@ -145,3 +150,130 @@ async fn test_list_attempts_by_endpoint() {
...
+    })
+    .await
+    .unwrap();
+}
+
+#[tokio::test]
+async fn test_attempt_failures_are_retried() {
+    let receiver = TestReceiver::start(axum::http::StatusCode::FORBIDDEN);
+
+    // HTTP non-200 failures:
```
tasn, line 216: Nice, I like it! I wouldn't have asked you to add what I'm just going to normally, but you made the tests really nice so it's actually very easy to add it!
Can you also add one for 500? Just asking as it's another line, so it's awesome. :)

server/svix-server/tests/e2e_attempt.rs
```diff
@@ -145,3 +150,130 @@ async fn test_list_attempts_by_endpoint() {
...
+    run_with_retries(|| async {
+        let list: ListResponse<MessageAttemptOut> = client
+            .get(
+                &format!("api/v1/app/{}/attempt/msg/{}", app_id, &msg.id),
+                StatusCode::OK,
+            )
+            .await
+            .unwrap();
+
+        if list.data.len() != 1 {
```
tasn, line 258: Same comments as for the other function.

server/svix-server/tests/e2e_attempt.rs
```diff
@@ -145,3 +150,130 @@ async fn test_list_attempts_by_endpoint() {
...
+    // HTTP non-200 failures:
+    test_attempt_retries(&receiver.endpoint, 403).await;
+    receiver.jh.abort();
+
+    // non-HTTP-related failures:
+    test_attempt_retries(&receiver.endpoint, 0).await;
+}
+
+#[tokio::test]
+async fn test_successful_attempt() {
```
tasn, line 225: Isn't this exactly almost the same as `test_attempt_retries` but with a 200 response? To be honest, I'm not 100% what's being tested in this test. Just a normal delivery path, no? If so, I'd rather have a different file for this that just tests e2e message delivery.

svix-james: It is indeed just checking for a succesful attempt (ie w/ no retries). I have refactored to remove the mostly-redundant code. This scenario is not previously tested as far as I can tell, so leaving it in.

server/svix-server/tests/e2e_attempt.rs
svix-james: Fixed.

server/svix-server/tests/e2e_attempt.rs
svix-james: Done.

server/svix-server/tests/e2e_attempt.rs
svix-james: Fixed

server/svix-server/tests/e2e_attempt.rs
svix-james: Done

PR #365, opened 2022-04-04
Tests: Recover failed webhooks

jaymell:
Add tests to ensure proper functioning of recover-failed-webhooks endpoint. This will validate that particular failed messages are retried an expected number of times based upon message creation time and the `since` time passed to the endpoint.

### DIFF
--- a/server/svix-server/tests/e2e_attempt.rs
+++ b/server/svix-server/tests/e2e_attempt.rs
@@ -14,12 +14,15 @@ use svix_server::{
 mod utils;
 
 use utils::{
-    common_calls::{create_test_app, create_test_endpoint, create_test_message},
+    common_calls::{
+        create_test_app, create_test_endpoint, create_test_message,
+        get_msg_attempt_list_and_assert_count,
+    },
     get_default_test_config, run_with_retries, start_svix_server, start_svix_server_with_cfg,
     TestReceiver,
 };
 
-use tokio::time::{sleep, Duration};
+use std::time::Duration;
 
 #[tokio::test]
 async fn test_list_attempted_messages() {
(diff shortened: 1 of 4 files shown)


server/svix-server/tests/e2e_attempt.rs
```diff
@@ -152,62 +151,6 @@ async fn test_list_attempts_by_endpoint() {
     receiver_2.jh.abort();
 }
 
-async fn try_message_attempts(
```
tasn, line 155: What's this?

tasn:
Let's have Daniel take a look, but please fix the title. You are not working on recover failed webhooks, but rather adding tests for it.

server/svix-server/tests/utils/common_calls.rs
```diff
@@ -208,3 +214,106 @@ pub async fn common_test_list<
...
+        &endp_id,
+        &client,
+        &cfg.retry_schedule,
+        msg_status,
+        status_code,
+        attempt_count,
+    );
+}
+
+pub async fn try_message_attempts_with_params(
```
svix-daniel, line 250: Nitpicky, but I'd probably add doc comments here -- especially since the `try_message_attempts` and `try_message_attempts_with_params` make assertions.

svix-daniel:
Also non-essential, but I usually like to leave assertions and unwraps in the function marked as `#[test]` so that when they panic, the line number given is actually in the test. A matter of preference though :)

jaymell:
I refactored based on comments and removed the `try_message_attempts` functions entirely since I think they were muddling the point.

server/svix-server/tests/utils/common_calls.rs
jaymell: At the cost of a bit more repetition, I removed these functions entirely.

svix-daniel (approved):

PR #366, opened 2022-04-06
Tests: Add additional endpoint integration tests

jaymell:
Adding additional end-to-end tests to bring rust tests to parity with python test suite.


tasn:
Why is it a draft? Let's just have @svix-daniel review it and get it in?

svix-daniel:
Approved but we should probably look into subdividing up the `e2e_endpoints` module at some point.

PR #367, opened 2022-04-07
Implement idempotency middleware

svix-daniel:
Implements idempotency middleware to take effect on POST request on the server.

### DIFF
--- a/server/svix-server/Cargo.toml
+++ b/server/svix-server/Cargo.toml
@@ -44,6 +44,7 @@ bb8-redis = "0.10.1"
 redis = { version = "0.21.5", features = ["tokio-comp"] }
 thiserror = "1.0.30"
 bytes = "1.1.0"
+blake2 = "0.10.4"
 # sea orm
 sea-orm = { version = "0.7.1", features = [ "sqlx-postgres", "runtime-tokio-rustls", "macros" ], default-features = false }
 sqlx = { version = "0.5", features = [ "runtime-tokio-rustls", "postgres", "migrate" ] }
--- a/server/svix-server/src/core/idempotency.rs
+++ b/server/svix-server/src/core/idempotency.rs
@@ -80,17 +80,26 @@ fn finished_serialized_response_to_reponse(
 }
 
 struct IdempotencyKey(String);
+
 impl IdempotencyKey {
     fn new(auth_token: &str, key: &str, url: &str) -> IdempotencyKey {
         let mut hasher = Blake2b512::new();
-        hasher.update(format!("{}:{}:{}", auth_token, key, url));
+
+        hasher.update(auth_token);
+        hasher.update(":");
+        hasher.update(key);
+        hasher.update(":");
+        hasher.update(url);
+
         let res = hasher.finalize();
         IdempotencyKey(base64::encode(&res))
     }
 }
+
 impl CacheKey for IdempotencyKey {
     const PREFIX_CACHE: &'static str = "SVIX_IDEMPOTENCY_CACHE";
 }
+
 impl AsRef<str> for IdempotencyKey {
     fn as_ref(&self) -> &str {
         &self.0
--- a/server/svix-server/src/core/idempotency.rs
+++ b/server/svix-server/src/core/idempotency.rs
@@ -38,7 +38,7 @@ enum SerializedResponse {
     Start,
     Finished {
         code: u16,
-        headers: Option<HashMap<String, String>>,
+        headers: Option<HashMap<String, Vec<u8>>>,
         body: Option<serde_json::Value>,
     },
 }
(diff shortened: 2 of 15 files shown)


server/svix-server/src/core/idempotency.rs
```diff
@@ -0,0 +1,426 @@
...
+
+/// The data structure containing all necessary components of a response ready to be (de)serialized
+/// from/into the cache
+#[derive(Deserialize, Serialize)]
+enum SerializedResponse {
+    Start,
+    Finished {
+        code: u16,
+        headers: Option<HashMap<String, Vec<u8>>>,
+        body: Option<serde_json::Value>,
```
tasn, line 42: You can't use a json here, it has to be `bytes`, otherwise you may actually return a slightly different response. It's not the end of the world if bytes are much harder, but I believe it's probably not harder?

server/svix-server/src/core/idempotency.rs
```diff
@@ -0,0 +1,426 @@
...
+    }
+
+    Ok(out)
+}
+
+struct IdempotencyKey(String);
+impl IdempotencyKey {
+    fn new(auth_token: &str, key: &str, url: &str) -> IdempotencyKey {
+        let mut hasher = Blake2b512::new();
+        hasher.update(format!("{}:{}:{}", auth_token, key, url));
```
tasn, line 87: I think it's probably more efficient to do:
```suggestion
        hasher.update(auth_token);
        hasher.update(":");
        hasher.update(key);
        ...
```
So no string formatting.

server/svix-server/src/core/idempotency.rs
```diff
@@ -0,0 +1,426 @@
...
+        let res = hasher.finalize();
+        IdempotencyKey(base64::encode(&res))
+    }
+}
+impl CacheKey for IdempotencyKey {
```
tasn, line 92: Nitpick: but the lack of empty lines everywhere is killing me. Let's do one empty line between related stuff, and two between unrelated (e.g. another struct).

server/svix-server/src/core/idempotency.rs
```diff
@@ -0,0 +1,426 @@
+// SPDX-FileCopyrightText: © 2022 Svix Authors
```
tasn, line 1: Also actually use it, and add an integration test that checks it works (by e.g. creating an application multiple times with the same key).

server/svix-server/Cargo.toml
```diff
@@ -44,6 +44,7 @@ bb8-redis = "0.10.1"
 redis = { version = "0.21.5", features = ["tokio-comp"] }
```
tasn, line 44: Maybe also upgrade axum to latest before this PR (if it's not already), just so we can make sure we don't have any issues.

server/svix-server/src/core/idempotency.rs
svix-daniel: `cargo fmt` will reduce two newlines to one, but I'm happy to do one line break consistently

tasn: Yeah, so at least that. :)

server/svix-server/src/core/idempotency.rs
tasn: You haven't fixed it?

server/svix-server/src/v1/endpoints/attempt.rs
```diff
@@ -688,7 +688,7 @@ pub fn router() -> Router {
                     .route("/attempt/", get(list_messageattempts))
                     .route("/attempt/:attempt_id/", get(get_messageattempt))
                     .route("/endpoint/", get(list_attempted_destinations))
-                    .route("/endpoint/:endp_id/resend/", post(resend_webhook))
+                    .route("/endpoint/:endp_id/resent/", post(resend_webhook))
```
tasn, line 691: Hmmm....?

server/svix-server/src/core/idempotency.rs
```diff
@@ -0,0 +1,615 @@
...
+                let auth = if let Some(Ok(auth)) =
+                    parts.headers.get("Authorization").map(|v| v.to_str())
+                {
+                    auth
+                } else {
+                    // No auth token -- pass off to service and do not cache
+                    return resolve_service(service, Request::from_parts(parts, body)).await;
+                };
+
+                let key = if let Some(Ok(key)) =
```
tasn, line 167: Tiny tiny comment, but: move this if above the auth if. The reason is that this if is much more likely to be false and cause an early return, so may as well have it before auth (but after the POST check).

It doesn't really matter much.

server/svix-server/src/core/idempotency.rs
```diff
@@ -0,0 +1,615 @@
...
+                    // No idempotency-key -- pass off to service and do not cache
+                    return resolve_service(service, Request::from_parts(parts, body)).await;
+                };
+
+                let uri = parts.uri.to_string();
+                let key = IdempotencyKey::new(auth, key, &uri);
+
+                // Check Redis now
+                match redis.get::<SerializedResponse>(&key).await {
+                    // If the value is a starting value, the operation is still in progress
```
tasn, line 181: The content of each arm of this match would benefit from being move to its own function the way it is now is very hard to read/follow.

server/svix-server/src/core/idempotency.rs
```diff
@@ -0,0 +1,688 @@
...
+                // normally in the event a key could not be created.
+                let key = if let Some(key) = get_key(&parts) {
+                    key
+                } else {
+                    return resolve_service(service, Request::from_parts(parts, body)).await;
+                };
+
+                // Set the [`SerializedResponse::Start`] lock if the key does not exist in the cache
+                // returning whether the value was set
+                let set = if let Ok(set) = redis
```
tasn, line 176: It took me multiple readings of this code and the function to understand that this variable means `lock_acquired`. Please rename it as such.

server/svix-server/src/core/idempotency.rs
```diff
@@ -0,0 +1,688 @@
...
+                                // as normal, but return 500 if the lock cannot be set
+                                if !matches!(
+                                    redis
+                                        .set_if_not_exists(
+                                            &key,
+                                            &SerializedResponse::Start,
+                                            expiry_starting(),
+                                        )
+                                        .await,
+                                    Ok(true)
```
tasn, line 229: This is not actually used and confused me a bit, but I guess it's better than not having it to indicate this path continues.

tasn (approved):

PR #370, opened 2022-04-08
Rust Webhook Signature

jaymell:
Add basic Rust Webhook implementation

### DIFF
--- a/server/svix-server/src/webhook.rs
+++ b/server/svix-server/src/webhook.rs
@@ -1,0 +1,324 @@
+// SPDX-FileCopyrightText: © 2022 Svix Authors
+// SPDX-License-Identifier: MIT
+
+use http::HeaderMap;
+use time::OffsetDateTime;
+
+#[derive(thiserror::Error, Debug)]
+pub enum WebhookError {
+    #[error("failed to parse timestamp")]
+    InvalidTimestamp,
+
+    #[error("invalid secret")]
+    InvalidSecret(#[from] base64::DecodeError),
+
+    #[error("invalid header {0}")]
+    InvalidHeader(&'static str),
+
+    #[error("signature timestamp too old")]
+    TimestampTooOldError,
+
+    #[error("signature timestamp too far in future")]
+    FutureTimestampError,
+
+    #[error("missing header {0}")]
+    MissingHeader(&'static str),
+
+    #[error("signature invalid")]
+    InvalidSignature,
+
+    #[error("payload invalid")]
+    InvalidPayload,
+}
+
+pub struct Webhook {
+    key: Vec<u8>,
+}
+
+const PREFIX: &str = "whsec_";
+const SVIX_MSG_ID_KEY: &str = "svix-id";
+const SVIX_MSG_SIGNATURE_KEY: &str = "svix-signature";
+const SVIX_MSG_TIMESTAMP_KEY: &str = "svix-timestamp";
+const UNBRANDED_MSG_ID_KEY: &str = "webhook-id";
+const UNBRANDED_MSG_SIGNATURE_KEY: &str = "webhook-signature";
+const UNBRANDED_MSG_TIMESTAMP_KEY: &str = "webhook-timestamp";
+const TOLERANCE_IN_SECONDS: i64 = 5 * 60;
+const SIGNATURE_VERSION: &str = "v1";
+
+impl Webhook {
+    pub fn new(secret: String) -> Result<Self, WebhookError> {
+        let secret = secret.strip_prefix(PREFIX).unwrap_or(&secret);
+        let key = base64::decode(secret)?;
+
+        Ok(Webhook { key })
+    }
+
+    pub fn verify(&self, payload: &[u8], headers: &HeaderMap) -> Result<(), WebhookError> {
+        let msg_id = Self::get_header(headers, SVIX_MSG_ID_KEY, UNBRANDED_MSG_ID_KEY, "id")?;
+        let msg_signature = Self::get_header(
+            headers,
+            SVIX_MSG_SIGNATURE_KEY,
+            UNBRANDED_MSG_SIGNATURE_KEY,
+            "signature",
+        )?;
+        let msg_ts = Self::get_header(
+            headers,
+            SVIX_MSG_TIMESTAMP_KEY,
+            UNBRANDED_MSG_TIMESTAMP_KEY,
+            "timestamp",
+        )
+        .and_then(Self::parse_timestamp)?;
+
+        Self::verify_timestamp(msg_ts)?;
+
+        let versioned_signature = self.sign(msg_id, msg_ts, payload)?;
+        let expected_signature = versioned_signature
+            .split_once(',')
+            .map(|x| x.1)
+            .ok_or(WebhookError::InvalidSignature)?;
+
+        msg_signature
+            .split(' ')
+            .filter_map(|x| x.split_once(','))
+            .filter(|x| x.0 == SIGNATURE_VERSION)
+            .any(|x| {
+                x.1.bytes()
+                    .zip(expected_signature.bytes())
+                    .fold(0, |acc, (a, b)| acc | (a ^ b))
+                    == 0
+            })
+            .then(|| ())
+            .ok_or(WebhookError::InvalidSignature)
+    }
+
+    pub fn sign(
+        &self,
+        msg_id: &str,
+        timestamp: i64,
+        payload: &[u8],
+    ) -> Result<String, WebhookError> {
+        let payload = std::str::from_utf8(payload).map_err(|_| WebhookError::InvalidPayload)?;
+        let to_sign = format!("{}.{}.{}", msg_id, timestamp, payload,);
+        let signed = hmac_sha256::HMAC::mac(to_sign.as_bytes(), &self.key);
+        let encoded = base64::encode(signed);
+
+        Ok(format!("{},{}", SIGNATURE_VERSION, encoded))
+    }
+
+    fn get_header<'a>(
+        headers: &'a HeaderMap,
+        svix_hdr: &'static str,
+        unbranded_hdr: &'static str,
+        err_name: &'static str,
+    ) -> Result<&'a str, WebhookError> {
+        headers
+            .get(svix_hdr)
+            .or_else(|| headers.get(unbranded_hdr))
+            .ok_or_else(|| WebhookError::MissingHeader(err_name))?
+            .to_str()
+            .map_err(|_| WebhookError::InvalidHeader(err_name))
+    }
+
+    fn parse_timestamp(hdr: &str) -> Result<i64, WebhookError> {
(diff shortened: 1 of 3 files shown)


server/svix-server/src/webhook.rs
```diff
@@ -0,0 +1,298 @@
+// SPDX-FileCopyrightText: © 2022 Svix Authors
+// SPDX-License-Identifier: MIT
+
+use chrono::Utc;
```
tasn, line 4: Use `time` not `chrono`. The latter is deprecated.

server/svix-server/src/webhook.rs
```diff
@@ -0,0 +1,298 @@
...
+use http::{HeaderMap, HeaderValue};
+
+#[derive(thiserror::Error, Debug)]
```
tasn, line 7: This structure is going to be exposed in our rust library. I don't know whether using `thiserror` is a good idea in a library or not. Let's keep it this way for now unless you or @svix-daniel have any thoughts.

server/svix-server/src/webhook.rs
```diff
@@ -0,0 +1,298 @@
+// SPDX-FileCopyrightText: © 2022 Svix Authors
```
tasn, line 1: We are going to move it to its own library (like the other libs) after you're done, so just keep that in mind when you design things.

server/svix-server/src/webhook.rs
```diff
@@ -0,0 +1,298 @@
...
+
+pub(crate) const PREFIX: &str = "whsec_";
+pub(crate) const SVIX_MSG_ID_KEY: &str = "svix-id";
+pub(crate) const SVIX_MSG_SIGNATURE_KEY: &str = "svix-signature";
+pub(crate) const SVIX_MSG_TIMESTAMP_KEY: &str = "svix-timestamp";
+pub(crate) const UNBRANDED_MSG_ID_KEY: &str = "webhook-id";
+pub(crate) const UNBRANDED_MSG_SIGNATURE_KEY: &str = "webhook-signature";
+pub(crate) const UNBRANDED_MSG_TIMESTAMP_KEY: &str = "webhook-timestamp";
+pub(crate) const TOLERANCE_IN_SECONDS: i64 = 5 * 60;
+pub(crate) const SIGNATURE_VERSION: &str = "v1";
```
tasn, line 43: Should just be private, not even pub-crate.

server/svix-server/src/webhook.rs
```diff
@@ -0,0 +1,298 @@
...
+                msg_id = headers.get(UNBRANDED_MSG_ID_KEY);
+                msg_signature = headers.get(UNBRANDED_MSG_SIGNATURE_KEY);
+                msg_ts = headers.get(UNBRANDED_MSG_TIMESTAMP_KEY);
+
+                if msg_id.is_none() || msg_signature.is_none() || msg_ts.is_none() {
+                    return Err(WebhookError::MissingHeaders);
+                }
+            }
+            (msg_id.unwrap(), msg_signature.unwrap(), msg_ts.unwrap())
+        };
```
tasn, line 69:
```suggestion
        let msg_id = headers.get(SVIX_MSG_ID_KEY).unwrap_or_else(|| headers.get(UNBRANDED_MSG_ID_KEY)).ok_or_else(|| Error::MissingHeader("ID"))?;
        let msg_signature = headers.get(SVIX_MSG_SIGNATURE_KEY).unwrap_or_else(|| headers.get(UNBRANDED_MSG_SIGNATURE_KEY)).ok_or_else(|| Error::MissingHeader("signature"))?;
        let msg_ts = headers.get(SVIX_MSG_TIMESTAMP_KEY).unwrap_or_else(|| headers.get(UNBRANDED_MSG_TIMESTAMP_KEY)).ok_or_else(|| Error::MissingHeader("timestamp"))?;
```

1. `mut` is a code smell. Don't assign, transform.
2. Don't unwrap, even if you're sure it's fine (other than in rare situations), it's a source for potential trouble.
3. Be specific in the error message regarding which header is missing. You don't have to be so explicit as I made it, you can just have it be `MissingHeaders("signature")` which I think is better anyway.

server/svix-server/src/webhook.rs
```diff
@@ -0,0 +1,298 @@
...
+
+        let msg_ts = Self::parse_timestamp(msg_ts)?;
+        Self::verify_timestamp(msg_ts)?;
+
+        let msg_id = msg_id.to_str()?;
+        let versioned_signature = self.sign(msg_id, msg_ts, payload)?;
+        let expected_signature = versioned_signature
+            .split_once(',')
+            .expect("Unexpected failure splitting signature")
```
tasn, line 78: Again, better to not panic, but rather return an error to caller.

server/svix-server/src/webhook.rs
```diff
@@ -0,0 +1,298 @@
...
+        &self,
+        msg_id: &str,
+        timestamp: i64,
+        payload: &[u8],
+    ) -> Result<String, WebhookError> {
+        let to_sign = format!(
+            "{}.{}.{}",
+            msg_id,
+            timestamp,
+            std::str::from_utf8(payload).expect("Payload not UTF-8")
```
tasn, line 101: Don't panic, return an error.

tasn (commented):
Made a few comments, looks good otherwise.

Oh, and please also call the PR `Webhook signature` and not just `Webhook`.

server/svix-server/src/webhook.rs
svix-daniel: `thiserror` is fine for libs in my opinion.

Something to note is that It's Rust convention to have errors in lowercase (minus proper nouns) so that it plays well with others in the `Debug` display.

server/svix-server/src/webhook.rs
jaymell: This is different from how the other libraries work, though. The java/go libs, for example, are explicit about checking for the presence/absence of all three of the headers for each flavor, i.e., not allowing mixing/matching of branded/unbranded headers, which is why I didn't take the approach you outlined.

jaymell: Which header should I report as missing in the error -- Svix or unbranded?

tasn: Just say "signature", you don't have to be explicit.

server/svix-server/src/webhook.rs
```diff
@@ -0,0 +1,319 @@
...
+    InvalidHeader(String),
+
+    #[error("signature timestamp too old")]
+    TimestampTooOldError,
+
+    #[error("signature timestamp too far in future")]
+    FutureTimestampError,
+
+    #[error("missing header {0}")]
+    MissingHeader(String),
```
tasn, line 25:
```suggestion
    MissingHeader(&'static str),
```

server/svix-server/src/webhook.rs
```diff
@@ -0,0 +1,319 @@
...
+}
+
+const PREFIX: &str = "whsec_";
+const SVIX_MSG_ID_KEY: &str = "svix-id";
+const SVIX_MSG_SIGNATURE_KEY: &str = "svix-signature";
+const SVIX_MSG_TIMESTAMP_KEY: &str = "svix-timestamp";
+const UNBRANDED_MSG_ID_KEY: &str = "webhook-id";
+const UNBRANDED_MSG_SIGNATURE_KEY: &str = "webhook-signature";
+const UNBRANDED_MSG_TIMESTAMP_KEY: &str = "webhook-timestamp";
+const ID: &str = "id";
```
tasn, line 45: Don't make it a const (the error messages). It's used once and should only be used once. Splitting it like this is confusing.

server/svix-server/src/webhook.rs
```diff
@@ -0,0 +1,319 @@
...
+        let msg_ts = headers
+            .get(SVIX_MSG_TIMESTAMP_KEY)
+            .or_else(|| headers.get(UNBRANDED_MSG_TIMESTAMP_KEY))
+            .ok_or_else(|| WebhookError::MissingHeader(TIMESTAMP.to_string()))?;
+
+        let msg_ts = Self::parse_timestamp(msg_ts)?;
+        Self::verify_timestamp(msg_ts)?;
+
+        let msg_id = msg_id
+            .to_str()
```
tasn, line 79: Move it to be part of the unwrapping above.

server/svix-server/src/webhook.rs
```diff
@@ -0,0 +1,319 @@
...
+            .map(|x| x.1)
+            .ok_or(WebhookError::InvalidSignature)?;
+
+        msg_signature
+            .to_str()
+            .map_err(|_| WebhookError::InvalidHeader(SIGNATURE.to_string()))?
+            .split(' ')
+            .filter_map(|x| x.split_once(','))
+            .filter(|x| x.0 == SIGNATURE_VERSION)
+            .any(|x| x.1 == expected_signature)
```
tasn, line 94: Use a constant time comparison function (google for one for rust).

server/svix-server/src/webhook.rs
```diff
@@ -0,0 +1,319 @@
...
+        let expected_signature = versioned_signature
+            .split_once(',')
+            .map(|x| x.1)
+            .ok_or(WebhookError::InvalidSignature)?;
+
+        msg_signature
+            .to_str()
+            .map_err(|_| WebhookError::InvalidHeader(SIGNATURE.to_string()))?
+            .split(' ')
+            .filter_map(|x| x.split_once(','))
```
tasn, line 92: Just map, no?

server/svix-server/src/webhook.rs
```diff
@@ -0,0 +1,319 @@
...
+
+impl Webhook {
+    pub fn new(secret: String) -> Result<Self, WebhookError> {
+        let secret = secret.strip_prefix(PREFIX).unwrap_or(&secret);
+        let key = base64::decode(secret)?;
+
+        Ok(Webhook { key })
+    }
+
+    pub fn verify(&self, payload: &[u8], headers: &HeaderMap) -> Result<(), WebhookError> {
```
tasn, line 59: Verify should return the payload (like in other libs).

jaymell: Will add -- note that it's inconsistent across libs. I was consulting java and go, neither of which return anything.

server/svix-server/src/webhook.rs
jaymell: `split_once` returns an Option, so `filter_map` will filter out `None` at same time.

server/svix-server/src/webhook.rs
tasn: Oh right, we only return it if we parse. OK, don't change it.

server/svix-server/src/webhook.rs
tasn: Ah oops, OK. :)

tasn (dismissed):

tasn (approved):

PR #373, opened 2022-04-10
Python libs: fix user agent to actually work.

tasn:


svix-james (approved):

tasn:
Grr, it dismissed it. :(((

PR #376, opened 2022-04-10
Endpoint key-rotation integration test

jaymell:
Test to validate that endpoint key rotation works as expected.

### DIFF
--- a/server/svix-server/tests/e2e_endpoint.rs
+++ b/server/svix-server/tests/e2e_endpoint.rs
@@ -20,6 +20,7 @@ use svix_server::{
         },
         utils::ListResponse,
     },
+    webhook::Webhook,
 };
 
 mod utils;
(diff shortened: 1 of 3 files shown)


server/svix-server/tests/e2e_endpoint.rs
```diff
@@ -553,7 +554,80 @@ async fn test_endpoint_rotate_max() {
...
+        .get(
+            &format!("api/v1/app/{}/endpoint/{}/secret/", app_id, endp.id),
+            StatusCode::OK,
+        )
+        .await
+        .unwrap();
+
+    assert_eq!(secret3_key, secret3.key);
+
+    let raw_payload = r#"{"test":"data1"}"#;
```
tasn, line 617: Just use the payload from the receiver. This can potentially break if serialization style fails. You can make it slightly more robust by making it into a `json!()` and then serializing it to a string and using that.

tasn (commented):
Made one small comment, looks great otherwise!

server/svix-server/tests/e2e_endpoint.rs
```diff
@@ -553,7 +554,81 @@ async fn test_endpoint_rotate_max() {
...
+    let payload = serde_json::from_str(raw_payload).unwrap();
+    let _msg = create_test_message(&client, &app_id, payload)
+        .await
+        .unwrap();
+
+    let last_headers = receiver.header_recv.recv().await.unwrap();
+    let last_body = receiver.data_recv.recv().await.unwrap().to_string();
```
tasn, line 624: Why convert to string if you are converting to bytes again after?

jaymell: Because I dont think you can go from serde_json::Value directly to bytes...let me investigate further

tasn: Ahhh, it's a json value? OK, all good. Approving.

tasn (approved):

PR #377, opened 2022-04-11
Add EndpointSecret Validation

jaymell:
Ensure that incoming endpoint secrets are validated; this also fixes a possible panic in deserialization when an invalid secret shorter than the key prefix is submitted.

### DIFF
--- a/server/svix-server/src/core/types.rs
+++ b/server/svix-server/src/core/types.rs
@@ -450,13 +450,39 @@ impl<'de> Deserialize<'de> for EndpointSecret {
         use serde::de::Error;
         String::deserialize(deserializer)
             .and_then(|string| {
-                base64::decode(&string[Self::PREFIX.len()..])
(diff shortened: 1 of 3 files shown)


jaymell:
This works, although it would be nice to not have to re-encode for validation

server/svix-server/src/core/types.rs
```diff
@@ -450,13 +450,39 @@ impl<'de> Deserialize<'de> for EndpointSecret {
...
+
+        if errors.is_empty() {
+            Ok(())
+        } else {
+            Err(errors)
+        }
+    }
+}
 impl From<EndpointSecret> for sea_orm::Value {
     fn from(v: EndpointSecret) -> Self {
```
tasn, line 487: Can we add a basic unit test for this in this file?

tasn (commented):
Looks great, one small comment.

tasn (approved):

PR #378, opened 2022-04-11
Python: set a 15s read timeout.

tasn:
Set the timeout to 10s, and fixed the type warning when instantiating the client.


tasn:
Redoing #374, I accidentally merged it into an existing branch and dropped everything. :|

svix-daniel (approved):

svix-daniel (approved):

PR #381, opened 2022-04-12
Validate input strings against control sequences

svix-dylan:
## Motivation

Validate that incoming strings do not contain control sequences.

## Solution

A custom Validator function was created in utils.rs that checks strings against the regex `r"[\x00-\x08]"`, and implemented on request struct fields of type String.


svix-james:
`EventTypeUpdate.description` should be validated as well

svix-daniel (approved):

PR #385, opened 2022-04-12
Rust: Streamline deserialization/validation of endpoint secrets

jaymell:


server/svix-server/src/core/types.rs
```diff
@@ -463,15 +463,11 @@ impl<'de> Deserialize<'de> for EndpointSecret {
 
 impl Validate for EndpointSecret {
     fn validate(&self) -> std::result::Result<(), ValidationErrors> {
-        let re = format!(
```
tasn, line 466: I don't fully understand what's going on here, but it looks like you're changing the code itself, not just the test (so commit message is wrong), but also unclear why you're making the changes.

server/svix-server/tests/e2e_endpoint.rs
```diff
@@ -669,16 +669,15 @@ async fn test_custom_endpoint_secret() {
 }
 
 #[tokio::test]
-#[ignore]
 async fn test_invalid_endpoint_secret() {
     let (client, _jh) = start_svix_server();
 
     let app_id = create_test_app(&client, "app1").await.unwrap().id;
 
-    let secret_invalid_prefix = "hwsec_C2FVsBQIhrscChlQIMV+b5sSYspob7oD".to_owned();
```
tasn, line 678: Why remove the invalid prefix test?

server/svix-server/src/core/types.rs
jaymell: Since the secret is converted to a `Vec<u8>` as part of deserialization, the regex validation of the key prefix was completely artificial.

server/svix-server/tests/e2e_endpoint.rs
jaymell: It was removed because of the way deserialization currently works. It doesn't validate the prefix but rather just strips off the length of the prefix from the incoming secret and decodes the rest of it. We could potentially add an additional equality check of the prefix itself during deserialization.

tasn (approved):

jaymell:
I've re-tooled deserialization and validation per comments and other discussion.

PR #386, opened 2022-04-13
Fix most Clippy pedantic lints in E2E tests

svix-daniel:


server/svix-server/tests/e2e_application.rs
```diff
@@ -234,7 +266,7 @@ async fn test_uid() {
             ApplicationIn {
                 name: "App 2".to_owned(),
                 uid: Some(ApplicationUid("app1".to_owned())),
-                ..Default::default()
+                ..ApplicationIn::default()
```
tasn, line 269: I'm not sure I prefer this.

tasn:
Let's not do this for now. I think pedantic does mostly undesirable things. I'm happy with the current linting.

PR #387, opened 2022-04-13
Rust: Refactor `test_recovery_expected_retry_counts` integration test

jaymell:
This should improve test stability by validating the number of msg
attempts before recording `since` times rather than simply sleeping
for an assumed-to-be-sufficient duration

### DIFF
--- a/server/svix-server/tests/e2e_endpoint.rs
+++ b/server/svix-server/tests/e2e_endpoint.rs
@@ -7,13 +7,13 @@ use std::{
 };
 
 use anyhow::Result;
-use chrono::{DateTime, Utc};
+use chrono::Utc;
 use reqwest::StatusCode;
 
 use svix_server::{
     core::types::{
-        ApplicationId, EndpointHeaders, EndpointId, EndpointSecret, EndpointUid, EventChannel,
-        EventChannelSet, EventTypeName, EventTypeNameSet, ExpiringSigningKeys,
+        ApplicationId, EndpointHeaders, EndpointSecret, EndpointUid, EventChannel, EventChannelSet,
+        EventTypeName, EventTypeNameSet, ExpiringSigningKeys,
     },
     v1::{
         endpoints::{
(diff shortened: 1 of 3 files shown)


server/svix-server/tests/e2e_endpoint.rs
```diff
@@ -507,7 +497,7 @@ async fn test_recovery_expected_retry_counts() {
         )
         .await;
 
-        sleep(Duration::from_millis(50)).await;
+        sleep(Duration::from_millis(500)).await;
```
tasn, line 500: Why not use run_with_retries?

server/svix-server/tests/e2e_endpoint.rs
```diff
@@ -425,80 +425,70 @@ async fn test_recovery_should_fail_if_start_time_too_old() {
...
+    for (i, (msg_1_retry_cnt, msg2_retry_cnt)) in [
+        // expected number of additional retry attempts for (msg1, msg2) if recover `since` is before msg 1
+        // (i.e., both messages will have one additional attempt after recovery):
+        (1, 1),
+        // expected values if recover `since` is before msg 2:
+        // (i.e., msg 2 will have one additional attempt after recovery):
+        (0, 1),
+        // expected values if recover `since` is after msg 2:
+        // (i.e., neither msg will have additional attempt after recovery):
+        (0, 0),
```
tasn, line 449: This part is very confusing and is obfuscating what's actually going on. If it's just for filtering the returned messages, why not literally just have the same create app, endpoint, msg sequence, and then just manually three `list` calls with different since. This is just very confusing and doesn't actually save any code/effort.

server/svix-server/tests/e2e_endpoint.rs
```diff
@@ -425,80 +425,70 @@ async fn test_recovery_should_fail_if_start_time_too_old() {
...
-    wait_for_msg_retries(&cfg.retry_schedule).await;
+        let before_msg_2 = Utc::now();
 
-    let after_msg_2 = Utc::now();
+        let msg_2 = create_test_message(&client, &app_id, serde_json::json!({"test": "data2"}))
+            .await
+            .unwrap();
 
-    receiver.jh.abort();
+        sleep(Duration::from_millis(100)).await;
```
tasn, line 483: Sleeps are not a good idea.

server/svix-server/tests/e2e_endpoint.rs
jaymell: The alternative is to increase number of retries and/or delay w/in `run_with_retries`, since the current fn by itself is not always enough.

tasn (approved):

PR #390, opened 2022-04-15
Rust: Support clustered redis

jaymell:


jaymell:
The main difficulty is that the `ManageConnection` trait cannot easily be handled with dynamic dispatch because its associated `Connection` type needs to be generic.

PR #392, opened 2022-04-21
Rust: Support clustered redis

jaymell:


tasn:
1. Branch conflicts.
2. There's no need to close a PR and open another, you can just push to the original.

server/svix-server/src/cfg.rs
```diff
@@ -60,7 +77,7 @@ retry_schedule = "5,300,1800,7200,18000,36000,36000"
 # The DSN for redis (can be left empty if not using redis)
 # redis_dsn = "redis://redis:6379"
 
-# What kind of message queue to use. Supported: memory, redis (must have redis_dsn configured).
+# What kind of message queue to use. Supported: memory, redis (must have redis_dsn configured), rediscluster (add multiple hosts to redis_dsn separated by commas).
```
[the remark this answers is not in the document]
jaymell: I still feel like we should support multiple hostnames. It's a trivial change, and perhaps some users would want to specify multiple in case a server is down.

server/svix-server/src/core/message_app.rs
```diff
@@ -4,6 +4,7 @@ use std::{
 };
 
 use chrono::{DateTime, FixedOffset};
+
```
svix-dylan, line 7: Nitpick but we can remove this newline and keep this file out of Changes

svix-dylan (commented):
I see some notes for things still being cleaned up but this looks good! Between this and #393 we'll have some conflicts at first, but they should be pretty easy to sort out.

server/svix-server/src/core/cache.rs
```diff
@@ -54,21 +52,26 @@ macro_rules! kv_def {
...
 pub(crate) use kv_def;
 
+use crate::redis::SvixConnectionManager;
+
 /// A Redis-based cache of data to avoid expensive fetches from PostgreSQL. Simply a wrapper over
 /// Redis.
 #[derive(Debug, Clone)]
 pub struct RedisCache {
-    redis: Pool<RedisConnectionManager>,
+    redis: SvixConnectionManager,
```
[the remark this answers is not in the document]
jaymell: `ConnectionManager` was completely wrong here, have since changed to `SvixRedisPool`

server/svix-server/src/cfg.rs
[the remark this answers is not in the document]
tasn: That's fair. Though please clarify the doc then to clarify that you should only pass the main one under normal circumstances and only pass multiple if you want.

Though to be honest, since it's such a trivial change, I'd rather we waited until a requirement comes from an actual customer before we do it. What do you think?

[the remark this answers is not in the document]
jaymell: That's fine. I do want to clarify the request here, though. Currently there are 2 configuration changes in this PR:
1. Change `redis_dsn` parameter to support multiple hosts
2. Add optional `cache_redis_dsn` and `queue_redis_dsn`, which also support multiple hosts.

Do you want to leave in support for 2) and just remove support for multiple hosts, or or remove everything and only support `redis_dsn` with a single host for now?

svix-dylan (commented):
Noting it looks like you still intend to put redis_cache type determination in a helper method? Re: FIXMEs in lib.rs, idempotency.rs, cache.rs.

server/svix-server/src/cfg.rs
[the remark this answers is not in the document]
tasn: I'd say: remove everything and only support redis_dsn with a single host.

Follow up PR should allow overriding just one of the redis's, but don't think it's related to this PR.

svix-dylan (approved):

server/svix-server/src/core/cache/redis.rs
```diff
@@ -112,12 +116,23 @@ mod tests {
...
+                    crate::redis::create_redis_pool(cfg.redis_dsn.as_ref().unwrap().as_str(), true)
+                        .await;
+                Some(mgr)
+            }
+            _ => {
+                let mgr = crate::redis::create_redis_pool(
+                    cfg.redis_dsn.as_ref().unwrap().as_str(),
+                    false,
+                )
+                .await;
```
tasn, line 131: Using true/false for this is very confusing. I actually missed it the first two times around. :|

server/svix-server/src/core/cache/redis.rs
```diff
@@ -169,11 +184,24 @@ mod tests {
...
+                Some(mgr)
+            }
+        };
```
tasn, line 202: This init code is duplicated many times over, probably better in a function.

server/svix-server/src/lib.rs
```diff
@@ -26,49 +26,52 @@ pub mod core;
...
-                panic!("Choosing queue_type=Redis requires setting a redis DSN.")
-            }))
-            .await
-        }
+    let redis_dsn = || {
+        cfg.redis_dsn
+            .as_ref()
+            .expect("Redis DSN not found")
+            .as_str()
     };
```
tasn, line 42: This is a bit weird, why have it as a function if it's just borrow in both places?

server/svix-server/src/queue/redis.rs
```diff
@@ -34,16 +36,15 @@ use super::{
 
 // FIXME: Change unwraps to have our own error type for the queue module entirely
 
-const MAIN: &str = "svix_queue_main";
-const PROCESSING: &str = "svix_queue_processing";
-const DELAYED: &str = "svix_queue_delayed";
+// FIXME: hash-tags here enable clustering but prevent proper sharding of lists
+const MAIN: &str = "{svix_queue}_main";
+const PROCESSING: &str = "{svix_queue}_processing";
+const DELAYED: &str = "{svix_queue}_delayed";
```
tasn, line 42: This change is quite problematic, because you may have items in the queue that you will now won't be pulling due to this (in the non-clustered case).

I'm talking about someone using Svix for a few weeks now, and now upgrading.

server/svix-server/src/queue/redis.rs
```diff
@@ -144,11 +165,14 @@ impl TaskQueueSend for RedisQueueProducer {
         let key = to_redis_key(&delivery);
         if let Some(timestamp) = timestamp {
             let _: () = pool
-                .zadd(DELAYED, key, timestamp.timestamp())
```
tasn, line 147: Looking at the worst experience here everywhere by using `redis::Cmd::zadd` instead of `pool.zadd`, I wonder if we should have just had our own wrappers for these functions instead of making the experience worst everywhere.

Can probably easily just implement it with a macro.

server/run-tests.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/sh -e
+
+# Run tests with various configurations:
+
+TEST_COMMAND="cargo test --all --all-features --all-targets -- --test-threads 1"
```
tasn, line 5: We don't want to limit to a single thread.

server/svix-server/src/queue/redis.rs
```diff
@@ -183,20 +213,21 @@ impl TaskQueueSend for RedisQueueProducer {
 }
 
 pub struct RedisQueueConsumer {
-    pool: Pool<RedisConnectionManager>,
+    pool: SvixRedisPool,
```
tasn, line 216: Calling things `SvixRedisPool` rather than `RedisPool` feels redundant.

Nitpick ^

server/svix-server/src/core/cache/redis.rs
jaymell: The reason it's true/false is b/c there are 2 different config enums, `CacheType` and `QueueType` for which this is applicable and wanted to easily reuse this.

server/svix-server/src/lib.rs
jaymell: Putting the `expect` in one place instead of in 4 different spots was the goal.

server/svix-server/src/core/cache/redis.rs
tasn: Yeah, I understand that, though having two functions with different names (even if they call the same private function internally that uses true/false) makes things much clearer. Here I easily missed it.

server/svix-server/src/lib.rs
tasn: Why not have:
```suggestion
    let redis_dsn =
        cfg.redis_dsn
            .as_ref()
            .expect("Redis DSN not found")
            .as_str()
    ;
```
This puts it just one in one place, no?

jaymell: Because this will panic if `redis_dsn` is not defined.

server/svix-server/src/queue/redis.rs
jaymell: Migrations should address this.

PR #393, opened 2022-04-22
Cache abstraction and MemoryCache implementation

svix-dylan:
Abstracts the cache module to support additional stores using enum_dispatch
for polymorphism, and implements a new store MemoryCache.

Drafting this PR to get general feedback, but also specifically regarding:
- MemoryCache's TTL implementation
- Idempotency integrity


server/svix-server/src/cfg.rs
```diff
@@ -128,6 +131,14 @@ pub enum QueueType {
     Redis,
 }
 
+#[derive(Clone, Debug, Deserialize)]
+#[serde(rename_all = "lowercase")]
+pub enum CacheType {
+    Memory,
+    Redis,
+    None,
```
tasn, line 139: Should we print a warning when people choose `None`? Something like "running with caching disabled which will negatively affect performance and cause things like idempotency not to work."?

server/svix-server/src/core/cache/redis.rs
```diff
@@ -0,0 +1,83 @@
+// SPDX-FileCopyrightText: © 2022 Svix Authors
```
tasn, line 1: Poor @svix-james is going to have clashes. 😢

server/svix-server/src/core/cache/memory.rs
```diff
@@ -0,0 +1,113 @@
...
+type State = HashMap<String, ValueWrapper>;
+type SharedState = Arc<Mutex<State>>;
+
+pub fn new() -> Cache {
+    let shared_state = Arc::new(Mutex::new(State::new()));
+
+    let shared_state_clone = shared_state.clone();
+    task::spawn(async move {
+        loop {
+            sleep(Duration::from_secs(30)).await;
```
tasn, line 42: Once every 5 minutes is probably sufficient.

server/svix-server/src/core/cache/memory.rs
```diff
@@ -0,0 +1,113 @@
...
+        ValueWrapper {
+            value,
+            ttl,
+            timer: Instant::now(),
+        }
+    }
+}
+
+type State = HashMap<String, ValueWrapper>;
+type SharedState = Arc<Mutex<State>>;
```
tasn, line 34: Use an `RwLock` not a mutex. We read from the cache often, we don't want to be locking everything when just reading.

I also wonder whether there's an atomic way of reading/writing so that a lock needs only to be acquired on writes. Not the end of the world, just a thought.

tasn (commented):
Made some comments.

server/svix-server/src/core/idempotency.rs
```diff
@@ -87,6 +87,7 @@ fn finished_serialized_response_to_reponse(
     Ok(out)
 }
 
+#[derive(Clone)]
```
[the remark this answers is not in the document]
svix-dylan: Leftover artifact from debugging; removed.

.github/workflows/server-ci.yml
```diff
@@ -64,6 +64,7 @@ jobs:
         SVIX_REDIS_DSN: "redis://localhost:6379"
         SVIX_QUEUE_TYPE: "redis"
         SVIX_JWT_SECRET: "test value"
+        SVIX_CACHE_TYPE: "memory"
```
svix-james, line 67: Should this be our default testing approach? I wonder if we should run tests twice for now in CI with `redis` specified as well until we have a better strategy for rolling through the various combinations.

jaymell: Resolving this for now, will take alternate approach to testing in future PR

svix-james:
As discussed elsewhere, I think we should consider disabling idempotency entirely if cache type is `None`.

jaymell (approved):

svix-james (approved):

PR #394, opened 2022-04-27
Multiple Commits Addressing Prior Comments

jaymell:
Addresses additional comments from #392.

### DIFF
--- a/server/run-tests.sh
+++ b/server/run-tests.sh
@@ -2,7 +2,7 @@
 
 # Run tests with various configurations:
 
-TEST_COMMAND="cargo test --all --all-features --all-targets -- --test-threads 1"
+TEST_COMMAND="cargo test --all --all-features --all-targets"
 
 # Common variables:
 export DATABASE_URL="postgresql://postgres:postgres@localhost:5432/postgres"
--- a/server/svix-server/src/queue/redis.rs
+++ b/server/svix-server/src/queue/redis.rs
@@ -37,9 +37,9 @@ use super::{
 // FIXME: Change unwraps to have our own error type for the queue module entirely
 
 // FIXME: hash-tags here enable clustering but prevent proper sharding of lists
-const MAIN: &str = "{svix_queue}_main";
-const PROCESSING: &str = "{svix_queue}_processing";
-const DELAYED: &str = "{svix_queue}_delayed";
+const MAIN: &str = "{queue}_svix_main";
+const PROCESSING: &str = "{queue}_svix_processing";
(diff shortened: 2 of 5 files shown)


server/run-tests.sh
```diff
@@ -2,7 +2,7 @@
```
tasn, line 2: The PR's title/description don't match a lot of the content.

jaymell: Should I broke this into separate PRs with titles/descriptions specific to the various changes?

tasn: You can have this PR as "fix the previous PR", but then have the commits separate and isolated doing their own thing, and change the title/description of the PR to say that this is a set of commits.

server/svix-server/src/queue/redis.rs
```diff
@@ -41,70 +41,60 @@ const MAIN: &str = "{svix_queue}_main";
 const PROCESSING: &str = "{svix_queue}_processing";
 const DELAYED: &str = "{svix_queue}_delayed";
```
tasn, line 42: Nitpick, but I'd change these to be: `{queue}_svix_main`.
I'm not entirely sure how the sharding works in redis and this feels more safe. With this the key is surely unique, and there's still a sharding key.

server/svix-server/src/queue/redis.rs
```diff
@@ -234,3 +215,123 @@ impl TaskQueueReceive for RedisQueueConsumer {
...
+async fn migrate_legacy_queues(pool: &mut PooledConnection<'_>) {
+    migrate_list(pool, LEGACY_MAIN, MAIN).await;
+    migrate_list(pool, LEGACY_PROCESSING, PROCESSING).await;
+    migrate_sset(pool, LEGACY_DELAYED, DELAYED).await;
+}
+
+async fn migrate_list(pool: &mut PooledConnection<'_>, legacy_queue: &str, queue: &str) {
+    let batch_size = 1000;
+    let mut x = 0;
+    loop {
```
tasn, line 228: It feels confusing to only wipe the queue once you migrated all of them, instead of after processing each batch. Also the way the condition for the wiping works is very implicit (oh, the last range fetch returned nothing it means we are past the end). I liked how you had it before better.

tasn (commented):
Looks great other than a tiny nitpick and the migration.
My suggestion is to remove the migration from this PR, fix the nitpick, and change nothing else. Once you do that we can just merge this and you can immediately iterate on the migration script based on the feedback.

tasn (approved):

PR #398, opened 2022-04-28
Rust: Migrate legacy redis queues on startup

jaymell:


server/svix-server/src/queue/redis.rs
```diff
@@ -51,9 +55,13 @@ pub async fn new_pair(pool: RedisPool) -> (TaskQueueProducer, TaskQueueConsumer)
         let batch_size: isize = 50;
         let task_validity_duration = chrono::Duration::from_std(TASK_VALIDITY_DURATION).unwrap();
 
-        let pool = worker_pool;
+        let mut pool = worker_pool.get().await.unwrap();
+
+        // drain legacy queues:
+        migrate_legacy_queues(&mut pool).await;
```
tasn, line 61: This never frees the pool connection, should be:
```suggestion
        let pool = worker_pool;
        {
          let mut pool = worker_pool.get().await.unwrap();

          // drain legacy queues:
          migrate_legacy_queues(&mut pool).await;
       }
```

server/svix-server/src/queue/redis.rs
```diff
@@ -206,3 +215,123 @@ impl TaskQueueReceive for RedisQueueConsumer {
...
+                "Migrating {} keys from queue {}",
+                legacy_keys.len(),
+                legacy_queue
+            );
+            let legacy_keys: Vec<(f64, String)> =
+                legacy_keys.into_iter().map(|(x, y)| (y, x)).collect();
+
+            let _: () = pool.zadd_multiple(queue, &legacy_keys).await.unwrap();
+        } else {
+            let _: () = pool.del(legacy_queue).await.unwrap();
```
tasn, line 271: Comments i made in the other PR still stand.

server/svix-server/src/queue/redis.rs
```diff
@@ -51,10 +55,17 @@ pub async fn new_pair(pool: RedisPool) -> (TaskQueueProducer, TaskQueueConsumer)
...
+        {
+            let pool = worker_pool.clone();
             let mut pool = pool.get().await.unwrap();
 
+            // drain legacy queues:
+            migrate_legacy_queues(&mut pool).await;
+        }
+
+        loop {
+            let mut pool = worker_pool.get().await.unwrap();
```
tasn, line 67: Why are we getting the pool here and didn't do before?

tasn (commented):
One comment, looks good otherwise.

server/svix-server/src/queue/redis.rs
svix-james: This is the code on `main` currently:
```
    let worker_pool = pool.clone();
    tokio::spawn(async move {
        // FIXME: enforce we only have one such worker (locking)
        let batch_size: isize = 50;
        let task_validity_duration = chrono::Duration::from_std(TASK_VALIDITY_DURATION).unwrap();

        let pool = worker_pool;
        loop {
            let mut pool = pool.get().await.unwrap();
```
I removed the (what I think is redundant) `let pool = worker_pool` assignment.

tasn: Ahh missed that, cool.

tasn (approved):

PR #400, opened 2022-04-29
Rust: Ensure Redis queue message IDs are refreshed when re-queuing

jaymell:
This prevents messages from being re-queued with the same ID, which can cause messages to be re-queued multiple times.

### DIFF
--- a/server/svix-server/src/queue/redis.rs
+++ b/server/svix-server/src/queue/redis.rs
@@ -48,7 +48,9 @@ const LEGACY_DELAYED: &str = "svix_queue_delayed";
 /// After this limit a task should be taken out of the processing queue and rescheduled
 const TASK_VALIDITY_DURATION: Duration = Duration::from_secs(45);
 
-fn refresh_key(msg: &str) -> String {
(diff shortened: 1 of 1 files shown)


server/svix-server/src/queue/redis.rs
```diff
@@ -265,7 +276,7 @@ async fn migrate_sset(pool: &mut PooledConnection<'_>, legacy_queue: &str, queue
         let _: () = pool.zadd_multiple(queue, &legacy_keys).await.unwrap();
     }
 }
-
+/*
```
tasn, line 279: Hmm??

server/svix-server/src/queue/redis.rs
```diff
@@ -48,6 +48,12 @@ const LEGACY_DELAYED: &str = "svix_queue_delayed";
 /// After this limit a task should be taken out of the processing queue and rescheduled
 const TASK_VALIDITY_DURATION: Duration = Duration::from_secs(45);
 
+fn refresh_key(msg: &str) -> String {
```
tasn, line 51: Needs a better name maybe (and probably a docstring to why it's needed).

PR #405, opened 2022-04-29
Support `payloadRetentionPeriod` for messages

svix-dylan:
This PR enables support for the `payloadRetentionPeriod` field as described in the [Svix API docs](https://api.svix.com/docs#operation/create_message_api_v1_app__app_id__msg__post). Requests to create a message can include the field `payloadRetentionPeriod` as an integer [5, 90] to indicate the number of days a message's payload data should be persisted (default: 90 days).

The PR also includes a worker that removes payload data for messages that are past their retention time.

Existing messages will be given a 90 day retention period, after which their payloads will be removed.


server/svix-server/migrations/20220428190104_message_expiration.down.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE message DROP COLUMN expiration;
```
tasn, line 1: Small note: why have two migrations rather than having both on the same one?

server/svix-server/src/worker.rs
```diff
@@ -326,6 +330,33 @@ pub async fn worker_loop(
     }
 }
 
+/// Nullifies the payload column for expired messages
+async fn clean_expired_messages(
```
tasn, line 334: I wouldn't put this in this file. I agree the name worker.rs is slightly confusing, but the intention is just for the webhook workers.

It's large enough already, and this worker is completely unrelated.

server/svix-server/src/worker.rs
```diff
@@ -326,6 +330,33 @@ pub async fn worker_loop(
...
+        .col_expr(message::Column::Payload, Expr::value(Value::Json(None)))
+        .filter(message::Column::Expiration.lte(Utc::now()))
+        .filter(Expr::col(message::Column::Payload).is_not_null())
+        .exec(pool)
+        .await
+}
+
+/// Runs every 5 minutes
+pub async fn expired_message_cleaner_loop(pool: &DatabaseConnection) -> Result<()> {
+    let pool = pool.clone();
```
tasn, line 347: When you clone it takes a connection from the pool, you want to do it inside of the loop, not outside.

server/svix-server/src/worker.rs
```diff
@@ -326,6 +330,33 @@ pub async fn worker_loop(
...
 }
 
+/// Nullifies the payload column for expired messages
+async fn clean_expired_messages(
+    pool: &DatabaseConnection,
+) -> std::result::Result<UpdateResult, DbErr> {
+    message::Entity::update_many()
+        .col_expr(message::Column::Payload, Expr::value(Value::Json(None)))
+        .filter(message::Column::Expiration.lte(Utc::now()))
+        .filter(Expr::col(message::Column::Payload).is_not_null())
```
tasn, line 340: Doesn't `message::Column::Payload.is_not_null()` (or something) not work?

server/svix-server/src/worker.rs
```diff
@@ -326,6 +330,33 @@ pub async fn worker_loop(
     }
 }
 
+/// Nullifies the payload column for expired messages
+async fn clean_expired_messages(
+    pool: &DatabaseConnection,
+) -> std::result::Result<UpdateResult, DbErr> {
+    message::Entity::update_many()
```
tasn, line 337: This scans the whole DB on every run. Adding an index on this model may be too expensive to be worth it.
I wonder whether there's a different strategy, like adding a new model and then just deleting the whole "content" item when expired? Though that may be expensive too.

Any thoughts?

server/svix-server/src/v1/endpoints/message.rs
```diff
@@ -60,23 +60,28 @@ pub struct MessageIn {
     pub uid: Option<MessageUid>,
     #[validate]
     pub event_type: EventTypeName,
-    pub payload: serde_json::Value,
+    pub payload: Option<serde_json::Value>,
```
tasn, line 63: Making the payload optional in MessageIn makes no sense. It's not optional, it has to exist.

server/svix-server/src/v1/endpoints/message.rs
```diff
@@ -87,8 +92,7 @@ pub struct MessageOut {
     #[serde(rename = "eventId")]
     pub uid: Option<MessageUid>,
     pub event_type: EventTypeName,
-    pub payload: serde_json::Value,
-
+    pub payload: Option<serde_json::Value>,
```
tasn, line 95: This change breaks the API (and compatibility with prod). Can just return some sort of a constant e.g. `{}` or better yet: `{"removed": true}`

server/svix-server/src/v1/endpoints/message.rs
```diff
@@ -204,7 +212,7 @@ async fn create_message(
         Some(app.clone()),
         app.id.clone(),
         app.org_id,
-        Duration::from_secs(30),
+        Duration::seconds(30).to_std().unwrap(),
```
tasn, line 215: Why was this change needed?

server/svix-server/src/worker.rs
```diff
@@ -326,6 +330,33 @@ pub async fn worker_loop(
...
+        .col_expr(message::Column::Payload, Expr::value(Value::Json(None)))
+        .filter(message::Column::Expiration.lte(Utc::now()))
+        .filter(Expr::col(message::Column::Payload).is_not_null())
+        .exec(pool)
+        .await
+}
+
+/// Runs every 5 minutes
+pub async fn expired_message_cleaner_loop(pool: &DatabaseConnection) -> Result<()> {
```
tasn, line 346:
```suggestion
pub async fn expired_message_cleaner_loop(pool: &DatabaseConnection) -> Result<!> {
```
I think this is what you want and then you can remove the `Ok(())`, as this function should never return a truthy value.

tasn (commented):
Not a full review, but I made some comments.

server/svix-server/src/v1/endpoints/message.rs
svix-dylan: I pulled in chrono::Duration, replacing std::time::Duration

server/svix-server/src/worker.rs
svix-dylan: Is your proposal to add a new table for storing payloads and their expirations, where we delete the entire row on expiration? I think that's reasonable, and brings the db scan from the number of every message to only the number of messages that still have payloads.

server/svix-server/src/v1/endpoints/message.rs
tasn:
```suggestion
        std::time::Duration::from_secs(30),
```

server/svix-server/src/expired_message_cleaner.rs
```diff
@@ -0,0 +1,33 @@
...
+use sea_orm::{
+    entity::prelude::*, sea_query::Expr, DatabaseConnection, DbErr, EntityTrait, UpdateResult,
+};
+use tokio::time::{sleep, Duration};
+
+/// Nullifies the payload column for expired messages
+async fn clean_expired_messages(
+    pool: &DatabaseConnection,
+) -> std::result::Result<UpdateResult, DbErr> {
+    message::Entity::update_many()
```
tasn, line 16: It wiped my comments, but I'm still concerned about this query being grossly inefficient without an index.

svix-dylan: @tasn I liked your previous comments about a new model. I can move `payload` and `expiration` timestamp to a new table, and delete the whole row on expiration. This way, db scanning is limited to n = the number of unexpired messages, instead of n = the sum of all messages, and the additional read/write overhead of this feature is isolated outside the main messages table.

svix-dylan: I've added a partial index, based on [this discussion and answer](https://stackoverflow.com/a/31966616), that will speed up the worker query by quickly filtering out non-NULL messages.

svix-dylan:
How do we want to handle existing messages, give them expirations that are 90 days from their `created_at` timestamps?

server/svix-server/migrations/20220428190104_message_expiration.up.sql
```diff
@@ -0,0 +1,5 @@
+BEGIN;
+ALTER TABLE message ADD COLUMN expiration timestamp with time zone;
+ALTER TABLE message ALTER COLUMN payload DROP NOT NULL;
+CREATE INDEX ON message (payload) WHERE payload IS NOT NULL;
```
tasn, line 4: 1. Explicitly set a name of the index (see in other SQL files how it's done).
2. The index should be on `expiration` not on `payload`.

server/svix-server/migrations/20220428190104_message_expiration.up.sql
```diff
@@ -0,0 +1,5 @@
+BEGIN;
```
tasn, line 1: I'm pretty sure it runs in a transaction anyway, you don't need to add one manually.

server/svix-server/migrations/20220428190104_message_expiration.up.sql
```diff
@@ -0,0 +1,5 @@
...
+ALTER TABLE message ADD COLUMN expiration timestamp with time zone;
```
tasn, line 2: There is no default value here and it's not nullable, so it's broken. You can probably define a default value of "now + 90 days" in SQL (server side calculated), or make it nullable.

server/svix-server/src/expired_message_cleaner.rs
```diff
@@ -0,0 +1,44 @@
...
+async fn clean_expired_messages(
+    pool: &DatabaseConnection,
+) -> std::result::Result<UpdateResult, DbErr> {
+    // this logic exists to gracefully migrate existing message attempt responses
+    message::Entity::update_many()
+        .col_expr(
+            message::Column::Expiration,
+            Expr::value(Utc::now() + Duration::days(90)),
+        )
+        .filter(message::Column::Expiration.is_null())
```
tasn, line 22: You actually made expiration not nullable.
You also want to check whether `payload` is null just so it hits the index.

server/svix-server/src/v1/endpoints/message.rs
```diff
@@ -61,22 +61,26 @@ pub struct MessageIn {
     #[validate]
     pub event_type: EventTypeName,
     pub payload: serde_json::Value,
-
     #[validate(custom = "validate_channels_msg")]
     #[validate]
     #[serde(skip_serializing_if = "Option::is_none")]
     pub channels: Option<EventChannelSet>,
+    #[validate(range(min = 5, max = 90))]
+    #[serde(rename = "payloadRetentionPeriod", default = "default_90")]
```
tasn, line 69: No need to rename, it automatically camelCases everything (probably if not, but it should.
```suggestion
    #[serde(default = "default_90")]
```

server/svix-server/migrations/20220428190104_message_expiration.up.sql
svix-dylan: 1. 👍 
2. Perhaps this index should be on both columns? Originally, `payload` was the column of concern, but since adding this business logic migration, `expiration` is as well as pointed out here.

server/svix-server/migrations/20220428190104_message_expiration.up.sql
svix-dylan: `expiration` is nullable unless I add a NOT NULL constraint, right?

However, assuming adding a default value of "now + 90 days" will add that default value for _existing_ rows before the migration, then that could be a clever workaround for the first update operation in the worker.

server/svix-server/src/expired_message_cleaner.rs
svix-dylan: `payload` is not nullable, but I think `expiration` can be null.

👍 on the payload condition so we can make use of that index (which would resolve my previous reply about adding an additional index specifically for the `expiration` column)

server/svix-server/src/v1/endpoints/message.rs
svix-dylan: Double checked to make sure and it does automatically camelCase!

server/svix-server/migrations/20220428190104_message_expiration.up.sql
svix-dylan: Confirmed, Postgres will apply a column's default value to existing rows when creating the column. I'll hold for your approval on this, but I think that setting the default expiration eliminates the need for the business logic update migration in the worker.

server/svix-server/migrations/20220428190104_message_expiration.up.sql
tasn: Still wrong, it should be:
```
CREATE INDEX ON message (expiration) WHERE payload IS NOT NULL; 
```

server/svix-server/migrations/20220428190104_message_expiration.up.sql
tasn: Yup, I agree.

My comment about the not-null was done in the wrong place, it was not null in the model code, which is what I meant. :)

server/svix-server/migrations/20220428190104_message_expiration.up.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE message ADD COLUMN expiration timestamp with time zone DEFAULT now() + INTERVAL '90 day';
```
tasn, line 1: Should be NOT NULL because you now have a default.

server/svix-server/src/expired_message_cleaner.rs
tasn:
```
     pub expiration: DateTimeWithTimeZone,
```
This is what I meant. It's not nullable in Rust.
Anyhow, it should also be not nullable in SQL so you can remove it.

server/svix-server/src/expired_message_cleaner.rs
tasn: Yeah, it should be fine now.

server/svix-server/src/expired_message_cleaner.rs
```diff
@@ -0,0 +1,33 @@
...
+    entity::prelude::*, sea_query::Expr, DatabaseConnection, DbErr, EntityTrait, UpdateResult,
+};
+use tokio::time::sleep;
+
+/// Nullifies the payload column for expired messages
+pub async fn clean_expired_messages(
+    pool: &DatabaseConnection,
+) -> std::result::Result<UpdateResult, DbErr> {
+    message::Entity::update_many()
+        .col_expr(message::Column::Payload, Expr::value(Value::Json(None)))
```
tasn, line 17:
```suggestion
        .col_expr(message::Column::Payload, Expr::value(None))
```

server/svix-server/src/v1/endpoints/message.rs
```diff
@@ -111,10 +113,11 @@ impl From<message::Model> for MessageOut {
         Self {
             uid: model.uid,
             event_type: model.event_type,
-            payload: model.payload,
-
+            payload: match model.payload {
+                Some(payload) => payload,
+                None => serde_json::json!({ "removed": true }),
```
tasn, line 118:
```suggestion
                None => serde_json::json!({ "expired": true }),
```

server/svix-server/src/expired_message_cleaner.rs
svix-dylan: Unfortunately this does not compile:
```
type inside `async fn` body must be known in this context
cannot infer type for type parameter `T` declared on the enum `Option`
```

jaymell (approved):

server/svix-server/src/expired_message_cleaner.rs
tasn: https://www.sea-ql.org/SeaORM/docs/basic-crud/update/#update-many

tasn: It's actually correct and the docs are wrong.

tasn (approved):

PR #412, opened 2022-04-29
Readme: Update links to dev guides

fcjr:
Just noticed the libs one is wrong and the server one doesn't autoscroll to the readme


tasn (approved):

tasn:
Oops, thanks!

PR #415, opened 2022-04-29
Rewrite Redis-based queue to use Streams functionality

svix-daniel:
Replaces the previous list-based approached with Redis streams which are meant to model queue structures.

### DIFF
--- a/server/svix-server/Cargo.toml
+++ b/server/svix-server/Cargo.toml
@@ -52,7 +52,7 @@ sqlx = { version = "0.5", features = [ "runtime-tokio-rustls", "postgres", "migr
 http = "0.2"
 time = { version = "0.3.9", features = [ "std" ]}
 futures = "0.3"
-redis_cluster_async = { git = "https://github.com/svix/redis-cluster-async" }
+redis_cluster_async = { git = "https://github.com/svix/redis-cluster-async", rev = "3d18fc2"}
 
 [dev-dependencies]
 anyhow = "1.0.56"
--- a/server/svix-server/Cargo.toml
+++ b/server/svix-server/Cargo.toml
@@ -51,8 +51,8 @@ sea-orm = { version = "0.7.1", features = [ "sqlx-postgres", "runtime-tokio-rust
 sqlx = { version = "0.5", features = [ "runtime-tokio-rustls", "postgres", "migrate" ] }
 http = "0.2"
 time = { version = "0.3.9", features = [ "std" ]}
-redis_cluster_async = "0.7.0"
 futures = "0.3"
+redis_cluster_async = { git = "https://github.com/svix/redis-cluster-async" }
 
 [dev-dependencies]
 anyhow = "1.0.56"
--- a/server/svix-server/Cargo.toml
+++ b/server/svix-server/Cargo.toml
@@ -52,7 +52,7 @@ sqlx = { version = "0.5", features = [ "runtime-tokio-rustls", "postgres", "migr
 http = "0.2"
 time = { version = "0.3.9", features = [ "std" ]}
 futures = "0.3"
-redis_cluster_async = { git = "https://github.com/svix/redis-cluster-async", rev = "3d18fc2"}
+redis_cluster_async = { git = "https://github.com/svix/redis-cluster-async.git", rev = "1fcd3ea" }
 
 [dev-dependencies]
 anyhow = "1.0.56"
--- a/server/svix-server/Cargo.toml
+++ b/server/svix-server/Cargo.toml
@@ -42,7 +42,7 @@ chrono = { version="0.4.19", features = ["serde"] }
 reqwest = { version = "0.11.9", features = ["json", "rustls-tls", "trust-dns"], default-features = false }
 bb8 = "0.7.1"
 bb8-redis = "0.10.1"
-redis = { version = "0.21.5", features = ["tokio-comp"] }
+redis = { version = "0.21.5", features = ["tokio-comp", "streams"] }
 thiserror = "1.0.30"
 bytes = "1.1.0"
 blake2 = "0.10.4"
--- a/server/svix-server/src/lib.rs
+++ b/server/svix-server/src/lib.rs
@@ -10,7 +10,6 @@ use cfg::{CacheType, QueueType};
 use std::{
     net::{SocketAddr, TcpListener},
     str::FromStr,
-    time::Duration,
 };
 use tower::ServiceBuilder;
 use tower_http::trace::TraceLayer;
@@ -63,11 +62,11 @@ pub async fn run(cfg: Configuration, listener: Option<TcpListener>) {
         match cfg.queue_type {
             QueueType::Redis => {
                 let pool = crate::redis::new_redis_pool(redis_dsn()).await;
-                queue::redis::new_pair(pool, Duration::from_secs(45), queue::redis::MAIN).await
+                queue::redis::new_pair(pool).await
             }
             QueueType::RedisCluster => {
                 let pool = crate::redis::new_redis_pool_clustered(redis_dsn()).await;
-                queue::redis::new_pair(pool, Duration::from_secs(45), queue::redis::MAIN).await
+                queue::redis::new_pair(pool).await
             }
             QueueType::Memory => queue::memory::new_pair().await,
         }
--- a/server/svix-server/src/lib.rs
+++ b/server/svix-server/src/lib.rs
@@ -9,7 +9,8 @@ use axum::{extract::Extension, Router};
 use cfg::{CacheType, QueueType};
 use std::{
     net::{SocketAddr, TcpListener},
-    str::FromStr, time::Duration,
+    str::FromStr,
+    time::Duration,
 };
 use tower::ServiceBuilder;
 use tower_http::trace::TraceLayer;
@@ -62,11 +63,11 @@ pub async fn run(cfg: Configuration, listener: Option<TcpListener>) {
         match cfg.queue_type {
             QueueType::Redis => {
                 let pool = crate::redis::new_redis_pool(redis_dsn()).await;
-                queue::redis::new_pair(pool, Duration::from_secs(45)).await
+                queue::redis::new_pair(pool, Duration::from_secs(45), queue::redis::MAIN).await
             }
             QueueType::RedisCluster => {
                 let pool = crate::redis::new_redis_pool_clustered(redis_dsn()).await;
-                queue::redis::new_pair(pool, Duration::from_secs(45)).await
+                queue::redis::new_pair(pool, Duration::from_secs(45), queue::redis::MAIN).await
             }
             QueueType::Memory => queue::memory::new_pair().await,
         }
--- a/server/svix-server/src/lib.rs
+++ b/server/svix-server/src/lib.rs
@@ -9,7 +9,7 @@ use axum::{extract::Extension, Router};
 use cfg::{CacheType, QueueType};
 use std::{
     net::{SocketAddr, TcpListener},
-    str::FromStr,
+    str::FromStr, time::Duration,
 };
 use tower::ServiceBuilder;
 use tower_http::trace::TraceLayer;
@@ -62,11 +62,11 @@ pub async fn run(cfg: Configuration, listener: Option<TcpListener>) {
         match cfg.queue_type {
             QueueType::Redis => {
                 let pool = crate::redis::new_redis_pool(redis_dsn()).await;
-                queue::redis::new_pair(pool).await
+                queue::redis::new_pair(pool, Duration::from_secs(45)).await
             }
             QueueType::RedisCluster => {
                 let pool = crate::redis::new_redis_pool_clustered(redis_dsn()).await;
-                queue::redis::new_pair(pool).await
+                queue::redis::new_pair(pool, Duration::from_secs(45)).await
             }
             QueueType::Memory => queue::memory::new_pair().await,
         }
--- a/server/svix-server/src/queue/redis.rs
+++ b/server/svix-server/src/queue/redis.rs
@@ -343,8 +343,8 @@ impl TaskQueueSend for RedisQueueProducer {
         Ok(())
     }
 
+    /// ACKing the delivery, XACKs the message in the queue so it will no longer be retried
     async fn ack(&self, delivery: TaskQueueDelivery) -> Result<()> {
-        // ACKing the delivery, XACKs the message in the queue so it will no longer be retried
         let mut pool = self.pool.get().await.unwrap();
         let processed: u8 = pool
             .query_async(Cmd::xack(
--- a/server/svix-server/src/queue/redis.rs
+++ b/server/svix-server/src/queue/redis.rs
@@ -55,9 +55,9 @@ const LEGACY_LEGACY_PROCESSING: &str = "svix_queue_processing";
 const LEGACY_LEGACY_DELAYED: &str = "svix_queue_delayed";
 
 /// Consumer group name constant
-const WORKERS_GROUP: &str = "svix_workers";
+const WORKERS_GROUP: &str = "svix_workers_group";
 /// Consumer group consumer name constant
-const WORKER_CONSUMER: &str = "svix_worker";
+const WORKER_CONSUMER: &str = "svix_workers_consumer";
 
 /// Special key for XADD command's which generates a real key automatically
 const GENERATE_STREAM_ID: &str = "*";
@@ -327,6 +327,7 @@ impl TaskQueueSend for RedisQueueProducer {
     }
 }
 
+#[derive(Debug, Clone)]
 pub struct RedisQueueConsumer {
     pool: RedisPool,
     main_queue_name: &'static str,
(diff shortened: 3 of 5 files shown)


server/svix-server/Cargo.toml
```diff
@@ -51,7 +51,7 @@ sea-orm = { version = "0.7.1", features = [ "sqlx-postgres", "runtime-tokio-rust
 sqlx = { version = "0.5", features = [ "runtime-tokio-rustls", "postgres", "migrate" ] }
 http = "0.2"
 time = { version = "0.3.9", features = [ "std" ]}
-redis_cluster_async = "0.7.0"
+redis_cluster_async = { git = "https://github.com/svix/redis-cluster-async" }
```
tasn, line 54:
```suggestion
redis_cluster_async = { git = "https://github.com/svix/redis-cluster-async", rev = "COMMIT_HASH" }
```
Set to a specific commit hash.

server/svix-server/src/lib.rs
```diff
@@ -62,11 +62,11 @@ pub async fn run(cfg: Configuration, listener: Option<TcpListener>) {
         match cfg.queue_type {
             QueueType::Redis => {
                 let pool = crate::redis::new_redis_pool(redis_dsn()).await;
-                queue::redis::new_pair(pool).await
+                queue::redis::new_pair(pool, Duration::from_secs(45)).await
```
tasn, line 65: Is this not a constant only for the testing? If so, I think it's better to just have a private module function that lets you fetch ignoring with a specific duration rather than having it always be set module-wide.

You don't usually want to uglify APIs just for tests. You want to have internal hooks.

server/svix-server/src/queue/redis.rs
```diff
@@ -36,17 +42,29 @@ use super::{
...
+const LEGACY_LEGACY_MAIN: &str = "svix_queue_main";
+const LEGACY_LEGACY_PROCESSING: &str = "svix_queue_processing";
+const LEGACY_LEGACY_DELAYED: &str = "svix_queue_delayed";
 
-/// After this limit a task should be taken out of the processing queue and rescheduled
-const TASK_VALIDITY_DURATION: Duration = Duration::from_secs(45);
+/// Consumer group name constant
+const WORKERS_GROUP: &str = "svix_workers";
+/// Consumer group consumer name constant
+const WORKER_CONSUMER: &str = "svix_worker";
```
tasn, line 62: The string values are too similar for vastly different things. You need to update the string value.
It's weird that one of the vars is plural and the other is singular.

server/svix-server/src/queue/redis.rs
```diff
@@ -56,61 +74,131 @@ fn regenerate_key(msg: &str) -> String {
...
+    {
+        let mut conn = pool
+            .get()
+            .await
+            .expect("Error retreiving connection from Redis pool");
 
-        {
-            let pool = worker_pool.clone();
-            let mut pool = pool.get().await.unwrap();
+        // TODO: Check error type
```
tasn, line 88: ...

server/svix-server/src/queue/redis.rs
```diff
@@ -217,25 +321,78 @@ pub struct RedisQueueConsumer {
...
+            legacy_queue
+        );
+
+        for key in legacy_keys {
+            let delivery = from_redis_key(&key);
+            let _: () = pool
+                .query_async(Cmd::xadd(
+                    queue,
+                    GENERATE_STREAM_ID,
+                    &[("data", serde_json::to_string(&delivery.task).unwrap())],
```
tasn, line 384: Make `"data"` a constant. You used constants for everything else. :)

server/svix-server/src/queue/redis.rs
```diff
@@ -167,40 +255,56 @@ impl TaskQueueSend for RedisQueueProducer {
...
-                key
+                delivery.id,
+                serde_json::to_string(&delivery.task).unwrap()
             );
         }
         Ok(())
     }
 
     async fn nack(&self, delivery: TaskQueueDelivery) -> Result<()> {
         // FIXME: do something else here?
```
tasn, line 297: Can you please open a ticket about this FIXME?

server/svix-server/src/queue/redis.rs
```diff
@@ -167,40 +255,56 @@ impl TaskQueueSend for RedisQueueProducer {
...
         let mut pool = self.pool.get().await.unwrap();
         let timestamp = delay.map(|delay| Utc::now() + chrono::Duration::from_std(delay).unwrap());
-        let delivery = TaskQueueDelivery::new(task, timestamp);
-        let key = to_redis_key(&delivery);
         if let Some(timestamp) = timestamp {
+            let delivery = TaskQueueDelivery::new(task, Some(timestamp));
+            let key = to_redis_key(&delivery);
             let _: () = pool
-                .zadd(DELAYED, key, timestamp.timestamp())
+                .zadd(LEGACY_DELAYED, key, timestamp.timestamp())
```
tasn, line 262: Why call it legacy delayed when it's clearly not legacy?

server/svix-server/src/queue/redis.rs
```diff
@@ -217,25 +321,78 @@ pub struct RedisQueueConsumer {
...
+                .await
+                .unwrap();
+
+            if resp.keys[0].ids.len() > 0 {
+                break resp;
+            }
+        };
+
+        let id = std::mem::take(&mut resp.keys[0].ids[0].id);
+        let map = std::mem::take(&mut resp.keys[0].ids[0].map);
```
tasn, line 344: `std::mem` is an anti-pattern, it's almost always scary.
```
let idtom = &mut resp.keys[0].ids[0];
let id = &mut idtom.id;
let map = &mut idtom.map;
```

server/svix-server/src/queue/redis.rs
```diff
@@ -56,61 +74,131 @@ fn regenerate_key(msg: &str) -> String {
...
+) -> (TaskQueueProducer, TaskQueueConsumer) {
+    // Create the stream and consumer group for the MAIN queue should it not already exist
+    {
+        let mut conn = pool
+            .get()
+            .await
+            .expect("Error retreiving connection from Redis pool");
 
-        {
-            let pool = worker_pool.clone();
```
tasn, line 67: Make the diff smaller. If you want to make this change, do it later, in a different PR.

server/svix-server/src/queue/redis.rs
```diff
@@ -1,4 +1,4 @@
-/// Redis queue implementation
+// Redis queue implementation
```
tasn, line 1: 🤔

server/svix-server/src/queue/redis.rs
```diff
@@ -40,46 +48,56 @@ use super::{
...
+const DELAYED: &str = "{queue}_svix_delayed";
 
-// FIXME: hash-tags here enable clustering but prevent proper sharding of lists
+// v2 KEY CONSTANTS
 const LEGACY_MAIN: &str = "{queue}_svix_main";
 const LEGACY_PROCESSING: &str = "{queue}_svix_processing";
-// TODO: Rename LEGACY_DELAYED
-const DELAYED: &str = "{queue}_svix_delayed";
 
+// v1 KEY CONSTANTS
```
tasn, line 65: Maybe instead of LEGACY_LEGACY just call it LEGACY_V1_MAIN and etc? And the other one LEGACY_V2?
Much better than LEGACY_LEGACY and documenting it in comments, no?

svix-james (approved):

svix-dylan (dismissed):

server/svix-server/src/queue/redis.rs
```diff
@@ -1,31 +1,43 @@
...
-///     AKA: Delayed
-///
-/// - Tasks in the queues are prefixed with a ksuid so that we can know the timestamp of when they
-/// should be executed.
-///
-/// The implementation spawns an additional worker that monitors both the zset delayed tasks and
-/// the tasks currently processing. It monitors the zset task set for tasks that should be
-/// processed now, and the currently processing queue for tasks that have timed out and should be
-/// put back on the main queue.
+//! Redis stream-based queue implementation
```
tasn, line 1: Why not `///`? I don't even remember what `//!` does tbh.

server/svix-server/src/queue/redis.rs
```diff
@@ -36,89 +48,222 @@ use super::{
...
-                    .map(|x| x.0)
-                    .map(|x| regenerate_key(&x))
+                let keys: Vec<(String, String)> = pool
+                    .zpopmin(delayed_queue_name, keys.len() as isize)
+                    .await
+                    .unwrap();
+                let tasks: Vec<&str> = keys
+                    .iter()
+                    // All information is stored in the key in which the ID and JSON formated task
+                    // are separated by a `|`. So, take the key, then take the part after the `|`
```
tasn, line 176: Isn't the ID external to the value, and the value is just there on its own as the data of key `data`?

server/svix-server/src/queue/redis.rs
```diff
@@ -36,89 +48,222 @@ use super::{
...
+                    .map(|x| &x.0)
+                    .map(|x| x.split('|').nth(1).expect("Improper key format"))
                     .collect();
-                let _: () = pool.rpush(MAIN, keys).await.unwrap();
+
+                // Then for each task, XADD them to ghe MAIN queue
+                for task in tasks {
```
tasn, line 182: Use a pipeline.

server/svix-server/src/queue/redis.rs
```diff
@@ -36,89 +48,222 @@ use super::{
...
+            let _ = cmd
+                .arg(&main_queue_name)
+                .arg(WORKERS_GROUP)
+                // Search only for IDs that have been idle for at least the pending_duration
+                .arg("IDLE")
+                .arg(pending_duration)
+                // And search for at most 1000 IDs from the minimum ID value to the maximum ID value
+                .arg("-")
+                .arg("+")
+                .arg(1000i16);
```
tasn, line 209: Make this a const. In general when you have values that affect behavior, let's have them somewhere more visible so that we know more easily what we can modify to affect performance and behavior.

server/svix-server/src/queue/redis.rs
```diff
@@ -36,89 +48,222 @@ use super::{
...
+const GENERATE_STREAM_ID: &str = "*";
+/// Special ID for XREADGROUP commands which reads any new messages
+const LISTEN_STREAM_ID: &str = ">";
+
+/// Each queue item has a set of KV pairs associated with it, for simplicity a sing key, "data" is
+/// used with the entire [`QueueTask`] as the value in serialized JSON
+const QUEUE_KV_KEY: &str = "data";
+
+/// Generates a [`TaskQueueProducer`] and a [`TaskQueueConsumer`] backed by Redis.
+pub async fn new_pair(pool: RedisPool) -> (TaskQueueProducer, TaskQueueConsumer) {
```
tasn, line 88: This is a nice way to have testability, gj. :)

server/svix-server/src/queue/redis.rs
```diff
@@ -166,41 +313,72 @@ fn from_redis_key(key: &str) -> TaskQueueDelivery {
...
+                serde_json::to_string(&delivery.task).unwrap()
             );
         }
         Ok(())
     }
 
     async fn nack(&self, delivery: TaskQueueDelivery) -> Result<()> {
-        // FIXME: do something else here?
-        let key = to_redis_key(&delivery);
+        // FIXME: NACKing will also XACK the message, but something else should be done here
```
tasn, line 367: TBH, I think nacking should probably not do anything for now (other than print an error) as we probably just want it to be picked up again once the pending timeout finishes?

server/svix-server/src/queue/redis.rs
svix-daniel: It's just the convention for documenting modules in whole. See https://doc.rust-lang.org/reference/comments.html

server/svix-server/src/queue/redis.rs
svix-daniel: The delayed queue is untouched as of now, so they're exactly the same

server/svix-server/src/queue/redis.rs
```diff
@@ -209,33 +387,91 @@ impl TaskQueueSend for RedisQueueProducer {
...
-            .arg(Direction::Left)
-            .arg(Direction::Right)
-            .arg(0u8);
-        let key: String = pool
-            .query_async(cmd)
-            .await
-            .map_err(|x| Error::Queue(x.to_string()))?;
+
+        // There is no way to make it await a message for unbounded times, so simply block for a short
+        // amount of time (to avoid locking) and loop if no messages were retreived
```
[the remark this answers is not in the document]
svix-daniel: The blocking actually seems to block and never yields control to other tasks. I figured a shorter block time was a simpler solution than spawning a new green thread.

server/svix-server/src/queue/redis.rs
tasn: Sweet.

server/svix-server/src/queue/redis.rs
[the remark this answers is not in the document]
tasn: This is a major issue even when only blocking for 500ms. This is unacceptable and a bug in the library it seems.

This will kill our throughput.

server/svix-server/src/queue/redis.rs
```diff
@@ -166,41 +313,72 @@ fn from_redis_key(key: &str) -> TaskQueueDelivery {
...
-        let processed: u8 = pool.lrem(PROCESSING, 1, &key).await.unwrap();
+        let processed: u8 = pool
+            .query_async(Cmd::xack(
+                self.main_queue_name,
+                WORKERS_GROUP,
+                &[delivery.id.as_str()],
+            ))
+            .await
+            .unwrap();
         if processed != 1 {
```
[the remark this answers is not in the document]
svix-daniel: It was in the previous version and I assumed it was there for a reason :). But no, it should be impossible for IDs to overlap.

server/svix-server/src/queue/redis.rs
```diff
@@ -36,89 +48,222 @@ use super::{
 
 // FIXME: Change unwraps to have our own error type for the queue module entirely
 
-// FIXME: hash-tags here enable clustering but prevent proper sharding of lists
-const MAIN: &str = "{queue}_svix_main";
-const PROCESSING: &str = "{queue}_svix_processing";
-const DELAYED: &str = "{queue}_svix_delayed";
+/// This is the key of the main queue. As a KV store, redis places the entire stream under this key.
+/// Confusingly, each message in the queue may have any number of KV pairs.
+const MAIN: &str = "svix_{queue}_v3_main";
```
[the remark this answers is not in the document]
tasn: You closed it by accident?

server/svix-server/src/queue/redis.rs
```diff
@@ -209,33 +375,98 @@ impl TaskQueueSend for RedisQueueProducer {
...
+            let delivery = from_redis_key(&key);
+            let _: () = pool
+                .query_async(Cmd::xadd(
+                    queue,
+                    GENERATE_STREAM_ID,
+                    &[(QUEUE_KV_KEY, serde_json::to_string(&delivery.task).unwrap())],
+                ))
+                .await
+                .unwrap();
+        }
```
tasn, line 462: Pipeline too, in batches of size X (whatever X is).

Pipeline is orders of magnitude faster, so it's especially important here.

tasn, line 462: Pipeline too, in batches of size X (whatever X is).

Pipeline is orders of magnitude faster, so it's especially important here.

server/svix-server/src/queue/redis.rs
```diff
@@ -166,41 +315,58 @@ fn from_redis_key(key: &str) -> TaskQueueDelivery {
...
     }
 
-    async fn nack(&self, delivery: TaskQueueDelivery) -> Result<()> {
-        // FIXME: do something else here?
-        let key = to_redis_key(&delivery);
-        let mut pool = self.pool.get().await.unwrap();
-        let _: () = pool.lrem(PROCESSING, 1, &key).await.unwrap();
-        tracing::error!("Failed processing msg: {}", key);
+    async fn nack(&self, _delivery: TaskQueueDelivery) -> Result<()> {
+        tracing::error!("NACKing is not yet supported");
```
tasn, line 369: Include the delivery ID for debugging.

tasn, line 369: Include the delivery ID for debugging.

server/svix-server/src/queue/redis.rs
```diff
@@ -166,41 +315,58 @@ fn from_redis_key(key: &str) -> TaskQueueDelivery {
...
+                .await
+                .unwrap();
         }
         tracing::trace!("RedisQueue: event sent > (delay: {:?})", delay);
         Ok(())
     }
 
     async fn ack(&self, delivery: TaskQueueDelivery) -> Result<()> {
-        let key = to_redis_key(&delivery);
+        // ACKing the delivery, XACKs the message in the queue so it will no longer be retried
```
tasn, line 347: Maybe make it a function doc?

tasn, line 347: Maybe make it a function doc?

tasn (commented):
Made tiny comments. Let have someone else do the final review and I think we can go in. Though I really really really want to test it properly to make sure nothing breaks.

svix-james (dismissed):

svix-james (dismissed):

svix-james (approved):

PR #416, opened 2022-05-02
Gracefully handle non-text endpoint response bodies

svix-dylan:
If parsing a message attempt's response body as a string fails, then base64 encode the response body bytes, and persist that.

### DIFF
--- a/server/svix-server/src/worker.rs
+++ b/server/svix-server/src/worker.rs
@@ -190,15 +190,17 @@ async fn dispatch(
                 MessageStatus::Fail
             };
             let http_error = res.error_for_status_ref().err();
-            let body = res
-                .text()
-                .await
-                .unwrap_or_else(|_| "BODY_NOT_TEXT".to_string());
+
+            let bytes = res.bytes().await.unwrap().to_vec();
+            let body = match std::str::from_utf8(&bytes) {
+                Ok(v) => v.to_owned(),
+                Err(_) => base64::encode(&bytes),
+            };
+
             let attempt = messageattempt::ActiveModel {
                 response_status_code: Set(status_code),
                 response: Set(body),
                 status: Set(status),
-
                 ..attempt
             };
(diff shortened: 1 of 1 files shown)


server/svix-server/src/worker.rs
```diff
@@ -190,15 +190,17 @@ async fn dispatch(
...
-            let body = res
-                .text()
-                .await
-                .unwrap_or_else(|_| "BODY_NOT_TEXT".to_string());
+
+            let bytes = res.bytes().await.unwrap().to_vec();
+            let body = match std::str::from_utf8(&bytes) {
+                Ok(v) => v.to_owned(),
+                Err(_) => base64::encode(&bytes),
+            };
```
tasn, line 198: Fallback to bytes if `.text()` fails. Going this two step, especially when moving to vec is not fun.

svix-dylan: @tasn unfortunately calling `.text()` consumes `res` -- I wasn't able to call the `.bytes()` method after calling `.text()`.

tasn: Any luck with getting rid of `to_vec()`?

svix-dylan: No luck. `.text()` is implemented [also with a to_vec()](https://github.com/seanmonstar/reqwest/blob/6ca5f3e50c979909b786a4f1e2c73611164254c7/src/async_impl/response.rs#L205). `.bytes()` returns bytes using `bytes::Bytes`, which does not seem to offer a quicker path to `String` ([docs](https://docs.rs/bytes/latest/bytes/struct.Bytes.html)).

tasn: So that's your answer right there, this is the right way of doing it, it seems.

tasn:
Just need to add tests now.

tasn (approved):

PR #417, opened 2022-05-04
Sqs queue

svix-james:


server/svix-server/src/cfg.rs
```diff
@@ -62,6 +62,9 @@ pub struct ConfigurationInner {
     /// What kind of message queue to use. Supported: memory, redis (must have redis_dsn configured).
     pub queue_type: QueueType,
 
+    /// SQS DSN (fully qualified queue DSN). Can be left empty if not using SQS
+    pub sqs_dsn: Option<String>,
```
tasn, line 66:
```suggestion
    pub queue_sqs_dsn: Option<String>,
```

server/svix-server/src/error.rs
```diff
@@ -28,6 +28,8 @@ pub enum Error {
     Validation(String),
     /// Any kind of HttpError
     Http(HttpError),
+    /// Timeout consuming from queue
+    QueueTimeout(String),
```
tasn, line 32: It should be more like Http errors, so:
```
Queue(QueueError)
```
And then the timeout would be one of their queue errors.

Though I think we don't need it anymore?

PR #418, opened 2022-05-04
Rust: TaskQueueReceive Receive to ReceiveAll

jaymell:
This modifies queue receivers to potentially return batches instead
of single messages.


jaymell:
Note that this change will offer better support for services, e.g., SQS, that can return batches of messages.

svix-dylan (approved):

tasn:
Let's get this bad boy in?

svix-dylan (approved):

PR #419, opened 2022-05-04
Preserve header key capitalization

svix-dylan:
Preserves the capitalization of header keys, removes a unit test that enforced lowercasing of keys, and adds an integration test that validates key capitalization is persisted as received.

### DIFF
--- a/server/svix-server/src/core/types.rs
+++ b/server/svix-server/src/core/types.rs
@@ -579,14 +579,14 @@ impl Validate for EndpointHeaders {
             if let Err(_e) = http::header::HeaderValue::try_from(v) {
                 errors.add(ALL_ERROR, ValidationError::new("Invalid Header Value."));
             }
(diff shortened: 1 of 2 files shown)


server/svix-server/src/core/types.rs
```diff
@@ -728,38 +726,15 @@ mod tests {
...
-        )]);
+        let hdr_map = HashMap::from([("User-Agent".to_string(), "true".to_owned())]);
         let endpoint_headers = EndpointHeaders(hdr_map);
         assert!(endpoint_headers.validate().is_err());
 
-        let hdr_map = HashMap::from([(
-            EndpointHeaders::FORBIDDEN_PREFIXES[0].to_owned(),
-            "true".to_owned(),
-        )]);
+        let hdr_map = HashMap::from([("User-Agent".to_string(), "true".to_owned())]);
```
tasn, line 733: This should test a broken prefix.

tasn (approved):

PR #425, opened 2022-05-06
Change to a more efficient first message dispatch

svix-dylan:
Introduces the queue type `MessageTaskBatch` for incoming messages so
we can concurrently send a message to multiple endpoints from within
`worker.rs`. The current queue type `MessageTask` is kept intact and 
still used for message retries and other scenarios where sending to
a single endpoint is the desired behavior.

Note: adds a dependency on the futures crate.


server/svix-server/src/worker.rs
```diff
@@ -291,6 +282,32 @@ fn bytes_to_string(bytes: bytes::Bytes) -> String {
...
+
+    // get individual MessageTasks per endpoint and collect their executions as futures
+    let futures: Vec<_> = queue_task
+        .get_message_tasks()
+        .into_iter()
+        .map(|msg_task| dispatch(cfg.clone(), db, cache.clone(), queue_tx, msg_task, &msg))
+        .collect();
+
+    // execute the MessageTasks concurrently; any single failure will bubble up to the invoking function
+    future::try_join_all(futures).await
```
svix-james, line 308: Can we change this to `join_all` to avoid unnecessarily cancelling other tasks?

svix-dylan: Been thinking about this on my flight. If we do this, then I need to manually check each result in the vector for error types -- this functionality is not provided.

I'm fine with that.. but the more I think about it, the more I'm not sure how much sense that makes? Like, the errors we can anticipate (non-200 responses, timeouts, etc), we've optimized for and they get requeued appropriately. Errors outside of what's anticipated, I think we should just fail the whole batch atomically (or as close to atomically as we can get). If this run crashes, we should nack without regard for attempting the remaining endpoints. They're all going to get rerun anyways, because this situation is going to result in a nack regardless.

Idk, that's my take. Curious to hear @svix-daniel's and @tasn's thoughts on the matter.

tasn: I'm with @svix-james here. I think we shouldn't cancel other tasks, but instead deal with them separately. It's not very hard to do so, and it's very useful.

This is without reviewing the rest of the PR. Just based on this small part and the discussion here.

server/svix-server/src/queue/mod.rs
```diff
@@ -40,11 +40,67 @@ impl MessageTask {
...
 }
 
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct MessageTaskBatch {
+    pub msg_id: MessageId,
+    pub app_id: ApplicationId,
+    pub endpoint_ids: Vec<EndpointId>,
+    pub trigger_type: MessageAttemptTriggerType,
+    pub attempt_count: u16,
```
tasn, line 50: Almost all of these are not needed.
- msg_id is needed
- app_id is needed I guess for fetching msg_id, but otherwise not really needed.
- endpoint_ids are not needed, you can just fetch the CreateMessageApp and get them back.
- attempt_count is always 0 for batch, so not needed.

server/svix-server/src/queue/mod.rs
```diff
@@ -79,6 +135,7 @@ impl TaskQueueConsumer {
     }
 }
 
+#[derive(Clone)]
```
tasn, line 138: I'm always weary about seeing clones, because well, they clone. Why is this needed all of a sudden?

server/svix-server/src/queue/mod.rs
```diff
@@ -40,11 +40,67 @@ impl MessageTask {
...
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 #[serde(tag = "type")]
 pub enum QueueTask {
     MessageV1(MessageTask),
+    MessageBatchV1(MessageTaskBatch),
+}
+
+impl QueueTask {
+    pub fn get_message_id(self) -> MessageId {
```
tasn, line 88:
```suggestion
    pub fn msg_id(self) -> MessageId {
```
Rust doesn't uses a `get_` prefix for getters (convention). Also, we call it msg_id everywhere.

server/svix-server/src/queue/mod.rs
```diff
@@ -40,11 +40,67 @@ impl MessageTask {
...
+        match self {
+            QueueTask::MessageV1(task) => task.msg_id,
+            QueueTask::MessageBatchV1(batch) => batch.msg_id,
+        }
+    }
+    pub fn get_message_tasks(self) -> Vec<MessageTask> {
```
tasn, line 94:
```suggestion
    pub fn msg_tasks(self) -> Vec<MessageTask> {
```

server/svix-server/src/v1/endpoints/message.rs
```diff
@@ -265,21 +265,27 @@ async fn create_message(
             ..Default::default()
         };
         msg_dests.push(msg_dest);
-        tasks.push(
-            MessageTask::new_task(
-                msg.id.clone(),
-                app.id.clone(),
-                endp.id, MessageAttemptTriggerType::Scheduled));
+        endpoint_ids.push(endp.id.clone());
```
tasn, line 268: Not needed, see other comment.

server/svix-server/src/v1/endpoints/message.rs
```diff
@@ -265,21 +265,27 @@ async fn create_message(
...
     }
     if !msg_dests.is_empty() {
         messagedestination::Entity::insert_many(msg_dests)
             .exec(db)
             .await?;
```
tasn, line 273: Don't save them here, only save them in the worker (to reduce latency here). You can just calculate them again easily there.

server/svix-server/src/queue/mod.rs
```diff
@@ -40,11 +40,67 @@ impl MessageTask {
...
+    ) -> QueueTask {
+        QueueTask::MessageBatchV1(Self {
+            msg_id,
+            app_id,
+            endpoint_ids,
+            attempt_count: 0,
+            trigger_type,
+        })
+    }
+    pub fn to_message_task(self, endpoint_id: &EndpointId) -> MessageTask {
```
tasn, line 68: I'm a bit suspicious of this function. Not sure why it's needed.

tasn (commented):
I haven't finished reviewing, but just wanted to share what I have in mind.

server/svix-server/src/queue/mod.rs
svix-dylan: This is ultimately so that I can reuse `dispatch` for both `MessageTask` and `MessageTaskBatch` by always exploding `MessageTaskBatch` into multiple `MessageTask`s per endpoint, and executing those concurrently in `dispatch`. To me, the logic for that conversion felt more natural implemented as a method on `MessageTaskBatch` than constructed ad-hoc somewhere in the worker.

server/svix-server/src/queue/mod.rs
svix-dylan: Good find, this was added during debugging but ultimately not needed, shouldn't be in this PR.

server/svix-server/src/queue/mod.rs
svix-dylan: But with your feedback of removing the endpoint_ids field completely and calculating them at work-time, this function probably becomes obsolete and can be removed.

tasn: I didn't mean at "work-time" in the sense that it has to be in `dispatch` but rather that you call the logic that does the math in the worker (after the queue), rather than before.

server/svix-server/src/v1/endpoints/message.rs
svix-dylan: Do we still want messagedestination rows? If we're going to just calculate this on the fly, I'm not sure what their use is; messageattempt acts as our record of what actually occurred.

But, if we're going to still want these rows, I feel like this is where that write belongs? Otherwise we are doing an upsert or conditional insert in the worker to only write these rows the first time a task is processed, which I think is awkward - the latency isn't _that bad_ is it?

tasn: We will eventually get rid of messagedestination because they aren't really needed. Though they are needed for now unfortunately (need to make a few more changes before we can get rid of them).

I think these can be written in the worker before you call each `dispatch`.

server/svix-server/src/worker.rs
```diff
@@ -291,6 +257,116 @@ fn bytes_to_string(bytes: bytes::Bytes) -> String {
...
+                endpoint.to_owned(),
+            )
+        })
+        .collect();
+
+    let join = future::join_all(futures).await;
+
+    let errs: Vec<_> = join.iter().filter(|x| x.is_err()).collect();
+    if !errs.is_empty() {
+        return Err(Error::Generic(format!(
```
tasn, line 361: I think we should re-add MessageTaskV1 tasks for the failed ones. No?

server/svix-server/src/v1/endpoints/message.rs
```diff
@@ -198,93 +193,36 @@ pub struct CreateMessageQueryParams {
...
     Extension(ref db): Extension<DatabaseConnection>,
     Extension(queue_tx): Extension<TaskQueueProducer>,
-    Extension(cache): Extension<Cache>,
     ValidatedQuery(CreateMessageQueryParams { with_content }): ValidatedQuery<
         CreateMessageQueryParams,
     >,
     ValidatedJson(data): ValidatedJson<MessageIn>,
     AuthenticatedApplication { permissions, app }: AuthenticatedApplication,
 ) -> Result<(StatusCode, Json<MessageOut>)> {
-    let create_message_app = CreateMessageApp::layered_fetch(
```
tasn, line 208: You do want it and need it. Don't remove it, just don't create messagedestinations from it.

It's used for error handling and to know whether there's even a need to create a task batch (what if there are no endpoints that match this message? E.g. due to event type filtering).

server/svix-server/src/worker.rs
```diff
@@ -73,87 +74,53 @@ fn generate_msg_headers(
 async fn dispatch(
     cfg: Configuration,
     db: &DatabaseConnection,
-    cache: Cache,
     queue_tx: &TaskQueueProducer,
+    payload: Option<Json>,
```
tasn, line 78: Why is payload Optional?

server/svix-server/src/v1/endpoints/message.rs
```diff
@@ -217,74 +216,57 @@ async fn create_message(
...
-                endp.id, MessageAttemptTriggerType::Scheduled));
-    }
-    if !msg_dests.is_empty() {
-        messagedestination::Entity::insert_many(msg_dests)
-            .exec(db)
+    let trigger_type = MessageAttemptTriggerType::Scheduled;
+    if create_message_app
+    .endpoints
+    .iter()
+    .any(|endpoint| {
```
svix-dylan, line 230: Looking for others thoughts here: we are including this check to see if this message even has any endpoints to send to, but it does not affect the response type. At the cost of bloating the queue, we could just remove this check, and let potentially noop messages in the queue, which the worker will resolve as noops and `ack` without issue. This will save us the network call(s) to fetch `create_message_app` here in this handler.

server/svix-server/src/worker.rs
svix-dylan: Another thing here though: I'm currently tracing all the errors, but this could be pretty messy in a scenario where a message has lots of endpoints and all or many of those dispatches failed in a way that dispatch didn't already handle (eg db connection issue, etc).

svix-james (dismissed):

server/svix-server/src/worker.rs
```diff
@@ -291,6 +257,117 @@ fn bytes_to_string(bytes: bytes::Bytes) -> String {
...
+        .await?
+        .ok_or_else(|| {
+            Error::Generic(format!(
+                "Unexpected: message doesn't exist {}",
+                queue_task.clone().msg_id()
+            ))
+        })?;
+    let payload = msg.payload.as_ref().expect("Message payload is NULL");
+
+    let msg_clone = msg.clone();
```
tasn, line 279: Cloning msg means cloning payload. I think you probably just want to clone app_id and org_id below?

server/svix-server/src/worker.rs
```diff
@@ -73,87 +74,53 @@ fn generate_msg_headers(
...
-            app_id, endp_id
-        ))
-    })?;
-
-    let body = serde_json::to_string(&msg.payload).expect("Error parsing message body");
+    let body = serde_json::to_string(&payload).expect("Error parsing message body");
     let headers = {
-        let keys: Vec<&EndpointSecret> = if let Some(ref old_keys) = endp.old_signing_keys {
-            iter::once(&endp.key)
+        let keys: Vec<&EndpointSecret> = if let Some(ref old_keys) = endpoint.old_signing_keys {
```
tasn, line 86: Call it endp so that the diff is small.

server/svix-server/src/worker.rs
```diff
@@ -291,6 +257,117 @@ fn bytes_to_string(bytes: bytes::Bytes) -> String {
...
+            .map(|endpoint| messagedestination::ActiveModel {
+                msg_id: Set(msg.id.clone()),
+                endp_id: Set(endpoint.id.clone()),
+                next_attempt: Set(Some(Utc::now().into())),
+                status: Set(MessageStatus::Sending),
+                ..Default::default()
+            });
+        messagedestination::Entity::insert_many(destinations)
+            .exec(db)
+            .await?;
```
tasn, line 318: Using a fetch first is very wasteful (another roundtrip, load on the db, etc), and it's also not needed.
You only need to create the destinations if the type is Batch. Otherwise they already exist.

server/svix-server/src/worker.rs
```diff
@@ -291,6 +257,117 @@ fn bytes_to_string(bytes: bytes::Bytes) -> String {
...
+
+    let futures: Vec<_> = endpoints
+        .iter()
+        .map(|endpoint| {
+            dispatch(
+                cfg.clone(),
+                db,
+                queue_tx,
+                payload.clone(),
+                queue_task.clone().to_msg_task(endpoint.clone().id),
```
tasn, line 329: Clone the ID, not the endpoint.

server/svix-server/src/worker.rs
```diff
@@ -291,6 +257,117 @@ fn bytes_to_string(bytes: bytes::Bytes) -> String {
...
+    }
+
+    let futures: Vec<_> = endpoints
+        .iter()
+        .map(|endpoint| {
+            dispatch(
+                cfg.clone(),
+                db,
+                queue_tx,
+                payload.clone(),
```
tasn, line 328: Can you make it a ref? All of this cloning is wasteful.

server/svix-server/src/worker.rs
tasn: OK, let's keep it this way, though it makes me a bit sad. I think we should recover when we can recover.

server/svix-server/src/worker.rs
```diff
@@ -291,6 +257,117 @@ fn bytes_to_string(bytes: bytes::Bytes) -> String {
...
+                queue_task.clone().to_msg_task(endpoint.clone().id),
+                endpoint.to_owned(),
+            )
+        })
+        .collect();
+
+    let join = future::join_all(futures).await;
```
tasn, line 335: > If any future returns an error then all other futures will be canceled and an error will be returned immediately.

From the docs for join_all. This is not good, it should not cancel others because it'll just put us in a weird state unnecessarily. I think @svix-james also mentioned it before.

server/svix-server/src/worker.rs
```diff
@@ -303,29 +380,28 @@ pub async fn worker_loop(
         match queue_rx.receive_all().await {
             Ok(batch) => {
                 for delivery in batch {
+                    let cfg = cfg.clone();
                     let pool = pool.clone();
-                    let queue_task = delivery.task.clone();
+                    let cache = cache.clone();
```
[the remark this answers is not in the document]
tasn: OK, I'll let this slide, but the reason I made this comment is that it cost me 60s of figuring out what the hell changed because of it. :)

server/svix-server/tests/e2e_attempt.rs
```diff
@@ -64,6 +64,8 @@ async fn test_list_attempted_messages() {
         .await
         .unwrap();
 
+    tokio::time::sleep(Duration::from_secs(2)).await;
```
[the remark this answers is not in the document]
svix-dylan: The network calls complete and unwrap whether or not messagedestination has been written yet; it's the actual `assert_eq!`s on their contents that fails.

server/svix-server/src/worker.rs
svix-dylan: I think this is for [`try_join_all`](https://docs.rs/futures/0.3.21/futures/future/fn.try_join_all.html), which we moved away from following @svix-james's comment.

server/svix-server/src/worker.rs
```diff
@@ -260,7 +226,7 @@ async fn dispatch(
                     .send(
                         QueueTask::MessageV1(MessageTask {
                             attempt_count: msg_task.attempt_count + 1,
-                            ..msg_task
+                            ..msg_task.clone()
```
tasn, line 229: Not the end of the world, but why is this needed all of a sudden?

server/svix-server/src/worker.rs
```diff
@@ -291,6 +257,89 @@ fn bytes_to_string(bytes: bytes::Bytes) -> String {
...
+
+    let futures: Vec<_> = endpoints
+        .iter()
+        .map(|endpoint| {
+            dispatch(
+                cfg.clone(),
+                db,
+                queue_tx,
+                payload,
+                queue_task.clone().to_msg_task(endpoint.id.clone()),
```
tasn, line 324: Why do we need the clone on queue_task?

svix-dylan: Unfortunately I wasn't able to figure out an alternative for this one - it's because of how it is used in this loop.

tasn: Ok.

server/svix-server/src/worker.rs
tasn: Wasn't this @svix-james's to *not* cancel them prematurely?

server/svix-server/tests/e2e_attempt.rs
```diff
@@ -64,37 +64,46 @@ async fn test_list_attempted_messages() {
...
-        .await
-        .unwrap();
-    let list_2: ListResponse<AttemptedMessageOut> = client
-        .get(
-            &format!("api/v1/app/{}/endpoint/{}/msg/", app_id, endp_id_2),
-            StatusCode::OK,
-        )
-        .await
-        .unwrap();
+    // And wait at most one second for all attempts to be processed
```
tasn, line 67:
```suggestion
```
We know what run_with_retries does, it doesn't help, and anyhow, don't assume it'll be a sec.

server/svix-server/src/worker.rs
svix-dylan: > If any future returns an error then all other futures will be canceled and an error will be returned immediately.

This is from the docs for `try_join_all`. `join_all` does not cancel prematurely, and I have to manually check the returned vector for error types (done below this line).

tasn: Man, I literally copy-pasted it from the doc, I don't see how I could have looked at the wrong place, but it looks like I have. 🤷🏻

tasn (approved):

PR #433, opened 2022-05-06
Tests: Remove queue overrides / limit concurrency

jaymell:
Ensure all backend variations are properly end-to-end tested
by removing in-memory queue override; this currently
requires limiting test concurrency to 1 because tests are
currently sharing backend services.


server/run-tests.sh
```diff
@@ -2,7 +2,7 @@
 
 # Run tests with various configurations:
 
-TEST_COMMAND="cargo test --all --all-features --all-targets"
+TEST_COMMAND="cargo test --all --all-features --all-targets -- --test-threads=1"
```
tasn, line 5: I could have sworn we had this in already. 🤔

server/testing-queue.conf
```diff
@@ -0,0 +1,17 @@
+include classpath("application.conf")
```
tasn, line 1: What dis?

server/run-tests.sh
jaymell: It was added, then removed in later commit.

server/testing-queue.conf
jaymell: Something that shouldn't have been added. Removed.

server/run-tests.sh
tasn: Ah cool.

tasn (approved):

PR #438, opened 2022-05-16
Write migration for the legacy processing queue to stream-based queue

svix-daniel:
This PR sets up a queue migration for the v2 processing queue and reinserts the members directly in the v3 main queue.

While it would have been preferable to migrate to the pending list of the main queue, this did not seem possible without introducing race conditions with other consumers from inserting it in the main queue first.

### DIFF
--- a/server/svix-server/src/queue/redis.rs
+++ b/server/svix-server/src/queue/redis.rs
@@ -23,6 +23,10 @@
 //! processed now, and the currently processing queue for tasks that have timed out and should be
 //! put back on the main queue.
 
+// This lint warns on `let _: () = ...` which is used throughout this file for Redis commands which
+// have generic return types. This is cleaner than the turbofish operator in my opinion.
+#![allow(clippy::let_unit_value)]
+
 use std::{num::NonZeroUsize, time::Duration};
 
 use axum::async_trait;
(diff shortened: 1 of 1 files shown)


server/svix-server/src/queue/redis.rs
```diff
@@ -23,6 +23,10 @@
 //! processed now, and the currently processing queue for tasks that have timed out and should be
 //! put back on the main queue.
 
+// This lint warns on `let _: () = ...` which is used throughout this file for Redis commands which
+// have generic return types. This is cleaner than the turbofish operator in my opinion.
+#![allow(clippy::let_unit_value)]
```
tasn, line 28: Why not have it global to all of our project?

server/svix-server/src/queue/redis.rs
```diff
@@ -23,6 +23,10 @@
 //! processed now, and the currently processing queue for tasks that have timed out and should be
```
tasn, line 23: This PR needs a better description. Explain the approach you took and why it's good.

server/svix-server/src/queue/redis.rs
svix-daniel: I generally agree with the lint when the turbofish isn't involved, ie. if there is a function that specifically doesn't return anything, it's cleaner to just `call_it()` instead of `let _: () = call_it()`. But it's only in this file where the warning has occurred because of generic return types on Redis commands.

If this pops up elsewhere I'd be okay with making it apply to the whole crate.

tasn (approved):
When you squash+merge it, make sure to include the description in the commit message.

PR #440, opened 2022-05-17
Allow PATCH deletion of header values via null

svix-dylan:
Allows deletion of headers via the PATCH endpoint by passing
a null value for the header key to delete.


server/svix-server/src/v1/endpoints/endpoint/mod.rs
```diff
@@ -216,18 +216,25 @@ impl From<EndpointHeaders> for EndpointHeadersOut {
...
+        }
+
         model.headers = if let Some(Some(mut hdrs)) = model.headers.take() {
-            hdrs.0.extend(self.headers.0);
+            hdrs.0.extend(headers.0);
             Set(Some(hdrs))
         } else {
-            Set(Some(self.headers))
+            Set(Some(headers))
         };
```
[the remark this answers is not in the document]
svix-dylan: Yes and silver lining testing also caught it :)

server/svix-server/src/v1/endpoints/endpoint/mod.rs
```diff
@@ -216,18 +216,31 @@ impl From<EndpointHeaders> for EndpointHeadersOut {
...
 impl ModelIn for EndpointHeadersPatchIn {
     type ActiveModel = endpoint::ActiveModel;
 
     fn update_model(self, model: &mut Self::ActiveModel) {
+        let headers: EndpointHeaders = self
+            .headers
+            .0
+            .iter()
+            .filter(|(_, v)| v.is_some())
+            .map(|(k, v)| (k.clone(), v.clone().unwrap()))
```
svix-daniel, line 231: Could you do a `filter_map` here instead of `.filter(...).map(...)`?

svix-dylan: 👀 did not know that was a thing! Done.

tasn:
Just clicked rebase in the UI here, wanted to see what it does. 😂

server/svix-server/src/core/types.rs
```diff
@@ -526,36 +526,72 @@ pub struct ExpiringSigningKey {
...
-    const FORBIDDEN_PREFIXES: &'static [&'static str] = &[
-        "x-amz-", "x-amzn-", "x-google", "x-goog-", "x-gfe", "x-amz-", "x-azure-", "x-fd-",
-        "x-svix-", "svix-",
-    ];
+    pub fn new() -> Self {
+        EndpointHeaders(HashMap::new())
+    }
+}
+
+impl FromIterator<(String, String)> for EndpointHeaders {
```
tasn, line 587: Why is this needed all of a sudden?

server/svix-server/src/core/types.rs
```diff
@@ -526,36 +526,72 @@ pub struct ExpiringSigningKey {
...
-        "content-language",
-        "content-location",
-    ];
-    const FORBIDDEN_PREFIXES: &'static [&'static str] = &[
-        "x-amz-", "x-amzn-", "x-google", "x-goog-", "x-gfe", "x-amz-", "x-azure-", "x-fd-",
-        "x-svix-", "svix-",
-    ];
+    pub fn new() -> Self {
+        EndpointHeaders(HashMap::new())
+    }
```
tasn, line 584: Why is this needed all of a sudden?

server/svix-server/src/core/types.rs
```diff
@@ -603,6 +622,35 @@ impl Validate for EndpointHeaders {
...
+    where
+        D: serde::Deserializer<'de>,
+    {
+        HashMap::deserialize(deserializer)
+            .map(|x: HashMap<String, Option<String>>| x.into_iter().map(|(k, v)| (k, v)).collect())
+            .map(EndpointHeadersPatch)
+    }
+}
+
+impl Validate for EndpointHeadersPatch {
```
tasn, line 640: We need a test for this validation that fails validation.

server/svix-server/src/v1/endpoints/endpoint/mod.rs
```diff
@@ -216,18 +216,30 @@ impl From<EndpointHeaders> for EndpointHeadersOut {
...
+            .collect();
+
         model.headers = if let Some(Some(mut hdrs)) = model.headers.take() {
-            hdrs.0.extend(self.headers.0);
+            hdrs.0.extend(headers.0);
+            for (k, v) in self.headers.0.iter() {
+                if v.is_none() {
+                    hdrs.0.remove(k);
+                }
+            }
```
tasn, line 239: Isn't there a way to filter a hashmap directly (functionally) rather than imperatively?
Or I'm also thinking at this point that just dropping extend from above and having:
```rust
for (k, v) in haeders.0 {
    if v.is_none() { remove(k) }
    else { add(k, v) }
```

svix-dylan: In my understanding we need the `.extend()` because that is what's combining existing headers (which may not be in the PATCH request) with the headers that are in the PATCH request. I might be wrong, but from my reading of the Python code we wanted to combine those two _and then_ remove anything explicitly set to NULL, as to not constrain the new headers values to only be what was in the PATCH request.

I've updated the code so that the block you commented on is performed using an `iter().for_each()`, lmk.

server/svix-server/src/core/types.rs
svix-dylan: This is so that I can construct a new `EndpointHeaders` type that is compatible with the SeaORM model, from the `EndpointHeadersPatch` type.

tasn: But why wasn't it needed before your changes?

server/svix-server/src/v1/endpoints/endpoint/mod.rs
svix-dylan: Ok so are you okay with this breaking from how the Python implementation behaves? The Python implementation preserves preexisting headers that are not mentioned in the PATCH request. This suggested change would overwrite those headers with only what is mentioned in the PATCH request.

svix-dylan: Ah yeah I see, misread it on the first go around.

server/svix-server/src/core/types.rs
svix-dylan: `EndpointHeadersPatch` did not exist before this PR, and it is also incompatible with the SeaORM model. This implementation exists so I can map `EndpointHeadersPatch` to `EndpointHeaders` for SeaORM.

tasn: This is not for Patch, this is for EndpointHeaders which did exist before. That's my confusion. Why is it only needed for something that existed already and not for the new addition.

server/svix-server/src/core/types.rs
```diff
@@ -603,6 +616,35 @@ impl Validate for EndpointHeaders {
...
+pub struct EndpointHeadersPatch(pub HashMap<String, Option<String>>);
+json_wrapper!(EndpointHeadersPatch);
+
+impl<'de> Deserialize<'de> for EndpointHeadersPatch {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        HashMap::deserialize(deserializer)
+            .map(|x: HashMap<String, Option<String>>| x.into_iter().map(|(k, v)| (k, v)).collect())
```
tasn, line 629: Does the second map actually do anything?

server/svix-server/src/v1/endpoints/endpoint/mod.rs
```diff
@@ -216,18 +216,30 @@ impl From<EndpointHeaders> for EndpointHeadersOut {
...
             Set(Some(hdrs))
         } else {
-            Set(Some(self.headers))
+            Set(Some(
+                self.headers
+                    .0
+                    .iter()
+                    .filter_map(|(k, v)| v.as_ref().map(|v| (k.clone(), v.clone())))
```
tasn, line 240: Does it actually work? We have a test for it?

server/svix-server/src/core/types.rs
svix-dylan: It is required [here](https://github.com/svix/svix-webhooks/blob/8eac6be878bc8641f97ab8176e1b38bc70a6c691/server/svix-server/src/v1/endpoints/endpoint/mod.rs#L241).

tasn: Why wasn't it needed before though? That's the confusion.

server/svix-server/src/core/types.rs
svix-dylan: Nope! Although I think this was here before I came along. Patched.

server/svix-server/src/core/types.rs
svix-dylan: Following up on our convo: we can't collect into `EndpointHeadersIn` unless we do the same implementation for _that_ type.

PR #441, opened 2022-05-17
Test queue with many consumers

svix-daniel:
With the integration tests, this adds one test which spreads 1000 message tasks across 20 producer, consumer pairs such as to begin to test how the system would respond to multiple worker instances running at once.

With this change, I also introduced an optional prefix to the Redis queue keys such that tests can be better isolated from production environments as this test assumes an empty queue.

### DIFF
--- a/server/svix-server/src/lib.rs
+++ b/server/svix-server/src/lib.rs
@@ -6,7 +6,7 @@
 
 use axum::{extract::Extension, Router};
 
-use cfg::{CacheType, QueueType};
+use cfg::CacheType;
 use std::{
     net::{SocketAddr, TcpListener},
     str::FromStr,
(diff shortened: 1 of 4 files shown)


server/svix-server/tests/queue.rs
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    // Add 50 test messages with unique message IDs to each producer for a total of 1000 unique
+    // messagues
+    for (index, (p, _c)) in producers_and_consumers.iter().enumerate() {
+        for num in 1..=50 {
+            p.send(
+                QueueTask::MessageV1(MessageTask {
+                    msg_id: MessageId(format!("{}", index * 50 + num)),
+                    app_id: ApplicationId("TeatApplicationId".to_owned()),
```
svix-james, line 63: spelling :stuck_out_tongue:

svix-james (approved):

server/svix-server/src/queue/mod.rs
```diff
@@ -6,13 +6,38 @@ use serde::{Deserialize, Serialize};
...
+    cfg::{Configuration, QueueType},
     core::types::{ApplicationId, EndpointId, MessageAttemptTriggerType, MessageId},
     error::Result,
 };
 
 pub mod memory;
 pub mod redis;
 
+pub async fn new_pair(
+    cfg: Configuration,
```
tasn, line 18: One comment (I know it's already merged): this is not something we like doing and I quite dislike it.

We don't pass large structures like this to functions if we can avoid it (not sure we can in this case) because it makes things painful to test and painful to reason about.

server/svix-server/src/lib.rs
```diff
@@ -58,19 +58,7 @@ pub async fn run(cfg: Configuration, listener: Option<TcpListener>) {
...
-                queue::redis::new_pair(pool).await
-            }
-            QueueType::RedisCluster => {
-                let pool = crate::redis::new_redis_pool_clustered(redis_dsn()).await;
-                queue::redis::new_pair(pool).await
-            }
-            QueueType::Memory => queue::memory::new_pair().await,
-        }
-    };
+    let (queue_tx, queue_rx) = queue::new_pair(cfg.clone(), None).await;
```
tasn, line 61: This is unnecessary cloning the cfg, no bueno. Please change it to a ref.

PR #444, opened 2022-05-18
Build/release ARM64 Docker images on release

svix-daniel:
In CI, on the publish of a release, this will now automatically build and push ARM64 builds to Docker Hub to on the latest tag.

### DIFF
--- a/.github/workflows/server-release.yml
+++ b/.github/workflows/server-release.yml
@@ -65,9 +65,6 @@ jobs:
       - name: Setup QEMU
         uses: docker/setup-qemu-action@v2
 
-      - name: Setup Buildx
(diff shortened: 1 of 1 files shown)


.github/workflows/server-release.yml
```diff
@@ -62,6 +62,12 @@ jobs:
     steps:
       - uses: actions/checkout@master
 
+      - name: Setup QEMU
+        uses: docker/setup-qemu-action@v2
+
+      - name: Setup Buildx
```
tasn, line 68: You're already setting it up below, just a different version. Change it below?

tasn (approved):

PR #445, opened 2022-05-18
Add soft and hard caps to pagination limits

svix-daniel:
Adds soft and hard caps to the limit for pagination. This is configured via a constant in the `v1::utils` module.

With soft caps no error is returned exceeding the cap, but list responses will use the cap instead of an exceeding value. With hard caps a 422 error is returned if the given limit exceeds the cap.

### DIFF
--- a/server/svix-server/src/v1/endpoints/application.rs
+++ b/server/svix-server/src/v1/endpoints/application.rs
@@ -90,7 +90,7 @@ async fn list_applications(
     pagination: ValidatedQuery<Pagination<ApplicationId>>,
     permissions: Permissions,
 ) -> Result<Json<ListResponse<ApplicationOut>>> {
-    let limit = pagination.limit;
+    let limit = pagination.capped_limit();
     let iterator = pagination.iterator.clone();
 
     let mut query = application::Entity::secure_find(permissions.org_id)
--- a/server/svix-server/src/v1/endpoints/application.rs
+++ b/server/svix-server/src/v1/endpoints/application.rs
@@ -5,8 +5,8 @@ use crate::{
     core::types::{ApplicationId, ApplicationIdOrUid, ApplicationUid},
     error::{HttpError, Result},
     v1::utils::{
-        validate_no_control_characters, EmptyResponse, ListLimit, ListResponse, ModelIn, ModelOut,
-        ValidatedJson, ValidatedQuery,
+        validate_no_control_characters, EmptyResponse, ListResponse, ModelIn, ModelOut,
+        PaginationLimit, ValidatedJson, ValidatedQuery,
     },
 };
 use axum::{
@@ -90,7 +90,7 @@ async fn list_applications(
     pagination: ValidatedQuery<Pagination<ApplicationId>>,
     permissions: Permissions,
 ) -> Result<Json<ListResponse<ApplicationOut>>> {
-    let ListLimit(limit) = pagination.limit;
+    let PaginationLimit(limit) = pagination.limit;
     let iterator = pagination.iterator.clone();
 
     let mut query = application::Entity::secure_find(permissions.org_id)
--- a/server/svix-server/src/v1/endpoints/application.rs
+++ b/server/svix-server/src/v1/endpoints/application.rs
@@ -5,7 +5,7 @@ use crate::{
     core::types::{ApplicationId, ApplicationIdOrUid, ApplicationUid},
     error::{HttpError, Result},
     v1::utils::{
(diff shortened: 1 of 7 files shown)


server/svix-server/tests/utils/common_calls.rs
```diff
@@ -223,7 +223,34 @@ pub async fn common_test_list<
         .await
         .unwrap();
 
-    Ok(())
+    // Test Limits
+    for i in 10..300 {
```
tasn, line 227: Does it potentially make all the tests super slow to test it everywhere? Should we maybe only test it for Application?

How much slower does it make things?

server/svix-server/src/v1/endpoints/application.rs
```diff
@@ -90,7 +90,7 @@ async fn list_applications(
     pagination: ValidatedQuery<Pagination<ApplicationId>>,
     permissions: Permissions,
 ) -> Result<Json<ListResponse<ApplicationOut>>> {
-    let limit = pagination.limit;
+    let limit = pagination.capped_limit();
```
tasn, line 93: See my other comments, the `endpoints/*.rs` files probably shouldn't need to change.

server/svix-server/src/v1/utils.rs
```diff
@@ -24,14 +24,63 @@ const fn default_limit() -> u64 {
...
+
+#[derive(Debug, Clone, Copy, PartialEq)]
+enum PaginationLimitCap {
+    HardCap,
+
+    // This is configured via constant, so SoftCap will only be used when PAGINATION_LIMIT_CAP_TYPE
+    // is set to it
+    #[allow(dead_code)]
+    SoftCap,
+}
```
tasn, line 41: No need for this, it should really just be a boolean.

server/svix-server/tests/utils/common_calls.rs
svix-daniel: Test weren't noticeably slower -- there's no delay on this one because ordering doesn't matter

server/svix-server/tests/utils/common_calls.rs
```diff
@@ -181,7 +181,7 @@ pub async fn common_test_list<
     assert!(!list.done);
 
     let list = client
-        .get::<ListResponse<ModelOut>>(&format!("{}?limit=500", path), StatusCode::OK)
+        .get::<ListResponse<ModelOut>>(&format!("{}?limit=50", path), StatusCode::OK)
```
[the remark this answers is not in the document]
svix-daniel: Everything passed for me, it was just checking what happens if the limit was over the count I think. Regardless it errors if you keep it at 500 because it returns 422 now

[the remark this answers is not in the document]
tasn: Yeah, it should return 422, that's why I'm surprised this works.

server/svix-server/src/v1/utils.rs
```diff
@@ -20,18 +20,59 @@ use crate::{
...
 
+#[derive(Debug)]
+pub struct ListLimit(pub u64);
+
+impl<'de> Deserialize<'de> for ListLimit {
+    fn deserialize<D>(deserializer: D) -> std::result::Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let lim = u64::deserialize(deserializer)?;
```
tasn, line 50:
```suggestion
        let limit = u64::deserialize(deserializer)?;
```
...

server/svix-server/tests/utils/common_calls.rs
tasn: Btw, why 10..300? That's a bit surprising.

server/svix-server/src/v1/utils.rs
```diff
@@ -20,18 +20,59 @@ use crate::{
...
+    #[validate]
     #[serde(default = "default_limit")]
-    pub limit: u64,
+    pub limit: ListLimit,
     #[validate]
     pub iterator: Option<T>,
 }
 
+#[derive(Debug)]
+pub struct ListLimit(pub u64);
```
tasn, line 43: ListLimit was not a good name (sorry), can you search/replace to make it PaginationLimit?

server/svix-server/tests/utils/common_calls.rs
svix-daniel: 10 were created earlier in the tests and 300 was the first satisfying number above the cap of 250 :stuck_out_tongue:

tasn (approved):

PR #446, opened 2022-05-18
Keep Redis queue worker alive after Redis interruption

svix-dylan:
If the Redis connection is interrupted, the background loop over the
delayed queue should gracefully fail instead of exiting. With this change,
it will continue to operate once the Redis connection is healthy again.

### DIFF
--- a/server/svix-server/src/queue/redis.rs
+++ b/server/svix-server/src/queue/redis.rs
@@ -185,10 +185,16 @@ async fn new_pair_inner(
 
             // First look for delayed keys whose time is up and add them to the main qunue
             let timestamp = Utc::now().timestamp();
-            let keys: Vec<String> = pool
+            let keys: Vec<String> = match pool
                 .zrangebyscore_limit(&delayed_queue_name, 0isize, timestamp, 0isize, batch_size)
                 .await
-                .unwrap();
+            {
+                Ok(res) => res,
+                Err(e) => {
+                    tracing::error!("{}", e);
+                    continue;
+                }
+            };
             if !keys.is_empty() {
                 // FIXME: needs to be a transaction
                 let keys: Vec<(String, String)> = pool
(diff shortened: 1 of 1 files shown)


server/svix-server/src/queue/redis.rs
```diff
@@ -106,6 +106,111 @@ pub async fn new_pair(
     .await
 }
 
+async fn background_task(
+    pool: RedisPool,
+    main_queue_name: String,
+    delayed_queue_name: String,
+    pending_duration: i64,
+) -> Result<()> {
+    // FIXME: enforce we only have one such worker via locking
```
tasn, line 115: This FIXME should be moved back to where this function is called from.

server/svix-server/src/queue/redis.rs
```diff
@@ -106,6 +106,111 @@ pub async fn new_pair(
...
+        // Then for each task, XADD them to ghe MAIN queue
+        let mut pipe = redis::pipe();
+        for task in tasks {
+            let _ = pipe.xadd(
+                &main_queue_name,
+                GENERATE_STREAM_ID,
+                &[(QUEUE_KV_KEY, task)],
+            );
+        }
+        let _: () = pool.query_async_pipeline(pipe).await.unwrap();
```
tasn, line 149: This still unwraps so it can kill the whole thing.

server/svix-server/src/queue/redis.rs
```diff
@@ -106,6 +106,111 @@ pub async fn new_pair(
...
+        let claimed: StreamClaimReply = pool
+            .query_async(Cmd::xclaim(
+                &main_queue_name,
+                WORKERS_GROUP,
+                WORKER_CONSUMER,
+                pending_duration,
+                &ids,
+            ))
+            .await
+            .unwrap();
```
tasn, line 184: This still unwraps so it can kill the whole thing.

server/svix-server/src/queue/redis.rs
```diff
@@ -106,6 +106,111 @@ pub async fn new_pair(
...
+async fn background_task(
+    pool: RedisPool,
+    main_queue_name: String,
+    delayed_queue_name: String,
+    pending_duration: i64,
+) -> Result<()> {
+    // FIXME: enforce we only have one such worker via locking
+    let batch_size: isize = 50;
+
+    let mut pool = pool.get().await.unwrap();
```
tasn, line 118: This still unwraps so it'll kill the whole thing, no?

tasn (commented):
I put some comments on likely-to-fail unwraps.

tasn (approved):

PR #447, opened 2022-05-19
Code coverage to GitHub Actions

svix-dylan:
Adds code coverage to our CI pipeline via the grcov tool
and GHA integration.


.github/workflows/server-ci.yml
```diff
@@ -42,6 +42,22 @@ jobs:
...
+    - name: grcov
+      uses: actions/checkout@v1
+      uses: actions-rs/toolchain@v1
+        with:
+          toolchain: nightly
+          override: true
+      uses: actions-rs/cargo@v1
+        with:
+          command: test
+          args: --all-features --no-fail-fast --manifest-path server/Cargo.toml --all -- --check
```
tasn, line 55: This runs the tests before the deps have been loaded (so many tests will fail), and also, because you are running it this way we are not hitting all of the different variations...

PR #448, opened 2022-05-19
Code coverage tooling

svix-dylan:


server/run-tests-with-grcov.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh -e
+
+# install deps
+cargo install grcov
+rustup component add llvm-tools-preview
```
tasn, line 5: This doesn't need to be run every time...

server/run-tests-with-grcov.sh
```diff
@@ -0,0 +1,20 @@
...
+
+# set env vars
+RUSTFLAGS="-Cinstrument-coverage"
+LLVM_PROFILE_FILE="svix-webhooks-%p-%m.profraw"
+RUSTC_BOOTSTRAP=1
```
tasn, line 10: Can you document what it does?

server/run-tests-with-grcov.sh
svix-dylan: These commands don't redownload or reinstall if the dependencies already exist on the system. Including them here just reduces manual installation. Can remove if wanted.

tasn: I think it's better if you added a section to the server README about code coverage with instructions to install this + how to run.

I'd probably also remove `open` from the script and just print out `open THE_PATH.html`

server/run-tests-with-grcov.sh
```diff
@@ -0,0 +1,20 @@
...
+# tell Rust to run with coverage instrumentation
+RUSTFLAGS=\"-Cinstrument-coverage\"
+# give grcov a profile name template for output files
+LLVM_PROFILE_FILE=\"svix-webhooks-%p-%m.profraw\"
+# put the compiler in nightly mode
+RUSTC_BOOTSTRAP=1
+
+# run tests
+./run-tests.sh
+" || true
```
tasn, line 14: Why are you doing this in a subscript like this?

server/run-tests-with-grcov.sh
```diff
@@ -1,20 +1,20 @@
...
-# back to stable
-RUSTC_BOOTSTRAP=0
+" || true
 
 # generate and open report output
 grcov . -s . --binary-path ./target/debug/ -t html --branch --ignore-not-existing -o ./target/debug/coverage/
-open ./target/debug/coverage/index.html
+
+LOCATION="coverage HTML file at $(pwd)/target/debug/coverage/index.html"
+echo "$LOCATION"
```
tasn, line 20:
```suggestion
echo "coverage HTML file at $(pwd)/target/debug/coverage/index.html"
```

server/run-tests-with-grcov.sh
svix-dylan: So that these variables are only set in this new session, and don't affect rustc in any subsequent commands a developer might run unrelated to this script.

server/run-tests-with-grcov.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh -e
```
jaymell, line 1: Is the `-e` needed here?

server/run-tests-with-grcov.sh
tasn: You don't need to do anything, it won't affect anything outside of the script.
Also, I just noticed you didn't run `export` so actually these won't even work at all.

server/run-tests-with-grcov.sh
```diff
@@ -0,0 +1,19 @@
...
+# tell Rust to run with coverage instrumentation
+RUSTFLAGS=\"-Cinstrument-coverage\"
+# give grcov a profile name template for output files
+LLVM_PROFILE_FILE=\"svix-webhooks-%p-%m.profraw\"
+# put the compiler in nightly mode
+RUSTC_BOOTSTRAP=1
+
+# run tests
+./run-tests.sh
+" || true
```
tasn, line 14:
```suggestion
# tell Rust to run with coverage instrumentation
export RUSTFLAGS="-Cinstrument-coverage"
# give grcov a profile name template for output files
export LLVM_PROFILE_FILE="svix-webhooks-%p-%m.profraw"

export TEST_NIGHTLY="+nightly"

# run tests
./run-tests.sh
```

And then in run-tests:
```
TEST_COMMAND="cargo ${TEST_NIGHTLY} test --all --all-features --all-targets -- --test-threads=1"
```

server/README.md
```diff
@@ -136,3 +136,9 @@ The easiest way to get these tests to pass is to:
 Alternatively, if you're only interested in running unit tests, you can just run `cargo test --lib`. These tests don't make any assumptions about the surrounding environment.
 
 To run only a specific test (e.g. only the application tests), you can use the `--test` flag to `cargo test` which supports common Unix glob patterns. For example: `cargo test --test '*app*'`.
+
+# Code Coverage
+To get a code coverage report, first make sure you have the following installations:
+- `cargo install grcov`
+- `rustup component add llvm-tools-preview`
```
tasn, line 143: This won't work if default is not set to nightly. Probably just need to say "use nightly" which will also simplify the rest.

server/run-tests-with-grcov.sh
tasn: Or better yet, just tell people to use nightly (see other comment), and then:
```
#!/bin/sh -e

# tell Rust to run with coverage instrumentation
export RUSTFLAGS="-Cinstrument-coverage"
# give grcov a profile name template for output files

export DIRNAME="./target/debug/coverage/$(date -u --iso-8601=seconds)"
mkdir -p "$DIRNAME"
export LLVM_PROFILE_FILE="$DIRNAME/svix-webhooks-%p-%m.profraw"

# run tests
./run-tests.sh

# generate and open report output
~/.cargo/bin/grcov "$DIRNAME" -s . --binary-path ./target/debug/ -t html --branch --ignore-not-existing -o "$DIRNAME"

echo "coverage HTML file at $(pwd)/target/debug/coverage/index.html"
```

tasn (commented):
Let's not worry about it for now, just dumped my thoughts.

server/README.md
svix-dylan: I ran both of these install commands on stable successfully

tasn: Yeah, they install fine, but I had issues. Though maybe my issues were unrelated.

svix-dylan:
Adding some thoughts here. Apologies about some confusion earlier in this PR, I was playing with multiple strategies and artifacts from both were left in the commits.

grcov uses one of two strategies to generate code coverage: source-based or "gcov"-based coverage (presumably via the gcc utility). Some internal concerns were raised that our code coverage reports were skewed due to the code's use of macros. Perhaps this is an effect of using source-based coverage (which this PR has only ever implemented, despite the presence of `RUSTC_BOOTSTRAP=1` and my own misinformed comments there)? If this is the case, we can attempt to use gcov-based coverage instead, though that will require code coverage be generated using rustc nightly.

PR #450, opened 2022-05-20
Additional testing

svix-dylan:
Adds testing that uses a small test server to verify message delivery, redelivery, and accurate response recording.

Not formally part of this PR but worth mentioning that along with #446, we have verified through manual testing that background worker loops will recover from both standalone Redis connection failures and Postgres connection failures.

TODO: manually test Redis Cluster connection failures. Local environment issues, will update once confirmed or provide a patch in this PR.

### DIFF
--- a/server/svix-server/tests/e2e_attempt.rs
+++ b/server/svix-server/tests/e2e_attempt.rs
@@ -94,9 +94,9 @@ async fn test_list_attempted_messages() {
             }
 
             // Assert order
-            assert_eq!(list.data[0].msg, msg_3);
-            assert_eq!(list.data[1].msg, msg_2);
-            assert_eq!(list.data[2].msg, msg_1);
+            if list.data[0].msg != msg_3 || list.data[1].msg != msg_2 || list.data[2].msg != msg_1 {
+                anyhow::bail!("test order failed");
+            }
         }
 
         Ok(())
--- a/server/svix-server/tests/utils/mod.rs
+++ b/server/svix-server/tests/utils/mod.rs
@@ -242,36 +242,45 @@ pub struct TestReceiver {
 
 impl TestReceiver {
     pub fn start(resp_with: axum::http::StatusCode) -> Self {
-        let listener = TcpListener::bind("127.0.0.1:0").unwrap();
-        let endpoint = format!("http://{}/", listener.local_addr().unwrap());
+        generate_axum_server(resp_with, None)
+    }
 
-        let (tx, data_recv) = mpsc::channel(32);
-        let (header_tx, header_recv) = mpsc::channel(32);
+    pub fn start_on_port(resp_with: axum::http::StatusCode, port: i32) -> Self {
+        generate_axum_server(resp_with, Some(port))
+    }
+}
 
-        let jh = tokio::spawn(async move {
-            axum::Server::from_tcp(listener)
-                .unwrap()
-                .serve(
-                    axum::Router::new()
-                        .route(
-                            "/",
-                            axum::routing::post(test_receiver_route).get(test_receiver_route),
-                        )
-                        .layer(axum::extract::Extension(tx))
-                        .layer(axum::extract::Extension(header_tx))
-                        .layer(axum::extract::Extension(resp_with))
-                        .into_make_service(),
-                )
-                .await
-                .unwrap();
-        });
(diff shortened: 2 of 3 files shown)


server/svix-server/tests/e2e_attempt.rs
```diff
@@ -92,11 +92,6 @@ async fn test_list_attempted_messages() {
             if list.data.len() != 3 {
                 anyhow::bail!("list len {}, not 3", list.data.len());
             }
-
-            // Assert order
-            assert_eq!(list.data[0].msg, msg_3);
-            assert_eq!(list.data[1].msg, msg_2);
-            assert_eq!(list.data[2].msg, msg_1);
```
svix-dylan, line 99: This part of the test is being removed here since we can no longer guarantee the order of this response data, and the test fails periodically on ‘main’ due to race condition. If the order is important here and we want to keep this test, we can either

- Revert the change that caused this issue (writing messagedestination in worker instead of the handler), or
- Examine/modify MessageAttempt response and its dependency on messagedestination

tasn: You don't need to verify order, but you can still check they all delivered.

tasn: Now that I think about it, why isn't order consistent anymore?

server/svix-server/tests/utils/mod.rs
```diff
@@ -242,36 +242,45 @@ pub struct TestReceiver {
...
-        }
+        generate_axum_server(resp_with, None)
+    }
+
+    pub fn start_on_port(resp_with: axum::http::StatusCode, port: i32) -> Self {
+        generate_axum_server(resp_with, Some(port))
+    }
+}
+
+fn generate_axum_server(resp_with: axum::http::StatusCode, port: Option<i32>) -> TestReceiver {
```
tasn, line 253: This is still an open topic (from the discussion with James too), just make the behavior controllable (heck, we can even just support passing a custom `Fn`) so that you can make it succeed/fail whenever you want. It can be as simple as calling a function that changes the "return status code" from 200 to 403.

server/svix-server/tests/e2e_attempt.rs
svix-dylan: So this test started being an issue when we moved the messagedestination write from the create message handler, to the worker - messagedestinations is no longer necessarily written in the order messages were received. If you take a look at the `list_attempted_messages` handler, there is a dependency on messagedestination.

That's as far as I've gotten on the issue. The test started being an issue after the handler->worker write was committed.

tasn: Ohh, sorry, I thought it was checking the *msgs*, not the attempted messages. All good. Yeah, we can no longer guarantee order, that's fine.

server/svix-server/tests/utils/mod.rs
svix-dylan: The hiccup here is, it does not seem that both the struct TestReceiver and an `axum.layer` consumption can share some structure indicating status code.

tasn: Mutex, RwLock, Arc, etc. There are many tools for sharing structures.

svix-james (approved):

server/svix-server/tests/utils/mod.rs
```diff
@@ -248,20 +258,24 @@ impl TestReceiver {
...
+
+        let routes = axum::Router::new()
+            .route(
+                "/",
+                axum::routing::post(test_receiver_route).get(test_receiver_route),
+            )
+            .layer(axum::extract::Extension(tx))
+            .layer(axum::extract::Extension(header_tx))
+            .layer(axum::extract::Extension(response_status_code.clone()))
+            .into_make_service();
```
tasn, line 273: This is an unrelated change that makes the diff annoying, move it back to where it was.

tasn:
Made one comment, but looks good otherwise.

server/svix-server/tests/utils/mod.rs
svix-dylan: I did not just do this, the compiler complains about response_status_code because the closure is `move`. I suppose I could clone response_status_code to another declared variable outside the closure and use that inside it.

I think this way looks better in the code at the end of the day, that way looks better for the diff.

tasn: OK.

tasn (approved):

PR #455, opened 2022-05-25
Csharp: make options optional for all list endpoinst.

tasn:
This is following the discussion in:
https://github.com/svix/svix-webhooks/pull/248\#issuecomment-1137223221


tasn:
@IAmArchaic, does this look right?

tasn:
Perfect, thanks!

PR #459, opened 2022-05-25
Add test for many consumers with delayed messages

svix-daniel:
Previously only many producers/consumers with non-delayed messages was tested. This simply repeats that test, but with a 500ms delay on every message to guarantee they go through the delayed queue without conflict.

### DIFF
--- a/server/svix-server/tests/queue.rs
+++ b/server/svix-server/tests/queue.rs
@@ -29,10 +29,7 @@ pub async fn get_pool(cfg: Configuration) -> RedisPool {
     }
 }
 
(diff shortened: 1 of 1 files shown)


tasn (approved):
1. Are you sure that there are many "scheduled workers" running in parallel and we don't accidentally only run one?
2. Do the events get evenly distributed between these scheduled workers (not the main workers, the scheduled ones that requeue the events)?

svix-daniel:
Via additional local testing which provided more insight to the internals than the test could get by itself, I have asserted that many of these workers run in parallel and that, while not evenly distributed as several are popped at a given moment, that multiple workers are processing keys simultaneously.

PR #461, opened 2022-05-25
Enable running tests in parallel

svix-daniel:
Previously tests could not run in parallel because multiple Redis queues would steal tasks from each other. This adds a unique prefix to test queue keys allowing them to run in parallel.

### DIFF
--- a/server/run-tests.sh
+++ b/server/run-tests.sh
@@ -2,7 +2,7 @@
 
 # Run tests with various configurations:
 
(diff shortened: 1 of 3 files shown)


tasn (approved):

tasn:
Looks like it took the test runtime down to 3m from 4:24m. At least compared to https://github.com/svix/svix-webhooks/pull/459

PR #477, opened 2022-05-31
Add basic status checks to health endpoint

svix-daniel:
Moves the existing health endpoint to `/heartbeart/` and adds a new health endpoint that performs basic checks on the status of PostgreSQL, the internal queue, and the internal cache.


tasn:
Lol, you have a typo in the description (it should be "heartbeat") that really threw me off for a few seconds.

server/svix-server/src/v1/endpoints/health.rs
```diff
@@ -1,12 +1,88 @@
...
+    Operational,
+    Error(String),
+}
+
+#[derive(Debug, Deserialize, Serialize)]
+pub struct HealthReport {
+    pg_status: HealthStatus,
+
+    queue_status: HealthStatus,
+    cache_status: HealthStatus,
```
tasn, line 30:
```suggestion
    database: HealthStatus,

    queue: HealthStatus,
    cache: HealthStatus,
```
1. status postfix is redundant.
2. pg is cryptic, let's call it database.

server/svix-server/src/v1/endpoints/health.rs
```diff
@@ -1,12 +1,88 @@
...
+
+    // Send a [`HealthCheck`] through the queue
+    let queue_status = if let Err(e) = queue_tx.send(QueueTask::HealthCheck, None).await {
+        HealthStatus::Error(e.to_string())
+    } else {
+        HealthStatus::Operational
+    };
+
+    // Set a cache value
+    let cache_status = if let Err(e) = cache
```
tasn, line 63: Caching it beats the point. :)

The whole idea is to get a status of the system.

server/svix-server/src/queue/mod.rs
```diff
@@ -91,33 +91,37 @@ impl MessageTaskBatch {
...
         match self {
-            QueueTask::MessageV1(task) => task.msg_id,
-            QueueTask::MessageBatch(batch) => batch.msg_id,
+            QueueTask::HealthCheck => None,
+            QueueTask::MessageV1(task) => Some(task.msg_id),
+            QueueTask::MessageBatch(batch) => Some(batch.msg_id),
         }
     }
-    pub fn to_msg_task(self, endpoint_id: EndpointId) -> MessageTask {
+    pub fn to_msg_task(self, endpoint_id: EndpointId) -> Option<MessageTask> {
```
tasn, line 107: Yuck...

Also, given that these functions are only used once each (based on your code), I think having the `match` be inline where they are used is MUCH more clear. I'd just have the match at the top of the worker and set *both* msg_id and trigger_type into local variables directly there.

server/svix-server/src/v1/endpoints/health.rs
```diff
@@ -1,12 +1,88 @@
...
+    queue::{QueueTask, TaskQueueProducer},
+};
+
+async fn heartbeat() -> StatusCode {
     StatusCode::NO_CONTENT
 }
 
+#[derive(Debug, Deserialize, Serialize)]
+pub enum HealthStatus {
+    Operational,
```
tasn, line 21:
```suggestion
    Ok,
```
More concise and more common in this kind of scenarios.

server/svix-server/src/v1/endpoints/health.rs
```diff
@@ -1,12 +1,88 @@
...
+    core::cache::{kv_def, Cache, CacheBehavior, CacheKey, CacheValue},
+    queue::{QueueTask, TaskQueueProducer},
+};
+
+async fn heartbeat() -> StatusCode {
     StatusCode::NO_CONTENT
 }
 
+#[derive(Debug, Deserialize, Serialize)]
+pub enum HealthStatus {
```
tasn, line 20: I think the problem with using an enum here is that I'm not sure this is actually what we want.

We probably want (in terms of the output json) something like:

```json
{
    "database": {
        "status": "ok",
        "information": "Can be an error string on error, and 'OK' on success."
    },
    "queue": ...,
    "cache": ...,
}
```
We can also potentially add "latency" to the HealthStatus later on, and stuff like that.

server/svix-server/src/v1/endpoints/health.rs
```diff
@@ -1,12 +1,88 @@
...
+async fn health(
+    Extension(pg): Extension<DatabaseConnection>,
+    Extension(queue_tx): Extension<TaskQueueProducer>,
+    Extension(cache): Extension<Cache>,
+) -> Json<HealthReport> {
+    // SELECT 1 FROM any table
+    let pg_status = if let Err(e) = pg
+        .execute(Statement::from_string(
+            DatabaseBackend::Postgres,
+            "SELECT 1 FROM endpoint".to_owned(),
```
tasn, line 46:
```suggestion
            "SELECT 1".to_owned(),
```
This is sufficient. :)

server/svix-server/src/v1/endpoints/health.rs
svix-daniel: Would you prefer a struct, or an internally tagged enum via Serde? Both would result in that format, except with the enum, you could skip the information field or anything else irrelevant to an Ok response.

tasn: I think I'd do a struct because we just want more control and we'll probably share the fields anyway, as in we always want information (the error message or a success string), we always want latency (when we add it), etc.

Status can just be an enum though.

server/svix-server/src/worker.rs
```diff
@@ -268,15 +270,30 @@ async fn process_task(
     queue_tx: &TaskQueueProducer,
     queue_task: QueueTask,
 ) -> Result<()> {
-    let msg = message::Entity::find_by_id(queue_task.clone().msg_id())
+    // FIXME: This function has a ridiculous amount of cloning, this should try to be eliminated
```
tasn, line 273: This FIXME doesn't belong in this PR.

server/svix-server/src/v1/endpoints/health.rs
```diff
@@ -1,12 +1,125 @@
...
+                ..
+            }
+        )
+    }
+}
+impl<O, E: std::error::Error> From<Result<O, E>> for HealthStatus {
+    fn from(res: Result<O, E>) -> Self {
+        match res {
+            Ok(_) => HealthStatus::new_ok(),
+            Err(e) => HealthStatus::new_error(e.to_string()),
```
tasn, line 60: We probably don't want this as it may contain private information (potentially).
So you need to have static strings: "the database timed out", "got an error", "whatever".

Though I think that maybe step one we can just drop information altogether. What do you think?

tasn (approved):

PR #484, opened 2022-06-06
Implement basic wait-for functionality

### LINKED ISSUE #334: Move the wait-for functionality directly into the daemon
Based on this comment: https://github.com/svix/svix-webhooks/issues/302#issuecomment-1067191833

The main advantage is that we can know the correct server was setup even when not using `wait-for` in docker.

Thoughts:
- Behavior should probably be controlled with a flag (timeout) - disabled, infinite, whatever.
- Check if the libs themselves can handle it, or whether we need to.
- Can the "is active" check be a real connection, or should we just do a TCP connection check?

svix-daniel:
On forming a connection to PostgreSQL or Redis and failing, the server will no longer immediately panic as long as the `--wait-for` flag is passed. Instead it will loop forming a connection until the configured timeout period (set via `--wait-for <duration in seconds>`) has been reached.

This has been tested manually and works as described. Programmatic testing is tricky with this PR, so it is not included.

Fixes #334


server/Dockerfile
```diff
@@ -34,15 +34,13 @@ RUN apt-get update ;\
 USER appuser
```
tasn, line 34: I'm commenting here so there's an easy thread.

I'm not a huge fan of this implementation. Look at all of the files and interfaces you had to change to make it, and how intrusive this change is. All of it in order to just make a small that was previously external and backend agnostic.

You now have to implement it for every new backend we add for cache, queue, and whatever, and the API is uglier for it.

I can't help but think that having literally just this in a function that's called from `main` would have been better:
```rust
for i in 1..=times_to_retry {
    if let Ok(_) = tcp_try_connect(hostname, port).await {
        break;
    }
    sleep(backoff_sleep).await;
}
```

server/svix-server/src/main.rs
```diff
@@ -115,5 +123,80 @@ async fn main() {
...
+
+            let sleep = tokio::time::sleep(Duration::from_secs(wait_for_seconds));
+            tokio::pin!(sleep);
+            loop {
+                tokio::select! {
+                    _ = tokio::net::TcpStream::connect((redis_host, redis_port)) => {
+                        break;
+                    },
+                    _ = &mut sleep => {
+                        panic!("Waiting for databases timed out");
```
tasn, line 194: Need to know which one.

server/svix-server/src/main.rs
```diff
@@ -115,5 +123,80 @@ async fn main() {
...
+                    .next()
+                    .expect("Expected at least one IP address")
+
+                }
+                url::Host::Ipv4(ipv4_addr) => ipv4_addr.into(),
+                url::Host::Ipv6(ipv6_addr) => ipv6_addr.into(),
+            };
+
+            let sleep = tokio::time::sleep(Duration::from_secs(wait_for_seconds));
+            tokio::pin!(sleep);
```
tasn, line 187: Unless I'm misunderstanding what's going on: no need to pin, just move it to line 193.

server/svix-server/src/main.rs
```diff
@@ -115,5 +123,80 @@ async fn main() {
...
+
+                let resp = resolver.lookup_ip(s).await.expect("DNS error");
+                resp.iter()
+                    .next()
+                    .expect("Expected at least one IP address")
+
+                }
+                url::Host::Ipv4(ipv4_addr) => ipv4_addr.into(),
+                url::Host::Ipv6(ipv6_addr) => ipv6_addr.into(),
+            };
```
tasn, line 184: Resolver code should be a helper function, it's exactly the same as the one above.

server/svix-server/src/lib.rs
```diff
@@ -31,6 +31,7 @@ pub mod expired_message_cleaner;
 pub mod queue;
 pub mod redis;
 pub mod v1;
+pub mod wait_for;
```
tasn, line 34: Use the mod from `main.rs`, no need to have it in lib.

server/svix-server/src/wait_for.rs
```diff
@@ -0,0 +1,78 @@
...
+
+pub async fn wait_for(cfg: &Configuration, wait_for_seconds: u64) {
+    if let Some(redis_dsn) = &cfg.redis_dsn {
+        tokio::join!(
+            wait_for_inner(&cfg.db_dsn, POSTGRES_NAME, POSTGRES_PORT, wait_for_seconds),
+            wait_for_inner(redis_dsn, REDIS_NAME, REDIS_PORT, wait_for_seconds),
+        );
+    } else {
+        wait_for_inner(&cfg.db_dsn, POSTGRES_NAME, POSTGRES_PORT, wait_for_seconds).await
+    }
```
tasn, line 24: Use an array of futures or something. This is redundant.

server/svix-server/src/wait_for.rs
```diff
@@ -0,0 +1,78 @@
...
+    let port = dsn.port().unwrap_or(default_port);
+
+    let host: IpAddr = match host {
+        url::Host::Domain(s) => resolve_host(s).await,
+        url::Host::Ipv4(ipv4_addr) => ipv4_addr.into(),
+        url::Host::Ipv6(ipv6_addr) => ipv6_addr.into(),
+    };
+
+    let sleep = tokio::time::sleep(Duration::from_secs(wait_for_seconds));
+    tokio::pin!(sleep);
```
tasn, line 47: Why can't this just be in line 59 without the pin?

server/svix-server/src/wait_for.rs
```diff
@@ -0,0 +1,78 @@
...
+            res = tokio::net::TcpStream::connect((host, port)) => {
+                if res.is_err() {
+                    continue;
+                } else {
+                    break;
+                }
+            }
+
+            _ = &mut sleep => {
+                panic!("Waiting for {} timed out", dependency_name);
```
tasn, line 60:
```suggestion
                panic!("Waiting for host={host} ({ip_addr}) port={port} timed out");
```

server/svix-server/src/wait_for.rs
tasn, line 24: Nitpick: this feels like application logic, rather than `wait_for` logic, so it feels like it should be in `main.rs`.

server/svix-server/src/wait_for.rs
svix-daniel: The pin is needed so that the sleep time is persisted between loops. If you put the sleep in the loop body then it'll loop forever when the connection if failing immediately cause it'll never time out.

server/svix-server/Cargo.toml
```diff
@@ -54,6 +54,7 @@ time = { version = "0.3.9", features = [ "std" ]}
 futures = "0.3"
```
tasn, line 54: Also: remove the wait-for from the docker file (both its usage, and copying it over). Remove it from the repo too.
Make sure though that we wait for the same time as before in the docker file.

server/Dockerfile
```diff
@@ -41,15 +41,13 @@ RUN apt-get update ;\
...
 COPY --from=build /app/target/release/svix-server /usr/local/bin/svix-server
 
 # Ignoring this lint, because it's an embedded shell script
 # hadolint ignore=DL3025
 CMD \
     set -ex ; \
     if [ ! -z "$WAIT_FOR" ]; then \
-        wait-for -t 15 pgbouncer:5432 -- echo "Postgres ready" ; \
-        wait-for -t 15 redis:6379 -- echo "Redis ready" ; \
+		WAIT_FOR_ARG="--wait-for 15"; \
```
tasn, line 51:
```suggestion
        WAIT_FOR_ARG="--wait-for 15"; \
```

tasn (approved):

server/svix-server/src/main.rs
```diff
@@ -115,5 +130,26 @@ async fn main() {
...
+        if let Some(redis_dsn) = &cfg.redis_dsn {
+            wait_for.push(wait_for_dsn(
+                redis_dsn,
+                REDIS_NAME,
+                REDIS_PORT,
+                wait_for_seconds,
+            ));
+        }
+
+        futures::future::join_all(wait_for).await;
```
tasn, line 151: Will this actually `panic` the main thread? I think no, as we are doing nothing with the `Result`.

Given my comment from Slack:

> Btw, I just realized, not that it matters, but waiting for both at the same time doesn't do anything, because you have to wait for both anyway, and waiting is instant if it exists.

We probably should just `await.expect()` on both of the `wait_for_dsn` inline. What do you think?

PR #488, opened 2022-06-07
Implement Dashboard Access Endpoint

svix-daniel:
Implements the `/auth/dashboard-access/:app_id/` endpoint and associated permissions changes. This makes it so there are tokens valid only for some endpoints. This is meant so that someone using the Svix service may create a token for the owner of an application under their organization to perform limited operations on what falls under the scope of that application.

### DIFF
--- a/server/svix-server/src/core/idempotency.rs
+++ b/server/svix-server/src/core/idempotency.rs
@@ -371,7 +371,7 @@ mod tests {
     use super::IdempotencyService;
     use crate::core::{
         cache,
-        security::generate_token,
+        security::generate_org_token,
         types::{BaseId, OrganizationId},
     };
 
@@ -445,7 +445,7 @@ mod tests {
         // Generate a new token so that keys are unique
         dotenv::dotenv().ok();
         let cfg = crate::cfg::load().unwrap();
-        let token = generate_token(&cfg.jwt_secret, OrganizationId::new(None, None), None)
+        let token = generate_org_token(&cfg.jwt_secret, OrganizationId::new(None, None))
             .unwrap()
             .to_string();
 
@@ -548,7 +548,7 @@ mod tests {
         dotenv::dotenv().ok();
         let cfg = crate::cfg::load().unwrap();
 
-        let token = generate_token(&cfg.jwt_secret, OrganizationId::new(None, None), None)
+        let token = generate_org_token(&cfg.jwt_secret, OrganizationId::new(None, None))
             .unwrap()
             .to_string();
 
@@ -638,7 +638,7 @@ mod tests {
         // Generate a new token so that keys are unique
         dotenv::dotenv().ok();
         let cfg = crate::cfg::load().unwrap();
-        let token = generate_token(&cfg.jwt_secret, OrganizationId::new(None, None), None)
+        let token = generate_org_token(&cfg.jwt_secret, OrganizationId::new(None, None))
             .unwrap()
             .to_string();
 
--- a/server/svix-server/src/core/idempotency.rs
+++ b/server/svix-server/src/core/idempotency.rs
@@ -445,7 +445,7 @@ mod tests {
         // Generate a new token so that keys are unique
         dotenv::dotenv().ok();
         let cfg = crate::cfg::load().unwrap();
-        let token = generate_token(&cfg.jwt_secret, OrganizationId::new(None, None))
+        let token = generate_token(&cfg.jwt_secret, OrganizationId::new(None, None), None)
             .unwrap()
             .to_string();
 
@@ -548,7 +548,7 @@ mod tests {
         dotenv::dotenv().ok();
         let cfg = crate::cfg::load().unwrap();
 
-        let token = generate_token(&cfg.jwt_secret, OrganizationId::new(None, None))
+        let token = generate_token(&cfg.jwt_secret, OrganizationId::new(None, None), None)
             .unwrap()
             .to_string();
 
@@ -638,7 +638,7 @@ mod tests {
         // Generate a new token so that keys are unique
         dotenv::dotenv().ok();
         let cfg = crate::cfg::load().unwrap();
-        let token = generate_token(&cfg.jwt_secret, OrganizationId::new(None, None))
+        let token = generate_token(&cfg.jwt_secret, OrganizationId::new(None, None), None)
             .unwrap()
             .to_string();
 
--- a/server/svix-server/src/core/security.rs
+++ b/server/svix-server/src/core/security.rs
@@ -209,32 +209,29 @@ where
     }
 }
 
-pub fn generate_token(
+pub fn generate_org_token(keys: &Keys, org_id: OrganizationId) -> Result<String> {
+    let claims = Claims::with_custom_claims(
+        CustomClaim { organization: None },
+        Duration::from_hours(24 * 365 * 10),
+    )
+    .with_issuer(env!("CARGO_PKG_NAME"))
+    .with_subject(org_id.0);
+    Ok(keys.key.authenticate(claims).unwrap())
+}
+
+pub fn generate_app_token(
     keys: &Keys,
     org_id: OrganizationId,
-    app_id: Option<ApplicationId>,
+    app_id: ApplicationId,
 ) -> Result<String> {
-    // If there is an app ID to encode, the the org is under the organization custom claim and the
-    // application is the subject
-    let claims = if let Some(app_id) = app_id {
-        Claims::with_custom_claims(
-            CustomClaim {
-                organization: Some(org_id.0),
-            },
-            Duration::from_hours(24 * 28),
-        )
-        .with_issuer(env!("CARGO_PKG_NAME"))
-        .with_subject(app_id.0)
-    }
-    // With no app ID to encode, the org field is blank while the sub is the organization ID
-    else {
-        Claims::with_custom_claims(
-            CustomClaim { organization: None },
-            Duration::from_hours(24 * 365 * 10),
-        )
-        .with_issuer(env!("CARGO_PKG_NAME"))
-        .with_subject(org_id.0)
-    };
+    let claims = Claims::with_custom_claims(
+        CustomClaim {
+            organization: Some(org_id.0),
+        },
+        Duration::from_hours(24 * 28),
+    )
+    .with_issuer(env!("CARGO_PKG_NAME"))
+    .with_subject(app_id.0);
     Ok(keys.key.authenticate(claims).unwrap())
 }
 
--- a/server/svix-server/src/core/security.rs
+++ b/server/svix-server/src/core/security.rs
@@ -19,7 +19,7 @@ use crate::{
     error::{Error, HttpError, Result},
 };
 
-use super::types::{ApplicationIdOrUid, OrganizationId};
+use super::types::{ApplicationId, ApplicationIdOrUid, OrganizationId};
 
 /// The default org_id we use (useful for generating JWTs when testing).
 pub fn default_org_id() -> OrganizationId {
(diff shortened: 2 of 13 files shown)


server/svix-server/src/core/security.rs
```diff
@@ -114,10 +209,32 @@ where
     }
 }
 
-pub fn generate_token(keys: &Keys, org_id: OrganizationId) -> Result<String> {
-    let claims = Claims::create(Duration::from_hours(24 * 365 * 10))
+pub fn generate_token(
```
tasn, line 212: Split (or make it private and wrap it) to two functions:
* generate_org_tokne
* generate_app_token

Because otherwise it's not clear (when calling this function) what's the level of access.

server/svix-server/src/v1/endpoints/auth.rs
```diff
@@ -0,0 +1,45 @@
...
+    Extension(cfg): Extension<Configuration>,
+    Path(_app_id): Path<ApplicationIdOrUid>,
+    AuthenticatedOrganization { permissions, app }: AuthenticatedOrganization,
+) -> Result<Json<DashboardAccessOut>> {
+    let token = generate_token(&cfg.jwt_secret, permissions.org_id, Some(app.id.clone()))?;
+
+    let login_key = serde_json::to_vec(&serde_json::json!({
+        "appId": app.id,
+        "token": token,
+        "region": "eu"
```
tasn, line 30: Add a comment that this is not actually used.

server/svix-server/src/v1/endpoints/auth.rs
```diff
@@ -0,0 +1,45 @@
...
+    }))
+    .map_err(|_| HttpError::internal_server_errer(None, None))?;
+
+    let login_key = base64::encode(&login_key);
+
+    let url = format!("https://app.svix.com/login#key={}", login_key);
```
tasn, line 36: nitpick: can you move the `https://app.svix.com` to a constant? It doesn't really matter, but just better to not hardcode stuff in the code.

server/svix-server/src/core/idempotency.rs
```diff
@@ -445,7 +445,7 @@ mod tests {
         // Generate a new token so that keys are unique
```
tasn, line 445: This PR is also missing tests.

server/svix-server/src/core/security.rs
```diff
@@ -77,11 +128,83 @@ where
...
+#[async_trait]
+impl<B> FromRequest<B> for AuthenticatedOrganization
+where
+    B: Send,
+{
+    type Rejection = Error;
+
+    async fn from_request(req: &mut RequestParts<B>) -> Result<Self> {
+        let permissions = Permissions::from_request(req).await?;
+        if permissions.type_ == KeyType::Application {
```
tasn, line 144:
```suggestion
        if permissions.type_ != KeyType::Organization {
```
What if we add a new type called `Foobar`? You've now given them org level access.

server/svix-server/src/core/security.rs
```diff
@@ -77,11 +128,83 @@ where
...
+impl<B> FromRequest<B> for OrganizationAuthenticatedApplication
+where
+    B: Send,
+{
+    type Rejection = Error;
+
+    async fn from_request(req: &mut RequestParts<B>) -> Result<Self> {
+        let permissions = Permissions::from_request(req).await?;
+
+        if permissions.type_ == KeyType::Application {
```
tasn, line 179:
```suggestion
        if permissions.type_ != KeyType::Organization {
```
Same as above.

server/svix-server/src/core/security.rs
```diff
@@ -77,11 +128,83 @@ where
...
+        .one(db)
+        .await?
+        .ok_or_else(|| HttpError::not_found(None, None))?;
+        Ok(OrganizationAuthenticatedApplication { permissions, app })
+    }
+}
+
 pub struct AuthenticatedApplication {
     pub permissions: Permissions,
     pub app: application::Model,
```
tasn, line 210: I can't make a comment on the correct line because Github sucks, though you need to add a check to verify that the app_id in the JWT is either None, or is the same as the path's app_id.

server/svix-server/src/v1/endpoints/application.rs
```diff
@@ -148,18 +151,26 @@ async fn get_application(
     Path(app_id): Path<ApplicationIdOrUid>,
     permissions: Permissions,
 ) -> Result<Json<ApplicationOut>> {
-    let app = application::Entity::secure_find_by_id_or_uid(permissions.org_id, app_id)
-        .one(db)
-        .await?
-        .ok_or_else(|| HttpError::not_found(None, None))?;
+    // If someone authorized for only applications calls this endpoint, it will only return their
+    // application regardless of input in the path
```
tasn, line 155: You should return the ID in the path (so change this back). The case where the ID doesn't match should be handled by the permission check as I mentioned in security.rs.

server/svix-server/src/v1/endpoints/application.rs
```diff
@@ -148,18 +151,26 @@ async fn get_application(
     Path(app_id): Path<ApplicationIdOrUid>,
     permissions: Permissions,
```
tasn, line 152:
```suggestion
    AuthenticatedApplication ...
```

server/svix-server/src/v1/endpoints/application.rs
```diff
@@ -148,18 +151,26 @@ async fn get_application(
...
+    .ok_or_else(|| HttpError::not_found(None, None))?;
     Ok(Json(app.into()))
 }
 
 async fn update_application(
     Extension(ref db): Extension<DatabaseConnection>,
     Path(app_id): Path<ApplicationIdOrUid>,
     ValidatedJson(data): ValidatedJson<ApplicationIn>,
-    permissions: Permissions,
+    AuthenticatedOrganization { permissions }: AuthenticatedOrganization,
```
tasn, line 173: It should be OrganizationAuthenticatedApp I think, and then you don't need to access permissions directly.

In general, accessing permissions directly should be less good now that we have the code that authenticates in the struct extension thing.

server/svix-server/src/v1/endpoints/event_type.rs
```diff
@@ -22,10 +26,6 @@ use serde::{Deserialize, Serialize};
 use svix_server_derive::{ModelIn, ModelOut};
 use validator::Validate;
 
-use crate::core::security::Permissions;
-use crate::db::models::eventtype;
-use crate::v1::utils::Pagination;
```
tasn, line 27: How come these two are no longer needed?

server/svix-server/src/v1/endpoints/event_type.rs
```diff
@@ -191,7 +191,7 @@ async fn create_event_type(
 async fn get_event_type(
     Extension(ref db): Extension<DatabaseConnection>,
     Path(evtype_name): Path<EventTypeName>,
-    permissions: Permissions,
+    AuthenticatedOrganization { permissions }: AuthenticatedOrganization,
```
tasn, line 194: OK, the naming is really confusing, we need to do a better job at naming these three Authenticated variants. Any suggestions?

tasn: Idea:
* AuthenticatedOrganization - keep it.
* AuthenticatedApplication - keep it.
* AuthenticatedOrganizationWithApplication - instead of OrganizationAuthentiacedApplication

Yeah, it's a bit long, but it's only used standalone on its own line in parameters list so maybe not a biggie. What do you think?

server/svix-server/src/core/idempotency.rs
tasn, line 445: This PR needs a better description.
https://github.com/svix/svix-webhooks/pull/488#issuecomment-1148902722

server/svix-server/src/core/security.rs
```diff
@@ -77,11 +128,89 @@ where
...
+        match permissions.type_ {
+            KeyType::Organization => {}
+            KeyType::Application => {
+                return Err(HttpError::unauthorized(
+                    None,
+                    Some(
+                        "You are only permitted to perform operations under the Application type"
+                            .to_owned(),
+                    ),
+                )
```
tasn, line 191: It should be a `permission_denied` (so 403) not a 401.
And just say "insufficient access".

server/svix-server/src/core/security.rs
```diff
@@ -110,14 +239,47 @@ where
...
+
+        if let Some(permitted_app_id) = &permissions.app_id {
+            if permitted_app_id != &app.id {
+                return Err(HttpError::unauthorized(
+                    None,
+                    Some(
+                        "You are only permitted to perform operations under the Application type"
+                            .to_owned(),
+                    ),
+                )
```
tasn, line 251: The error we currently return is 404, which makes sense as the app doesn't exist (for all this auth token knows).
We can however return a 403 like above. Either works for me, though not this.

server/svix-server/src/core/security.rs
```diff
@@ -110,14 +239,47 @@ where
...
-pub fn generate_token(keys: &Keys, org_id: OrganizationId) -> Result<String> {
-    let claims = Claims::create(Duration::from_hours(24 * 365 * 10))
-        .with_issuer(env!("CARGO_PKG_NAME"))
-        .with_subject(org_id.0);
+pub fn generate_org_token(keys: &Keys, org_id: OrganizationId) -> Result<String> {
+    let claims = Claims::with_custom_claims(
+        CustomClaim { organization: None },
+        Duration::from_hours(24 * 365 * 10),
+    )
+    .with_issuer(env!("CARGO_PKG_NAME"))
```
tasn, line 265: Nit (since I already made other comments): have a constant variable called `JWT_ISSUER` that stores this.

server/svix-server/src/v1/endpoints/application.rs
```diff
@@ -145,10 +148,10 @@ async fn create_application(
 
 async fn get_application(
     Extension(ref db): Extension<DatabaseConnection>,
-    Path(app_id): Path<ApplicationIdOrUid>,
-    permissions: Permissions,
+    Path(_app_id): Path<ApplicationIdOrUid>,
```
tasn, line 151: If you are not using it, you can just remove it.

server/svix-server/src/v1/endpoints/application.rs
```diff
@@ -157,15 +160,13 @@ async fn get_application(
 
 async fn update_application(
     Extension(ref db): Extension<DatabaseConnection>,
-    Path(app_id): Path<ApplicationIdOrUid>,
+    Path(_app_id): Path<ApplicationIdOrUid>,
```
tasn, line 163: If you are not using it, you can just remove it.
```suggestion
```

server/svix-server/src/v1/endpoints/application.rs
```diff
@@ -176,7 +177,7 @@ async fn update_application(
 async fn delete_application(
     Extension(ref db): Extension<DatabaseConnection>,
     Path(app_id): Path<ApplicationIdOrUid>,
-    permissions: Permissions,
+    AuthenticatedOrganization { permissions }: AuthenticatedOrganization,
```
tasn, line 180: Why not also change this to AuthenticatedOrganizationWithApp and remove the manual fetch inside like you did above?

server/svix-server/src/v1/endpoints/attempt.rs
```diff
@@ -652,10 +652,10 @@ async fn resend_webhook(
     Extension(ref db): Extension<DatabaseConnection>,
     Extension(queue_tx): Extension<TaskQueueProducer>,
     Path((_app_id, msg_id, endp_id)): Path<(ApplicationIdOrUid, MessageIdOrUid, EndpointIdOrUid)>,
-    AuthenticatedApplication {
+    AuthenticatedOrganizationWithApplication {
```
tasn, line 655: This should actually remain AuthenticatedApplication. Applications are allowed to send their own.

server/svix-server/src/v1/endpoints/application.rs
tasn: And everywhere else where this applies.

tasn (commented):
Looks great, there's just one pending comment.

tasn (approved):

PR #493, opened 2022-06-08
Add endpoint stats endpoint

svix-daniel:
Adds the endpoint stats endpoint which returns metrics on the status of messages for that endpoint.

### DIFF
--- a/server/svix-server/src/v1/endpoints/endpoint/mod.rs
+++ b/server/svix-server/src/v1/endpoints/endpoint/mod.rs
@@ -9,8 +9,8 @@ use crate::{
     core::{
         security::AuthenticatedApplication,
         types::{
-            ApplicationIdOrUid, EndpointId, EndpointIdOrUid, EndpointUid, EventChannelSet,
-            EventTypeNameSet, MessageStatus,
+            ApplicationIdOrUid, BaseId, EndpointId, EndpointIdOrUid, EndpointUid, EventChannelSet,
+            EventTypeNameSet, MessageEndpointId, MessageStatus,
         },
     },
     db::models::messagedestination,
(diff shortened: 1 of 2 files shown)


server/svix-server/src/v1/endpoints/endpoint/mod.rs
```diff
@@ -262,6 +274,66 @@ impl ModelIn for EndpointHeadersPatchIn {
...
+            .collect::<HashMap<_, _>>();
+
+        Ok(Json(EndpointStatsOut {
+            success: query_out.remove(&MessageStatus::Success).unwrap_or(0),
+            pending: query_out.remove(&MessageStatus::Pending).unwrap_or(0),
+            fail: query_out.remove(&MessageStatus::Fail).unwrap_or(0),
+            sending: query_out.remove(&MessageStatus::Sending).unwrap_or(0),
+        }))
+    } else {
+        Err(HttpError::unauthorized(None, None).into())
```
tasn, line 333: No need to put the whole body in an if/else, just do the fetch/return-on-failure short-circuit like is done literally everywhere else.

Also, the error should be `not_found` not `unauthorized`...

tasn (commented):
Small comments, looks good otherwise.

server/svix-server/src/v1/endpoints/endpoint/mod.rs
```diff
@@ -262,6 +274,65 @@ impl ModelIn for EndpointHeadersPatchIn {
...
+                messagedestination::Column::Id.gte(MessageEndpointId::start_id(
+                    chrono::Utc::now() - chrono::Duration::days(28),
+                )),
+            )
+            .into_model::<EndpointStatsQueryOut>()
+            .all(db)
+            .await
+            .map_err(|e| {
+                tracing::error!("ERROR {}", e);
+                e
```
tasn, line 321: Why is this needed? We don't do it anywhere else.

tasn (approved):

PR #503, opened 2022-06-10
Add changing region in dashboard access endpoint

svix-daniel:
Creates a configuration variable for the region in which the dashboard access region can be changed.

### DIFF
--- a/server/svix-server/src/cfg.rs
+++ b/server/svix-server/src/cfg.rs
@@ -105,6 +105,28 @@ pub struct ConfigurationInner {
     pub api_enabled: bool,
     /// Should this instance run the message worker
     pub worker_enabled: bool,
+
+    #[serde(default = "Default::default")]
+    pub internal: InternalConfig,
+}
+
+#[derive(Clone, Debug, Deserialize)]
+pub struct InternalConfig {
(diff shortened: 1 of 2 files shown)


server/svix-server/src/cfg.rs
```diff
@@ -105,6 +105,28 @@ pub struct ConfigurationInner {
     pub api_enabled: bool,
     /// Should this instance run the message worker
     pub worker_enabled: bool,
+
+    #[serde(default = "Default::default", flatten)]
```
tasn, line 109: Why is this needed?

svix-daniel: It was so the structure doesn't have to appear in the default config file, but I'll check how that works with flatten

server/svix-server/src/v1/endpoints/auth.rs
```diff
@@ -25,8 +25,7 @@ async fn dashboard_access(
     let login_key = serde_json::to_vec(&serde_json::json!({
         "appId": app.id,
         "token": token,
-        // Region is unused
-        "region": "eu"
+        "region": cfg.internal.region,
```
tasn, line 28: Is it missing an &

tasn (approved):

tasn (approved):

PR #504, opened 2022-06-12
Server: fix caching dependencies for docker builds

tasn:
The problem is that in order to build deps we need the full source tree.
This fix just adds a dummy source tree so that we can build and cache
the deps before doing a full build.

This is still an incomplete solution. It caches all of the deps, but it
still doesn't cache the building of the main binary. So it's 10x better
than what we have now, but there's still 10x more to go.

One day Cargo will hopefully fix it:
https://github.com/rust-lang/cargo/issues/2644

Solution from:
https://stackoverflow.com/questions/58473606/cache-rust-dependencies-with-docker-build

### DIFF
--- a/server/.dockerignore
+++ b/server/.dockerignore
@@ -1,5 +1,7 @@
 # will have compiled files and executables
 target/
+# We don't want it to bust our docker build cache
+Cargo.lock
 
 /tmp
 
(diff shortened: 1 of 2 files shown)


LukeMathWalker:
I bumped into this PR quite randomly and I realise that the solution got lost in that issue in `cargo`'s repository - a fairly polished solution to this problem already exists as a cargo sub-command, [`cargo-chef`](https://github.com/lukeMathWalker/cargo-chef). It might save you some fiddling, especially as your project grows in complexity.

PR #508, opened 2022-06-14
Send operational webhooks from worker and endpoint API

svix-daniel:
Sends "operational webhooks" about the status of endpoints and messages as per the models defined here: https://api.svix.com/docs#tag/Webhooks.

### DIFF
--- a/server/svix-server/config.default.toml
+++ b/server/svix-server/config.default.toml
@@ -5,6 +5,12 @@
 # The address to listen on
 listen_address = "0.0.0.0:8071"
 
+# The address of the Svix server to use for sending operational webhooks (disabled when omitted/null)
+# Operational webhooks (otherwise known as "incoming webhooks"), are webhooks
+# send from the Svix server to you, to let you know when some events happen.
+# For a list of supported events please refer to: https://api.svix.com/docs#tag/Webhooks
+# operational_webhook_address = "http://127.0.0.1:8071"
+
 # The JWT secret for authentication - should be secret and securely generated
 # jwt_secret = "8KjzRXrKkd9YFcNyqLSIY8JwiaCeRc6WK4UkMnSW"
 
--- a/server/svix-server/src/core/operational_webhooks.rs
+++ b/server/svix-server/src/core/operational_webhooks.rs
@@ -1,0 +1,181 @@
+// SPDX-FileCopyrightText: © 2022 Svix Authors
+// SPDX-License-Identifier: MIT
+
+//! Module defining an interface for sending webhook events about the service.
+
+use std::sync::Arc;
+
+use chrono::{DateTime, Utc};
+use http::StatusCode;
+use serde::Serialize;
+use svix::api::{MessageIn, Svix, SvixOptions};
+
+use super::{
+    security::{generate_management_token, Keys},
+    types::{
+        ApplicationId, ApplicationUid, EndpointId, EndpointUid, MessageAttemptId, MessageId,
+        MessageUid, OrganizationId,
+    },
+};
+use crate::{
+    db::models::messageattempt,
+    error::{HttpError, Result},
+};
+
+/// Sent when an endpoint has been automatically disabled after continuous failures.
+#[derive(Debug, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct EndpointDisabledEvent<'a> {
+    pub app_id: &'a ApplicationId,
+    pub app_uid: Option<&'a ApplicationUid>,
+    pub endpoint_id: &'a EndpointId,
+    pub endpoint_uid: Option<&'a EndpointUid>,
+    pub fail_since: DateTime<Utc>,
+}
+
+/// Sent when an endpoint is created, updated, or deleted
+#[derive(Debug, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct EndpointEvent<'a> {
+    pub app_id: &'a ApplicationId,
+    pub app_uid: Option<&'a ApplicationUid>,
+    pub endpoint_id: &'a EndpointId,
+    pub endpoint_uid: Option<&'a EndpointUid>,
+}
+
+#[derive(Debug, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct MessageAttempetLast<'a> {
+    pub id: &'a MessageAttemptId,
+    pub response_status_code: i16,
+    pub timestamp: DateTime<Utc>,
+}
+
+impl<'a> From<&'a messageattempt::Model> for MessageAttempetLast<'a> {
+    fn from(attempt: &'a messageattempt::Model) -> Self {
+        Self {
+            id: &attempt.id,
+            response_status_code: attempt.response_status_code,
+            timestamp: attempt.created_at.into(),
+        }
+    }
+}
+
+/// Sent when a message delivery has failed (all of the retry attempts have been exhausted) as a
+/// "message.attempt.exhausted" type or after it's failed four times as a "message.attempt.failing"
+/// event.
+#[derive(Debug, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct MessageAttemptEvent<'a> {
+    pub app_id: &'a ApplicationId,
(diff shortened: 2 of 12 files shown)


server/svix-server/src/core/operational_webhooks.rs
```diff
@@ -0,0 +1,120 @@
...
+    endpoint_id: EndpointId,
+    endpoint_uid: Option<EndpointUid>,
+    fail_since: DateTime<Utc>,
+}
+
+/// Sent when an endpoint is created, updated, or deleted
+#[derive(Deserialize, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct EndpointEvent {
+    app_id: ApplicationId,
```
tasn, line 31: Can you alias these so that we use the correct event name when using it?

server/svix-server/src/worker.rs
```diff
@@ -227,6 +233,22 @@ async fn dispatch(
                 };
                 let _msg_dest = msg_dest.update(db).await?;
 
+                if attempt_count == 4 {
```
tasn, line 236: Make the 4 a const at the top of this file.

server/svix-server/config.default.toml
```diff
@@ -5,6 +5,12 @@
 # The address to listen on
 listen_address = "0.0.0.0:8071"
 
+# The address to send operational webhooks to. When not present, operational webhooks will not be
+# sent. When present, the API server with the given URL will be used to send operational webhooks.
+# This should be the address of the Svix API server that this application is running. It is
+# configurable for the case of a worker running independently of an API server.
```
tasn, line 11:
```suggestion
# The address of the Svix server to use for sending operational webhooks (disabled when omitted/null)
# Operational webhooks (otherwise known as "incoming webhooks"), are webhooks
# send from the Svix server to you, to let you know when some events happen.
# For a list of supported events please refer to: https://api.svix.com/docs#tag/Webhooks
```

server/svix-server/tests/utils/mod.rs
```diff
@@ -233,12 +238,14 @@ pub fn start_svix_server() -> (TestClient, tokio::task::JoinHandle<()>) {
...
 ) -> (TestClient, tokio::task::JoinHandle<()>) {
-    let cfg = Arc::new(cfg.clone());
+    let cfg = cfg.clone();
 
     let token = generate_org_token(&cfg.jwt_secret, OrganizationId::new(None, None)).unwrap();
     let listener = TcpListener::bind("127.0.0.1:0").unwrap();
     let base_uri = format!("http://{}", listener.local_addr().unwrap());
 
+    let cfg = Arc::new(cfg);
+
```
tasn, line 248: Turn it back to what it was.

server/svix-server/tests/utils/mod.rs
```diff
@@ -223,7 +223,12 @@ pub fn get_default_test_config() -> ConfigurationInner {
     let _ = dotenv::dotenv();
     let cfg = svix_server::cfg::load().unwrap();
 
-    cfg.as_ref().clone()
+    let mut cfg = cfg.as_ref().clone();
+
+    // Ensure operational webhooks are turned off for tests by default
+    cfg.operational_webhook_address = None;
```
tasn, line 229: Why would you need to be explicit about it? Just leave it how it was, it'll use the default value.

server/svix-server/src/worker.rs
```diff
@@ -80,6 +84,8 @@ async fn dispatch(
     payload: &Json,
     msg_task: MessageTask,
     endp: CreateMessageEndpoint,
+    create_message_app: CreateMessageApp,
+    op_webhook_sender: Option<&OperationalWebhookSender>,
```
tasn, line 88: I wouldn't make it optional, but rather just make it a no-op to send.
Yeah, I know you'll have a few extra clones when passing values to it, but honestly, it's not a biggie. We can always just change the structures to use references instead of strings if this ever becomes a concern.

To be honest though, we probably want the struct to use lifetimes and references. It'll save you from all of the clones and code changes you had to do.

server/svix-server/src/error.rs
```diff
@@ -92,6 +92,13 @@ impl From<DbErr> for Error {
     }
 }
 
+impl From<svix::error::Error> for Error {
+    fn from(err: svix::error::Error) -> Self {
+        tracing::error!("Error sending operational webhook: {:?}", err);
+        HttpError::internal_server_errer(None, None).into()
+    }
+}
+
```
tasn, line 101: We don't want to propagate this error outside. We want to just log an error with what we would have sent (so we can recover it from the logs) and continue.

server/svix-server/src/core/security.rs
```diff
@@ -248,6 +248,14 @@ pub fn generate_org_token(keys: &Keys, org_id: OrganizationId) -> Result<String>
     Ok(keys.key.authenticate(claims).unwrap())
 }
 
+pub fn generate_management_token(keys: &Keys) -> Result<String> {
+    let claims =
+        Claims::with_custom_claims(CustomClaim { organization: None }, Duration::from_mins(10))
+            .with_issuer(JWT_ISSUER)
+            .with_subject("org_0000SvixManagement0000");
```
tasn, line 255: Make the subject a constant. Btw, is it a valid ksuid? Does `0000SvixManagement0000` actually works as a ksuid and you can get the timestamp, etc from it?

server/svix-server/src/core/operational_webhooks.rs
```diff
@@ -0,0 +1,134 @@
...
+                    payload_retention_period: None,
+                },
+                None,
+            )
+            .await
+        {
+            Ok(_) => {}
+            // Ignore 404s because not every org will have an associated application
+            // TODO: Change to `svix::error::Error::Http` when lib fixed
+            Err(svix::error::Error::Validation(svix::error::HttpErrorContent {
```
tasn, line 125: Update the lib and change this.

server/svix-server/src/core/operational_webhooks.rs
```diff
@@ -0,0 +1,134 @@
...
+        match svix_api
+            .message()
+            .create(
+                recipient_org_id.to_string(),
+                MessageIn {
+                    channels: None,
+                    event_id: None,
+                    event_type,
+                    payload,
+                    payload_retention_period: None,
```
tasn, line 116:
```suggestion
                    event_type,
                    payload,
                   ..MessageIn::default()
```

server/svix-server/src/core/operational_webhooks.rs
```diff
@@ -0,0 +1,134 @@
...
+/// event.
+#[derive(Debug, Deserialize, Serialize, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct MessageAttemptEvent {
+    pub app_id: ApplicationId,
+    pub app_uid: Option<ApplicationUid>,
+    pub msg_id: MessageId,
+    pub msg_event_id: Option<MessageAttemptId>,
+    pub endpoint_id: EndpointId,
+    pub last_attempt: DateTime<Utc>,
```
tasn, line 52: Make all of them use lifetimes so that we don't need to clone when creating them.

server/svix-server/src/core/operational_webhooks.rs
```diff
@@ -0,0 +1,181 @@
...
+    pub app_uid: Option<&'a ApplicationUid>,
+    pub endpoint_id: &'a EndpointId,
+    pub endpoint_uid: Option<&'a EndpointUid>,
+    pub fail_since: DateTime<Utc>,
+}
+
+/// Sent when an endpoint is created, updated, or deleted
+#[derive(Debug, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct EndpointEvent<'a> {
```
svix-dylan, line 39: Should this structure include a field for event_type?

tasn: No, it's automatically done in the enum (search this file for `endpoint.updated` for example).
The enum automatically puts this (the data) into `data`) and the name into `type`.

svix-dylan (dismissed):

svix-dylan (approved):

PR #513, opened 2022-06-15
Server: support passing an org_id to the JWT generate command.

tasn:
This lets people developing locally (or in general) the ability to
generate different tokens for different org_ids (e.g. for operational
webhooks) easily, without having to do JWT signing themselves.

Related: #508

### DIFF
--- a/server/svix-server/src/main.rs
+++ b/server/svix-server/src/main.rs
@@ -6,6 +6,8 @@
 
 use dotenv::dotenv;
 use std::process::exit;
+use svix_server::core::types::OrganizationId;
+use validator::Validate;
 
 use svix_server::core::security::{default_org_id, generate_org_token};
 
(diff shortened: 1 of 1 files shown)


server/svix-server/src/main.rs
```diff
@@ -90,11 +102,12 @@ async fn main() {
         println!("Migrations run");
         exit(0);
     } else if let Some(Commands::Jwt {
-        command: JwtCommands::Generate,
+        command: JwtCommands::Generate { org_id },
     }) = &args.command
     {
+        let org_id = org_id.clone().unwrap_or_else(default_org_id);
```
tasn, line 108: I was getting issues when trying to avoid the clone, and tbh, it doesn't really matter. It's a manual place that just runs once.

svix-daniel:
```
else if let Some(Commands::Jwt {
        command: JwtCommands::Generate { org_id },
    }) = &mut args.command
  {
        let org_id = org_id.take().unwrap_or_else(default_org_id);
```

This might work to avoid the clone, but it looks a little funky

svix-daniel: Actually is there any reason you can't just take the whole command? It's the last in the chain.

svix-daniel (dismissed):

svix-daniel (approved):

PR #519, opened 2022-06-21
fix: Endpoint spelling in some error messages and comments

daniel-deutsch-benchling:
## Motivation

I noticed an error message I had received contained a typo, which looks to have been copy-pasted a couple times around the repository.

## Solution

Fixed typos.


tasn (approved):
Whoops. Nice catch, and thanks for the PR.

At least we are consistent always calling it `enpoint` instead of `endpoint`. It looks like someone needs to replace their `d` key. 😂

PR #521, opened 2022-06-22
Server: Add Cache support for getting/setting Strings and raw bytes

jaymell:
Currently, all values are assumed to be de/serializable as json,
which is not appropriate for all circumstances. This functionality
enables us to store Strings and raw bytes without json de/serialization and
still allowing for type safety of cached data

### DIFF
--- a/server/svix-server/src/core/cache/memory.rs
+++ b/server/svix-server/src/core/cache/memory.rs
@@ -12,7 +12,7 @@ use serde_json;
 use std::collections::HashMap;
 use std::sync::Arc;
 
-use super::{Cache, CacheBehavior, CacheKey, CacheValue, Result};
+use super::{Cache, CacheBehavior, CacheKey, CacheValue, Error, Result, StringCacheValue};
 
 #[derive(Debug)]
 struct ValueWrapper {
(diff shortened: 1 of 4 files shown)


server/svix-server/src/core/cache/memory.rs
```diff
@@ -56,17 +56,52 @@ pub struct MemoryCache {
...
-            .transpose()?)
+            .map(|x| serde_json::from_str(&x).map_err(|e| e.into()))
+            .transpose()
+    }
+
+    async fn get_string<T: StringCacheValue>(&self, key: &T::Key) -> Result<Option<T>> {
+        self.get_raw(key.as_ref())
+            .await
+            .map(|x| x.try_into().map_err(|_| Error::DeserializationOther))
+            .transpose()
```
tasn, line 104: If you add `get_raw`, `set_raw` and the likes to the trait, you can then save yourself from having to implement `get_string` and the likes in each backend, and just have a default implementation in the trait itself.
Essentially reducing all of this duplicate but identical code.

Does that make sense?

server/svix-server/src/core/cache/memory.rs
```diff
@@ -8,21 +8,24 @@ use tokio::{
...
-use serde_json;
 use std::collections::HashMap;
 use std::sync::Arc;
 
-use super::{Cache, CacheBehavior, CacheKey, CacheValue, Result};
+use super::{Cache, CacheBehavior, CacheKey, Result};
+
+// Apparently traits needed by types in default implementation aren't recognized:
+#[allow(unused_imports)]
+use super::{CacheValue, StringCacheValue};
```
tasn, line 18: This doesn't make sense... I'm sure it was something else and removing this will work.

server/svix-server/src/core/cache/redis.rs
```diff
@@ -11,7 +11,6 @@ use super::{Cache, CacheBehavior, CacheKey, Error, Result};
 
 // Apparently traits needed by types in default implementation aren't recognized:
 #[allow(unused_imports)]
```
tasn, line 13:
```suggestion
```

server/svix-server/src/core/cache/redis.rs
```diff
@@ -5,11 +5,12 @@ use std::time::Duration;
...
 
-use serde_json;
-
 use crate::redis::{PoolLike, PooledConnectionLike, RedisPool};
 
-use super::{Cache, CacheBehavior, CacheKey, CacheValue, Error, Result};
+use super::{Cache, CacheBehavior, CacheKey, Error, Result};
+
+// Apparently traits needed by types in default implementation aren't recognized:
+#[allow(unused_imports)]
```
tasn, line 13:
```suggestion
```

tasn (approved):

PR #522, opened 2022-06-24
Implement basic OpenTelemetry support

### LINKED ISSUE #344: Start using the OpenTelemetry crate for telemtry
https://github.com/open-telemetry/opentelemetry-rust

Also: don't forget to update the docs/README to explain how to use it with datadog/newrelic/sentry.

More info / related attempts:
* https://github.com/tokio-rs/axum/issues/750
* https://github.com/tokio-rs/axum/pull/769
* https://github.com/tower-rs/tower-http/pull/180
* https://github.com/EmbarkStudios/server-framework/blob/main/src/middleware/trace/opentelemetry.rs
* Maybe https://github.com/davidB/sandbox_axum_observability

svix-daniel:
This PR integrates OpenTelemetry support and creates middleware for Axum that ensures that the spans created for each request follow OpenTelemetry conventions.

Fixes #344

### DIFF
--- a/server/svix-server/src/db/mod.rs
+++ b/server/svix-server/src/db/mod.rs
@@ -10,6 +10,7 @@ pub mod models;
 
 static MIGRATIONS: sqlx::migrate::Migrator = sqlx::migrate!();
 
+#[tracing::instrument(level = "trace")]
 async fn connect(cfg: &Configuration) -> sqlx::Pool<sqlx::Postgres> {
     tracing::debug!("DB: Initializing pool");
     if DbBackend::Postgres.is_prefix_of(&cfg.db_dsn) {
@@ -23,10 +24,12 @@ async fn connect(cfg: &Configuration) -> sqlx::Pool<sqlx::Postgres> {
     }
 }
 
+#[tracing::instrument(level = "trace")]
 pub async fn init_db(cfg: &Configuration) -> DatabaseConnection {
     SqlxPostgresConnector::from_sqlx_postgres_pool(connect(cfg).await)
 }
 
+#[tracing::instrument(level = "trace")]
 pub async fn run_migrations(cfg: &Configuration) {
     let db = connect(cfg).await;
     MIGRATIONS.run(&db).await.unwrap();
--- a/server/svix-server/src/db/mod.rs
+++ b/server/svix-server/src/db/mod.rs
@@ -10,7 +10,6 @@ pub mod models;
 
 static MIGRATIONS: sqlx::migrate::Migrator = sqlx::migrate!();
 
-#[tracing::instrument(level = "trace")]
 async fn connect(cfg: &Configuration) -> sqlx::Pool<sqlx::Postgres> {
     tracing::debug!("DB: Initializing pool");
     if DbBackend::Postgres.is_prefix_of(&cfg.db_dsn) {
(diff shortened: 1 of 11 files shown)


svix-dylan (approved):

server/svix-server/src/main.rs
```diff
@@ -90,19 +94,58 @@ async fn main() {
...
+            let stdout_layer = tracing_subscriber::fmt::layer()
                 .event_format(fmt)
-                .fmt_fields(json_fields)
-                .with_env_filter(filter)
+                .fmt_fields(json_fields);
+
+            tracing_subscriber::Registry::default()
+                .with(otel_layer)
+                .with(stdout_layer)
+                .with(tracing_subscriber::EnvFilter::from_default_env())
```
tasn, line 148: So it does both stdout and opentelemetry? I almost think it should be one or the other, no? I really have no idea. Maybe ask @svix-james, I think he should know better.

I'm fine with having both, I was just wondering.

server/svix-server/src/main.rs
```diff
@@ -90,19 +94,58 @@ async fn main() {
         );
     }
 
+    let otel_layer = cfg.opentelemetry_address.as_ref().map(|addr| {
```
tasn, line 97: Shouldn't otel_layer be optional and only added if needed?

server/svix-server/src/main.rs
```diff
@@ -90,19 +94,58 @@ async fn main() {
...
+    // either formatted normally or in JSON format
     match cfg.log_format {
         cfg::LogFormat::Default => {
-            tracing_subscriber::fmt::init();
+            let stdout_layer = tracing_subscriber::fmt::layer();
+            tracing_subscriber::Registry::default()
+                .with(otel_layer)
+                .with(stdout_layer)
+                .with(tracing_subscriber::EnvFilter::from_default_env())
+                .init();
```
tasn, line 135: This looks like exactly the same code as below, shouldn't this be shared between the two log formats?
I'm thinking:
```rust
// Is it stdout or stderr?
let stdout_layer = match cfg.log_format {
  ...
};
// Probably also dependent on whether the otel_layer exists
tracing_subscriber::Registry::default()
                .with(otel_layer)
                .with(stdout_layer)
                .with(tracing_subscriber::EnvFilter::from_default_env())
                .init();
```

tasn:
I haven't really reviewed the code, as Dylan already has, I just reviewed the docs + main.rs.

server/svix-server/src/main.rs
svix-daniel: `otel_layer` here is an `Option` as is the `opentelemetry_address`.

server/svix-server/src/main.rs
svix-daniel: The `tracing_subscriber`layers are different types depending on log format, so the initialization has to occur there.

server/svix-server/src/main.rs
tasn: Oh, adding layers supports options? (the `with()` below)

server/svix-server/src/main.rs
tasn: It looks identical to me. The only difference I see is stdout_layer, which my comment above addresses.

server/svix-server/src/main.rs
svix-daniel: Yup!

server/svix-server/src/lib.rs
```diff
@@ -35,12 +38,14 @@ pub mod redis;
 pub mod v1;
 pub mod worker;
 
+#[tracing::instrument(name = "app_start", level = "trace")]
```
tasn, line 41:
```suggestion
#[tracing::instrument(name = "app_start", level = "trace", skipall)]
```
Same for the rest of the functions.

Also, I'm not quite sure why we would want to instrument these functions in the first place. I get instrumenting the migration one, but why the rest of these?

server/svix-server/src/db/mod.rs
```diff
@@ -10,6 +10,7 @@ pub mod models;
 
 static MIGRATIONS: sqlx::migrate::Migrator = sqlx::migrate!();
 
+#[tracing::instrument(level = "trace")]
```
tasn, line 13: Here too, don't instrument parameters...

tasn (approved):

PR #525, opened 2022-06-24
Server: add support for asymmetric signatures

### LINKED ISSUE #501: Consider IETF HTTP Message Signatures draft
## Feature Request
Move svix from using HMAC endpoint signing to the [IETF HTTP Message Signatures](https://datatracker.ietf.org/doc/draft-ietf-httpbis-message-signatures/) draft. The IETF draft decreases the number of secrets in a system which is very appealing. Would svix consider implementing this; even on a long-term time frame?

An example implementation is [here](https://github.com/pyauth/http-message-signatures) in Python.

tasn:
We now support both symmetric and asymmetric signature schemes at the same time
for different endpoints, and even rotate the signing key for a specific
endpoint from symmetric to asymmetric and vice-versa. By default, the Svix
server will still generate symmetric secrets for endpoints when one isn't set
explicitly, though this is now configurable, and you can turn on asymmetric
mode by setting the following setting:

```toml
default_signature_type = "ed25519"
```

Once this is set, all secrets will be generated to be asymmetric by default.

Regardless of whether the config is set, you can explicitly set secrets (and
thus the signature scheme) to be symmetric or asymmetric explicitly by passing
secrets with the prefix `whsec_` for symmetric and `whpk_` for asymmetric.

The signature scheme used for the asymmetric signatures is `ed25519(m)`.

Please note, that asymmetric signatures are a few orders of magnitude
slower than symmetric signatures, and also more complex. So only use it
if it makes sense for your environment.

Fixes #501


tasn:
Missing items:
- [ ] Documenting it in the README
- [ ] The config should also have a warning that it's slower and symmetric is recommended
- [ ] Implement a command line command that prints the pubkey for the private keys set, or for a private key passed.

tasn:
I think it's ready. :)

svix-dylan (approved):

svix-dylan (approved):

PR #534, opened 2022-06-30
Server: add support for encrypting webhook secrets

tasn:
Webhook secrets are used for signing the webhook messages and they are
stored in the main database and cache. Normally they are stored there in
the clear, which means that people with access to the database could
potentially retrieve them and forge webhook messages.

With this change, we are adding support for an optional secret key that
can be used to encrypt the webhook secrets.

This change is fully backwards compatible when encryption is turned off,
and you can also change from non-encrypted to encrypted whenever you
want and everything will continue to work.

Changing the encryption key, or moving from encrypted to non-encrypted
is not supported, as without the key you won't be able to view endpoint
secrets. We could potentially add a script that migrates the database
from one to the other in the future if needed, but it's probably a very
rare requirement.

### DIFF
--- a/server/svix-server/src/core/security.rs
+++ b/server/svix-server/src/core/security.rs
@@ -8,7 +8,6 @@ use axum::{
     extract::{Extension, FromRequest, Path, RequestParts, TypedHeader},
     headers::{authorization::Bearer, Authorization},
 };
-use ed25519_compact::*;
 
 use jwt_simple::prelude::*;
 use sea_orm::DatabaseConnection;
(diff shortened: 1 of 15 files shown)


server/svix-server/src/core/security.rs
```diff
@@ -333,3 +336,57 @@ impl PartialEq for AsymmetricKey {
...
+
+    pub fn new_noop() -> Self {
+        Self(None)
+    }
+
+    pub fn new(key: [u8; 32]) -> Self {
+        Self(Some(Key::from_slice(&key).to_owned()))
+    }
+
+    pub fn encrypt(&self, data: &[u8]) -> crate::error::Result<Vec<u8>> {
```
svix-dylan, line 354: I see we have e2e tests which probably suffice but this structure looks like it easily supports unit tests on `encrypt` and `decrypt` as well if we wanted to be thorough. Not prohibitive.

svix-dylan (dismissed):

server/svix-server/src/core/security.rs
tasn: Done.

PR #538, opened 2022-07-01
Patch builds to work with new OpenSSL dependency

svix-dylan:
Patches our builds to work with a recently introduced OpenSSL dependency via redis-rs. This PR includes devtool installations in our Dockerfile, and temporarily disables linux-musl builds until redis-rs supports rustls (tracking in redis-rs/redis-rs#601). As an interim replacement for linux-musl, this PR builds linux-gnu binaries.

### DIFF
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -1,7 +1,7 @@
 # Base build
 FROM rust:1.60-slim-bullseye AS build
 
-RUN apt-get update && apt-get install -y build-essential=12.* checkinstall=1.* zlib1g-dev=1:* --no-install-recommends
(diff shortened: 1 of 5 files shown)


server/Dockerfile
```diff
@@ -1,6 +1,8 @@
 # Base build
 FROM rust:1.60-slim-bullseye AS build
 
+RUN apt-get update && apt-get install -y build-essential checkinstall zlib1g-dev
```
tasn, line 4: The link complains about lack of pinning, you can add an ignore to this, no need to pin them.

tasn (approved):

tasn (approved):

PR #555, opened 2022-07-12
Update libraries to accept raw webhook secret

svix-dylan:
This PR updates our libraries to optionally take a raw webhook secret.

### DIFF
--- a/csharp/Svix/Webhook.cs
+++ b/csharp/Svix/Webhook.cs
@@ -31,14 +31,9 @@ namespace Svix
             this.key = Convert.FromBase64String(key);
         }
 
-        public WebhookRaw(byte[] key)
+        public Webhook(byte[] key)
         {
             this.key = key;
-        }
-
-        public WebhookRaw(string key)
-        {
-            this.key = Encoding.ASCII.GetBytes(string);
         }
 
         public void Verify(string payload, WebHeaderCollection headers)
--- a/csharp/Svix/Webhook.cs
+++ b/csharp/Svix/Webhook.cs
@@ -31,6 +31,16 @@ namespace Svix
             this.key = Convert.FromBase64String(key);
         }
 
+        public WebhookRaw(byte[] key)
+        {
+            this.key = key;
+        }
+
+        public WebhookRaw(string key)
+        {
+            this.key = Encoding.ASCII.GetBytes(string);
+        }
+
         public void Verify(string payload, WebHeaderCollection headers)
         {
             string msgId = headers.Get(SVIX_ID_HEADER_KEY);
--- a/go/webhook.go
+++ b/go/webhook.go
@@ -39,6 +39,18 @@ func NewWebhook(secret string) (*Webhook, error) {
 	}, nil
 }
 
+func NewWebhookFromRaw(secret []byte) (*Webhook, error) {
+	return &Webhook {
+		key: secret
+	}, nil
+}
+
+func NewWebhookFromRawString(secret string) (*Webhook, error) {
+	return &Webhook {
+		key: []byte(secret)
+	}, nil
+}
+
 // Verify validates the payload against the svix signature headers
 // using the webhooks signing secret.
 //
--- a/go/webhook.go
+++ b/go/webhook.go
@@ -39,13 +39,13 @@ func NewWebhook(secret string) (*Webhook, error) {
 	}, nil
 }
 
-func NewWebhookFromRaw(secret []byte) (*Webhook, error) {
+func NewWebhookRaw(secret []byte) (*Webhook, error) {
 	return &Webhook {
 		key: secret,
 	}, nil
 }
 
-func NewWebhookFromRawString(secret string) (*Webhook, error) {
+func NewWebhookRaw(secret string) (*Webhook, error) {
 	return &Webhook {
 		key: []byte(secret),
 	}, nil
--- a/go/webhook.go
+++ b/go/webhook.go
@@ -41,13 +41,13 @@ func NewWebhook(secret string) (*Webhook, error) {
 
 func NewWebhookFromRaw(secret []byte) (*Webhook, error) {
 	return &Webhook {
-		key: secret
+		key: secret,
 	}, nil
 }
 
 func NewWebhookFromRawString(secret string) (*Webhook, error) {
 	return &Webhook {
-		key: []byte(secret)
+		key: []byte(secret),
 	}, nil
 }
 
--- a/go/webhook.go
+++ b/go/webhook.go
@@ -45,7 +45,7 @@ func NewWebhookRaw(secret []byte) (*Webhook, error) {
 	}, nil
 }
 
-func NewWebhookRaw(secret string) (*Webhook, error) {
+func NewWebhookRawString(secret string) (*Webhook, error) {
 	return &Webhook {
 		key: []byte(secret),
 	}, nil
--- a/go/webhook.go
+++ b/go/webhook.go
@@ -45,12 +45,6 @@ func NewWebhookRaw(secret []byte) (*Webhook, error) {
 	}, nil
 }
 
-func NewWebhookRawString(secret string) (*Webhook, error) {
-	return &Webhook {
-		key: []byte(secret),
-	}, nil
-}
-
 // Verify validates the payload against the svix signature headers
 // using the webhooks signing secret.
 //
--- a/java/lib/src/main/java/com/svix/Webhook.java
+++ b/java/lib/src/main/java/com/svix/Webhook.java
@@ -35,11 +35,11 @@ public final class Webhook {
 		}
 		this.key = Base64.getDecoder().decode(sec);
 	}
-	
+
 	public Webhook(final byte[] secret) {
 		this.key = secret;
 	}
-	
+
 	public WebhookFromRawString(final String secret) {
 		this.key = secret.getBytes();
 	}
--- a/java/lib/src/main/java/com/svix/Webhook.java
+++ b/java/lib/src/main/java/com/svix/Webhook.java
@@ -35,6 +35,14 @@ public final class Webhook {
 		}
 		this.key = Base64.getDecoder().decode(sec);
 	}
+	
+	public Webhook(final byte[] secret) {
+		this.key = secret;
+	}
+	
+	public WebhookFromRawString(final String secret) {
+		this.key = secret.getBytes();
+	}
 
 	public void verify(final String payload, final HttpHeaders headers) throws WebhookVerificationException {
 		Optional<String> msgId = headers.firstValue(SVIX_MSG_ID_KEY);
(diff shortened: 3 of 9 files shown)


javascript/src/index.ts
```diff
@@ -590,19 +590,25 @@ export interface WebhookUnbrandedRequiredHeaders {
   "webhook-signature": string;
 }
 
+export interface WebhookOptions {
+  raw?: boolean;
```
tasn, line 594:
```suggestion
  format?: "raw";
```
We may support other formats going forward.

javascript/src/index.ts
```diff
@@ -590,18 +590,28 @@ export interface WebhookUnbrandedRequiredHeaders {
...
     }
-    if (secret.startsWith(Webhook.prefix)) {
+    if (!options?.raw && secret instanceof String && secret.startsWith(Webhook.prefix)) {
       secret = secret.substring(Webhook.prefix.length);
     }
-    this.key = base64.decode(secret);
+    if (options?.raw && secret instanceof Uint8Array) {
+      this.key = secret;
+    } else if (options?.raw && secret instanceof String) {
+      this.key = Uint8Array.from(atob(secret.toString()), (c) => c.charCodeAt(0));
```
tasn, line 611: Isn't `atob` the poor man's base64.decode? We shouldn't decode in this situation.

javascript/src/index.ts
```diff
@@ -590,18 +590,34 @@ export interface WebhookUnbrandedRequiredHeaders {
   "webhook-signature": string;
 }
 
+const SECRET_FORMAT_RAW = "raw";
+
+export interface WebhookOptions {
+  format?: string;
```
tasn, line 596:
```suggestion
  format?: "raw";
```

javascript/src/index.ts
```diff
@@ -590,18 +590,34 @@ export interface WebhookUnbrandedRequiredHeaders {
...
+      secret instanceof String &&
+      secret.startsWith(Webhook.prefix)
+    ) {
       secret = secret.substring(Webhook.prefix.length);
     }
-    this.key = base64.decode(secret);
+    if (options?.format === SECRET_FORMAT_RAW && secret instanceof Uint8Array) {
+      this.key = secret;
+    } else if (options?.format === SECRET_FORMAT_RAW && secret instanceof String) {
+      this.key = Uint8Array.from(secret.toString(), (c) => c.charCodeAt(0));
```
tasn, line 617: why the to string? It should already be a string.

javascript/src/index.ts
```diff
@@ -590,18 +590,34 @@ export interface WebhookUnbrandedRequiredHeaders {
...
+    } else {
+      this.key = base64.decode(secret.toString());
```
tasn, line 619: why the to string? It should already be a string.

javascript/src/index.ts
```diff
@@ -590,18 +590,34 @@ export interface WebhookUnbrandedRequiredHeaders {
...
   private static prefix = "whsec_";
   private readonly key: Uint8Array;
 
-  constructor(secret: string) {
+  constructor(secret: string | Uint8Array, options?: WebhookOptions) {
     if (!secret) {
       throw new Error("Secret can't be empty.");
     }
-    if (secret.startsWith(Webhook.prefix)) {
+    if (
```
tasn, line 607: These are a lot of unnecessary ifs that check the same thing, it should be:

```typescript
if (options.format === "raw") {
   // .. all of the raw handling
} else {
   // What's currently there
}
```

svix-dylan: What, something like this?

```javascript
    if (options?.format !== "raw" && secret instanceof String) {
      if (secret.startsWith(Webhook.prefix)) {
        secret = secret.substring(Webhook.prefix.length);
      }
      this.key = base64.decode(secret as string);
    } else {
      if (options?.format === "raw" && secret instanceof Uint8Array) {
        this.key = secret;
      } else if (options?.format === "raw" && secret instanceof String) {
        this.key = Uint8Array.from(secret, (c) => c.charCodeAt(0));
      }
    }
```

We also need the compiler to know that we are always setting this.key to _something_ or it complains.

javascript/src/index.ts
```diff
@@ -590,18 +590,33 @@ export interface WebhookUnbrandedRequiredHeaders {
...
+        this.key = Uint8Array.from(secret, (c) => c.charCodeAt(0));
+      }
+    } else {
+      if (!(secret instanceof String)) {
+        throw new Error("Expected secret to be of type string");
+      }
+      if (secret.startsWith(Webhook.prefix)) {
+        secret = secret.substring(Webhook.prefix.length);
+      }
+      this.key = base64.decode(secret as string);
```
tasn, line 618:
```suggestion
      this.key = base64.decode(secret);
```

svix-dylan: Typescript requires this or .toString(). It's not picking up on the if block above, and so it infers secret can possibly still be `UInt8Array` here.

java/lib/src/main/java/com/svix/Webhook.java
```diff
@@ -35,6 +35,14 @@ public Webhook(final String secret) {
...
 		this.key = Base64.getDecoder().decode(sec);
 	}
+	
+	public Webhook(final byte[] secret) {
+		this.key = secret;
+	}
+	
+	public WebhookFromRawString(final String secret) {
+		this.key = secret.getBytes();
+	}
```
tasn, line 45: Webhook should remain as is, and then you should have `WebhookRaw` that has two variants, the raw string and raw bytes.

javascript/src/index.ts
tasn: Weird, as it's wrong. :(

OK.

kotlin/lib/src/main/kotlin/Webhook.kt
```diff
@@ -100,11 +100,20 @@ class Webhook(secret: String) {
...
-            sec = sec.substring(SECRET_PREFIX.length)
+    constructor(secret: String) {
+        if (secret.startsWith(SECRET_PREFIX)) {
+            secret = secret.substring(SECRET_PREFIX.length)
         }
-        key = Base64.getDecoder().decode(sec)
+        key = Base64.getDecoder().decode(secret)
+    }
+
+    constructor(secret: String, format: String) {
```
tasn, line 110: You can just have a creator function `Webhook.fromRaw()`

php/src/Webhook.php
```diff
@@ -16,6 +16,12 @@ public function __construct($secret)
         $this->secret = base64_decode($secret);
     }
 
+    public static function fromRaw($secret) {
+        $obj = new self(); 
+        $obj->secret = $secret;
+        return $obj;
+    }
```
svix-dylan, line 23: @tasn does this look right? When I googled PHP's `base64_decode` function, it looks like it returns a string anyways, not `[]byte`. So in PHP at least, we only handle strings, is this correct?

kotlin/lib/src/main/kotlin/Webhook.kt
svix-dylan: What does that look like? This is a tricky search to me / Kotlin noob, as it appears there have been multiple approaches through Kotlin's existence. And is this overloaded constructor approach invalid?

php/src/Webhook.php
tasn: I think that's correct, yeah.

kotlin/lib/src/main/kotlin/Webhook.kt
tasn: It's just confusing that it feels very different to everything else we do in all of the languages.

Can't you just define a static function (or maybe a static function on the companion object) that does it?

python/svix/webhooks.py
```diff
@@ -29,6 +29,16 @@ def __init__(self, whsecret: str, *, enc_key: t.Optional[str] = None):
...
         self._enc_key = base64.b64decode(enc_key) if enc_key is not None else None
 
+    def from_raw(whsecret: bytes):
+        wh = Webhook()
+        wh._whsecret = whsecret
+        return wh
+
+    def from_raw(whsecret: str):
+        wh = Webhook()
+        wh._whsecret = str.encode(whsecret)
```
tasn, line 39:
```suggestion
        wh._whsecret = whsecret.encode()
```
I think that's the python way of doing it.

go/webhook.go
```diff
@@ -39,6 +39,18 @@ func NewWebhook(secret string) (*Webhook, error) {
...
+	return &Webhook {
+		key: secret,
+	}, nil
+}
+
+func NewWebhookFromRawString(secret string) (*Webhook, error) {
+	return &Webhook {
+		key: []byte(secret),
+	}, nil
+}
```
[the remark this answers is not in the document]
svix-dylan: @tasn per CI, looks like Go doesn't like funcs of same name.

python/svix/webhooks.py
```diff
@@ -29,6 +29,16 @@ def __init__(self, whsecret: str, *, enc_key: t.Optional[str] = None):
         self._whsecret = base64.b64decode(whsecret)
         self._enc_key = base64.b64decode(enc_key) if enc_key is not None else None
 
+    def from_raw(whsecret: bytes):
```
tasn, line 32:
```suggestion
    @staticmethod
    def from_raw(whsecret: bytes) -> Webhook:
```

go/webhook.go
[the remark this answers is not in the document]
tasn: OK, so great. :)

kotlin/lib/src/main/kotlin/Webhook.kt
svix-dylan: Ok I will take a look at it -- although I feel like this still offers an ergonomic syntax:

```kotlin
Webhook("secret")
Webhook("secret", "raw")
Webhook(bytes)
```

In the meantime, tests are passing on all the other libraries!! Although I don't have runtimes/compilers so they're not verified yet

tasn: I think using a string for Raw rather than an enum is also not very natural. Actually, I have another thought, will make it elsewhere.

csharp/Svix/Webhook.cs
```diff
@@ -31,6 +31,16 @@ public Webhook(string key)
             this.key = Convert.FromBase64String(key);
```
tasn, line 31: JS is not one of these languages, but quite a few languages have an easy `as_bytes()`, I wonder if we should just have a raw that accepts bytes and not have a string variant for literally all of the languages other than JS?

svix-dylan:
@tasn js is the only lib that still supports rawString+rawBytes. All other libs now only support bytes as the only raw option, as an overload where available.

tasn (approved):

PR #556, opened 2022-07-12
Add graceful shutdown of service

svix-daniel:
Adds graceful shutdown to the three main components of the service including
the API server, the worker, and the expired message cleaner.

### DIFF
--- a/server/svix-server/src/lib.rs
+++ b/server/svix-server/src/lib.rs
@@ -7,9 +7,11 @@
 use axum::{extract::Extension, Router};
 
 use cfg::CacheType;
+use lazy_static::lazy_static;
 use std::{
     net::{SocketAddr, TcpListener},
     str::FromStr,
+    sync::atomic::{AtomicBool, Ordering},
     time::Duration,
 };
 use tower::ServiceBuilder;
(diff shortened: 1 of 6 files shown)


server/svix-server/src/worker.rs
```diff
@@ -504,47 +504,67 @@ pub async fn worker_loop(
...
-                            tracing::error!("Error executing task: {}", err);
-                            queue_tx
-                                .nack(delivery)
-                                .await
-                                .expect("Error sending 'nack' to Redis after task execution error");
-                        } else {
-                            queue_tx.ack(delivery).await.expect(
-                                "Error sending 'ack' to Redis after successful task execution",
-                            );
+        tokio::select! {
```
tasn, line 510: Using select will mean that the first to finish "wins", which in turns mean that a signal is not a graceful shutdown, it's violently killing the async task.

It should just be an `if` in a place where we are OK with killing (sort of like a "yield"). So just at the beginning of every loop for example, check the value and see if we are good to stop. We don't even need a watch, it can literally just be a variable that we check. Heck, it can even be a global variable with `lazy_static!` so we don't even need to pollute the code.

server/svix-server/src/lib.rs
```diff
@@ -124,15 +157,18 @@ pub async fn run_with_prefix(
...
                 } else {
                     tracing::debug!("API: Listening on {}", listen_address);
                     axum::Server::bind(&listen_address)
                         .serve(app.into_make_service())
+                        .with_graceful_shutdown(graceful_shutdown_handler())
                         .await
                 }
             } else {
                 tracing::debug!("API: off");
+                graceful_shutdown_handler().await;
```
tasn, line 171: API is off here, you don't need it, just remove this line.

server/svix-server/src/queue/redis.rs
```diff
@@ -462,36 +462,34 @@ impl TaskQueueReceive for RedisQueueConsumer {
 
             // There is no way to make it await a message for unbounded times, so simply block for a short
             // amount of time (to avoid locking) and loop if no messages were retreived
-            let resp = loop {
```
tasn, line 465: Why did the redis code even need to change?

svix-daniel: With the loop it only broke upon successfully receiving a task. Now it delivers an empty `Vec` after the `BLOCK` period on Redis such that the worker is guaranteed to shutdown eventually.

server/svix-server/src/lib.rs
tasn: OK, if it actually works than we are all good. :)

server/svix-server/src/queue/redis.rs
tasn: Cool, that makes sense, so just the other comment.

tasn (commented):
Made one comment, feel free to ignore it and then I can just approve.

tasn (approved):

PR #564, opened 2022-07-13
Add retries to queue

svix-daniel:
Uses `run_with_retries` on the `send`, `ack`, `nack`, and `receive_all` methods
on the `TaskQueueProducer` and `TaskQueueConsumer` such as to ensure transient
errors do not result in an error response.

### DIFF
--- a/server/svix-server/src/queue/memory.rs
+++ b/server/svix-server/src/queue/memory.rs
@@ -1,4 +1,4 @@
-use std::time::Duration;
+use std::{sync::Arc, time::Duration};
 
 use axum::async_trait;
 use chrono::Utc;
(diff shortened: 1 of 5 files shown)


server/svix-server/src/queue/memory.rs
```diff
@@ -26,10 +26,10 @@ pub struct MemoryQueueProducer {
 
 #[async_trait]
 impl TaskQueueSend for MemoryQueueProducer {
-    async fn send(&self, msg: QueueTask, delay: Option<Duration>) -> Result<()> {
+    async fn send(&self, msg: &QueueTask, delay: Option<Duration>) -> Result<()> {
         let tx = self.tx.clone();
         let timestamp = delay.map(|delay| Utc::now() + chrono::Duration::from_std(delay).unwrap());
-        let delivery = TaskQueueDelivery::new(msg, timestamp);
+        let delivery = TaskQueueDelivery::new(msg.clone(), timestamp);
```
tasn, line 32: You're cloning a potentially fat thing, no way around it?

svix-james (approved):

PR #566, opened 2022-07-18
Add support for disabling endpoints on repeated failure

svix-daniel:
If a given endpoint has only failed for a duration exceeding the configured
grace period, then this PR will disable the endpoint such as to avoid flooding
the queue with tasks to endpoints which only ever fail.

### DIFF
--- a/server/svix-server/src/cfg.rs
+++ b/server/svix-server/src/cfg.rs
@@ -66,6 +66,14 @@ where
     }
 }
 
+fn deserialize_hours<'de, D>(deserializer: D) -> std::result::Result<Duration, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let hours = u64::deserialize(deserializer)?;
+    Ok(Duration::from_secs(60 * 60 * hours))
+}
+
 const DEFAULTS: &str = include_str!("../config.default.toml");
 
 pub type Configuration = Arc<ConfigurationInner>;
@@ -138,6 +146,17 @@ pub struct ConfigurationInner {
     /// How long to wait when making a request (in seconds)
     #[validate(range(min = 1, max = 30))]
     pub worker_request_timeout: u16,
+
+    /// How long of a period to wait after the first dispatch failure to disable an endpoint. If a
+    /// message is successfully sent during this time, then the endpoint will not disable.
+    #[serde(deserialize_with = "deserialize_hours")]
+    pub dispatch_disable_grace_period: Duration,
+
+    /// How long of a period to wait after a first failure has been cached to remove the record. If
+    /// a consistently failing endpoint is called only once within this period, then it will never
+    /// be disabled.
+    #[serde(deserialize_with = "deserialize_hours")]
+    pub dispatch_disable_expiration_period: Duration,
 
     // Execution mode
     /// Should this instance run the API
--- a/server/svix-server/src/worker.rs
+++ b/server/svix-server/src/worker.rs
@@ -4,11 +4,13 @@
 use crate::cfg::Configuration;
 
 use crate::core::cryptography::Encryption;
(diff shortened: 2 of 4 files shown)


svix-james (approved):

server/svix-server/src/worker.rs
```diff
@@ -351,22 +474,71 @@ async fn dispatch(
...
+                    &msg_task.endpoint_id,
+                    chrono::Duration::from_std(cfg.dispatch_disable_grace_period)
+                        .expect("Expiration period exceeds maximum"),
+                    cfg.dispatch_disable_expiration_period,
+                )
+                .await?
+                {
+                    EndpointDisableStatus::DoNotDisable => {
+                        // Send operational webhook
+                        op_webhook_sender
```
tasn, line 489: This one needs to be sent regardless of whether we are disabling the endpoint.

server/svix-server/src/worker.rs
```diff
@@ -43,6 +46,122 @@ const USER_AGENT: &str = concat!("Svix-Webhooks/", env!("CARGO_PKG_VERSION"));
...
+/// If the key value pair does not already exist in the cache, indicating that the endpoint never
+/// stopped responding, no operation is performed.
+async fn process_success_cache(
+    cache: &Cache,
+    app_id: &ApplicationId,
+    endp_id: &EndpointId,
+) -> Result<()> {
+    let key = FailureCacheKey::new(app_id, endp_id);
+
+    // XXX: Make sure to test deleting a value that does not exist doesn't cause an error response
```
tasn, line 104: Code is no place for TODO items. These kind of comments are useful for stuff specific to people changing the code, but this is not the case here.

server/svix-server/src/cfg.rs
```diff
@@ -139,6 +147,17 @@ pub struct ConfigurationInner {
     #[validate(range(min = 1, max = 30))]
     pub worker_request_timeout: u16,
 
+    /// How long of a period to wait after the first dispatch failure to disable an endpoint. If a
+    /// message is successfully sent during this time, then the endpoint will not disable.
+    #[serde(deserialize_with = "deserialize_hours")]
+    pub dispatch_disable_grace_period: Duration,
```
tasn, line 153: This value is not useful, and should be removed.

server/svix-server/src/cfg.rs
```diff
@@ -139,6 +147,17 @@ pub struct ConfigurationInner {
...
+
+    /// How long of a period to wait after a first failure has been cached to remove the record. If
+    /// a consistently failing endpoint is called only once within this period, then it will never
+    /// be disabled.
+    #[serde(deserialize_with = "deserialize_hours")]
+    pub dispatch_disable_expiration_period: Duration,
```
tasn, line 159: You need a new config value: `disable_endpoint_in` which disables after X hours.

This shouldn't be a config, just set it to 60-80% of the new one we'll add.

server/svix-server/src/worker.rs
```diff
@@ -43,6 +46,122 @@ const USER_AGENT: &str = concat!("Svix-Webhooks/", env!("CARGO_PKG_VERSION"));
...
+#[derive(Deserialize, Serialize)]
+pub struct FailureCacheValue {
+    pub first_failure_at: DateTimeUtc,
+}
+
+/// The key is simply a formatted [`String`] containing the [`ApplicationId`] and [`EndpointId`]
+/// of the endpoint which exhausted the retry schedule.
+pub struct FailureCacheKey {
+    key: String,
+}
```
tasn, line 67: Why not use the `kv_def` macro?

server/svix-server/src/worker.rs
tasn, line 489: This one should be sent regardless of whether we are disabling...

PR #568, opened 2022-07-19
Add optional cache prefix parameter to kv_def

svix-daniel:
While the `kv_def` macro was there previously, we did not use it in some places
because of the need for custom prefixes. This PR augments the `kv_def` and
`string_kv_def` macros to take an optional string literal to be used as the
`PREFIX_CACHE` constant for the `CacheKey` trait.

Additionally this replaces the two manual implementations of the cache traits
with this new macro syntax.

### DIFF
--- a/server/svix-server/src/core/cache/mod.rs
+++ b/server/svix-server/src/core/cache/mod.rs
@@ -51,9 +51,9 @@ pub trait StringCacheValue: ToString + TryFrom<String> + Send + Sync {
     type Key: CacheKey;
 }
 
-/// A macro that creates a [`CacheKey`] and ties it to any value that implements
-/// [`DeserializeOwned`] and [`Serialize`]
-macro_rules! kv_def {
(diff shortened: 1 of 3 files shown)


server/svix-server/src/core/cache/mod.rs
```diff
@@ -54,6 +54,25 @@ pub trait StringCacheValue: ToString + TryFrom<String> + Send + Sync {
 /// A macro that creates a [`CacheKey`] and ties it to any value that implements
 /// [`DeserializeOwned`] and [`Serialize`]
 macro_rules! kv_def {
+    ($key_id:ident, $val_struct:ident, $lit_prefix:literal) => {
```
tasn, line 57: This is very redundant, can't you just make the other macro use this one? (even recursively using the same name, but also a different name if you can't use the same)

tasn (approved):

PR #569, opened 2022-07-20
Eliminate period in which dispatches could fail during improperly timed crashes

svix-daniel:
Prior to these changes, there was an error where pending tasks being reinserted
into the queue were ACKed before the reinsertion meaning an improperly timed
crash could lead to lost message tasks. These changes reorder the ACKing and
reinsertion and perform them in a Redis pipeline such as to reduce the chance of
data loss on a server crash.

### DIFF
--- a/server/svix-server/src/queue/redis.rs
+++ b/server/svix-server/src/queue/redis.rs
@@ -56,8 +56,6 @@ use super::{
 /// Confusingly, each message in the queue may have any number of KV pairs.
 const MAIN: &str = "{queue}_svix_v3_main";
 
-// TODO: Migrate DELAYED queue to use streams too
-
 /// The key for the DELAYED queue in which scheduled messages are placed. This is the same DELAYED
 /// queue as v2 of the queue implementation.
 const DELAYED: &str = "{queue}_svix_delayed";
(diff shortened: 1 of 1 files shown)


server/svix-server/src/queue/redis.rs
```diff
@@ -181,29 +179,29 @@ async fn background_task(
...
+                        } else {
+                            None
+                        }
+                    })
+                    .collect::<Vec<(&str, &[u8])>>(),
+            );
         }
+
+        // Acknowledge all the stale ones so the pending queue is cleared
+        let _ = pipe.xack(&main_queue_name, WORKERS_GROUP, &ids);
```
tasn, line 202: Can you change this to happen outside of the pipeline and after line 204?

The problem is that a pipeline is not a transaction so it may partially fail, and we want to make sure that xack only pass when the other one passes.

server/svix-server/src/queue/redis.rs
```diff
@@ -181,29 +179,29 @@ async fn background_task(
             ))
```
tasn, line 179: Take a look into xautoclaim

tasn (approved):

PR #570, opened 2022-07-21
Improve delayed queue processing to support concurrent workers

svix-daniel:
Introduces a lock on the delayed queue processing background task to allow
concurrent workers.

### DIFF
--- a/server/svix-server/src/queue/redis.rs
+++ b/server/svix-server/src/queue/redis.rs
@@ -60,6 +60,9 @@ const MAIN: &str = "{queue}_svix_v3_main";
 /// queue as v2 of the queue implementation.
 const DELAYED: &str = "{queue}_svix_delayed";
 
+/// The key for the lock guarding the delayed queue background task.
+const DELAYED_LOCK: &str = "{queue}_svix_delayed_lock";
+
 // v2 KEY CONSTANTS
 const LEGACY_V2_MAIN: &str = "{queue}_svix_main";
 const LEGACY_V2_PROCESSING: &str = "{queue}_svix_processing";
@@ -100,6 +103,7 @@ pub async fn new_pair(
         prefix.unwrap_or_default(),
         MAIN,
         DELAYED,
+        DELAYED_LOCK,
     )
     .await
 }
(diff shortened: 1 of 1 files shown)


server/svix-server/src/queue/redis.rs
```diff
@@ -32,6 +32,7 @@ use std::{num::NonZeroUsize, sync::Arc, time::Duration};
 use axum::async_trait;
 
 use chrono::Utc;
+use rand::{distributions::Alphanumeric, Rng};
```
tasn, line 35: Remove rand and use set nx=true as the locking mechanism.

server/svix-server/src/queue/redis.rs
```diff
@@ -114,39 +118,75 @@ async fn background_task(
...
+    let mut cmd = redis::cmd("SET");
+    cmd.arg(DELAYED_LOCK)
+        .arg(&rand_value)
+        .arg("NX")
+        .arg("PX")
+        .arg(5000);
+    let _: () = pool.query_async(cmd).await?;
+
+    // Then check that it was set, ensuing nothing else had the lock first
+    let lock_check: String = pool.get(DELAYED_LOCK).await?;
```
tasn, line 146: This is not how it's done, the previous command will let you know if it was set or not.

server/svix-server/src/queue/redis.rs
```diff
@@ -114,39 +117,60 @@ async fn background_task(
...
-                GENERATE_STREAM_ID,
-                &[(QUEUE_KV_KEY, task)],
-            );
+        if !keys.is_empty() {
+            // Delay every key to be scheduled ten seconds in the future so if the worker crashes
+            // during this block, the messages aren't lost
+            let mut pipe = redis::pipe();
+            for key in &keys {
+                let _ = pipe.zincr(&delayed_queue_name, key, 10_000);
+            }
```
tasn, line 145: There's no need for this anymore because you're locking access to this so if the worker crashes they'll just be picked up again.

server/svix-server/src/queue/redis.rs
```diff
@@ -319,6 +342,8 @@ async fn new_pair_inner(
                 tokio::time::sleep(Duration::from_millis(500)).await;
                 continue;
             };
+
+            sleep(Duration::from_millis(500)).await;
```
tasn, line 346: You've now rate limited this worker effectively to once every 0.5s, we should only sleep when nothing is found, not when something is found. The reason for this is because we have a batch size so we may not get everything at once.

server/svix-server/src/queue/redis.rs
```diff
@@ -114,39 +117,60 @@ async fn background_task(
...
+
+            // Finally remove the tasks from the delayed queue so they aren't resent
+            let _: () = pool
+                .query_async(Cmd::zrem(&delayed_queue_name, keys))
+                .await?;
         }
-        let _: () = pool.query_async_pipeline(pipe).await?;
-    } else {
-        // Wait for half a second before attempting to fetch again if nothing was found
-        sleep(Duration::from_millis(500)).await;
```
tasn, line 149: Things are very messy with the two unrelated tasks happening in the same function (both the xpending and the zrange). I think what we should do is split them to two separate functions and honestly just loop over them separately with different sleep schedules, etc. This is just very confusing and it's easy to put in mistakes (like you have with moving this sleep).

svix-daniel: See https://github.com/svix/svix-webhooks/pull/573

tasn (approved):

PR #589, opened 2022-07-26
Introduce svix.NullableString utility

SokratisVidros:
## Motivation

The MessageIn struct has an EventId field of type `openapi.NullableString`. The `openapi.NullableString` is defined in an internal package, so it's impossible to be imported by the end developer.

## Solution

This PR introduces the `svix.NullableString` public utility to generate an `openapi.NullableString` struct to be used as an `EventId`.

### DIFF
--- a/go/svix.go
+++ b/go/svix.go
@@ -4,8 +4,8 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
-	"time"
 	"strings"
+	"time"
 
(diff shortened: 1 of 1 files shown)


tasn:
Oh, lol, I just opened a ticket following your report on Slack, and didn't see you already had a PR open. :)

tasn (approved):

PR #594, opened 2022-07-27
Consolidate configuration validation to the cfg module

svix-daniel:
Makes miscellaneous improvements to the `ConfigurationInner` validation to provide clearer error messages and avoid validation code scattered elsewhere in the codebase.

The main change is to the Redis DSN validation and the associated helper methods for fetching the key value store backend used by both the queue and the cache.

This also changes the type of the `listen_address` from a `String` to a `SocketAddr` such as to avoid parsing it elsewhere in the code.

### DIFF
--- a/server/svix-server/src/cfg.rs
+++ b/server/svix-server/src/cfg.rs
@@ -1,7 +1,7 @@
 // SPDX-FileCopyrightText: © 2022 Svix Authors
 // SPDX-License-Identifier: MIT
 
-use std::sync::Arc;
+use std::{borrow::Cow, collections::HashMap, net::SocketAddr, sync::Arc};
 
 use figment::{
     providers::{Env, Format, Toml},
@@ -12,7 +12,7 @@ use std::time::Duration;
 use crate::{core::cryptography::Encryption, core::security::Keys, error::Result};
 use serde::{Deserialize, Deserializer};
 use tracing::Level;
-use validator::Validate;
+use validator::{Validate, ValidationError};
 
 fn deserialize_jwt_secret<'de, D>(deserializer: D) -> std::result::Result<Keys, D::Error>
 where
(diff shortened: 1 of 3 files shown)


server/svix-server/src/cfg.rs
```diff
@@ -162,6 +168,83 @@ pub struct ConfigurationInner {
...
 
+fn validate_config_complete(
+    config: &ConfigurationInner,
+) -> std::result::Result<(), ValidationError> {
+    match config.cache_type {
+        CacheType::None | CacheType::Memory => {}
+        CacheType::Redis | CacheType::RedisCluster => {
+            if config.redis_dsn.is_none() {
+                return Err(ValidationError {
+                    code: Cow::from("unset redis_dsn"),
```
tasn, line 179:
```suggestion
                    code: Cow::from("missing_field"),
```
Probably something like this? I don't think the code should actually be something specific here.

server/svix-server/src/cfg.rs
```diff
@@ -162,6 +168,83 @@ pub struct ConfigurationInner {
...
+            }
+        }
+    }
+
+    match config.queue_type {
+        QueueType::Memory => {}
+        QueueType::Redis | QueueType::RedisCluster => {
+            if config.redis_dsn.is_none() {
+                return Err(ValidationError {
+                    code: Cow::from("unset redis_dsn"),
```
tasn, line 194:
```suggestion
                    code: Cow::from("missing_field"),
```
Probably something like this? I don't think the code should actually be something specific here.

server/svix-server/src/queue/mod.rs
```diff
@@ -6,7 +6,7 @@ use serde::{Deserialize, Serialize};
 use svix_ksuid::*;
 
 use crate::{
-    cfg::{Configuration, QueueType},
+    cfg::{Configuration, QueueBackend},
```
tasn, line 9: I expected this same change in the cache backend, is it not needed there?

tasn (approved):

PR #596, opened 2022-07-28
Allow overriding redis_dsn for separate queue and cache DSNs

svix-daniel:
This adds two configuration values, `redis_dsn` and `cache_dsn` which both will override the default `redis_dsn` when set. This is such that you can use separate Redis instances/clusters for the cache and queue.

### DIFF
--- a/server/svix-server/config.default.toml
+++ b/server/svix-server/config.default.toml
@@ -45,7 +45,9 @@ db_dsn = "postgresql://postgres:postgres@pgbouncer/postgres"
 # Higher values can significantly increase performance if your database can handle it.
 db_pool_max_size = 20
 
-# The DSN for redis (can be left empty if not using redis)
+# The defualt DSN for redis. `queue_dsn` and `cache_dsn` with take precedence over this value.
+# (can be left empty if not using redis or if Redis is configured through the queue and/or cache
+# specific DSNs)
 redis_dsn = "redis://redis:6379"
 
 # The maximum number of connections for the Redis pool. Minimum value of 10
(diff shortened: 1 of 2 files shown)


server/svix-server/config.default.toml
```diff
@@ -45,25 +45,32 @@ db_dsn = "postgresql://postgres:postgres@pgbouncer/postgres"
 # Higher values can significantly increase performance if your database can handle it.
 db_pool_max_size = 20
 
-# The DSN for redis (can be left empty if not using redis)
+# The defualt DSN for redis. Deprecated for `queue_dsn` and `cache_dsn`. (can be left empty if not
```
tasn, line 48: Not deprecated anymore..

server/svix-server/src/cfg.rs
```diff
@@ -206,37 +214,36 @@ impl ConfigurationInner {
...
+            QueueType::Memory => QueueBackend::Memory,
+            QueueType::Redis => QueueBackend::Redis(dsn()),
+            QueueType::RedisCluster => QueueBackend::RedisCluster(dsn()),
         }
     }
 
     /// Fetches the configured backend information for the cache, or `None` if the [`CacheType`] is
     ///  `None`. May panic is the configuration has not been validated
     pub fn cache_backend(&self) -> CacheBackend<'_> {
+        let dsn = || {
```
tasn, line 235: Why don't just make it a struct function and then also use it in the validation function?

tasn (dismissed):

tasn (approved):

PR #599, opened 2022-07-29
Implement PATCH endpoints for relevant models

svix-daniel:
This PR implements PATCH endpoints for all models in which updating via PUT is
supported -- `application`s,`event_type`s, and `endpoint`s.


server/svix-server/src/v1/utils.rs
```diff
@@ -256,6 +256,208 @@ fn validation_errors(
         .collect()
 }
 
+/// This is an enum that will wrap every nullable field for a PATCH request. Nonnullable fields can
```
tasn, line 259: Maybe move all of this logic into its own utils submodule?

server/svix-server/src/v1/endpoints/application.rs
```diff
@@ -38,7 +41,7 @@ pub struct ApplicationIn {
     )]
     pub name: String,
 
-    #[validate(range(min = 1, message = "Application rate limits must be at least 1 if set"))]
+    #[validate(custom = "validate_rate_limit")]
```
tasn, line 44: I honestly think it was better before, the same way you did it for name below...

server/svix-server/src/v1/utils.rs
```diff
@@ -256,6 +256,208 @@ fn validation_errors(
...
+/// None via PATCHing while allowing omitted fields to be skipped when updating.
+///
+/// NOTE: You must tag these fields with `#[serde(default)]` in order for the serialization to work
+/// correctly.
+#[derive(Debug)]
+pub enum NullablePatchField<T> {
+    Absent,
+    None,
+    Some(T),
+}
```
tasn, line 271: To make it much simpler, it should just be:
```suggestion
pub enum NullablePatchField<T> {
    Absent,
    Some(T),
}
```
And then the T can be e.g. String or Option<String>

server/svix-server/src/v1/endpoints/event_type.rs
```diff
@@ -70,6 +71,30 @@ impl ModelIn for EventTypeUpdate {
...
+
+    #[serde(default, skip_serializing_if = "NullablePatchField::is_absent")]
+    schemas: NullablePatchField<serde_json::Value>,
+}
+
+impl ModelIn for EventTypePatch {
+    type ActiveModel = eventtype::ActiveModel;
+
+    fn update_model(self, model: &mut Self::ActiveModel) {
+        patch_field_non_nullable!(self, model, description);
```
tasn, line 92: Random idea, but to make it less error-prone we could maybe have:
```rust
let { description, deleted, schemas } = self;
```
Or whatever the syntax for deconstruction is, and then we'll be forced to use all of the field members below, no?

tasn (approved):
I think we can probably reduce a lot of this duplication, but I'm not currently sure how (short of a proc macro)...

PR #600, opened 2022-07-30
Ruby library: fix region detection

gadirlih:
## Motivation
Region detection fails to detect region automatically

## Solution
Swap the places of the operands around the || operator. That way if the region is detectable using AUTH_TOKEN it will use that, otherwise it will use default url.

### DIFF
--- a/ruby/lib/svix/svix.rb
+++ b/ruby/lib/svix/svix.rb
@@ -5,7 +5,7 @@ module Svix
         attr_accessor :debug
         attr_accessor :server_url
 
-        def initialize(debug = false, server_url = "https://api.svix.com")
+        def initialize(debug = false, server_url = nil)
             @debug=debug
             @server_url=server_url
         end
@@ -22,7 +22,6 @@ module Svix
 
         def initialize(auth_token, options = SvixOptions.new)
 
-            regional_url = nil
             region = auth_token.split(".").last
             if region == "us"
                 regional_url = "https://api.us.svix.com"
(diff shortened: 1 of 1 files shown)


ruby/lib/svix/svix.rb
```diff
@@ -32,7 +32,7 @@ def initialize(auth_token, options = SvixOptions.new)
                 regional_url = "https://api.in.svix.com"
             end
 
-            uri = URI(options.server_url || regional_url)
+            uri = URI(regional_url || options.server_url)
```
tasn, line 35: This is on purpose, the option server URL is higher priority and should be. Unless I'm missing something?

gadirlih: You are right. If the option server URL has the highest priority then we need to keep everything as it is, and make default value of the server_url nil and instead add the default url in the else condition after line 32

tasn: Oh, the options have a default value? If so, yeah, indeed a bug. Can you revert back this line and make the other one be null like you suggested?

gadirlih: Done

tasn (approved):
Thanks a lot, looks great!

tasn:
We'll release a new version in the coming days!

gadirlih:
Great!

tasn:
New version is being released, thanks again!

gadirlih:
Great! Thanks for the update!

PR #604, opened 2022-08-02
Typo in error.rs

areski:
Just a typo error


tasn (approved):
Whoops, nice catch. Thanks!

PR #605, opened 2022-08-02
Upsert behavior on PUT methods

### LINKED ISSUE #603: Support upserting entities on PUT requests
## Feature Request
It seems that today PUT requests don't create resources that don't already exist (unless they were previously deleted).

Could we support creating resources on all PUT requests?

### Motivation
The current behavior is a bit out of line with standard rest behavior that expects PUT requests to create a resource if it doesn't already exist https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PUT.

### Proposal

Enable upserting from PUT endpoints.

The resources that would see the most upserting are probably:
1. Applications
2. Endpoints
3. Event types

### Alternatives

It's possible to get the same behavior today with try/catch'ing a POST and PUT, but this is extra boilerplate that doesn't offer any distinct advantage or flexibility for many use cases.

svix-dylan:
This PR enables support for creation of entities during PUT requests if the provided ID did not match an existing record. Affected entities are Application, Endpoint, and Event Type. PUT requests against those entities will now return 200 to indicate the record existed and was modified, or 201 to indicate the record was created.

Closes #603.

### DIFF
--- a/server/svix-server/src/v1/endpoints/application.rs
+++ b/server/svix-server/src/v1/endpoints/application.rs
@@ -9,7 +9,7 @@ use crate::{
             AuthenticatedApplication, AuthenticatedOrganization,
             AuthenticatedOrganizationWithApplication,
         },
-        types::{ApplicationId, ApplicationUid},
+        types::{ApplicationId, ApplicationIdOrUid, ApplicationUid},
     },
     db::models::application,
     error::{HttpError, Result},
@@ -24,7 +24,7 @@ use crate::{
     },
 };
 use axum::{
-    extract::Extension,
+    extract::{Extension, Path},
     routing::{get, post},
     Json, Router,
 };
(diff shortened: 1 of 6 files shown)


server/svix-server/src/v1/endpoints/application.rs
```diff
@@ -249,16 +249,34 @@ async fn get_application(
...
 
-    let ret = app.update(db).await?;
-    Ok(Json(ret.into()))
+    let (status_code, ret) = match app {
+        Some(app) => {
+            let mut app: application::ActiveModel = app.into();
+            data.update_model(&mut app);
+            let ret = app.update(db).await?;
+
+            (StatusCode::OK, ret)
```
tasn, line 265: Why not just return from here (and the other place)?

server/svix-server/src/v1/endpoints/endpoint/crud.rs
```diff
@@ -125,36 +125,60 @@ pub(super) async fn update_endpoint(
...
+    let app_uid = app.uid;
 
-    let ret = endp.update(db).await?;
+    let (status_code, ret) = match endp {
+        Some(endp) => {
+            let mut endp: endpoint::ActiveModel = endp.into();
+            data.update_model(&mut endp);
+            let ret = endp.update(db).await?;
+
+            (StatusCode::OK, ret)
```
tasn, line 146: Why not just return from here and the other place?

server/svix-server/src/v1/endpoints/endpoint/crud.rs
```diff
@@ -125,36 +125,60 @@ pub(super) async fn update_endpoint(
...
+    let op_event = EndpointEvent {
+        app_id: &ret.app_id,
+        app_uid: app_uid.as_ref(),
+        endpoint_id: &ret.id,
+        endpoint_uid: ret.uid.as_ref(),
+    };
+
+    let op_webhook = match status_code {
+        StatusCode::OK => OperationalWebhook::EndpointUpdated(op_event),
+        StatusCode::CREATED => OperationalWebhook::EndpointCreated(op_event),
```
tasn, line 173: This is way too implicit, it's using external values to this function (the returned state) as internal control flow. The fact that this not the right way is exposed by the `panic!` below. You wouldn't need to panic.

You can just move the operation webhook above (yeah, duplicate for each, it's a minor duplication anyway, and they are inherently different things that just happen to look similar now but maybe change (e.g. we may add "changed fields" to the changed event), so de-duplication them doesn't reduce complexity, but rather adds some.

server/svix-server/src/v1/endpoints/endpoint/crud.rs
```diff
@@ -125,36 +125,60 @@ pub(super) async fn update_endpoint(
...
+
+            (StatusCode::OK, ret)
+        }
+        None => {
+            let ret = endpoint::ActiveModel {
+                app_id: Set(app.id),
+                key: Set(generate_secret(
+                    &cfg.encryption,
+                    &cfg.default_signature_type,
+                )?),
```
tasn, line 154: This is wrong, see the create method... You are always overriding the key, you should support using it.

I think you may have done it because EndpointUpdate doesn't have key? Though it's what I said on Slack, you need to change it to use EndpointIn.

At this point I'm thinking: why not just call the "create endpoint" function directly from here instead of duplicating it wrong?

server/svix-server/src/v1/endpoints/event_type.rs
```diff
@@ -258,17 +258,34 @@ async fn update_event_type(
...
-    data.update_model(&mut evtype);
+    let (status_code, ret) = match evtype {
+        Some(evtype) => {
+            let mut evtype: eventtype::ActiveModel = evtype.into();
+            data.update_model(&mut evtype);
+            let ret = evtype.update(db).await?;
 
-    let ret = evtype.update(db).await?;
-    Ok(Json(ret.into()))
+            (StatusCode::OK, ret)
```
tasn, line 273: Same comment about status code and just returning from here.

tasn (approved):

PR #606, opened 2022-08-02
Distinguish caching values and storing values persistently

svix-daniel:
Before these changes we had used the concept of the cache for features such as idempotency and others which have persistent data. This conflicts with the definition of caching, which is meant to work with ephemeral data. Thus these changes distinguish the concepts with a `SharedStore` and a `Cache` structure, which internally, both use a `KvBackend` implementor.

Note that this is not exposed in the configuration, but this split is needed as additional functionality unique to persistent storage or genuine caching is implemented.

Additionally this updates prior uses of the cache to utilize the structure most appropriate.


server/svix-server/config.default.toml
```diff
@@ -65,8 +65,7 @@ queue_type = "redis"
 # What kind of cache to use. Supported: memory, redis, rediscluster, none.
 # Redis backends must have a redis_dsn or cache_dsn configured.
 # The memory backend is recommended if you only have one instance running (not including workers). If you have
-# multiple API servers running, please use the redis backend or some functionality, (e.g. Idempotency)
-# may fail to work correctly.
+# multiple API servers running, please use the redis backend or some cached items will not be mirrored.
```
tasn, line 68: I think the previous wording was better. But you can merge the two, just change the "fail to work correctly" to "fail to work correctly because the information won't be shared across instances."

server/svix-server/src/cfg.rs
```diff
@@ -136,17 +136,17 @@ pub struct ConfigurationInner {
...
-    /// queue_dsn configured).
+    /// queue_dsn configured), and rediscluster (must also have a DSN set).
     pub queue_type: QueueType,
     /// The DSN for the Redis-backed queue. Overrides `redis_dsn`. (can be left empty if not using
     /// redis)
     pub queue_dsn: Option<String>,
 
-    /// What kind of cache to use. Supported: memory, redis (must have redis_dsn or cache_dsn
-    /// configured), none.
+    /// What kind of cache/persistent store to use. Supported: memory, redis (must have redis_dsn or
```
tasn, line 145: I'll remove the `/persistent store` -> you can just just `cache/persistent store` to `persistent cache`.

server/svix-server/src/cfg.rs
```diff
@@ -136,17 +136,17 @@ pub struct ConfigurationInner {
...
+    /// cache_dsn configured), rediscluster (must also have a DSN set) and none.
     pub cache_type: CacheType,
-    /// The DSN for the Redis-backed cache. Overrides `redis_dsn`. (can be left empty if not using
-    /// redis)
+    /// The DSN for the Redis-backed cache/persistent store. Overrides `redis_dsn`. (can be left
```
tasn, line 148: Same comment.

server/svix-server/src/cfg.rs
```diff
@@ -243,6 +247,30 @@ impl ConfigurationInner {
             CacheType::RedisCluster => CacheBackend::RedisCluster(self.cache_dsn().expect(err)),
         }
     }
+
+    pub fn shared_store_backend(&self) -> SharedStoreBackend<'_> {
+        let err = "Called [`shared_store_backend`] before validating configuration";
+
+        match self.shared_store_type() {
+            SharedStoreType::None => SharedStoreBackend::None,
```
tasn, line 255: None is not allowed for the shared store. So I guess set it to at least memory if the cache is set to None.

server/svix-server/src/cfg.rs
```diff
@@ -279,6 +307,14 @@ pub enum CacheBackend<'a> {
     RedisCluster(&'a str),
 }
 
+#[derive(Debug, Eq, PartialEq)]
+pub enum SharedStoreBackend<'a> {
+    None,
```
tasn, line 312: None not allowed.

server/svix-server/src/cfg.rs
```diff
@@ -311,6 +347,15 @@ pub enum CacheType {
     None,
 }
 
+#[derive(Clone, Copy, Debug, Deserialize)]
+#[serde(rename_all = "lowercase")]
+pub enum SharedStoreType {
+    Memory,
+    Redis,
+    RedisCluster,
+    None,
```
tasn, line 356: No None.

server/svix-server/src/core/cache.rs
```diff
@@ -0,0 +1,24 @@
+// SPDX-FileCopyrightText: © 2022 Svix Authors
+// SPDX-License-Identifier: MIT
+
+use enum_dispatch::enum_dispatch;
+
+use super::kv_backend::{memory::MemoryKv, none::NoKvBackend, redis::RedisKv, KvBackend};
```
tasn, line 6: kv is a very confusing name, and the fact that it's called "MemoryKv" vs "KvBackend" and "NoKvBackend" is even more confusing. I'd at least prefix all of them with Kv if you must, though anyhow, as said, we need a different name.

server/svix-server/src/core/idempotency.rs
```diff
@@ -375,7 +377,7 @@ mod tests {
     ) -> (JoinHandle<()>, String, Arc<Mutex<usize>>) {
         dotenv::dotenv().ok();
 
-        let cache = cache::memory::new();
+        let store = SharedStore::MemorySharedStore(MemoryKv::new());
```
tasn, line 380: This feels very redundant, if we already have SharedStore::MemorySharedStore, why not have it create a new memory kv on its own (internally)?

server/svix-server/src/v1/endpoints/health.rs
```diff
@@ -67,16 +71,41 @@ pub struct HealthReport {
...
+    HealthCheckCacheValue,
+    "SVIX_CACHE_HEALTH_CHECK"
+);
+impl HealthCheckCacheKey {
+    pub fn new(id: &str) -> HealthCheckCacheKey {
+        HealthCheckCacheKey(format!("{}{}", Self::KEY_PREFIX, id))
+    }
+}
+
+kv_def!(
```
tasn, line 93: Kinda sucks that you can reuse the same kv for both cache and store, so you're not accidentally protected against mistakes there. Can we maybe have two variants to this macro that just wrap this one with a newtype?

server/svix-server/src/lib.rs
```diff
@@ -86,15 +88,38 @@ pub async fn run_with_prefix(
...
+            Cache::RedisCache(RedisKv::new(mgr))
+        }
+    };
+
+    tracing::debug!("Store type: {:?}", cfg.shared_store_type());
+    let store = match cfg.shared_store_backend() {
+        SharedStoreBackend::None => {
+            tracing::warn!("With no shared store backend, some features such as idempotency and disabling endpoints on repeated failures will not work");
+            SharedStore::None(NoKvBackend)
+        }
```
tasn, line 111:
```suggestion
```

server/svix-server/src/lib.rs
```diff
@@ -86,15 +88,38 @@ pub async fn run_with_prefix(
...
+        SharedStoreBackend::Memory => {
+            tracing::warn!("With a memory-backed shared store, features such as idempotency and disabling endpoints on repeated failure will not work with multiple services running and will wipe on shutdown");
```
tasn, line 113:
```suggestion
            tracing::warn!("When using a memory backed shared store, features such as idempotency and disabling endpoints on repeated failure will not work correctly across multiple instances and instance restarts.");
```

server/svix-server/src/core/cache.rs
tasn: Maybe even a KeyValueStore a better name than kv (you can still use kv for the macro to keep it short).

server/svix-server/config.default.toml
```diff
@@ -64,9 +64,9 @@ queue_type = "redis"
 
 # What kind of cache to use. Supported: memory, redis, rediscluster, none.
 # Redis backends must have a redis_dsn or cache_dsn configured.
-# The memory backend is recommended if you only have one instance running (not including workers). If you have
-# multiple API servers running, please use the redis backend or some functionality, (e.g. Idempotency)
-# may fail to work correctly.
+# The memory backend is recommended if you only have one instance running, if you have multple
+# services running, then pleas use the redis backend or features like idempotency will fail to work
```
tasn, line 68:
```suggestion
# services running, then please use the redis backend or features like idempotency will fail to work
```

server/svix-server/src/core/kv_backend/none.rs
```diff
@@ -5,43 +5,38 @@ use std::time::Duration;
...
-
-pub fn new() -> Cache {
-    tracing::warn!("Running with caching disabled will negatively affect performance. Idempotency is not supported without a cache.");
-    NoCache {}.into()
-}
+use super::{KeyValueStoreBackend, Result, StringValue, Value};
 
 #[derive(Clone)]
-pub struct NoCache;
+pub struct NoKeyValueStore;
```
tasn, line 11: If you are already renaming it, please rename it to KeyValueStoreNone to be consistent with the rest.

server/svix-server/src/core/message_app.rs
```diff
@@ -181,11 +181,11 @@ impl TryFrom<endpoint::Model> for CreateMessageEndpoint {
...
 }
 
-kv_def!(AppEndpointKey, CreateMessageApp);
+cache_kv_def!(AppEndpointKey, CreateMessageApp, "SVIX_CACHE_APP_v3");
 impl AppEndpointKey {
     // FIXME: Rewrite doc comment when AppEndpointValue members are known
     /// Returns a key for fetching all cached endpoints for a given organization and application.
     pub fn new(org: OrganizationId, app: ApplicationId) -> AppEndpointKey {
-        AppEndpointKey(format!("{}_APP_v3_{}_{}", Self::PREFIX_CACHE, org, app))
+        AppEndpointKey(format!("_APP_v3_{}_{}", org, app))
```
tasn, line 189: This looks wrong? Shouldn't it be:
```suggestion
        AppEndpointKey(format!("{}_{}", org, app))
```
And I'm not sure where the _ between the prefix and the key go.

I gotta say, I'm not sure I'm a huge fan of having the prefix defined separately than the function, it makes it very hard to know what's going on. It's much easier to just see:
```rust
         AppEndpointKey(format!("SVIX_CACHE_APP_v3_{}_{}", org, app))
```
All in one place.

server/svix-server/src/worker.rs
```diff
@@ -47,42 +48,42 @@ const USER_AGENT: &str = concat!("Svix-Webhooks/", env!("CARGO_PKG_VERSION"));
...
-/// Otherwise, it is stored in the cache with an expiration.
+/// Otherwise, it is stored in the store with an expiration.
 #[derive(Deserialize, Serialize)]
-pub struct FailureCacheValue {
+pub struct FailureValue {
     pub first_failure_at: DateTimeUtc,
 }
 
-kv_def!(FailureCacheKey, FailureCacheValue, "SVIX_FAILURE_CACHE");
+store_kv_def!(FailureKey, FailureValue, "FAILURES_");
```
tasn, line 65: Missing the Svix prefix.

After seeing this, I'm 100% convinced about my other comment of splitting the prefix being confusing.

tasn (commented):
I made some comments, though I think the may problem is that this PR is just too big.

For example, I'd split the fact that cache can't be null (it's checked in idempotency) to its own PR, and just get that in.
I'd then have another PR that splits the cache and store to two, but everything will still use cache.
I'd then have another PR (or better yet, commit on the second PR) to start using the store.

It's just really really hard to understand what's going on and why a change is needed and what may be a mistake.

I'm also 95% certain we need to remove the prefix from the cache/store structs and just have the key fully defined in the new() function that we already have for each of them anyway.

PR #610, opened 2022-08-05
Server: Fix validation errors

jaymell:
Previously many of our custom validation errors were passing messages
as error codes, which caused them not to be shown by our current
validation error handling. Fixed this and added method to simplify
proper instantiation of validation errors.


tasn (approved):
Sorry, I thought I approved it already.

PR #611, opened 2022-08-05
Add Message payload validation

jaymell:
Message payload should be a valid JSON object


tasn:
Looks great from what I was able to parse from the diff, but you need to rebase on top of the other branch (or merge, and then on top of main) for it to look good.

tasn (approved):

PR #612, opened 2022-08-10
Update libraries to support multiple MessageStatus filters when getting messages

svix-dylan:
Currently, filtering by message status when listing attempted messages is constrained to filtering by just one status (eg "MessageStatus.Success"). However, it would be useful to filter by multiple statuses, particularly for statuses like MessageStatus.Pending and MessageStatus.Sending. This PR updates our libraries to support that behavior.


svix-jplatte:
We're well past the point where we can just ship breaking changes like this.

PR #614, opened 2022-08-11
Filter expired keys so they aren't used to sign webhooks

### LINKED ISSUE #613: Messages are signed using old (expired) signing keys
It's not a serious issue, but when you rotate a key it's still used for signing even if it has already expired.
The problem is in worker.rs:
```rust
        let keys: Vec<&EndpointSecretInternal> = if let Some(ref old_keys) = endp.old_signing_keys {
            iter::once(&endp.key)
                .chain(old_keys.0.iter().map(|x| &x.key))
                .collect()
        } else {
            vec![&endp.key]
        };
```
As you can see it just uses old_signing_keys without first trying to filter the endpoint keys.

svix-dylan:
We currently allow expired rotated keys to be used for signing webhooks. This PR filters old_signing_keys based on expiration, so that only unexpired keys can be used for signing.

Fixes #613.

### DIFF
--- a/server/svix-server/src/core/message_app.rs
+++ b/server/svix-server/src/core/message_app.rs
@@ -206,3 +206,52 @@ impl AppEndpointKey {
         AppEndpointKey(format!("{}_APP_v3_{}_{}", Self::PREFIX_CACHE, org, app))
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::core::cryptography::Encryption;
+    use crate::core::types::{EndpointSecret, ExpiringSigningKey};
+
+    #[test]
+    fn test_get_valid_signing_keys() {
+        let key = EndpointSecretInternal::from_endpoint_secret(
+            EndpointSecret::Symmetric(base64::decode("MfKQ9r8GKYqrTwjUPD8ILPZIo2LaLaSw").unwrap()),
+            &Encryption::new_noop(),
+        )
+        .unwrap();
+
+        let unexpired_old_key = ExpiringSigningKey {
+            key: key.clone(),
+            expiration: Utc::now()
+                + chrono::Duration::hours(ExpiringSigningKeys::OLD_KEY_EXPIRY_HOURS),
+        };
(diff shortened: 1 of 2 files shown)


server/svix-server/src/worker.rs
```diff
@@ -151,7 +151,7 @@ fn sign_msg(
     timestamp: i64,
     body: &str,
     msg_id: &MessageId,
-    endpoint_signing_keys: &[&EndpointSecretInternal],
+    endpoint_signing_keys: Vec<EndpointSecretInternal>,
```
tasn, line 154: Why change this? This feels wholly unnecessary.

server/svix-server/src/worker.rs
```diff
@@ -532,6 +520,24 @@ async fn dispatch(
     Ok(())
 }
 
+fn filter_expired_keys(
```
tasn, line 523: It doesn't really filter, it combines the key with old key and then filters.

Have you considered just adding a function on the CreateMessageEndpoint struct that's called `valid_signing_keys()` or something like that which just gets it? Feels like something that's specific to that struct anyway.

svix-dylan: I took this approach because I could unit-test the functionality without having to instantiate an endpoint. But yeah, this would probably be better off as a method of CreateMessageEndpoint. One moment I'll have it wired up that way.

server/svix-server/src/core/message_app.rs
```diff
@@ -157,6 +157,23 @@ pub struct CreateMessageEndpoint {
     pub deleted: bool,
 }
 
+impl CreateMessageEndpoint {
+    pub fn get_valid_signing_keys(&self) -> Vec<&EndpointSecretInternal> {
```
tasn, line 161:
```suggestion
    pub fn valid_signing_keys(&self) -> Vec<&EndpointSecretInternal> {
```
Getters in rust don't have the word get.

tasn (dismissed):

tasn (approved):

PR #619, opened 2022-08-16
Alter request ID field in HTTP request logs to use unique KsuidMs

svix-daniel:
Prior request IDs were only reported when a specific header value was sent in the request. These changes alter the code so a unique KsuidMs is generated per request and used as a request ID when this header value is not set.

### DIFF
--- a/server/svix-server/src/core/otel_spans.rs
+++ b/server/svix-server/src/core/otel_spans.rs
@@ -11,6 +11,7 @@ use opentelemetry::trace::TraceContextExt;
 use svix_ksuid::{KsuidLike, KsuidMs};
 use tower_http::{
     classify::ServerErrorsFailureClass,
+    request_id::RequestId,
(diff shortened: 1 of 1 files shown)


server/svix-server/src/core/otel_spans.rs
```diff
@@ -60,8 +60,8 @@ impl<B> MakeSpan<B> for AxumOtelSpanCreator {
 
         let request_id = request
             .extensions()
-            .get::<RequestId>()
-            .and_then(|id| id.header_value().to_str().ok())
+            .get::<KsuidMs>()
+            .map(ToString::to_string)
```
tasn, line 64: Why do we need this extra struct and request_ids.rs file? Why not literally just do:
```rust
let request_id = KsuidMs::generate(None ,None);
```
?

tasn (approved):

PR #643, opened 2022-08-30
Override retry schedule per event type

svix-dylan:
Allows a user to set a specific retry schedule, of variable length, per event type. Introduces a migration to add a column to the event type table (retry_schedule), a new type (RetrySchedule), a retry schedule CRUD (GET, PUT), and a modification to the worker to override the global retry schedule when an event type has a custom retry schedule defined.

Add or update an event type's custom retry schedule (the array `retrySchedule` can be of variable length, integers represent seconds):
```sh
curl -X "PUT" "[SERVER_URL]/api/v1/event-type/[EVENT_NAME]/retry-schedule/" \
     -H 'Authorization: Bearer abc' \
     -H 'Content-Type: application/json' \
     -d $'{
  "retrySchedule": [10, 30, 60]
}'
```

Remove an event type's custom retry schedule:
```sh
curl -X "PUT" "[SERVER_URL]/api/v1/event-type/[EVENT_NAME]/retry-schedule/" \
     -H 'Authorization: Bearer abc' \
     -H 'Content-Type: application/json' \
     -d $'{
  "retrySchedule": null
}'
```

Get an event type's current custom retry schedule:
```sh
curl -X "GET" "[SERVER_URL]/api/v1/event-type/[EVENT_NAME]/retry-schedule/" \
     -H 'Authorization: Bearer abc' \
```


server/svix-server/src/core/types.rs
```diff
@@ -427,6 +428,20 @@ impl Validate for EventTypeNameSet {
...
 }
 
+#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
+pub struct RetrySchedule(pub Vec<u64>);
+json_wrapper!(RetrySchedule);
+
+impl RetrySchedule {
+    pub fn as_durations(&self) -> Vec<Duration> {
+        self.0
+            .clone()
```
svix-dylan, line 438: We can remove this clone if we are ok consuming self. Consuming self happens to work out fine in this PR as its implemented, but left it this way in case it could be useful in the future. Happy to remove.

svix-dylan:
Needs: tests

server/svix-server/src/core/types.rs
```diff
@@ -427,6 +428,20 @@ impl Validate for EventTypeNameSet {
     }
 }
 
+#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
+pub struct RetrySchedule(pub Vec<u64>);
```
tasn, line 432:
```suggestion
pub struct RetrySchedule(pub Vec<Duration>);
```
Will simplify everything.

server/svix-server/src/worker.rs
```diff
@@ -401,10 +401,15 @@ async fn dispatch(
         Err((attempt, err)) => {
             let attempt = attempt.insert(db).await?;
 
+            let retry_schedule = match endp.retry_schedule {
+                Some(retry_schedule) => retry_schedule.as_durations(),
+                None => cfg.retry_schedule.clone(),
```
tasn, line 406:
```suggestion
            let retry_schedule = match endp.retry_schedule {
                Some(retry_schedule) => retry_schedule.as_durations(),
                None => cfg.retry_schedule.clone(),
```
Don't clone, just borrow...

server/svix-server/src/core/types.rs
svix-dylan: Unfortunately not -- this approach is simpler than serializing/deserializing Durations.

server/svix-server/src/worker.rs
svix-dylan: This is trickier than it looks, at least to me. I'm not sure how to make `retry_schedule.to_durations()` -- or rather, `iter.collect()` -- return a reference.

server/svix-server/src/v1/endpoints/retry_schedule.rs
```diff
@@ -0,0 +1,75 @@
...
+#[serde(rename_all = "camelCase")]
+struct RetryScheduleInOut {
+    retry_schedule: Option<RetrySchedule>,
+}
+
+async fn get_retry_schedule(
+    Extension(ref db): Extension<DatabaseConnection>,
+    Path(evtype_name): Path<EventTypeName>,
+    AuthenticatedOrganization { permissions }: AuthenticatedOrganization,
+) -> Result<(StatusCode, Json<RetryScheduleInOut>)> {
```
tasn, line 34:
```suggestion
) -> Result<Json<RetryScheduleInOut>> {
```

server/svix-server/src/v1/endpoints/retry_schedule.rs
```diff
@@ -0,0 +1,75 @@
+// SPDX-FileCopyrightText: © 2022 Svix Authors
```
tasn, line 1: I'd put these in the same file as the other event_type stuff, just at the bottom. Or at the very least name it event_type_retry_schedule. Though this doesn't conform with the rest of our organization, so putting it in the other file makes more sense.

server/svix-server/src/v1/endpoints/retry_schedule.rs
```diff
@@ -0,0 +1,75 @@
...
+
+    Ok((StatusCode::OK, Json(retry_schedule)))
+}
+
+async fn update_retry_schedule(
+    Extension(ref db): Extension<DatabaseConnection>,
+    Path(evtype_name): Path<EventTypeName>,
+    ValidatedJson(data): ValidatedJson<RetryScheduleInOut>,
+    AuthenticatedOrganization { permissions }: AuthenticatedOrganization,
+) -> Result<(StatusCode, Json<RetryScheduleInOut>)> {
```
tasn, line 53:
```suggestion
) -> Result<Json<RetryScheduleInOut>> {
```

server/svix-server/src/v1/endpoints/retry_schedule.rs
```diff
@@ -0,0 +1,75 @@
...
+
+    Ok((StatusCode::OK, Json(retry_schedule)))
+}
+
+pub fn router() -> Router {
+    Router::new().route(
+        "/event-type/:event_type_name/retry-schedule/",
+        get(get_retry_schedule).put(update_retry_schedule),
+    )
+}
```
tasn, line 75: Missing tests, both for the endpoints and for the worker with an overriden retry schedule.

server/svix-server/src/worker.rs
```diff
@@ -401,10 +421,18 @@ async fn dispatch(
         Err((attempt, err)) => {
             let attempt = attempt.insert(db).await?;
 
+            let retry_schedule_override =
+                get_retry_schedule_override(db, cache, org_id, event_type_name).await?;
+
+            let retry_schedule = match retry_schedule_override {
+                Some(retry_schedule) => retry_schedule.to_durations(),
+                None => cfg.retry_schedule.to_owned(),
+            };
```
tasn, line 430: `to_owned()` is essentially a clone() here. It'll also copy the data...

Just do something like:
```rust
let retry_schedule = retry_schedule_override.map(|x| x.to_durations()).unwrap_or(cfg.retry_schedule);
```

server/svix-server/src/worker.rs
```diff
@@ -546,6 +574,43 @@ fn bytes_to_string(bytes: bytes::Bytes) -> String {
...
+        .map_err(|e| Error::from(e.to_string()))?
+    {
+        return Ok(retry_schedule.retry_schedule);
+    };
+
+    let retry_schedule =
+        eventtype::Entity::secure_find_by_name(org_id.to_owned(), event_type_name.to_owned())
+            .one(db)
+            .await?
+            .unwrap()
```
tasn, line 597: ... unwrap...

server/svix-server/src/worker.rs
svix-dylan: Yeah, I know to_owned is often just a clone. That code above does not work either, because cfg.retry_schedule is behind an Arc and does not implement Copy. Although I do like it in that its terser than a match haha.

server/svix-server/src/worker.rs
```diff
@@ -546,6 +574,43 @@ fn bytes_to_string(bytes: bytes::Bytes) -> String {
...
+            .retry_schedule;
+
+    cache
+        .set(
+            &cache_key,
+            &RetryScheduleCacheValue {
+                retry_schedule: retry_schedule.clone(),
```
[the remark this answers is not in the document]
svix-dylan: Yes -- I also return `retry_schedule` at the end of the function.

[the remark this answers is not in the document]
tasn: I'm pretty sure you're incorrect, you can just do:
```rust
let retry_schedule_value = RetryScheduleCacheValue { retry_schedule };
cach.set(...., &retry_schedule_value);
...
Ok(retry_schedule_value.retry_schedule).
```

server/svix-server/src/v1/endpoints/retry_schedule.rs
tasn: Why did you resolve these? They are not fixed.

svix-dylan: I'm working through them locally and thought I'd push sooner.

server/svix-server/src/worker.rs
svix-dylan: We don't want `&Option<Vec<Duration>>`, we want `&Vec<Duration>`. The problem is that `Vet<Duration>` is created inside the `to_durations()` method. The compiler suggests binding it with `let` so that it lives longer, but we'd have to unwrap it from the Option first -- which is when we fallback on `cfg.retry_schedule`.

server/svix-server/src/v1/endpoints/retry_schedule.rs
tasn: 👍🏻

server/svix-server/src/worker.rs
tasn: I had a small mistake there, here's the full code:
```rust
let retry_schedule_orig = retry_schedule_override.map(|x| x.to_durations());
let retry_schedule = retry_schedule_orig.as_deref().unwrap_or(&cfg.retry_schedule);
```

server/svix-server/src/worker.rs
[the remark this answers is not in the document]
svix-dylan: Unfortunately no, cannot move out of retry_schedule_value.retry_schedule which is behind a shared reference.

server/svix-server/src/worker.rs
svix-dylan: Oh neat. I will remember this for the future.

server/svix-server/src/worker.rs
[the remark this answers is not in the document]
tasn: I'm pretty sure you have a small error there, but ANYHOW, the only reason why we even need to think about this and deal with this is because we have the wrong type in the first place...
It's trivial to do the serialize and desrialize for duration.
```rust
            let buf: Vec<u64> = Vec::deserialize(deserializer)?;
            Ok(buf.into_iter().map(|x| Duration::new(x, 0)).collect())
```
This is roughly the deserialize function. I'm sure the serialize one is even simpler. We already have it (kinda, though with hacks because of legacy stuff) in cfg.rs, this will be solved if the typing was just correct.

server/svix-server/src/worker.rs
tasn: It's quite possible you can even reuse the name (so no need for the _orig nonsense), I just didn't want to potentially have it not work.

server/svix-server/src/worker.rs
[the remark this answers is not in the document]
tasn: OK, I mixed up two topics it seems, ignore my previous message. :)

Though you can definitely make it work, just put the `cache.set` inside a block of curlys `{}`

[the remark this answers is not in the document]
svix-dylan: I'll take another look at the typing/serialization regardless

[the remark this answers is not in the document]
tasn: It's really easy. You can take a look at types.rs for inspiration, we do it for a lot of types.

server/svix-server/src/core/types.rs
```diff
@@ -427,6 +428,35 @@ impl Validate for EventTypeNameSet {
...
+impl Serialize for RetrySchedule {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        let mut seq = serializer.serialize_seq(Some(self.0.len()))?;
+        for e in self.clone().0 {
+            seq.serialize_element(&e.as_secs())?;
+        }
+        seq.end()
```
tasn, line 444: This is fine, but you can literally do the reverse of what we did in deserialize (simpler):
```rust
let buf: Vec<u64> = self.0.map(|x| x.as_secs()).collect();
buf.serialize(serializer)
```

tasn (approved):

PR #646, opened 2022-09-06
Correct request ID handling in tracing spans

svix-daniel:
This fixes a bug where the `x-request-id` header is not used and corrects the functionality to support the `svix-req-id` header when creating the spans for each request to the API server.

`x-request-id` is used before `svix-req-id`, and if both are unset, then the request ID will be a randomly generated `KsuidMs`.


tasn:
Why wasn't it used? Where was the bu?

svix-daniel:
> Why wasn't it used? Where was the bug?

Essentially the way that we were fetching it required an additional extra layer to set the `RequestId` extension. I figured just grabbing it in the `AxumOtelSpanCreator` was simpler than an additional layer just to set the `RequestId` when we don't even forward it anywhere.

tasn (approved):

PR #652, opened 2022-09-13
Server: Improve error logging

jaymell:
Ensure that all server-side errors are logged at error level.
Also fix spelling error.


jaymell:
Note that these changes will cause >1 error log per error in cases where the error is already logged before returning response. The goal here is to avoid ever missing 500 errors in DEBUG level logs.

tasn (approved):

Issue #574, opened 2022-07-22
Suggestion: use the error_chain create for better error handling


tasn:
We can potentially use the https://docs.rs/error-chain/latest/error_chain/ crate for better error handling and stack traces.
I'm not sure how needed it is, because with Rust most issues are enforced at compile time, but we could still have dynamo or something errors and it would be better to know where they came from.

svix-gabriel:
Right now I'm leaning on `anyhow`, which gives us access to backtraces or manually tracing using `.with_context(...)`. It's easy to implement `From<anyhow::Error>` for our custom error type in `crate::error::Error`, but the downside is that we'd need to add `.with_context(...)` calls everywhere we want to add tracing.

I've experimented with a couple of other approaches, and the issue I keep coming back to is returning the `anyhow::Error` from an axum handler. I think that if we want to reliably and automatically capture backtraces, without using `.with_context(...)`, we need the error type that we return from our handlers to wrap an `anyhow::Error` in one way or another. I think this necessitates a non-trivial refactor of the `crate::error::Error` type.

Maybe there's another approach I'm not thinking of - I guess I want to double check that carving up the `crate:error::Error` struct to wrap an `anyhow::Error` is a reasonable approach.

PR #653, opened 2022-09-14
Add file/line information to main error type with macros

svix-gabriel:
This closes https://github.com/svix/svix-webhooks/issues/574, by adding diagnostic information to the main error type along with some helpful macros.

## Motivation

For internal server errors, we want to visibility into both the underlying error, in addition to where the error originated from. e.g. for code like this
```rust
fn foo() -> crate::error::Result<()> {
   let app = Application.fetch(...)?;
   app.bar = 12;
   app.save()?;
   Ok(())
}
```
Just looking at the logs, it can be hard to diagnose whether the underlying db error came from `Application.fetch(...)?` or `app.save()?`.

We want to be able to quickly correlate an error to a specific line, while still keeping error handling relatively boilerplate free.

## Solution

First, we extend the `error::Error` type with some diagnostic information - literally just the line number and file name.

To avoid really boilerplate heavy code, most of the `error::Error` initialization logic is now handled by macros. `queue_err!`, `generic_err!`, etc. This makes it pretty easy to add the file/line information without making error initialization excessively verbose (in some cases, we're actually able to make error initialization easier than before).

One point of friction is that we can't easily attach `location` information when propagating errors using `?` - which we lean on pretty heavily when converting `DbErr`, `RedisError`, `bb8::RunError`, etc. Adding the location information to an error should be easy, but it should also be something that is difficult to forget to do.

To circumvent this problem, I dropped the `From<...> for Error` implementations for `DbErr`, `RedisError`, etc. These were replaced with a very similar trait to just transforms `std::result::Result<T, E>` into `crate::error::Result<T>` for `DbErr`, `RedisError`, etc. There's then a `ctx!` macro that handles the conversion, and adds the `location` information. Now error handling looks closer too

```rust
fn foo() -> crate::error::Result<()> {
   let app = ctx!(Application.fetch(...))?;
   app.bar = 12;
   ctx!(app.save())?;
   Ok(())
}
```

So, marginally more boilerplate. But forgetting to include `ctx!` now becomes a compiler error, forcing developers to add the diagnostic information using `ctx!`.

In addition, `ctx!` adds tracing information if the error is already a `crate::result::Result`. 

What does this end up looking like in practice? If you have code like this
```rust
fn fallible_a() -> std::result::Result<(), DbErr> {
    Err(DbErr::Custom("db borked!".into()))
}

fn fallible_b() -> Result<()> {
    ctx!(fallible_a()) // line 182 in application.s
}

// . . .

ctx!(fallible_b())?; // line 190 in application.rs
```
when the error is transformed into a response from the axum handlers, we get logs like
```
2022-09-14T14:13:04.156518Z ERROR HTTP request{...}: svix_server::error: type: Database("Custom Error: db borked!"), location: ["svix-server/src/v1/endpoints/application.rs:182", "svix-server/src/v1/endpoints/application.rs:190"]
```

### DIFF
--- a/server/svix-server/src/core/cryptography.rs
+++ b/server/svix-server/src/core/cryptography.rs
@@ -8,7 +8,8 @@ use chacha20poly1305::{Key, XChaCha20Poly1305, XNonce};
 use ed25519_compact::*;
 use rand::Rng;
 
-use crate::error::{Error, Result};
+use crate::err_generic;
+use crate::error::Result;
 
 // Asymmetric Signature keys
 #[derive(Clone, Eq)]
@@ -20,14 +21,13 @@ impl AsymmetricKey {
     }
 
     pub fn from_slice(bytes: &[u8]) -> Result<Self> {
-        Ok(AsymmetricKey(KeyPair::from_slice(bytes).map_err(|_| {
-            Error::Generic("Failed parsing key.".to_string())
-        })?))
+        Ok(AsymmetricKey(
+            KeyPair::from_slice(bytes).map_err(|_| err_generic!("Failed parsing key."))?,
+        ))
     }
 
     pub fn from_base64(b64: &str) -> Result<Self> {
-        let bytes =
-            base64::decode(b64).map_err(|_| Error::Generic("Failed parsing base64".to_string()))?;
+        let bytes = base64::decode(b64).map_err(|_| err_generic!("Failed parsing base64"))?;
 
         Self::from_slice(bytes.as_slice())
     }
@@ -74,7 +74,7 @@ impl Encryption {
             let nonce = XNonce::from_slice(&nonce);
             let mut ciphertext = cipher
                 .encrypt(nonce, data)
-                .map_err(|_| crate::error::Error::Generic("Encryption failed".to_string()))?;
+                .map_err(|_| err_generic!("Encryption failed"))?;
             let mut ret = nonce.to_vec();
             ret.append(&mut ciphertext);
             Ok(ret)
@@ -90,7 +90,7 @@ impl Encryption {
             let ciphertext = &ciphertext[Self::NONCE_SIZE..];
             cipher
                 .decrypt(XNonce::from_slice(nonce), ciphertext)
-                .map_err(|_| crate::error::Error::Generic("Encryption failed".to_string()))
+                .map_err(|_| err_generic!("Encryption failed"))
         } else {
             Ok(ciphertext.to_vec())
         }
--- a/server/svix-server/src/core/types.rs
+++ b/server/svix-server/src/core/types.rs
@@ -19,7 +19,7 @@ use std::ops::Deref;
 use svix_ksuid::*;
 use validator::{Validate, ValidationErrors};
 
-use crate::v1::utils::validation_error;
+use crate::{err_generic, v1::utils::validation_error};
 
 use super::cryptography::{AsymmetricKey, Encryption};
 
@@ -470,8 +470,8 @@ impl EndpointSecretMarker {
     fn from_u8(v: u8) -> crate::error::Result<Self> {
         let encrypted = (v & Self::ENCRYPTED_FLAG) != 0;
         let v = v & !Self::ENCRYPTED_FLAG;
-        let type_ = EndpointSecretType::try_from(v)
-            .map_err(|_| crate::error::Error::Generic("Invalid marker value".to_string()))?;
+        let type_ =
+            EndpointSecretType::try_from(v).map_err(|_| err_generic!("Invalid marker value"))?;
 
         Ok(Self { type_, encrypted })
     }
@@ -542,9 +542,7 @@ impl EndpointSecretInternal {
     fn from_vec(v: Vec<u8>) -> crate::error::Result<Self> {
         // Legacy had exact size
         match v.len() {
-            0..=Self::KEY_SIZE_MINUS_ONE => {
-                Err(crate::error::Error::Generic("Value too small".to_string()))
-            }
+            0..=Self::KEY_SIZE_MINUS_ONE => Err(err_generic!("Value too small")),
             Self::KEY_SIZE => Ok(Self {
                 marker: EndpointSecretMarker {
                     type_: EndpointSecretType::Hmac256,
@@ -608,9 +606,7 @@ impl EndpointSecretInternal {
             if encryption.enabled() {
                 encryption.decrypt(&self.key)?
             } else {
-                return Err(crate::error::Error::Generic(
-                    "main_secret unset, can't decrypt key".to_string(),
-                ));
+                return Err(err_generic!("main_secret unset, can't decrypt key"));
             }
         } else {
             self.key.to_vec()
--- a/server/svix-server/src/error.rs
+++ b/server/svix-server/src/error.rs
@@ -16,8 +16,180 @@ use serde_json::json;
 pub type Result<T> = std::result::Result<T, Error>;
 
 /// The error type returned from the Svix API
+#[derive(Debug)]
+pub struct Error {
+    // the file name and line number of the error. Used for debugging non Http errors
+    pub trace: Vec<&'static str>,
+    pub typ: ErrorType,
+}
+
+impl Error {
+    pub fn generic(s: impl fmt::Display, location: &'static str) -> Self {
+        Self {
+            trace: Self::init_trace(location),
+            typ: ErrorType::Generic(s.to_string()),
+        }
+    }
+
+    pub fn database(s: impl fmt::Display, location: &'static str) -> Self {
+        Self {
+            trace: Self::init_trace(location),
+            typ: ErrorType::Database(s.to_string()),
+        }
+    }
+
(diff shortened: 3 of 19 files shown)


server/svix-server/src/core/cryptography.rs
```diff
@@ -8,7 +8,8 @@ use chacha20poly1305::{Key, XChaCha20Poly1305, XNonce};
 use ed25519_compact::*;
 use rand::Rng;
 
-use crate::error::{Error, Result};
+use crate::error::Result;
+use crate::generic_err;
```
svix-gabriel, line 12: I couldn't figure out how to scope the macros in `error.rs` in the actual `error` module, so we could call `error::generic!(...)` instead of `generic_err!(...)`. Suggestions there would be helpful.

server/svix-server/src/error.rs
```diff
@@ -54,53 +72,130 @@ impl error::Error for Error {
...
 
-impl<T: error::Error + 'static> From<bb8::RunError<T>> for Error {
-    fn from(err: bb8::RunError<T>) -> Self {
-        Error::Queue(err.to_string())
-    }
+/// Returns a [&'static str] of the file.rs\[<line_number>\].
+#[macro_export]
+macro_rules! location {
+    () => {
+        concat!(file!(), "[", line!(), "]")
```
tasn, line 94:
```suggestion
        concat!(file!(), ":", line!())
```
More standard?

server/svix-server/src/error.rs
```diff
@@ -16,35 +16,53 @@ use serde_json::json;
...
-    Queue(String),
-    /// Database error
-    Validation(String),
-    /// Any kind of HttpError
-    Http(HttpError),
+#[derive(Debug)]
+pub struct Error {
+    // the file name and line number of the error. Used for debugging non Http errors
+    pub location: &'static str,
+    pub error_type: ErrorType,
```
tasn, line 23: Nitpick:
```suggestion
    pub type_: ErrorType,
```
or `t` or `typ`. The word error is redundant.

server/svix-server/src/error.rs
```diff
@@ -16,35 +16,53 @@ use serde_json::json;
...
-    /// Queue error
-    Queue(String),
-    /// Database error
-    Validation(String),
-    /// Any kind of HttpError
-    Http(HttpError),
+#[derive(Debug)]
+pub struct Error {
+    // the file name and line number of the error. Used for debugging non Http errors
+    pub location: &'static str,
```
tasn, line 22: Correct me if I'm wrong, but we are not maintaining the error chain? E.g. we are missing location (and type) of the caller? Maybe it makes sense to have a linked-list type of thing of vectors (or heck, a pre-allocated vec of size 10 that should always be sufficient) so we keep the trace?

tasn: Every time we call `ctx!` btw, obviously it'll be lost if we don't.

svix-gabriel: Yeah, good catch.

Should be updated now so that `ctx!` also adds the tracing (I ended up renaming the `IntoSvixError` trait, hopefully it makes a bit more sense now).

PR description should also be updated showing what the end "trace" result looks like.

I didn't go through and add `ctx!(...)` to _every_ single relevant callsite, though I can if you want (it'll be a bit tedious). Unfortunately, I'm not sure how to make forgetting to call `ctx!` on a `crate::error::Result` a compiler error like I was when forgetting to call `ctx!` on a normal `Result`.

server/svix-server/src/core/cryptography.rs
tasn: Oh right, macros are always crate-wide?

So maybe at least prefix it with err? So `err_generic` so it's at least namespaced like this, even if poorly.

server/svix-server/src/error.rs
tasn: Nah, no need to add it everywhere for now.

Btw, I remember you preferred `e!`, why the change? (I don't care much either way)

svix-gabriel: Haha, I still _personally_ prefer `e!` just for brevity. But I also think `ctx!` is more descriptive / it's easier to infer what `ctx!` is doing.

Maybe not though 🤷.

tasn: I think I agree. :)

server/svix-server/src/core/cryptography.rs
```diff
@@ -74,7 +74,7 @@ impl Encryption {
             let nonce = XNonce::from_slice(&nonce);
             let mut ciphertext = cipher
                 .encrypt(nonce, data)
-                .map_err(|_| crate::error::Error::Generic("Encryption failed".to_string()))?;
+                .map_err(|_| crate::err_generic!("Encryption failed"))?;
```
svix-daniel, line 77: Is the `crate::` prefix necessary here (or line 93)?

server/svix-server/src/core/types.rs
```diff
@@ -471,7 +471,7 @@ impl EndpointSecretMarker {
         let encrypted = (v & Self::ENCRYPTED_FLAG) != 0;
         let v = v & !Self::ENCRYPTED_FLAG;
         let type_ = EndpointSecretType::try_from(v)
-            .map_err(|_| crate::error::Error::Generic("Invalid marker value".to_string()))?;
+            .map_err(|_| crate::err_generic!("Invalid marker value"))?;
```
svix-daniel, line 474: Is there a reason not to `use crate::err_generic` and get rid of the prefix? I imagine that macro name is pretty unique. (Also lines 545 and 609)

server/svix-server/src/error.rs
```diff
@@ -16,35 +16,59 @@ use serde_json::json;
...
-    Generic(String),
-    /// Database error
-    Database(String),
-    /// Queue error
-    Queue(String),
-    /// Database error
-    Validation(String),
-    /// Any kind of HttpError
-    Http(HttpError),
+#[derive(Debug)]
```
svix-daniel, line 19: Not a big concern, but how does the debug print look like in the terminal? It might make more sense to implement `std::fmt::Debug` instead of deriving it?

Should there be alternative formats depending on log format?

Again, not super important, at least for the first draft.

server/svix-server/src/error.rs
```diff
@@ -54,53 +78,139 @@ impl error::Error for Error {
...
-    fn from(err: DbErr) -> Error {
-        if let DbErr::Query(ref err_str) = err {
-            // Surely there's a better way
-            if err_str.contains("duplicate key value violates unique constraint") {
-                Error::Http(HttpError::conflict(None, None))
+impl<T> Traceable<T> for std::result::Result<T, DbErr> {
+    fn trace(self, location: &'static str) -> Result<T> {
+        self.map_err(|err| {
+            let typ = if let DbErr::Query(ref err_str) = err {
+                // Surely there's a better way
```
svix-daniel, line 162: Without changes to SeaORM, I don't think there is one. I'd probably remove the comment.

svix-daniel:
Overall, these changes are pretty good. It's a shame that the syntax can't be nicer, but I understand it'd require a lot more effort to get it to look more natural.

One question:
> In addition, ctx! adds tracing information if the error is already a crate::result::Result.
Is there any way to enforce adding a `ctx!` invocation easily? It'd be useful to get the traceback as bulky as possible (within reason), so I'd want to get it to as close to the top as possible in its use.

I have one really hacky idea for that, but it'd be terribly hard to do, so I'll leave the topic be if there's no obvious solution.

server/svix-server/src/core/types.rs
tasn: I think having some prefix is nice. Because we'll also have `database!` which is less unique...

svix-gabriel: I didn't have a specific reason to use the `crate::`, other than it's what was already there. Removed.

server/svix-server/src/core/cryptography.rs
svix-gabriel: Nope (same reason as before). Removed

svix-gabriel:
> Is there any way to enforce adding a ctx! invocation easily?

Yeah, I definitely agree having some kind of compile time enforcement for calling `ctx!` would be nice. I can't think of any reasonable approach though

server/svix-server/src/core/otel_spans.rs
```diff
@@ -128,7 +128,7 @@ impl<B> MakeSpan<B> for AxumOtelSpanCreator {
         );
 
         if let Some(key) = idempotency_key {
-            span.record("idempotency_key", &key);
+            span.record("idempotency_key", key);
```
tasn, line 131: Why all of these unrelated changes?

svix-gabriel: I was getting clippy failures in the build https://github.com/svix/svix-webhooks/actions/runs/3136424848/jobs/5093319099#step:8:727

I can remove these diffs, but I didn't think the PR would be mergable unless they passed

svix-gabriel: Fixed

svix-daniel (approved):

PR #659, opened 2022-09-30
Further reorganization of security module

svix-dylan:
Small changes to security module to support downstream
refactoring.


svix-gabriel (approved):

svix-gabriel:
Maybe just add some info to the pr / commit message on the motivation? I know what `This should be the last small refactoring needed.` means from context, but I could easily see forgetting in a year

PR #668, opened 2022-10-13
Update the event-type CRUD to more strictly enforce Schema type

svix-gabriel:
The current event-type CRUD is too permissive with the `schema` field. It accepts any json (such as `1`, "foobar", etc.), but it should only accept json objects (such as `{"foo":"bar"}`).

This PR updates the event-type CRUD groups.

In addition, I wanted to support calling `json_wrapper!` outside of types.rs. `types.rs` is already quite large, and I think it makes a bit more sense for types (such as `eventtype::Schema`) to be colocated with other types they interact with.

Finally, I updated the error logging to add trace information for 4xxs. This wasn't strictly necessary, but it made diagnosing the source of 4xxs when testing locally much easier.

### DIFF
--- a/server/svix-server/src/core/types.rs
+++ b/server/svix-server/src/core/types.rs
@@ -89,6 +89,7 @@ macro_rules! enum_wrapper {
     };
 }
 
+#[macro_export]
 macro_rules! json_wrapper {
     ($name_id:ty) => {
         impl From<$name_id> for sea_orm::Value {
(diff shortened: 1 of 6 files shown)


server/svix-server/src/core/types.rs
```diff
@@ -89,6 +89,8 @@ macro_rules! enum_wrapper {
     };
 }
 
+// LEFT OFF HERE
```
svix-gabriel, line 92:
```suggestion
```

svix-dylan (approved):

PR #674, opened 2022-10-18
Simplify security.rs and authorization logic

svix-gabriel:
This PR refactors `security.rs` and our auth* related axum extractors.

## Motivation

* Cut down on lines-of-code
* Make permission/authorization parsing a bit more intuitive for axum handlers
* Better leverage the type system to restrict axum handlers only to the permission-level information they need.

## Solution

Some specifics
* Many of our extractors had redundant information, or largely unused fields. This PR removes them
* The `Permissions` struct has been simplified a bit, namely by moving things into an enum (as opposed to using the `type_` enum just as a tag.
* Removed the axum extractor for `Permissions`. We often chain authorization logic on top of the `Permissions`. However, the axum extractor for `Permissions` only authenticates requests, it doesn't authorize them. 
* Fixes a discrepancy between `list_event_types` and `get_event_type`. `list_event_types` was not checking for Organization privileges like `get_event_types` was.
* `AuthenticatedOrganization`, `AuthenticatedApp`, etc, were also simplified a bit, and moved into a separate `permissions.rs` module.

### DIFF
--- a/server/svix-server/src/core/mod.rs
+++ b/server/svix-server/src/core/mod.rs
@@ -7,6 +7,7 @@ pub mod idempotency;
 pub mod message_app;
 pub mod operational_webhooks;
 pub mod otel_spans;
+pub mod permissions;
 pub mod run_with_retries;
 pub mod security;
 pub mod types;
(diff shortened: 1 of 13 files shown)


svix-dylan (commented):
Permissions module seems misnamed, since the Permissions struct is defined in security.rs. Is there a better name?

svix-dylan (approved):

PR #675, opened 2022-10-19
Handle failure to read a response body after dispatch

svix-daniel:
It was noticed that there was an `expect` call on a fallible method to retrieve a response body from a given response to webhook dispatch. While it is rare for this method to fail, it is possible that a panic occurred before these changes.

With these changes, it now records such failures as if the whole request was a failure in that a failed message attempt is recorded with the error message as the response and a reported status code of 0.

### DIFF
--- a/server/svix-server/src/worker.rs
+++ b/server/svix-server/src/worker.rs
@@ -334,23 +334,35 @@ async fn dispatch(
             };
             let http_error = res.error_for_status_ref().err();
 
-            let bytes = res
-                .bytes()
-                .await
-                .expect("Could not read endpoint response body");
-            let body = bytes_to_string(bytes);
+            let attempt = match res.bytes().await {
+                Ok(bytes) => {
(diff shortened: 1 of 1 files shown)


server/svix-server/src/worker.rs
```diff
@@ -334,21 +334,32 @@ async fn dispatch(
             };
             let http_error = res.error_for_status_ref().err();
 
-            let bytes = res
-                .bytes()
-                .await
-                .expect("Could not read endpoint response body");
-            let body = bytes_to_string(bytes);
+            match res.bytes().await {
```
svix-james, line 337: It's not clear to me that this should be treated as a failed attempt. Presumably we got back a successful status, but for whatever reason reading the body timed out. Why not log/record the failure reading the body but treat it as a success? Otherwise, we will surely be sending duplicate messages when we retry.

svix-james (approved):

PR #677, opened 2022-10-25
Update event-type schemas to be stricter

svix-gabriel:
This updates the /event-type endpoints "schemas" parsing to match our API more closely.

## Motivation

In our API, we specifically parse event-type schemas as a map of version numbers to Json Schemas. While we don't enforce the version numbers, we do enforce that the Json schemas are, well, valid JSON schemas!

The open-source repo was allowing arbitrary JSON objects, which is too lax.

## Solution

We use the `jsonschemas` library to parse Json Schemas. This library has a compatible license (MIT), no known vulnerabilities, is actively maintained, with modest popularity (283 stars in Github, a little over half a million downloads).

We still need to (de)serialize the `event_type::Schema`, and `jsonschema::JSONSchema` doesn't implement any of the (de)serialization traits. Consequently, the type definition is still a wrapper over a Hash of strings to Json, and we manually implement deserialize to enforce that that the json schema is always correct.

### DIFF
--- a/server/Cargo.lock
+++ b/server/Cargo.lock
@@ -29,6 +29,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "ahash"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "57e6e951cfbb2db8de1828d49073a113a29fd7117b1596caa781a258c7e38d72"
+dependencies = [
+ "cfg-if",
+ "getrandom",
+ "once_cell",
+ "serde",
+ "version_check",
+]
+
+[[package]]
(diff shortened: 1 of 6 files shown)


svix-dylan (approved):

svix-daniel (approved):

svix-dylan:
Almost forgot, we are supposed to provide some information about open source libraries when we add them to deps. [Here are some examples](https://github.com/svix/monorepo-private/pulls?q=is%3Apr+label%3A%22open+source+review%22+is%3Aclosed).

svix-gabriel:
Thanks @svix-dylan, I updated the PR message

PR #679, opened 2022-10-26
Add queue clearing logic to the Redis migrations

svix-daniel:
As a followup to #671, this change adds logic like the Redis migration functions which clears the queue of acknowledged, stale tasks up to the lowest pending ID. This is run every thirty seconds until there has been a successful trimming of tasks.

The following is a brief explanation of how Redis streams work:

If you `XREADGROUP` it'll be added to the PEL (pending entries list) and thus show up in `XPENDING` until it has been `XACK`ed. But it'll never show up in `XREADGROUP` with that specific consumer group again.

Once `XACK` ed it removes the reference in the PEL, and will show up in neither `XPENDING` or `XREADGROUP`.

It will however remain in the stream and show up via `XREAD` regardless of whether it is `ACK`ed, but we don't do this anywhere. We instead use `XAUTOCLAIM` on items that have been read but not `XACK`ed for more than 45 seconds and reinsert them into the queue.

While it doesn't show up in `XREADGROUP`, it's still there consuming memory until the item has been `XDEL`ed from the stream.

The bug was that we would `XACK` tasks, but not `XDEL` them at any point up until the changes of #671. However, #671 only fixes tasks that are acknowledged after updating the server instance. This PR is meant to clear stale queue tasks that were acknowledged before updating.

The approach here is to use `XPENDING` to get the lowest ID that has been read but not ACKed (knowing the server reads from the stream by order of entry), and to `XTRIM` up to that point exclusive. Should there be no items in the `XPENDING` return, then the logic will loop with a 30 second delay until there are entries in the PEL at the time of reading. This is an imperfect solution, but it is still likely that the PEL will have at least one entry within a reasonable amount of time given a high-load queue.

Given there isn't high load and `XPENDING` never returns a non-empty array, it is assumed that this bug has not consumed enough memory to make clearing the queue automatically a great necessity.

### DIFF
--- a/server/svix-server/src/queue/redis.rs
+++ b/server/svix-server/src/queue/redis.rs
@@ -33,7 +33,7 @@ use axum::async_trait;
 
 use chrono::Utc;
 use redis::{
-    streams::{StreamClaimReply, StreamId, StreamReadOptions, StreamReadReply},
+    streams::{StreamClaimReply, StreamId, StreamPendingReply, StreamReadOptions, StreamReadReply},
     Cmd, FromRedisValue, RedisResult, RedisWrite, ToRedisArgs,
 };
 use tokio::time::sleep;
@@ -343,9 +343,22 @@ async fn new_pair_inner(
     let mqn = main_queue_name.clone();
     let dqn = delayed_queue_name.clone();
 
+    // Clear stale [`QueueTask`]s once per startup (see #679 for context)
+    tokio::spawn({
+        let pool = pool.clone();
+        let main_queue_name = main_queue_name.clone();
+
+        async move {
+            if let Err(e) = clear_acked(pool, &main_queue_name).await {
+                tracing::error!("Error clearing ACKed queue items: {}", e);
+            }
+        }
+    });
+
     // Migrate v1 queues to v2 and v2 queues to v3 on a loop with exponential backoff.
     tokio::spawn({
         let pool = pool.clone();
+
         async move {
             let delays = [
                 // 11.25 min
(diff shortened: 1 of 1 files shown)


svix-james (dismissed):

server/svix-server/src/queue/redis.rs
```diff
@@ -588,6 +594,31 @@ impl TaskQueueReceive for RedisQueueConsumer {
     }
 }
 
+async fn clear_acked(pool: RedisPool, queue: &str) -> Result<()> {
+    let mut conn = ctx!(pool.get().await)?;
+
+    let xpending_resp: StreamPendingReply =
+        ctx!(conn.query_async(Cmd::xpending(queue, WORKERS_GROUP)).await)?;
```
tasn, line 601: You want to add a count (limit) to the number your return, you only want 1 anyway...

server/svix-server/src/queue/redis.rs
```diff
@@ -588,6 +594,31 @@ impl TaskQueueReceive for RedisQueueConsumer {
...
+
+    match xpending_resp {
+        StreamPendingReply::Empty => Ok(()),
+        StreamPendingReply::Data(resp) => {
+            // Get the smallest unACKed task that's been read, and know that any task with an ID
+            // less than that has already been ACKed
+            let min = resp.start_id;
```
tasn, line 608: I'd rather add a safety mechanism where this item ID will have to be at least one week old too, so we really don't delete anything recent.

tasn: Maybe there's no need actually, I guess that XPENDING by definition gets us the oldest one.

server/svix-server/src/queue/redis.rs
```diff
@@ -33,7 +33,7 @@ use axum::async_trait;
 
 use chrono::Utc;
```
tasn, line 34: I'm still unclear about the relationship of all of these things.

For example, if I:
1. Add a message.
2. XREADGROUP it
3. XACX it

Will it show in XREADGROUP again? Will it show in XPENDING? Will it show in neither?

I get the idea of using XPENDING, it sounds clever and correct, but can you maybe document all of this in the PR description too (and thus in the commit message)?

tasn (approved):

PR #689, opened 2022-11-04
Create subcommand for wiping an organization's data

svix-daniel:
These changes make it possible to completely wipe the PostgreSQL data related to a given organization.

It is invoked like `svix-server  wipe --yes-i-know-what-im-doing org_<id here>`.

While this has been tested to work, the test is ignored such as to reduce the time it takes for CI to run the test suite.

NOTE: Wiping organizations may take a while.

### DIFF
--- a/server/svix-server/src/db/mod.rs
+++ b/server/svix-server/src/db/mod.rs
@@ -1,12 +1,16 @@
 // SPDX-FileCopyrightText: © 2022 Svix Authors
 // SPDX-License-Identifier: MIT
 
-use sea_orm::{DatabaseConnection, DbBackend, SqlxPostgresConnector};
+use sea_orm::{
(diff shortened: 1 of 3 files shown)


server/svix-server/src/db/mod.rs
```diff
@@ -31,3 +35,102 @@ pub async fn run_migrations(cfg: &Configuration) {
     let db = connect(cfg).await;
     MIGRATIONS.run(&db).await.unwrap();
 }
+
+/// Wipe an organization from existence in a way that ensures the operation can be tried again on
+/// failure.
+pub async fn wipe_org(cfg: &Configuration, org_id: OrganizationId) {
+    let db = init_db(cfg).await;
```
tasn, line 42: Use a transaction and commit periodically.

The problem you have is that every seaorm call is making a new db connection (or trying from the pool) and initing it, which is slow.

tasn (approved):

PR #692, opened 2022-11-09
Add tracing to run-tests.sh

svix-gabriel:
This moves the tracing setup outside of `main.rs` so that tests can use tracing. This emits server logs in failed test runs, which makes diagnosing failed tests much easier, depending on the reason for failing.

There is a small functional change - instead of using `SubscriberInitExt::init`, we now use `SubscriberInitExt::try_init`. init will panic if the subscriber was already setup, whereas try_init will return an `Err(...)`. This doesn't matter in prod, but in tests we end up calling setup_tracing more than once. AFAICT we can safely ignore this error, so we do.

### DIFF
--- a/server/svix-server/src/lib.rs
+++ b/server/svix-server/src/lib.rs
@@ -6,7 +6,10 @@
 
 use axum::{extract::Extension, Router};
 
+use cfg::ConfigurationInner;
 use lazy_static::lazy_static;
+use opentelemetry::runtime::Tokio;
+use opentelemetry_otlp::WithExportConfig;
 use std::{
     net::TcpListener,
     sync::atomic::{AtomicBool, Ordering},
@@ -15,6 +18,7 @@ use std::{
 use tower::ServiceBuilder;
 use tower_http::cors::{AllowHeaders, Any, CorsLayer};
 use tower_http::trace::TraceLayer;
+use tracing_subscriber::{prelude::*, util::SubscriberInitExt};
 
 use crate::{
     cfg::{CacheBackend, Configuration},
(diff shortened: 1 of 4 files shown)


server/svix-server/src/lib.rs
```diff
@@ -15,6 +18,7 @@ use std::{
 use tower::ServiceBuilder;
 use tower_http::cors::{AllowHeaders, Any, CorsLayer};
 use tower_http::trace::TraceLayer;
+use tracing_subscriber::{prelude::__tracing_subscriber_SubscriberExt, util::SubscriberInitExt};
```
tasn, line 21:
```suggestion
use tracing_subscriber::{prelude::*, util::SubscriberInitExt};
```
I think importing this explicitly just looks weird, but anyhow, it wasn't imported before, is it even needed?
Also, are you even sure

server/svix-server/src/lib.rs
```diff
@@ -15,6 +18,7 @@ use std::{
 use tower::ServiceBuilder;
 use tower_http::cors::{AllowHeaders, Any, CorsLayer};
 use tower_http::trace::TraceLayer;
+use tracing_subscriber::{prelude::*, util::SubscriberInitExt};
```
tasn, line 21: Why is this prelude even needed now?

svix-gabriel: We still need `__tracing_subscriber_SubscriberExt`, for
```rust
tracing_subscriber::Registry::default()
                .with(...)
```
which comes from the prelude

svix-dylan (approved):

PR #701, opened 2022-11-14
Bump sea_orm to 10.2 and sqlx to 6.2

svix-gabriel:
This bumps sea_orm and sqlx to 10.2 and 6.2, respectively. In addition, we clean up some macro hygiene with related type macros that we use for deriving a lot of sea_orm relevant traits.

## Motivation

Aside from the general benefits of upgrading, our current version of sea_orm doesn't support `ON CONFLICT ...` operations, which we need for some upcoming features. So we need to upgrade sea_orm first.

The `DbErr` in sea_orm also exposes the underlying `sqlx::Error` type, so we had to bump the sqlx version as well.

Finally, the macros in `types.rs` weren't fully qualifying their dependencies, which made them difficult to use in the rest of the crate. Since I had to fix a bunch of compile time errors in `types.rs`, I went ahead and fixed the macro hygiene as well.

## Solution

* Bumped the versions
* Ran `cargo update`
* Fixed compile errors
* Ran tests
* Profit!!!

### DIFF
--- a/server/svix-server/src/core/types.rs
+++ b/server/svix-server/src/core/types.rs
@@ -9,11 +9,7 @@ use num_enum::{IntoPrimitive, TryFromPrimitive};
 use rand::Rng;
 
 use regex::Regex;
-use sea_orm::{
-    entity::prelude::*,
-    sea_query::{ColumnType, Nullable, ValueType, ValueTypeErr},
-    TryFromU64, TryGetError, TryGetable,
-};
+
 use serde::{Deserialize, Serialize};
 use std::ops::Deref;
 use svix_ksuid::*;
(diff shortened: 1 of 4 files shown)


server/svix-server/src/core/types.rs
```diff
@@ -135,6 +143,10 @@ macro_rules! json_wrapper {
             fn column_type() -> sea_orm::sea_query::ColumnType {
                 sea_orm::sea_query::ColumnType::JsonBinary
             }
+
+            fn array_type() -> sea_orm::sea_query::ArrayType {
+                sea_orm::sea_query::ArrayType::Json
```
tasn, line 148: This doesn't look right. Should it be Json?

tasn (commented):
Looks good. Just need someone to review.

server/svix-server/src/core/types.rs
svix-gabriel: I think so - sea_orm itself uses `ArrayType::Json` for the json Value type - which made me think that's what our json wrapper type should use as well.

I might just be wildly misunderstanding what this method is meant to do though 🤷

tasn: Oh, I think then what you want is JsonBinary.

Essentially what it does (I think), is indicate how to tell PG to store this in an array. Since it's json we probably just want to use Json directly rather than a PG array of JSONs. Though use JsonBinary so that it matches the type we use for the column.

(unless there's no json binary, then json is fine)>

svix-gabriel: Yeah there's no JsonBinary for ArrayType, unfortunately
https://docs.rs/sea-query/latest/sea_query/value/enum.ArrayType.html

tasn: Weird, OK.

svix-daniel (approved):

PR #704, opened 2022-11-15
Server: Worker refactor

jaymell:
* Add webhook-http-client, which is based on hyper, enabling several
  critical improvements:
  * Default filtering of endpoints that resolve to private
    address space, with ability to add CIDR exceptions as needed
  * Case-sensitive HTTP1 headers -- although HTTP spec states that
    headers are not case-sensitive, in practice many widely
    deployed web-servers expect to support case-sensitive headers.
    This requires a fork of hyper, which currently has but does
    not expose support for case-sensitive headers.
  * Assurance that the `host` header appears first in outgoing HTTP requests.
    Although HTTP spec does not prescribe an order for headers, in
    our experience, many web-servers will fail if the `host` header does
    not appear first in request headers. We ensure this by manually
    adding the header ourselves rather than relying on hyper to do so.
* Use `openssl` instead of `rusttls` for outbound webhooks. In our experience,
  many widely deployed webservers use "weak" ciphers that rusttls refuses to
  support, leaving us with no choice but to rely on openssl libraries.
* Support for soft-limiting of concurrent worker tasks. Unbounded concurrent worker
  tasks can easily overwhelm a system. This sets a default of 500 tasks with support
  for configuring other limits or setting unlimited (`0`) concurrent tasks
* Refactor message dispatch code for clarity. Dispatch code is now broken into
  separate methods: `prepare_dispatch`, `make_http_call`,
  `handle_successful_dispatch`, and `handle_failed_dispatch` with the dispatch itself
  having different types depending on its stage in the dispatch process.
  Other function names have also been changed for improved clarity.
* Prevent worker from shutting down when there are active tasks.
* Avoid multiple DB calls for message-destination insert/retrieval

### DIFF
--- a/server/svix-server/src/core/webhook_http_client.rs
+++ b/server/svix-server/src/core/webhook_http_client.rs
@@ -1,0 +1,764 @@
+use std::{
+    collections::HashMap,
+    future::Future,
+    net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr},
+    pin::Pin,
+    str::FromStr,
+    sync::Arc,
+    task::Poll,
+    time::{Duration, Instant},
+};
+
+use axum::headers::{authorization::Credentials, Authorization};
+use bytes::Bytes;
+use futures::{future::BoxFuture, FutureExt};
+use http::header::HeaderName;
+use http::{HeaderMap, HeaderValue, Method, Response, StatusCode, Version};
+use hyper::{
+    client::connect::{dns::Name, HttpConnector},
+    ext::HeaderCaseMap,
+    Body, Client, Uri,
+};
+use hyper_openssl::HttpsConnector;
+use ipnet::IpNet;
+use openssl::ssl::{SslConnector, SslMethod};
+use serde::Serialize;
+use thiserror::Error;
+use tokio::sync::Mutex;
+use tower::Service;
+use trust_dns_resolver::{
+    error::ResolveError, lookup_ip::LookupIpIntoIter, AsyncResolver, TokioConnection,
+    TokioConnectionProvider, TokioHandle,
+};
+
+pub type CaseSensitiveHeaderMap = HashMap<String, HeaderValue>;
+
+#[derive(Debug, Error)]
+pub enum Error {
+    #[error("failure response: {0}")]
+    FailureStatus(StatusCode),
+
+    #[error("making requests to local IP addresses is forbidden and blocked")]
+    BlockedIp,
+    #[error("error resolving name: {0}")]
+    ResolveError(#[from] ResolveError),
+
+    #[error("request timed out")]
+    TimedOut,
+
+    #[error("error forming request: {0}")]
+    InvalidHttpRequest(http::Error),
+    #[error("error making request: {0}")]
+    FailedRequest(hyper::Error),
+}
+
+pub struct WebhookClient {
+    client: Client<HttpsConnector<NonLocalConnector>, Body>,
+    whitelist_nets: Arc<Vec<IpNet>>,
+}
+
+impl WebhookClient {
+    pub fn new(
+        whitelist_nets: Option<Arc<Vec<IpNet>>>,
+        whitelist_names: Option<Arc<Vec<String>>>,
+    ) -> Self {
+        let whitelist_nets = whitelist_nets.unwrap_or_else(|| Arc::new(Vec::new()));
+        let whitelist_names = whitelist_names.unwrap_or_else(|| Arc::new(Vec::new()));
+
+        let mut connector = HttpConnector::new_with_resolver(NonLocalDnsResolver::new(
+            whitelist_nets.clone(),
+            whitelist_names,
+        ));
+        connector.enforce_http(false);
+
+        let ssl = SslConnector::builder(SslMethod::tls()).expect("SslConnector build failed");
+        let https = HttpsConnector::with_connector(NonLocalConnector { connector }, ssl)
+            .expect("HttpsConnector build failed");
+
+        let client: Client<_, hyper::Body> = Client::builder()
+            .http1_ignore_invalid_headers_in_responses(true)
+            .http1_title_case_headers(true)
+            .build(https);
+
+        Self {
+            client,
+            whitelist_nets,
+        }
+    }
+
+    pub async fn execute(&self, request: Request) -> Result<Response<Body>, Error> {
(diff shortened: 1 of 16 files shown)


server/svix-server/src/worker.rs
```diff
@@ -76,7 +76,7 @@ impl FailureCacheKey {
 ///
 /// If the key value pair does not already exist in the cache, indicating that the endpoint never
 /// stopped responding, no operation is performed.
-async fn process_success_cache(
+async fn process_endpoint_success(
     cache: &Cache,
     app_id: &ApplicationId,
     endp_id: &EndpointId,
```
tasn, line 82: Since you're refactoring, I'd probably also pass `org_id` here and fix the cache key for failure to include the org_id. I know app_id is globally unique, but having good segmentation between different orgs is probably wanted.

server/svix-server/src/worker.rs
```diff
@@ -99,12 +99,10 @@ async fn process_success_cache(
 ///
 /// All cache values are set with an expiration time greater thah the grace period, so occasional
 /// failures will not cause an endpoint to be disabled.
-async fn process_failure_cache(
+async fn process_endpoint_failure(
     cache: &Cache,
```
tasn, line 103: Same here about org_id

server/svix-server/src/worker.rs
jaymell: Should we have a migration here to accommodate prior keys?

server/svix-server/tests/worker.rs
```diff
@@ -124,6 +124,7 @@ async fn test_no_redirects_policy() {
 /// tests don't take too long, thes grace period and expiration period will be reconfigured to be
 /// on the order of seconds
 #[tokio::test]
+#[ignore = "reason"]
```
tasn, line 127: ?

jaymell: Removed.

server/svix-server/src/worker.rs
```diff
@@ -62,11 +75,40 @@ pub struct FailureCacheValue {
...
+    }
+}
+
+impl std::fmt::Display for FailureCacheValue {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.first_failure_at)
+    }
+}
+
+impl crate::core::cache::CacheValue for FailureCacheValue {
```
jaymell, line 93: I don't know why the macro is not successfully handling this. Need to fix. :neutral_face:

server/svix-server/src/error.rs
```diff
@@ -422,3 +434,18 @@ impl IntoResponse for HttpError {
...
     }
 }
+
+impl From<ErrorType> for Error {
+    fn from(typ: ErrorType) -> Self {
+        Self { trace: vec![], typ }
+    }
+}
+
+impl From<crate::core::webhook_http_client::Error> for Error {
```
tasn, line 444: Not sure what this is for. More of a bookmark for myself than anything.

jaymell: b/c the webhook-http-client has its own Error type

server/svix-server/src/core/webhook_http_client.rs
```diff
@@ -0,0 +1,764 @@
...
+        let whitelist_nets = whitelist_nets.unwrap_or_else(|| Arc::new(Vec::new()));
+        let whitelist_names = whitelist_names.unwrap_or_else(|| Arc::new(Vec::new()));
+
+        let mut connector = HttpConnector::new_with_resolver(NonLocalDnsResolver::new(
+            whitelist_nets.clone(),
+            whitelist_names,
+        ));
+        connector.enforce_http(false);
+
+        let ssl = SslConnector::builder(SslMethod::tls()).expect("SslConnector build failed");
```
jaymell, line 74: Remove panic?

server/svix-server/src/core/webhook_http_client.rs
```diff
@@ -0,0 +1,764 @@
...
+        let https = HttpsConnector::with_connector(NonLocalConnector { connector }, ssl)
+            .expect("HttpsConnector build failed");
```
jaymell, line 76: Remove panic?

server/svix-server/src/core/webhook_http_client.rs
```diff
@@ -0,0 +1,764 @@
...
+        }
+    }
+
+    pub fn build(self) -> Result<Request, RequestBuildError> {
+        self.validate()?;
+
+        let custom_headers = self.headers.unwrap_or_default();
+
+        let uri = self.uri.unwrap();
+        let authority = uri.authority().expect("Missing authority");
```
jaymell, line 374: Remove panic?

server/svix-server/src/core/webhook_http_client.rs
```diff
@@ -0,0 +1,764 @@
...
+
+        if let Some(user_agent) = self.user_agent {
+            headers.insert(http::header::USER_AGENT, user_agent);
+        }
+
+        if let Some(auth_header) = self.basic_auth {
+            if !headers.contains_key(http::header::AUTHORIZATION) {
+                headers.insert(
+                    http::header::AUTHORIZATION,
+                    HeaderValue::from_bytes(&auth_header).unwrap(),
```
jaymell, line 405: Remove panic?

server/svix-server/src/core/webhook_http_client.rs
```diff
@@ -0,0 +1,764 @@
...
+        }
+
+        Ok(Request {
+            method: self.method.unwrap_or(Method::POST),
+            uri,
+            headers,
+            header_names: self.header_names,
+            body: self.body,
+            timeout: self.timeout,
+            version: self.version.unwrap(),
```
jaymell, line 417: Remove panic?

server/svix-server/src/worker.rs
```diff
@@ -174,148 +197,161 @@ fn generate_msg_headers(
...
+        headers.insert("svix-signature".to_owned(), signatures_str);
     }
-
+    headers.insert(
+        "user-agent".to_owned(),
+        USER_AGENT.to_string().parse().unwrap(),
+    );
+    headers.insert(
+        "content-type".to_owned(),
+        "application/json".parse().unwrap(),
```
jaymell, line 228: Remove panic?

tasn (approved):

PR #705, opened 2022-11-16
Adding metadata to application CRUD group

svix-gabriel:
This adds the `metadata` field to the application CRUD group.

## Motivation

The `metadata` field is just a JSON dictionary with string only values. This field was added to our public API, but never to our open source solution. This PR corrects that.

## Solution

We add the field to our `/apps` CRUD group. `Metadata` itself is a separate joined table, instead of a column on the apps field, because metadata can be quite large (up to 4096 bytes).

Because it's a separate model, the sea_orm queries end up being pretty tedious. I tried to keep things efficient and query for both records in a single round trip where possible, but in some places we needed two roundtrips.

A side effect is that the endpoints were starting to get fairly complicated, so I tried to offload a lot of the logic to the relevant `db::models` where appropriate to offset this.

Finally, the `metadata` field is entirely optional, but we default to rendering `{}` if it's unspecified. Additionally, we don't create empty `metadata` records. It ended up being _much_ simpler to just pass around an `applicationmetadata::[Model|ActiveModel]` around, and intelligently upsert only IFF the underlying data wasn't empty.

### DIFF
--- a/server/svix-server/src/core/permissions.rs
+++ b/server/svix-server/src/core/permissions.rs
@@ -7,17 +7,28 @@ use sea_orm::DatabaseConnection;
 
 use crate::{
     ctx,
-    db::models::application,
+    db::models::{application, applicationmetadata},
     error::{Error, HttpError, Result},
 };
 
 use super::{
-    security::{permissions_from_bearer, AccessLevel},
-    types::{ApplicationIdOrUid, OrganizationId},
+    security::{permissions_from_bearer, AccessLevel, Permissions},
+    types::{ApplicationId, ApplicationIdOrUid, OrganizationId},
 };
 
 pub struct Organization {
     pub org_id: OrganizationId,
+}
+
+impl Permissions {
+    fn check_app_is_permitted(&self, app_id: &ApplicationId) -> Result<()> {
+        if let Some(ref permitted_app_id) = self.app_id() {
+            if permitted_app_id != app_id {
+                return Err(HttpError::not_found(None, None).into());
+            }
+        }
+        Ok(())
+    }
 }
 
 #[async_trait]
@@ -63,11 +74,7 @@ where
         )?
         .ok_or_else(|| HttpError::not_found(None, None))?;
 
-        if let Some(permitted_app_id) = permissions.app_id() {
-            if permitted_app_id != app.id {
-                return Err(HttpError::not_found(None, None).into());
-            }
-        }
+        permissions.check_app_is_permitted(&app.id)?;
 
         Ok(Self { app })
     }
@@ -98,6 +105,36 @@ where
         )?
         .ok_or_else(|| HttpError::not_found(None, None))?;
         Ok(OrganizationWithApplication { app })
+    }
+}
+
+pub struct ApplicationWithMetadata {
+    pub app: application::Model,
+    pub metadata: applicationmetadata::Model,
+}
+
+#[async_trait]
+impl<B> FromRequest<B> for ApplicationWithMetadata
+where
+    B: Send,
+{
+    type Rejection = Error;
+
+    async fn from_request(req: &mut RequestParts<B>) -> Result<Self> {
+        let permissions = permissions_from_bearer(req).await?;
+
+        let Path(ApplicationPathParams { app_id }) =
+            ctx!(Path::<ApplicationPathParams>::from_request(req).await)?;
+        let Extension(ref db) = ctx!(Extension::<DatabaseConnection>::from_request(req).await)?;
+        let (app, metadata) = ctx!(
+            application::Model::fetch_with_metadata(db, permissions.org_id(), app_id.to_owned())
+                .await
+        )?
+        .ok_or_else(|| HttpError::not_found(None, None))?;
+
+        permissions.check_app_is_permitted(&app.id)?;
+
+        Ok(Self { app, metadata })
     }
 }
 
--- a/server/svix-server/src/core/types/metadata.rs
+++ b/server/svix-server/src/core/types/metadata.rs
@@ -1,0 +1,39 @@
+use std::collections::HashMap;
+
+use crate::json_wrapper;
+use serde::{Deserialize, Serialize};
+
+pub const MAX_METADATA_SIZE: usize = 4096;
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Default)]
+pub struct Metadata(HashMap<String, String>);
+
+json_wrapper!(Metadata);
+
+impl Metadata {
+    pub fn is_empty(&self) -> bool {
+        self.0.is_empty()
+    }
+}
+
+impl<'de> Deserialize<'de> for Metadata {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let inner: Option<HashMap<String, String>> = Deserialize::deserialize(deserializer)?;
+        let metadata = inner.unwrap_or_default(); // coerce `null` to `{}`
+
+        let size = serde_json::to_string(&metadata)
+            .map(|blob| blob.len())
+            .map_err(|_| serde::de::Error::custom("metadata is not valid json"))?;
+
+        if size > MAX_METADATA_SIZE {
+            return Err(serde::de::Error::custom(format!(
+                "metadata must be less than or equal to {MAX_METADATA_SIZE} bytes"
+            )));
+        }
+
+        Ok(Self(metadata))
+    }
+}
--- a/server/svix-server/src/db/models/application.rs
+++ b/server/svix-server/src/db/models/application.rs
@@ -4,9 +4,13 @@
 use crate::core::types::{
     ApplicationId, ApplicationIdOrUid, ApplicationUid, BaseId, OrganizationId,
 };
+use crate::{ctx, error};
 use chrono::Utc;
 use sea_orm::ActiveValue::Set;
 use sea_orm::{entity::prelude::*, Condition};
+use sea_orm::{ConnectionTrait, QueryOrder, QuerySelect};
+
+use super::applicationmetadata;
 
 #[derive(Clone, Debug, PartialEq, Eq, DeriveEntityModel)]
 #[sea_orm(table_name = "application")]
@@ -28,6 +32,8 @@ pub enum Relation {
     Endpoint,
     #[sea_orm(has_many = "super::message::Entity")]
     Message,
+    #[sea_orm(has_one = "super::applicationmetadata::Entity")]
+    Metadata,
 }
 
 impl Related<super::endpoint::Entity> for Entity {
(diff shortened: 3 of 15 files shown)


server/svix-server/src/core/types/metadata.rs
```diff
@@ -0,0 +1,39 @@
...
+
+impl<'de> Deserialize<'de> for Metadata {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let inner: Option<HashMap<String, String>> = Deserialize::deserialize(deserializer)?;
+        let metadata = inner.unwrap_or_default(); // coerce `null` to `{}`
+
+        let size = serde_json::to_string(&metadata)
```
tasn, line 27: Slightly inefficient, probably doesn't matter at the moment, though let's at least put a FIXME here so it sticks out.

server/svix-server/src/v1/endpoints/application.rs
```diff
@@ -84,24 +89,30 @@ pub struct ApplicationPatch {
...
+    #[serde(default, skip_serializing_if = "UnrequiredField::is_absent")]
+    pub metadata: UnrequiredField<Metadata>,
 }
 
 impl ModelIn for ApplicationPatch {
-    type ActiveModel = application::ActiveModel;
+    type ActiveModel = (application::ActiveModel, applicationmetadata::ActiveModel);
 
-    fn update_model(self, model: &mut Self::ActiveModel) {
+    fn update_model(self, (app, app_metadata): &mut Self::ActiveModel) {
```
tasn, line 100: This syntax looks a bit weird, does it even work? What's the type of app? Metadata?

tasn: Oh, I just saw it, it's very confusing.

tasn: I feel like it just needs to be split, there's no real value merging them both into this `update_model`, as metadata is really just a one field, no?

server/svix-server/src/core/permissions.rs
```diff
@@ -101,6 +101,44 @@ where
...
+            application::Model::get_with_application_metadata(
+                db,
+                permissions.org_id(),
+                app_id.to_owned()
+            )
+            .await
+        )?
+        .ok_or_else(|| HttpError::not_found(None, None))?;
+
+        if let Some(permitted_app_id) = permissions.app_id() {
```
tasn, line 132: What's scary to me is that this logic is now duplicated, so it's another place to update this logic once we change something.

server/svix-server/src/db/models/application.rs
```diff
@@ -42,6 +49,88 @@ impl Related<super::message::Entity> for Entity {
...
 }
 
+impl Related<super::applicationmetadata::Entity> for Entity {
+    fn to() -> RelationDef {
+        Relation::Metadata.def()
+    }
+}
+
+impl Model {
+    pub async fn get_application_metadata(
```
tasn, line 59: The convention that we follow is `fetch_`, not `get_`. I think at least, no?

server/svix-server/src/db/models/applicationmetadata.rs
```diff
@@ -0,0 +1,111 @@
...
+            id,
+            data,
+            ..Self::new()
+        }
+    }
+
+    /// Upserts the record if it's new or updated, AND data is nonempty. Otherwise the record is
+    /// ignored or destroyed as appropriate.
+    pub async fn upsert_or_delete(self, db: &impl ConnectionTrait) -> error::Result<Metadata> {
+        // No linked app, so nothing worth saving
```
tasn, line 77: This sounds like an error case, not an ignorable case. When would this happen under normal circumstances?

server/svix-server/src/db/models/applicationmetadata.rs
```diff
@@ -0,0 +1,111 @@
...
+        if self.id.is_not_set() {
+            return Ok(Metadata::default());
+        }
+
+        let exists_in_db = self.id.is_unchanged();
```
tasn, line 82: This feels very implicit and error-prone. It's also trying to regenerate context here implicitly, from something that would be known by the caller.

The caller can just delete if empty and exists, and can upsert otherwise. This function doesn't add much, but it adds potential failure cases/places.

server/svix-server/src/db/models/application.rs
```diff
@@ -60,13 +149,56 @@ impl ActiveModelBehavior for ActiveModel {
...
+}
+
 impl Entity {
     pub fn secure_find(org_id: OrganizationId) -> Select<Entity> {
         Self::find()
             .filter(Column::OrgId.eq(org_id))
             .filter(Column::Deleted.eq(false))
     }
 
+    pub fn secure_find_many(
```
tasn, line 186: Secure find already does many, so the name is confusing, though I don't fully understand the reason for it to exist. It's logic that should only exist in one place (list apps), so why abstract it?

server/svix-server/src/v1/endpoints/application.rs
```diff
@@ -48,26 +48,31 @@ pub struct ApplicationIn {
...
     pub uid: Option<ApplicationUid>,
+
+    #[serde(default)]
+    pub metadata: Metadata,
 }
 
 // FIXME: This can and should be a derive macro
 impl ModelIn for ApplicationIn {
-    type ActiveModel = application::ActiveModel;
+    type ActiveModel = (application::ActiveModel, applicationmetadata::ActiveModel);
```
tasn, line 58: I think that why biggest grip with this whole change is that it tries to take unrelated items that we update and fetch (app and metadata), and force-treat them as if they are one model. They are very related, and tied together, they could have maybe been a property on the model, but they aren't, and trying to force it feels off and causes a lot of confusing code.

server/svix-server/src/core/types/metadata.rs
svix-gabriel: Yeah, the only other approach I could think was manually iterating through the metadata and calculating it's size by hand - but that feels clunky and error prone.

I'm really skeptical this will have any noticeable impact on performance so I kept this as is.

server/svix-server/src/v1/endpoints/application.rs
svix-gabriel: I ended up keeping this as is. The models are closely related, at least as far as the API is concerned. I don't really agree that this is confusing (it's quite explicit), and makes it obvious that the two models need to be updated together.

server/svix-server/src/db/models/application.rs
```diff
@@ -42,6 +49,88 @@ impl Related<super::message::Entity> for Entity {
...
+        let metadata = ctx!(query.one(db).await)?
+            .map(applicationmetadata::ActiveModel::from)
+            .unwrap_or_else(|| {
+                applicationmetadata::ActiveModel::new_for_app(self.id.clone(), None)
+            });
+
+        Ok(metadata)
+    }
+
+    pub async fn get_many_with_metadata(
```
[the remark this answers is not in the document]
svix-gabriel: Updated some, but I kept a lot of these functions in. They make the endpoints much easier to grok - and they capture very straightforward behavior.

I don't agree that they're complex - but if you _do_ think they're complex, then that's exactly the reason to keep them as separate functions. The models are DB abstractions, the whole point is to offload complexity out of the endpoints and into the layer better suited to handle them.

server/svix-server/src/db/models/applicationmetadata.rs
svix-gabriel: Partially updated - although I kept the function in. We call this in multiple places, so I strongly disagree that it doesn't add much value.

server/svix-server/src/db/models/applicationmetadata.rs
svix-gabriel: Updated - see comment below

server/svix-server/src/db/models/application.rs
```diff
@@ -42,21 +49,112 @@ impl Related<super::message::Entity> for Entity {
...
 }
 
+impl Related<super::applicationmetadata::Entity> for Entity {
+    fn to() -> RelationDef {
+        Relation::Metadata.def()
+    }
+}
+
+impl Model {
+    pub async fn fetch_application_metadata(
```
tasn, line 59:
```suggestion
    pub async fn fetch_or_create_metadata(
```

server/svix-server/src/db/models/application.rs
```diff
@@ -42,21 +49,112 @@ impl Related<super::message::Entity> for Entity {
...
+                .await
+        )?;
+
+        Ok(results.into_iter().map(|(app, metadata)| {
+            let metadata = metadata.map(|m| m.metadata()).unwrap_or_default();
+            (app, metadata)
+        }))
+    }
+
+    pub async fn fetch_with_metadata(
```
tasn, line 98: I can't get over this function and the next one. They are similarly named, no idea what one does vs the other, etc.

I'd just drop this one, it's so confusing.

server/svix-server/src/db/models/application.rs
```diff
@@ -42,21 +49,112 @@ impl Related<super::message::Entity> for Entity {
...
             updated_at: Set(timestamp.into()),
             deleted: Set(false),
             ..ActiveModelTrait::default()
         }
     }
 
-    fn before_save(mut self, _insert: bool) -> Result<Self, DbErr> {
-        self.updated_at = Set(Utc::now().into());
-        Ok(self)
+    pub async fn fetch_with_application_metadata(
```
tasn, line 143: Another fetch_with_application_metadata?

server/svix-server/src/db/models/applicationmetadata.rs
```diff
@@ -0,0 +1,95 @@
...
+        }
+    }
+
+    /// Upserts the record if it's new or updated, AND data is nonempty. Otherwise the record is
+    /// ignored or destroyed as appropriate.
+    pub async fn upsert_or_delete(self, db: &impl ConnectionTrait) -> error::Result<Metadata> {
+        let data = self.data.clone().take().unwrap_or_default();
+
+        // can't delete the record if there is no ID
+        if data.is_empty() && !self.id.is_not_set() {
```
tasn, line 73: OK, I now see what was bothering me above, and this place shows it.

We can't on one hand say `create` requires an id, and on the other, have Metadata objects lying around that don't have `id` set.

Remove the `is_not_set`, and just assume it's always set. How to assume that? Remove the default implementation, and instead always create with a specific `id` when you create a fake one.

Though honestly, not sure about the value of creating a fake one always, it's fine to return `None` and handle it outside.

server/svix-server/src/db/models/application.rs
```diff
@@ -42,21 +49,112 @@ impl Related<super::message::Entity> for Entity {
...
+            Entity::secure_find_by_id_or_uid(org_id, id_or_uid)
+                .find_also_related(applicationmetadata::Entity)
+                .one(db)
+                .await
+        )
+    }
+}
+
 impl ActiveModelBehavior for ActiveModel {
-    fn new() -> Self {
```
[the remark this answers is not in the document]
svix-gabriel: It's implemented by default, I opted to remove because we don't use it anywhere and I wanted `fn create` to be the defactor / obvious way to initialize a new app.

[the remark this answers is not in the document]
tasn: Yeah, but it's implemented by default without the `created_at` and the things that we want, no? That's why we've used it elsewhere. Maybe let's have it here for good measure, and also use it in `create`? Just to make sure we don't use it by accident.

server/svix-server/src/db/models/applicationmetadata.rs
svix-gabriel: I originally tried passing around `Option<Metadata>` / `Option<applicationmetadata::Model>` everywhere. It's fine from a correctness standpoint, but it _really_ adds a lot of cruft/boilerplate and conditional logic in the endpoints that are a huge PITA to deal with - and it makes the control flow a nightmare because you have to constantly worry about None values.

Just having a default `{}`, and then _one_ function that intelligently upserts is much simpler.

In practice `!self.id.is_not_set()` is probably never true - so I'll go ahead and removed the check.

server/svix-server/src/db/models/application.rs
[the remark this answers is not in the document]
svix-gabriel: It's noise, and wasted code.

Whether or not we override `::new()`, inserting an `application::ActiveModel::new()` record will _always_ encounter a runtime error because it's missing some required fields (like an org_id).

In other words, overriding ::new() does nothing to protect us against runtime errors. But keeping the function defined just encourages folks to use it over `create(...)`. Why bother with the extra code / complexity?

server/svix-server/src/db/models/applicationmetadata.rs
svix-gabriel: That is how it is currently written. I didn't override `ActiveModel::new()` - `ActiveModel::create(...)` ensures the ID is set.

We just have an optimization in the upsert where we don't insert empty data records. That's all this is, there's no extra risk of a runtime error than what was there before.

server/svix-server/src/db/models/application.rs
svix-gabriel: Updated, I think I was able to simplify things a bit

server/svix-server/src/db/models/application.rs
svix-gabriel: Updated

tasn (approved):

PR #728, opened 2022-12-17
Change Pg connection to be lazy

svix-gabriel:
This changes the connection pool to postgres to open connections lazily.

## Motivation

This might improve application startup a bit - both in speed and in reducing the likelihood of crashes on start up.

## Solution

Use the [connect_lazy](https://docs.rs/sqlx/latest/sqlx/struct.Pool.html#method.connect_lazy) function instead of `connect`, and remove now-unnecessary async/await calls.

### DIFF
--- a/server/svix-server/src/db/mod.rs
+++ b/server/svix-server/src/db/mod.rs
@@ -14,32 +14,31 @@ use models::{application, endpoint, eventtype, message, messageattempt, messaged
 
 static MIGRATIONS: sqlx::migrate::Migrator = sqlx::migrate!();
 
-async fn connect(dsn: &str, max_pool_size: u16) -> sqlx::Pool<sqlx::Postgres> {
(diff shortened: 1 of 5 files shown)


tasn:
I think we do the same in redis too, connect on start?

But anyhow, I think there's value in actually checking connection at start. Maybe we should change it to be more intentional, but if we don't do it we don't actually ever know it can connect to PG. This can mean that a new service can become healthy when it's in fact not healthy.

The health check won't catch it, as we don't actually check we can connect to PG in the health check ping (which is what should be used) because it would mean that PG goes down would fail health checks for the server which will in turn bring the whole service down when the issue is just PG.

tasn:
Btw, we already have `--wait-for` to reduce the likelihood of crashes on startup because it verifies it can connect to the TCP socket (not actually negotiate with PG though).

PR #731, opened 2022-12-20
Simplify Endpoint CRUD

svix-gabriel:
## Motivation

While adding metadata to the endpoint CRUD group, I noticed some repetitive and boilerplatey logic that made the endpoints harder to read and extend.

## Solution

This DRYs up some of the `EndpointEvent` and secret generation logic. As a little bonus, the endpoint secret generation code should be a bit more performant since we're no longer cloning the entire input or secret.

### DIFF
--- a/server/svix-server/src/v1/endpoints/endpoint/mod.rs
+++ b/server/svix-server/src/v1/endpoints/endpoint/mod.rs
@@ -6,16 +6,18 @@ mod recovery;
 mod secrets;
 
 use crate::{
+    cfg::DefaultSignatureType,
     core::{
+        cryptography::Encryption,
         permissions,
         types::{
-            ApplicationIdOrUid, BaseId, EndpointId, EndpointIdOrUid, EndpointUid, EventChannelSet,
-            EventTypeNameSet, MessageEndpointId, MessageStatus,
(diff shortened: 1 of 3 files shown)


server/svix-server/src/v1/endpoints/endpoint/mod.rs
```diff
@@ -160,6 +163,19 @@ pub struct EndpointIn {
     pub key: Option<EndpointSecret>,
 }
 
+impl EndpointIn {
+    pub fn take_or_generate_key(
+        &mut self,
+        cfg: &Configuration,
```
tasn, line 169: I'd be explicit as for what it receives. Encryption and default signature type, rather than passing the whole config.

server/svix-server/src/v1/endpoints/endpoint/mod.rs
```diff
@@ -160,6 +163,19 @@ pub struct EndpointIn {
     pub key: Option<EndpointSecret>,
 }
 
+impl EndpointIn {
+    pub fn take_or_generate_key(
```
tasn, line 167:
```suggestion
    pub fn key_take_or_generate(
```
So that it autocompletes nicely in the editor when you type `key`

tasn (approved):

PR #732, opened 2022-12-20
Add Metadata field to Endpoint CRUD

svix-gabriel:
## Motivation

A while ago, we added `metadata` (just a dict of str:str) to applications and endpoints to our app. We also added this field to our application CRUD in the OSS, so it's about time we did the same for endpoints.

## Solution

This updates the CRUD group for /endpoints to return/configure `metadata`. Most of this pretty straightforward, with a couple of notes
* We don't persist empty Metadata records, just to save space (the model abstracts over this detail pretty well)
* The endpoint/crud methods were getting fat. Since `put` upserts `Endpoint`s, there was some common "update endpoint" / "create endpoint" functions we could extract
* Finally, `EndpointOut` is also used by MessageAttemptOut. I didn't want to add the `metadata` field to the message endpoint, so there's a bit of extra indirection in the different `EndpointOut` models that are kind of annoying

### DIFF
--- a/server/svix-server/src/v1/endpoints/endpoint/crud.rs
+++ b/server/svix-server/src/v1/endpoints/endpoint/crud.rs
@@ -1,11 +1,11 @@
-use std::collections::HashSet;
+use std::{collections::HashSet, mem};
 
 use axum::{
     extract::{Extension, Path},
     Json,
 };
 use hyper::StatusCode;
-use sea_orm::{entity::prelude::*, ActiveValue::Set, QueryOrder};
+use sea_orm::{entity::prelude::*, ActiveValue::Set, QueryOrder, TransactionTrait};
 use sea_orm::{ActiveModelTrait, DatabaseConnection, QuerySelect};
 use url::Url;
 
@@ -21,10 +21,10 @@ use crate::{
         },
     },
     ctx,
-    db::models::{endpoint, eventtype},
+    db::models::{application, endpoint, endpointmetadata, eventtype},
     error::{HttpError, Result, ValidationErrorItem},
     v1::utils::{
-        patch::{UnrequiredField, UnrequiredNullableField},
+        patch::{patch_field_non_nullable, UnrequiredField, UnrequiredNullableField},
         EmptyResponse, ListResponse, ModelIn, ModelOut, Pagination, PaginationLimit, ValidatedJson,
         ValidatedQuery,
     },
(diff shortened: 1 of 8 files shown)


server/svix-server/src/v1/endpoints/endpoint/mod.rs
```diff
@@ -330,14 +340,11 @@ pub struct EndpointOut {
     #[serde(rename = "filterTypes")]
     pub event_types_ids: Option<EventTypeNameSet>,
     pub channels: Option<EventChannelSet>,
-
-    pub id: EndpointId,
```
svix-gabriel, line 334: Annoying that this had to be removed - but in order for `ModelOut` to work, you need an `id` field. And it was easier if `EndpointOut impl ModelOut` instead of `EndpointOutNoMetadata`. (And since `EndpointNoMetadata` is embedded and flattened in `EndpointOut`, both structs can't have an `id` field)

tasn: I don't fully understand this. Shouldn't this just be common-endpoint-out?

Also, why do we even need an EndpointOutNoMetadata? I don't see it really used anywhere. EndpointOut should just not have metadata and that's it.

server/svix-server/src/v1/endpoints/endpoint/mod.rs
```diff
@@ -348,14 +355,32 @@ impl From<endpoint::Model> for EndpointOut {
...
     }
 }
 
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize, ModelOut)]
+#[serde(rename_all = "camelCase")]
+pub struct EndpointOut {
+    #[serde(flatten)]
+    pub ep: EndpointOutNoMetadata,
+    pub id: EndpointId,
+    pub metadata: Metadata,
```
tasn, line 370: This shouldn't be returned normally, only when metadata is specifically asked for.

svix-gabriel: In our standard API we always return this field. We should definitely be consistent (I also don't see the value in hiding this field normally regardless)

server/svix-server/src/v1/endpoints/endpoint/mod.rs
svix-gabriel: Why shouldn't EndpointOut return the metadata field. We always return this field in our normal API - beyond that it seems like unnecessary complexity to hide this field normally.

See the PR description for `EndpointNoMetadata` got added - one of the message attempt endpoints uses `EndpointOut`, and we _don't_ normally return the metadata there.

tasn: OK, I get this comment now, though I'd at least call it EndpointOutCommon or something.

server/svix-server/src/v1/endpoints/endpoint/crud.rs
```diff
@@ -47,156 +47,172 @@ pub(super) async fn list_endpoints(
...
     )?
     .ok_or_else(|| HttpError::not_found(None, None))?;
-    Ok(Json(endp.into()))
+
+    let metadata = metadata.map(|m| m.data).unwrap_or_default();
+
+    Ok(Json((endp, metadata).into()))
+}
+
+async fn update_endp_data(
```
tasn, line 136:
```suggestion
async fn update_endp_from_data(
```
I guess?

Same for `create`

server/svix-server/src/v1/endpoints/endpoint/crud.rs
```diff
@@ -47,156 +47,172 @@ pub(super) async fn list_endpoints(
...
-                .send_operational_webhook(
-                    &app.org_id,
-                    OperationalWebhook::EndpointCreated(EndpointEvent::new(app_uid.as_ref(), &ret)),
-                )
-                .await?;
-
-            Ok((StatusCode::CREATED, Json(ret.into())))
-        }
+    if let Some((mut endp, mut metadata)) = models {
+        metadata.data = Set(mem::take(&mut data.metadata));
```
tasn, line 179: mem::take is scary, why do we need it?

svix-gabriel: So all it does is just replace the value with it's `default()` (which in this case is just an empty map), and return the old value. It's not `unsafe` - I definitely wouldn't have used it if it was.

The advantage is just that it saves a `clone`, which for `metadata` seems worthwhile.

server/svix-server/src/v1/endpoints/endpoint/mod.rs
tasn: Oh we do, I thought we didn't.

tasn (approved):

PR #733, opened 2022-12-21
Fix TS string compare

arjunyel:
## Motivation

Fix broken JavaScript string compare

## Solution

Used function from the Deno standard Library

### DIFF
--- a/javascript/src/index.ts
+++ b/javascript/src/index.ts
@@ -51,6 +51,7 @@ import {
 } from "./openapi/index";
 export * from "./openapi/models/all";
 export * from "./openapi/apis/exception";
+import { timingSafeEqual } from "./timing_safe_equal";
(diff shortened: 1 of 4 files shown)


javascript/src/index.ts
```diff
@@ -302,20 +303,21 @@ class Endpoint {
...
   ): Promise<void> {
-    return this.api.replayMissingWebhooksApiV1AppAppIdEndpointEndpointIdReplayMissingPost({
-      appId,
-      endpointId,
-      replayIn,
-      ...options,
-    })
-    .then(() => Promise.resolve());
+    return this.api
+      .replayMissingWebhooksApiV1AppAppIdEndpointEndpointIdReplayMissingPost({
```
tasn, line 314: Unrelated change

tasn (approved):

PR #735, opened 2022-12-21
Fix table in endpointmetadata model declaration

svix-gabriel:


tasn (approved):
Yeah, I think it's fine because this hasn't been actually released as a version.

PR #748, opened 2023-01-03
Endpoint.URL - change to parse, not validate

svix-gabriel:
## Motivation

Right now, we call `Url::parse(...)` multiple times on `EndpointIn`'s URL field. This isn't a huge inconvenience, but it's definitely more complicated and inefficient than things need to be.

## Solution

This PR changes the EndpointIn field to be a URL type instead of a String - leveraging the URL crate's serde (de)serialization to only `Url::parse` once.

The only annoying downside is that `Url` doesn't have a default implementation - so we had to role our own. However, this ended up cleaning up a lot of test logic (since that's all that uses `EndpointIn::default()`), so more likely a win.

In addition, `Url::to_string()` normalizes the URL domain, down-casing the string and adding a trailing `/`. This meant updating a few tests, although this shouldn't be a functionally breaking change for anyone.

### DIFF
--- a/server/svix-server/src/v1/endpoints/endpoint/mod.rs
+++ b/server/svix-server/src/v1/endpoints/endpoint/mod.rs
@@ -96,37 +96,26 @@ fn validate_channels_endpoint_unrequired_nullable(
     }
 }
 
-pub fn validate_url(val: &str) -> std::result::Result<(), ValidationError> {
-    match Url::parse(val) {
-        Ok(url) => {
-            let scheme = url.scheme();
-            if scheme == "https" || scheme == "http" {
-                Ok(())
-            } else {
-                Err(validation_error(
-                    Some("url"),
-                    Some("Endpoint URL schemes must be http or https"),
-                ))
-            }
-        }
-
-        Err(_) => Err(validation_error(
+pub fn validate_url(url: &Url) -> std::result::Result<(), ValidationError> {
+    let scheme = url.scheme();
+    if scheme == "https" || scheme == "http" {
+        Ok(())
+    } else {
+        Err(validation_error(
             Some("url"),
-            Some("Endpoint URLs must be valid"),
-        )),
+            Some("Endpoint URL schemes must be http or https"),
+        ))
     }
 }
 
-fn validate_url_unrequired(
-    val: &UnrequiredField<String>,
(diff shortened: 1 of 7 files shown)


server/svix-server/src/v1/endpoints/endpoint/mod.rs
```diff
@@ -203,7 +210,7 @@ impl ModelIn for EndpointIn {
         model.description = Set(description);
         model.rate_limit = Set(rate_limit.map(|x| x.into()));
         model.uid = Set(uid);
-        model.url = Set(url);
+        model.url = Set(url.into());
```
tasn, line 213: You can probably also change the database model to be a Url (or a wrapper of it most likely), and then you don't even need the from/into.

svix-gabriel: I'm not opposed in principle, although in this case I don't think it's worth the effort. `url::Url` doesn't already have the traits for being used as a column value w/sea_orm, so we'd have to add a wrapper type, all the sea_orm traits boilerplate, and still call `.into()` here.

Since we don't interact/mutate the URL outside of `EndpointIn`, I don't think the wrapper type would be much use yet (although it might be worth it in the future).

server/svix-server/src/v1/endpoints/endpoint/mod.rs
```diff
@@ -96,37 +96,28 @@ fn validate_channels_endpoint_unrequired_nullable(
...
-                Ok(())
-            } else {
-                Err(validation_error(
-                    Some("url"),
-                    Some("Endpoint URL schemes must be http or https"),
-                ))
-            }
-        }
+pub fn validate_url(url: &Url) -> std::result::Result<(), ValidationError> {
+    tracing::warn!("validating url = {:?}", url.to_string());
```
svix-james, line 100: Why is there a `warn` log here?

svix-gabriel: Because I forgot to delete after debugging locally. Deleting . . .

svix-james (approved):

server/svix-server/src/v1/endpoints/endpoint/mod.rs
tasn: 👍🏻

PR #753, opened 2023-01-04
Change operational webhook types to own data

svix-gabriel:


server/svix-server/src/core/operational_webhooks.rs
```diff
@@ -25,47 +25,47 @@ use crate::{
...
-    pub fn new(app_uid: Option<&'a ApplicationUid>, endp: &'a endpoint::Model) -> Self {
+impl EndpointEvent {
+    pub fn new(app_uid: Option<&ApplicationUid>, endp: &endpoint::Model) -> Self {
         Self {
-            app_id: &endp.app_id,
-            app_uid,
-            endpoint_id: &endp.id,
-            endpoint_uid: endp.uid.as_ref(),
+            app_id: endp.app_id.clone(),
+            app_uid: app_uid.map(Clone::clone),
```
jaymell, line 50: Can we change this (and other occurrences) to used `.cloned` instead of `.map`? IMO is a bit cleaner.

svix-gabriel: TIL about that that method. Agreed it's much cleaner, should be updated now.

svix-james (approved):

PR #761, opened 2023-01-10
Switch axum `Extension` usage to `State`

### LINKED ISSUE #726: Server: change use of Axum `Extension` with `State`
We probably want to just pass this state to the worker too, and have a global state of connection pools, etc.

svix-andor:
## Motivation
`State` is the new, type-safe way of passing around global state between handlers in axum. `Extension`s were only resolved at runtime, therefore if you forgot to pass an extension with `layer()` it'd only fail at runtime. With `State` a handler can't be registered on the router if its state can't pass all the required parameters to the handler.

## Solution
Replace all usage of `Extension` with state.

Closes #726


svix-gabriel (approved):
We should just make sure to squash the commits before merging (might have to force push, in which case I can reapprove).

svix-gabriel (approved):

PR #763, opened 2023-01-13
Integrate aide for automatic OpenAPI generation

svix-andor:
Aide is a library that automatically generates an OpenAPI schema by integrating with axum, extracting API information based on axum routing and extractors.

## Motivation
Replace the currently static OpenAPI definition with an automatically generated one, based on code. Among other benefits, it's useful to ensure we don't make breaking changes to the API by accident.

## Solution
`aide` was chosen as it is lightweight and has the most features out of the box for us. Alternatives considered:
* [`utoipa`](https://github.com/juhaku/utoipa) - integrates with axum like aide but relies heavily on code generation with a lot of annotations that aide is able to derive automatically out of the box
* [`okapi`](https://github.com/GREsau/okapi) - does not integrate with axum out of the box

While aide is newer and perhaps less mature it has fewer dependencies. Nearly all of its direct dependencies are crates we already include other than `schemars`, and `dyn-clone` for aide-macros, the latter being optional. `cargo audit` reports no known vulnerabilities in its dependency tree that we don't already depend on (`time=0.1.45`) otherwise.

### DIFF
--- a/server/Cargo.lock
+++ b/server/Cargo.lock
@@ -51,6 +51,39 @@ dependencies = [
 ]
 
 [[package]]
+name = "aide"
+version = "0.9.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "befdff0b4683a0824fc8719ce639a252d9d62cd89c8d0004c39e2417128c1eb8"
+dependencies = [
+ "aide-macros",
+ "axum",
+ "bytes",
+ "cfg-if",
+ "http",
+ "indexmap",
+ "schemars",
+ "serde",
+ "serde_json",
+ "thiserror",
+ "tower-layer",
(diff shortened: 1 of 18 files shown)


svix-gabriel:
Having had time to go through the libraries being used, in general this change seems good. And I think the approach of trying to push the scaffolding first makes sense since there's such a large blast radius, diff-wise.

That being said - since the docs generated from this change aren't quite ready yet, could we keep the old documentation that shows up /docs? That way this won't break any existing workflow, but we can keep iterating on the new doc generation until it's ready?

svix-andor:
Yep, that makes sense. I added back the existing docs.

svix-gabriel (approved):

PR #765, opened 2023-01-17
Add feature flag gating to event types

svix-andor:
## Motivation

Organizations might want to introduce new event types without first exposing them to their users. These changes allow adding an optional `feature_flag` to event types. If an event type has a feature flag set then only application tokens which have the corresponding feature flag set can see the event type. This does *not* filter receiving these events on endpoints though.

## Solution

Event types can now be gated behind a feature flag field. If an event type has a featue flag set then it won't be returned to application access clients that don't have the appropriate event type feature flag claim set in their token.

Organization access tokens can always see all event types.

### DIFF
--- a/server/svix-server/migrations/20230116095641_add_event_type_feature_flags.down.sql
+++ b/server/svix-server/migrations/20230116095641_add_event_type_feature_flags.down.sql
@@ -1,0 +1,2 @@
+-- Add down migration script here
+ALTER TABLE ONLY eventtype DROP COLUMN feature_flag;
--- a/server/svix-server/src/core/permissions.rs
+++ b/server/svix-server/src/core/permissions.rs
@@ -14,11 +14,12 @@ use crate::{
 
 use super::{
     security::{permissions_from_bearer, AccessLevel, Permissions},
-    types::{ApplicationId, ApplicationIdOrUid, OrganizationId},
+    types::{ApplicationId, ApplicationIdOrUid, FeatureFlagSet, OrganizationId},
 };
 
 pub struct ReadAll {
     pub org_id: OrganizationId,
+    pub feature_flags: AllowedFeatureFlags,
 }
 
 #[async_trait]
@@ -28,7 +29,14 @@ impl FromRequestParts<AppState> for ReadAll {
     async fn from_request_parts(parts: &mut Parts, state: &AppState) -> Result<Self> {
         let permissions = ctx!(permissions_from_bearer(parts, state).await)?;
         let org_id = permissions.org_id();
-        Ok(Self { org_id })
+        let feature_flags = match permissions.access_level {
+            AccessLevel::Organization(_) => AllowedFeatureFlags::All,
+            AccessLevel::Application(_, _) => AllowedFeatureFlags::Some(permissions.feature_flags),
+        };
+        Ok(Self {
+            org_id,
+            feature_flags,
+        })
     }
 }
 
@@ -156,4 +164,13 @@ impl FromRequestParts<AppState> for ApplicationWithMetadata {
 #[derive(serde::Deserialize)]
 struct ApplicationPathParams {
     app_id: ApplicationIdOrUid,
+}
+
+/// Denotes what features gated by feature flags the requester has access to.
+pub enum AllowedFeatureFlags {
+    /// Requester has access to all features regardless of flags. No checking
+    /// of feature flags should be done for this request.
+    All,
+    /// Requester has access to a limited set of features. The set may be empty.
+    Some(FeatureFlagSet),
 }
--- a/server/svix-server/src/core/security.rs
+++ b/server/svix-server/src/core/security.rs
@@ -19,7 +19,7 @@ use crate::{
     AppState,
 };
 
-use super::types::{ApplicationId, OrganizationId};
+use super::types::{ApplicationId, FeatureFlagSet, OrganizationId};
 
 /// The default org_id we use (useful for generating JWTs when testing).
 pub fn default_org_id() -> OrganizationId {
@@ -38,6 +38,7 @@ pub enum AccessLevel {
 
 pub struct Permissions {
     pub access_level: AccessLevel,
+    pub feature_flags: FeatureFlagSet,
 }
 
 impl Permissions {
@@ -60,6 +61,13 @@ impl Permissions {
 pub struct CustomClaim {
     #[serde(rename = "org", default, skip_serializing_if = "Option::is_none")]
     pub organization: Option<String>,
+
+    #[serde(
+        rename = "feature_flags",
+        default,
+        skip_serializing_if = "FeatureFlagSet::is_empty"
+    )]
+    pub feature_flags: FeatureFlagSet,
 }
 
 pub async fn permissions_from_bearer(parts: &mut Parts, state: &AppState) -> Result<Permissions> {
@@ -100,6 +108,7 @@ pub fn permissions_from_jwt(claims: JWTClaims<CustomClaim>) -> Result<Permission
 
             Ok(Permissions {
                 access_level: AccessLevel::Application(org_id, app_id),
+                feature_flags: claims.custom.feature_flags,
             })
         } else {
             Err(
@@ -119,6 +128,7 @@ pub fn permissions_from_jwt(claims: JWTClaims<CustomClaim>) -> Result<Permission
         })?;
         Ok(Permissions {
             access_level: AccessLevel::Organization(org_id),
+            feature_flags: claims.custom.feature_flags,
         })
     } else {
         Err(
@@ -132,7 +142,10 @@ const JWT_ISSUER: &str = env!("CARGO_PKG_NAME");
 
 pub fn generate_org_token(keys: &Keys, org_id: OrganizationId) -> Result<String> {
     let claims = Claims::with_custom_claims(
-        CustomClaim { organization: None },
+        CustomClaim {
+            organization: None,
+            feature_flags: Default::default(),
+        },
         Duration::from_hours(24 * 365 * 10),
     )
     .with_issuer(JWT_ISSUER)
@@ -141,10 +154,15 @@ pub fn generate_org_token(keys: &Keys, org_id: OrganizationId) -> Result<String>
 }
 
 pub fn generate_management_token(keys: &Keys) -> Result<String> {
-    let claims =
-        Claims::with_custom_claims(CustomClaim { organization: None }, Duration::from_mins(10))
-            .with_issuer(JWT_ISSUER)
-            .with_subject(management_org_id());
+    let claims = Claims::with_custom_claims(
+        CustomClaim {
+            organization: None,
+            feature_flags: Default::default(),
+        },
+        Duration::from_mins(10),
+    )
+    .with_issuer(JWT_ISSUER)
+    .with_subject(management_org_id());
     Ok(keys.key.authenticate(claims).unwrap())
 }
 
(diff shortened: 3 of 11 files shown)


server/svix-server/migrations/20230116095641_add_event_type_feature_flags.down.sql
```diff
@@ -0,0 +1,2 @@
+-- Add down migration script here
+ALTER TABLE ONLY eventtype DROP COLUMN feature_flags;
```
tasn, line 2:
```suggestion
ALTER TABLE ONLY eventtype DROP COLUMN feature_flag;
```

server/svix-server/src/core/permissions.rs
```diff
@@ -157,3 +157,36 @@ impl FromRequestParts<AppState> for ApplicationWithMetadata {
 struct ApplicationPathParams {
     app_id: ApplicationIdOrUid,
 }
+
+pub enum EventTypeFeatureFlagPermissions {
```
tasn, line 161: This looks overly specific and verbose. I don't like the idea of having a specific permission extractor for whatever function we want, and I like it even less having it in permissions.rs rather next to the rest of the code.

Can't we just also extract the JWT extra info and pass it to the function that uses it?

server/svix-server/Cargo.toml
```diff
@@ -54,7 +54,7 @@ bytes = "1.1.0"
 blake2 = "0.10.4"
 chacha20poly1305 = "0.9.0"
 # sea orm
-sea-orm = { version = "0.10.2", features = [ "sqlx-postgres", "runtime-tokio-rustls", "macros", "with-chrono", "with-json" ], default-features = false }
+sea-orm = { version = "0.10.2", features = [ "sqlx-postgres", "runtime-tokio-rustls", "macros", "with-chrono", "with-json", "postgres-array" ], default-features = false }
```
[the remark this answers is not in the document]
svix-andor: What do you mean? Support for postgres array types is behind a feature flag (ha) in sea-orm, this is needed.

server/svix-server/src/core/permissions.rs
tasn: I see you put it in the Permissions struct, shouldn't we just use that directly?

server/svix-server/src/v1/endpoints/auth.rs
```diff
@@ -16,11 +17,24 @@ pub struct DashboardAccessOut {
...
+pub struct DashboardAccessIn {
+    #[validate]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub feature_flags: Option<EventTypeFeatureFlagSet>,
+}
+
 async fn dashboard_access(
     State(AppState { cfg, .. }): State<AppState>,
     permissions::OrganizationWithApplication { app }: permissions::OrganizationWithApplication,
+    ValidatedJson(json): ValidatedJson<DashboardAccessIn>,
```
tasn, line 30:
```suggestion
    ValidatedJson(data): ValidatedJson<DashboardAccessIn>,
```
I think we call it `data` everywhere else.

server/svix-server/src/v1/endpoints/event_type.rs
```diff
@@ -192,6 +213,10 @@ async fn list_event_types(
         query = query.filter(eventtype::Column::Name.gt(iterator));
     }
 
+    if let Some(flags) = flags.allowed_feature_flags() {
+        query = query.filter(eventtype::Column::FeatureFlag.is_in(flags.0.iter().cloned()));
```
tasn, line 217: Do we really need to clone? Why not into_iter? Or rather, can we just steal this?

server/svix-server/Cargo.toml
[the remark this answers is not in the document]
tasn: But you're not using postgres arrays in this PR...

server/svix-server/src/core/permissions.rs
svix-andor: The answer to this depends on how we want to do permission checks with axum extractors. Currently, we use this "mixin" style, where the different permission levels are different structs, see `permissions::{ReadAll, Application, Organization}`, these are higher level axum extractors that build on top of `Permissions`.

server/svix-server/Cargo.toml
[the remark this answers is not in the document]
svix-andor: Ahh... 🙈

server/svix-server/src/v1/endpoints/event_type.rs
svix-andor: `into_iter` won't work as is. `allowed_feature_flags` returns a borrow into `flags` and `into_iter` would try to consume it. I guess we can make `allowed_feature_flags` take `self` because it's unlikely you'd want to use the original after calling that.

server/svix-server/src/core/permissions.rs
tasn: Can't the application extractor also return the feature flags when exist (for example)?

server/svix-server/src/core/security.rs
```diff
@@ -60,6 +61,9 @@ impl Permissions {
 pub struct CustomClaim {
     #[serde(rename = "org", default, skip_serializing_if = "Option::is_none")]
     pub organization: Option<String>,
+
+    #[serde(rename = "evt_ff", default, skip_serializing_if = "Option::is_none")]
+    pub event_feature_flags: Option<EventTypeFeatureFlagSet>,
```
svix-gabriel, line 66: Does this need to be an `Option`? Could we get away with an empty set as the Default?

server/svix-server/src/core/types/mod.rs
```diff
@@ -437,6 +437,26 @@ impl Validate for EventChannelSet {
     }
 }
 
+string_wrapper!(EventTypeFeatureFlag);
+
+impl Validate for EventTypeFeatureFlag {
```
svix-gabriel, line 442: I know we use this pattern a lot (relying on manually `Validate`ing), but I think a better pattern is just to forbid invalid types from being initialized at all.

`types::metadata` does this - where `serde::Deserialize` is manually implemented to just outright reject invalid values. This way, we don't have to worry about manually validating the type in the API definitions.

I think this would also let you avoid having to define a separate `EventTypeFeatureFlagSet`. We can just use `HashSet<EventTypeFeatureFlag>`, and not worry about invalid types still getting initialized.

server/svix-server/src/v1/endpoints/auth.rs
```diff
@@ -16,11 +17,24 @@ pub struct DashboardAccessOut {
     pub token: String,
 }
 
+#[derive(Deserialize, Serialize, Validate, JsonSchema)]
+pub struct DashboardAccessIn {
+    #[validate]
+    #[serde(skip_serializing_if = "Option::is_none")]
```
svix-gabriel, line 23: Can we change this to just be a `EventTypeFeatureFlagSet` (or `HashSet<EventTypeFeatureFlag>` if you go that route - basically just drop the `Option`). I think Serde should be able to still default it to an empty set if nothing is passed in.

server/svix-server/src/core/permissions.rs
svix-andor: Is it ok if I just drop the `FromRequestParts` impl from `EventTypeFeatureFlagPermissions` and put it as a field on `Application`? Because simply putting `EventTypeFeatureFlagSet` by itself into `Application` is not enough. We need to know if this `Application` authorization came from an organization token or an app token, otherwise in the handler we can't differentiate and allow organization tokens access regardless of flags.

server/svix-server/src/core/permissions.rs
```diff
@@ -157,3 +157,36 @@ impl FromRequestParts<AppState> for ApplicationWithMetadata {
...
+    Organization,
+    Application(EventTypeFeatureFlagSet),
+}
+
+#[async_trait]
+impl FromRequestParts<AppState> for EventTypeFeatureFlagPermissions {
+    type Rejection = Error;
```
svix-gabriel, line 168: Just to add to the above - it seems fine to me if instead of a new type,  just add the `EventTypeFeatureFlagSet` directly to the `ReadAll` struct (or whichever `permissions::` struct needs it). The endpoints that don't need to read the flags can just ignore with `..`, which I think we do a lot of already.

server/svix-server/src/core/permissions.rs
tasn: I don't really remember how the permissions are structured, though sure, it sounds reasonable.

server/svix-server/src/core/security.rs
svix-andor: Yes, empty set as default works, I only made it an `Option` because I'd expect this will not be used for most requests, therefore:
1. Keeps the JWT smaller. Not a big deal but would save an `"evt_ff":[]`.
2. Performance. If it's not an option then a `HashSet` will be allocated even when not used. But that might very well be premature optimization.

Will change to non-`Option`al 👍

tasn: @svix-andor, not serializing an empty evt_ff in tokens is a worthy goal. I dunno whether I'd want to sacrifice it.

svix-gabriel: I think empty collections usually don't allocate, but point 1 (keeping the JWT smaller) seems reasonable.

I'm ok if you leave as is.

server/svix-server/src/core/security.rs
```diff
@@ -60,6 +61,9 @@ impl Permissions {
 pub struct CustomClaim {
     #[serde(rename = "org", default, skip_serializing_if = "Option::is_none")]
     pub organization: Option<String>,
+
+    #[serde(rename = "evt_ff", default, skip_serializing_if = "Option::is_none")]
```
tasn, line 65: It's not just an event type feature flag, so `evt_ff` is very misleading.

I'd literally just call it: `feature_flags`. Yeah, it's maybe a bit longer (by a few chars), though it'll only be used by ones that actually have it, so it's a not-that-important optimization.

server/svix-server/src/core/security.rs
tasn, line 66: As said above, it's not only for events. It's a general feature flag mechanism. Name it accordingly.

server/svix-server/src/core/security.rs
svix-andor: Renamed to `feature_flags`

server/svix-server/src/core/security.rs
svix-andor: Same as above, renamed to `feature_flags`

server/svix-server/src/core/permissions.rs
svix-andor: Yep, did exactly that, I'll resolve this thread so we can keep this topic in one convo.

server/svix-server/src/core/security.rs
tasn: What I said was in addition to your comment. As in, change `skip_serializing_if` to check for HashMap::is_empty or whatever.

server/svix-server/src/core/types/mod.rs
svix-andor: As discussed on slack split `string_wrapper!` into two macros to do this (because that did a `derive(Deserialize)`) and implemented `Deserialize` manually which calls `validate_limited_str`. Also, just made `FeatureFlagSet` a type  alias for `HashSet<FeatureFlag>`.

server/svix-server/src/v1/endpoints/auth.rs
svix-andor: Done.

server/svix-server/src/core/security.rs
svix-andor: Done as discussed:
* No longer an option, just a direct `FeatureFlagSet`
* Added serde `default` annotation
* Changed to `skip_serializing_if = "FeatureFlagSet::is_empty"`

server/svix-server/src/core/permissions.rs
svix-andor: Done, now it's just an enum on `ReadAll` (for some reason I was mixing up `Application` and `ReadAll`)

PR #771, opened 2023-01-19
Make request body to dashboard-access optional again

svix-andor:
## Motivation
PR #765 added support for attaching feature flags to auth tokens. This added a request body requirement to the dashboard access endpoint which is a backwards-incompatible change for old client libraries. 

## Solution
Change it back so that a body is optional, keeping backwards compatibility.


svix-gabriel (approved):
Maybe change the PR title? "Revert" feels like a misnomer. But Otherwise LGTM

svix-gabriel (approved):

PR #774, opened 2023-01-20
Fix event type filtering involving feature flags

svix-andor:
## Motivation
The previous logic was incorrectly filtering out event types with no feature flag set.

## Solution
Apply the correct condition to the database query.

### DIFF
--- a/server/svix-server/src/v1/endpoints/event_type.rs
+++ b/server/svix-server/src/v1/endpoints/event_type.rs
@@ -215,7 +215,7 @@ async fn list_event_types(
     }
 
     if let permissions::AllowedFeatureFlags::Some(flags) = feature_flags {
-        query = query.filter(eventtype::Column::FeatureFlag.is_in(flags.into_iter()));
+        query = eventtype::Entity::filter_feature_flags(query, flags);
     }
 
     Ok(Json(EventTypeOut::list_response_no_prev(
@@ -280,7 +280,7 @@ async fn get_event_type(
 ) -> Result<Json<EventTypeOut>> {
     let mut query = eventtype::Entity::secure_find_by_name(org_id, evtype_name);
     if let permissions::AllowedFeatureFlags::Some(flags) = feature_flags {
-        query = query.filter(eventtype::Column::FeatureFlag.is_in(flags.into_iter()));
+        query = eventtype::Entity::filter_feature_flags(query, flags);
     }
     let evtype = ctx!(query.one(db).await)?.ok_or_else(|| HttpError::not_found(None, None))?;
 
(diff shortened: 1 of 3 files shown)


server/svix-server/src/v1/endpoints/event_type.rs
```diff
@@ -280,7 +283,10 @@ async fn get_event_type(
 ) -> Result<Json<EventTypeOut>> {
     let mut query = eventtype::Entity::secure_find_by_name(org_id, evtype_name);
     if let permissions::AllowedFeatureFlags::Some(flags) = feature_flags {
-        query = query.filter(eventtype::Column::FeatureFlag.is_in(flags.into_iter()));
+        let condition = sea_orm::Condition::any()
```
svix-gabriel, line 286: Seeing the beefier condition - maybe we could move this as a method on the `eventtype::Entity`? I know we don't do that all the time, although when the sea_orm queries get more verbose I think it's generally cleaner (code reuse arguments aside)

tasn: I agree.

Just to add another point to this: splitting it to a separate condition variable is unnecessarily verbose and makes things less clear, not more clear.

svix-andor: I moved the filtering into an associated function on the entity

server/svix-server/src/v1/endpoints/event_type.rs
```diff
@@ -280,7 +280,7 @@ async fn get_event_type(
 ) -> Result<Json<EventTypeOut>> {
     let mut query = eventtype::Entity::secure_find_by_name(org_id, evtype_name);
     if let permissions::AllowedFeatureFlags::Some(flags) = feature_flags {
-        query = query.filter(eventtype::Column::FeatureFlag.is_in(flags.into_iter()));
+        query = eventtype::Entity::filter_feature_flags(query, flags);
```
svix-gabriel, line 283: AFAICT we haven't really done anything like this, but FWIW I like it. It's certainly more composable than the individual methods on the Entity.

server/svix-server/tests/e2e_event_type.rs
```diff
@@ -238,50 +253,38 @@ async fn test_event_type_feature_flags() {
...
+        if should_see {
+            // If the client is expected to see both event types it should be able to retrieve it
+            let got_et: EventTypeOut = client.get(&path, StatusCode::OK).await.unwrap();
+            assert_eq!(et, got_et);
+
+            // ... and see it in the list.
+            assert_eq!(list.data.len(), 2);
+            assert!(list.data.contains(&et));
+        } else {
+            // If the client is not supposed to see it it shouldn't be able to rertieve it
```
svix-gabriel, line 280:
```suggestion
            // If the client is not supposed to see it it shouldn't be able to retrieve it
```

svix-gabriel (approved):

PR #777, opened 2023-01-20
implement clone derive macro for Sivx client

imabdulbasit:
## Motivation
Configuration struct derives clone macro so Sivx struct can also derive it. This would be helpful to clone Sivx client when sharing it between threads without wrapping it in Arc.

## Solution
Add Clone derive macro

### DIFF
--- a/rust/src/api.rs
+++ b/rust/src/api.rs
@@ -25,6 +25,7 @@ pub struct SvixOptions {
     pub server_url: Option<String>,
 }
 
+#[derive(Clone)]
 pub struct Svix {
     cfg: Configuration,
 }


tasn:
I'm not against cloning it per-se, but it feels wasteful to clone it and all of the inner structure for it. I almost feel like if you want to do it, we should at least have the inner object wrapped in an Arc, no?

imabdulbasit:
> I'm not against cloning it per-se, but it feels wasteful to clone it and all of the inner structure for it. I almost feel like if you want to do it, we should at least have the inner object wrapped in an Arc, no?

Inner structure also has Clone implemented :)

tasn:
There's a `Configuration` structure, no? It probably includes multiple things in it.

imabdulbasit:
> There's a `Configuration` structure, no? It probably includes multiple things in it.

Yes but Configuration structure implements Clone so Svix struct can also implement clone

tasn:
Yeah, but that's what I'm saying, it ends up cloning the whole Configuration structure...

Anyhow, I can go either way. @svix-gabriel, do you have any thoughts on this?

svix-gabriel:
I definitely see the value in implementing `Clone`, but I agree with the suggestion that this should wrap the Configuration struct in an `Arc<...>` internally. `Clone` will be more performant if it's just incrementing a reference count, rather than copying all of the underlying strings and other configuration data.

svix-jplatte:
Superseded by  #1493.

PR #783, opened 2023-01-24
Make `wait_for` timeout early and retry

### LINKED ISSUE #781: Fix `wait-for` to have a short timeout (e.g 2s) for each connection attempt
The problem that we are facing now (without it), is that if the first attempt hangs for whatever reason, the whole wait for would fail, even though trying again would have worked just fine.

It's a simple fix.

svix-andor:
## Motivation
It's better to retry the connection attempt after a few seconds than waiting for the entire duration of `wait_for_seconds`. In cloud and containerized environments it's not uncommon to start running before the network stack has been brought up, causing the first connection attempt to hang until the full `wait_for` duration has been exhausted, even though retrying after a few seconds would have resulted in a successful connection.

## Solution
Wrap connection attempt in a `tokio::time::timeout`.

Resolves #781

### DIFF
--- a/server/svix-server/src/wait_for.rs
+++ b/server/svix-server/src/wait_for.rs
@@ -16,13 +16,22 @@ pub async fn wait_for_dsn(
     let sleep = tokio::time::sleep(Duration::from_secs(wait_for_seconds));
     tokio::pin!(sleep);
 
+    const ATTEMPT_TIMEOUT: Duration = Duration::from_secs(3);
+    const RETRY_WAIT: Duration = Duration::from_millis(500);
+
     loop {
         tokio::select! {
-            res = tokio::net::TcpStream::connect((host.to_string(), port)) => {
-                if res.is_err() {
-                    continue;
-                } else {
-                    break;
+            // Attempt the connection with a timeout
+            res = tokio::time::timeout(ATTEMPT_TIMEOUT, tokio::net::TcpStream::connect((host.to_string(), port))) => {
+                match res {
+                    // Connection attempt succeeded
+                    Ok(Ok(_)) => break,
+                    // Connection failed before the timeout was reached _or_ timed out
+                    Ok(Err(_)) | Err(_) => {
+                        tracing::debug!("{dependency_name} connection attempt failed, retrying in {RETRY_WAIT:?}...");
+                        tokio::time::sleep(RETRY_WAIT).await;
+                        continue;
+                    },
                 }
             }
 


server/svix-server/src/wait_for.rs
```diff
@@ -16,10 +16,13 @@ pub async fn wait_for_dsn(
...
     tokio::pin!(sleep);
 
+    const ATTEMPT_TIMEOUT: Duration = Duration::from_secs(3);
+
     loop {
         tokio::select! {
-            res = tokio::net::TcpStream::connect((host.to_string(), port)) => {
+            res = tokio::time::timeout(ATTEMPT_TIMEOUT, tokio::net::TcpStream::connect((host.to_string(), port))) => {
                 if res.is_err() {
+                    tracing::debug!("{dependency_name} connection attempt failed after {:?}, retrying...", ATTEMPT_TIMEOUT);
```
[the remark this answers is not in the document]
svix-andor: I added a simple exponential backoff to the connection attempt if it fails before hitting the deadline.

server/svix-server/src/wait_for.rs
```diff
@@ -16,13 +16,31 @@ pub async fn wait_for_dsn(
...
         tokio::select! {
-            res = tokio::net::TcpStream::connect((host.to_string(), port)) => {
-                if res.is_err() {
-                    continue;
-                } else {
-                    break;
+            res = tokio::time::timeout(ATTEMPT_TIMEOUT, tokio::net::TcpStream::connect((host.to_string(), port))) => {
+                match res {
+                    Ok(Ok(_)) => break,
+                    Ok(Err(_)) => {
```
tasn, line 29: Going to make a comment that I don't know if it can be solved, but I wanted to bring this up: the code is very confusing at the moment. You have timeouts in three different places and the `Ok(Ok(_))` is indecipherable (should at least have comments explaining what which option is).

I don't know, but I wonder if it's possible to simplify it somehow, something like: an overall timeout enforced like the select is now, and then just wrap the TcpConnect with a timeout, so that every time we attempt this attempt has a timeout for itself (of e.g. 3s). Then after every attempt (or timeout) just sleep for 500ms, that's it. (it's looped)

Does that make sense?
```

Comment (svix-andor) [reply server/svix-server/src/wait_for.rs:29]: If you mean setting some TCP socket-level option: that's not possible with tokio `TcpStream`s, it's a bit antithetical with async. I can simplify by treating both the timeout-error `Err(_)` and the `Ok(Err(_))` in the same branch and just apply a fixed 500ms sleep.

Comment (tasn) [reply server/svix-server/src/wait_for.rs:29]: I more meant:
```rust
pub async fn andors_tcp_connect(
    dsn: &str,
) {
    tokio::select! {
        res = TcpConnect(...) => { return true }

       _request_timeout => { return false }
   }
}
```

And then just use this function as a timeoutable TCP request function.

You can then do in the main code that we already have:
```rust
if andors_tcp_connect(dsn) {
    break
} else {
    sleep(500ms).await
}
```
Get what I mean?


This will make the code very easy to parse and follow.

svix-andor: I simplified it so that both the early-error and timeout cases are handled in one branch, and it uses a simple fixed retry.

svix-gabriel (approved):

Issue #756, opened 2023-01-09
Add Date Range Option To Endpoint Get Stats


enchorb:
## Feature Request

Add `before` and `after` to `endpoint.getStats`

### Motivation

Would be useful for billing at the end of a month and getting all the usage for an endpoint in that interval

tasn:
Thanks for the report, and yeah, I full agree!

tasn:
@svix-gabriel is already working on adding `since` and `until` to this. Just tagging him so he's also aware of this ticket.

svix-gabriel:
Moving this to the backlog as we still need to update the SDKs. However, the new query params are available on the `stats` endpoint (see [docs](https://api.svix.com/docs#tag/Endpoint/operation/get_endpoint_stats_api_v1_app__app_id__endpoint__endpoint_id__stats__get)), so they should be usable outside.

tasn:
Thanks @svix-gabriel for the update!

@enchorb, it seems like we added this all way back in Jan! We just forgot to update here because we haven't updated the libs.

tasn:
Estimate: 1

PR #784, opened 2023-01-24
Add since/until query params to endpoint stats

svix-gabriel:
## Motivation

For https://github.com/svix/svix-webhooks/issues/756, we want to give folks the ability to control the lookback period to get endpoint stats for. The original issue suggested "before/after", but most of our newer endpoints have used "since/until", so that's what we went with here.

## Solution

Just adds the new query params. This endpoint also didn't have any E2E tests, so I went ahead and added some. They aren't perfect integration tests, since we create the messagedestination records manually. But it at least tests the query being run (and a true integration test would be much flakier since we can't really control the state of messagedestinations from outside the worker).

### DIFF
--- a/server/svix-server/src/v1/endpoints/endpoint/mod.rs
+++ b/server/svix-server/src/v1/endpoints/endpoint/mod.rs
@@ -36,10 +36,10 @@ use aide::axum::{
     ApiRouter,
 };
 use axum::{
-    extract::{Path, State},
+    extract::{Path, Query, State},
     Json,
 };
-use chrono::{DateTime, Utc};
+use chrono::{DateTime, Duration, Utc};
 use schemars::JsonSchema;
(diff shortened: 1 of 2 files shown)


server/svix-server/src/v1/endpoints/endpoint/mod.rs
```diff
@@ -474,12 +474,45 @@ impl ModelIn for EndpointHeadersPatchIn {
...
+        if until > Utc::now() {
+            return Err(HttpError::bad_request(
+                Some("invalid_range".into()),
+                Some("'until' cannot be in the future".into()),
+            )
+            .into());
+        }
+
+        let since = self.since.unwrap_or(until - Duration::days(28));
+        if until - since > Duration::days(28) {
```
svix-andor, line 496: I would probably play extra-safe with this and add a bit of an epsilon to the allowed window in the condition here, because this can be weird due to leap seconds or other calendar weirdness.

svix-andor (approved):

tasn (approved):

PR #785, opened 2023-01-24
Add org_id and app_id to main tracing span

svix-daniel:
This change simply adds two additional fields, the `org_id` and `app_id` (if any) associated with a request, to the main `tracing` span created in `core::otel_spans` allowing the relevant organization to be recorded in the event of an error.

### DIFF
--- a/server/svix-server/src/core/otel_spans.rs
+++ b/server/svix-server/src/core/otel_spans.rs
@@ -125,6 +125,8 @@ impl<B> MakeSpan<B> for AxumOtelSpanCreator {
             request_id = %request_id,
             trace_id = %trace_id,
             idempotency_key = tracing::field::Empty,
+            org_id = tracing::field::Empty,
+            app_id = tracing::field::Empty,
         );
 
         if let Some(key) = idempotency_key {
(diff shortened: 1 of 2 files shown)


server/svix-server/src/core/permissions.rs
```diff
@@ -71,6 +74,8 @@ impl FromRequestParts<AppState> for Organization {
             _ => return Err(HttpError::permission_denied(None, None).into()),
         };
 
+        tracing::Span::current().record("org_id", org_id.to_string());
```
tasn, line 77: Instead of having this everywhere, shouldn't this just be in permissions_from_bearer so that it's centralized?

server/svix-server/src/core/otel_spans.rs
```diff
@@ -125,6 +125,7 @@ impl<B> MakeSpan<B> for AxumOtelSpanCreator {
             request_id = %request_id,
             trace_id = %trace_id,
             idempotency_key = tracing::field::Empty,
```
tasn, line 127: Since you are at it, also add `app_id`, and populate it when it's an app token.

server/svix-server/src/core/security.rs
```diff
@@ -126,6 +133,7 @@ pub fn permissions_from_jwt(claims: JWTClaims<CustomClaim>) -> Result<Permission
                 Some("`sub' is not a valid organization id.".to_string()),
             )
         })?;
+
```
tasn, line 136:
```suggestion
```

svix-james (approved):

PR #786, opened 2023-01-25
Add replay method to libraries

### LINKED ISSUE #764: Replay missing in Rust library
It seems like the Rust library is missing the ability to replay missing messages.

svix-andor:
It was only implemented in JS and missed for the other libs.

Fixes #764

### DIFF
--- a/java/lib/src/main/java/com/svix/Endpoint.java
+++ b/java/lib/src/main/java/com/svix/Endpoint.java
@@ -12,6 +12,7 @@ import com.svix.models.EndpointSecretOut;
 import com.svix.models.EndpointSecretRotateIn;
 import com.svix.models.ListResponseEndpointOut;
 import com.svix.models.RecoverIn;
+import com.svix.models.ReplayIn;
 import com.svix.models.EndpointStats;
 
 public final class Endpoint {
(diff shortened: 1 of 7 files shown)


java/lib/src/main/java/com/svix/Endpoint.java
```diff
@@ -128,4 +129,20 @@ public EndpointStats getStats(final String appId, final String endpointId) throw
 			throw Utils.wrapInternalApiException(e);
 		}
 	}
+
+	public void replay(final String appId, final String endpointId, final ReplayIn replayIn) throws ApiException {
+		try {
```
svix-gabriel, line 134: Just looking at how `recover` is implemented, shouldn't this be
```java
this.replay(appId, endpointId, replayIn, new PostOptions())
```

svix-andor: Good catch, fix'd

java/lib/src/main/java/com/svix/Endpoint.java
```diff
@@ -128,4 +129,20 @@ public EndpointStats getStats(final String appId, final String endpointId) throw
...
+			this.replay(appId, endpointId, replayIn, new PostOptions());
+		} catch (com.svix.internal.ApiException e) {
```
svix-gabriel, line 136: Granted it's been a while since I've written Java, but the `try { } catch {}` here is redundant, no?

svix-andor: Yep, also done.

svix-gabriel (approved):

PR #791, opened 2023-01-26
Add endpoints to erase payload and response

svix-andor:
These endpoints can be used to erase the payload of a message, and the response body of attempts. They can be useful in cases where a message was inadvertently sent with sensitive data that it shouldn't have contained, or an endpoint returned a response body with something of the same sort.

Erasing message payloads is technically OK, the worker already handles the case where there's no payload (due to payload expiration) and the response body is not currently used by svix, it is only retained for troubleshooting purposes.

### DIFF
--- a/server/svix-server/src/v1/endpoints/attempt.rs
+++ b/server/svix-server/src/v1/endpoints/attempt.rs
@@ -26,7 +26,7 @@ use crate::{
     AppState,
 };
 use aide::axum::{
-    routing::{get, post},
+    routing::{delete, get, post},
     ApiRouter,
 };
 use axum::{
@@ -37,7 +37,10 @@ use chrono::{DateTime, Utc};
 
 use hyper::StatusCode;
 use schemars::JsonSchema;
-use sea_orm::{entity::prelude::*, sea_query::Expr, DatabaseConnection, QueryOrder, QuerySelect};
+use sea_orm::{
+    entity::prelude::*, sea_query::Expr, DatabaseConnection, IntoActiveModel, QueryOrder,
+    QuerySelect,
+};
 use serde::{Deserialize, Serialize};
 
 use svix_server_derive::ModelOut;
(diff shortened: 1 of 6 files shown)


server/svix-server/src/v1/endpoints/attempt.rs
```diff
@@ -720,6 +754,11 @@ pub fn router() -> ApiRouter<AppState> {
             get(get_messageattempt),
             openapi_tag("Message Attempt"),
         )
+        .api_route_with(
+            "/app/:app_id/msg/:msg_id/attempt/:attempt_id/response/",
```
tasn, line 758:
```suggestion
            "/app/:app_id/msg/:msg_id/attempt/:attempt_id/content/",
```
Probably let's make it consistent with msg? I'm on the fence here to be honest, but I think calling both `content` or maybe we should change both to `body`?

svix-andor: I'm leaning toward `content`.

tasn: Yeah, you're right. Especially because I just remembered we have `with_content` as a query param in various places.

server/svix-server/src/v1/endpoints/attempt.rs
```diff
@@ -707,6 +710,37 @@ async fn resend_webhook(
...
+        MessageAttemptId,
+    )>,
+    permissions::OrganizationWithApplication { app }: permissions::OrganizationWithApplication,
+) -> Result<StatusCode> {
+    let msg = ctx!(
+        message::Entity::secure_find_by_id_or_uid(app.id, msg_id)
+            .one(db)
+            .await
+    )?
+    .ok_or_else(|| HttpError::not_found(None, None))?;
```
svix-gabriel, line 727: We're definitely inconsistent about this, but I think in situations like this it's helpful to add a bit more information to the `not_found(...)` error. As it is now, if a request 404s it's not clear if that's because the app_id was wrong, the msg_id was wrong, or the attempt_id was wrong.

svix-andor: I added a message (but not a separate code) to each of the errors to say if it errored on the message or the message attempt.

svix-gabriel (approved):

Issue #792, opened 2023-01-27
Mark error message about local network blocking more useful and document it in README


tasn:
The error is currently:
> making requests to local IP addresses is forbidden and blocked

We should probably make it more useful like:
> Requests to this IP range is blocked (in the server configuration).

And update the README to have a section about SSRF, this, and how to enable it (a paragraph is enough).

PR #793, opened 2023-01-27
Clarify error message and documentation around filtered IP addresses

svix-daniel:
Solves #792 by clarifying the error message surrounding a filtered internal IP address, making it clear that this can be configured. Additionally this changes the  README to include a short paragraph on how to add subnets to the whitelist.

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,6 +238,8 @@ There are two configuration variables `db_pool_max_size` and `redis_pool_max_siz
 
 They default to a max size of 20, but higher values can significantly increase performance if your database can handle it.
 
+### SSRF Attacks and Internal IP Addresses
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -238,6 +238,8 @@ There are two configuration variables `db_pool_max_size` and `redis_pool_max_siz
 
 They default to a max size of 20, but higher values can significantly increase performance if your database can handle it.
 
+### SSRF Attacks and Internal IP Addresses
+To prevent SSRF attackks, message dispatches to internal IP addresses are blocked by default. However we undestand that this doesn't meet the needs of every user say, for example, the service can only be accessed internally. To bypass these restrictions, see the `whitelist_subnets` configuration option, which accepts an array of CIDR-notation subnets to allow messages to be dispatched to.
```
svix-gabriel, line 242:
```suggestion
To prevent SSRF attackks, message dispatches to internal IP addresses are blocked by default. However we understand that this doesn't meet the needs of every user say, for example, the service can only be accessed internally. To bypass these restrictions, see the `whitelist_subnets` configuration option, which accepts an array of CIDR-notation subnets to allow messages to be dispatched to.
```

svix-gabriel (approved):

PR #797, opened 2023-02-01
Change all path extractors to use structs

svix-andor:
This has 2 benefits:
1. It is a little safer because tuple params are extracted based simply on position so it's possible to make a mistake simply by putting the types in the wrong order. It still doesn't check that the path placeholders match the extractors at compile-time, though.
2. Aide can match field names to URL placeholder names this way, as a result including the path parameters in the OpenAPI params section.

### DIFF
--- a/server/svix-server/src/v1/endpoints/application.rs
+++ b/server/svix-server/src/v1/endpoints/application.rs
@@ -4,7 +4,7 @@
 use crate::{
     core::{
         permissions,
-        types::{metadata::Metadata, ApplicationId, ApplicationIdOrUid, ApplicationUid},
+        types::{metadata::Metadata, ApplicationId, ApplicationUid},
     },
     ctx,
     db::models::{application, applicationmetadata},
@@ -17,8 +17,8 @@ use crate::{
             UnrequiredNullableField,
         },
         validate_no_control_characters, validate_no_control_characters_unrequired,
-        validation_error, EmptyResponse, ListResponse, ModelIn, ModelOut, Pagination,
-        PaginationLimit, ValidatedJson, ValidatedQuery,
+        validation_error, ApplicationPath, EmptyResponse, ListResponse, ModelIn, ModelOut,
+        Pagination, PaginationLimit, ValidatedJson, ValidatedQuery,
     },
     AppState,
 };
@@ -263,7 +263,7 @@ async fn get_application(
 
 async fn update_application(
     State(AppState { ref db, .. }): State<AppState>,
-    Path(app_id): Path<ApplicationIdOrUid>,
+    Path(ApplicationPath { app_id }): Path<ApplicationPath>,
     permissions::Organization { org_id }: permissions::Organization,
     ValidatedJson(data): ValidatedJson<ApplicationIn>,
 ) -> Result<(StatusCode, Json<ApplicationOut>)> {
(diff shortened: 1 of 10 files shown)


server/svix-server/src/v1/endpoints/application.rs
```diff
@@ -39,6 +39,17 @@ use serde::{Deserialize, Serialize};
 use svix_server_derive::ModelOut;
 use validator::{Validate, ValidationError};
 
+#[derive(Deserialize, JsonSchema)]
+struct ApplicationIdOrUidParam {
```
tasn, line 43: Nit-pick: but no need to be too verbose here. The "IdOrUid" is implied because it's a path param and we always accept both.

`ApplicationPath` would be enough.
Then for endpoint:
`ApplicationEndpointPath`
Then for attempt:
`ApplicationMsgAttemptPath`

I think this is consistent and easy.

server/svix-server/src/v1/utils/mod.rs
```diff
@@ -462,6 +462,32 @@ pub fn get_unix_timestamp() -> u64 {
         .as_secs()
 }
 
+#[derive(Deserialize, JsonSchema)]
```
tasn, line 465: Copy over the ApplicationPath one here too if you have all the rest.

server/svix-server/src/v1/utils/mod.rs
```diff
@@ -462,6 +462,32 @@ pub fn get_unix_timestamp() -> u64 {
...
+pub struct AppEndpointParams {
```
tasn, line 466: Either "App" and "Endp" or "Application" and "Endpoint".
I think the latter, and I think we should use it consistently in all of them.

server/svix-server/src/v1/endpoints/application.rs
```diff
@@ -39,6 +39,17 @@ use serde::{Deserialize, Serialize};
...
+    app_id: ApplicationIdOrUid,
+}
+
+impl Into<ApplicationIdOrUid> for ApplicationIdOrUidParam {
```
tasn, line 47: Don't special case it, just do it the same as the rest...

server/svix-server/src/v1/utils/mod.rs
```diff
@@ -462,6 +465,42 @@ pub fn get_unix_timestamp() -> u64 {
...
+
+#[derive(Deserialize, JsonSchema)]
+pub struct ApplicationMsgAttemptPath {
+    pub app_id: ApplicationIdOrUid,
+    pub msg_id: MessageIdOrUid,
+    pub attempt_id: MessageAttemptId,
+}
+
+#[derive(Deserialize, JsonSchema)]
+pub struct ApplicationPath {
```
tasn, line 495: Nit-pick: put this first. We also go from "large" (most encompassing) to "small".

server/svix-server/src/v1/utils/mod.rs
```diff
@@ -462,6 +465,42 @@ pub fn get_unix_timestamp() -> u64 {
...
 }
 
+#[derive(Deserialize, JsonSchema)]
+pub struct ApplicationEndpointPath {
+    pub app_id: ApplicationIdOrUid,
+    pub endpoint_id: EndpointIdOrUid,
+}
+
+#[derive(Deserialize, JsonSchema)]
+pub struct ApplicationMsgPath {
```
tasn, line 475: I'm on the fence whether this should be `Message` or kept `Msg`. I think we call it `msg` everywhere, like really everywhere, so it's probably fine. Though it does stick out a bit. :)

svix-andor: Yep, I went with `Msg` because it's so widely used, will stick to that.

tasn (approved):
TBH, it looks much better than the tuples!

PR #803, opened 2023-02-03
Fix `/attempt/endpoint`'s broken `?channel=` query

svix-gabriel:
## Motivation

`/attempt/endpoint` is borked.

When `?channel=` is specified, an improper join happens which causes the Endpoint to fail with a 500 error.

Also, the actual query on a specific channel just straight up doesn't work. `channel ?? ?` is neither the proper operation, nor is `?` even used for params in postgres. You need a `$N`.

## Solution

Fix the damn queries! And update the tests so we catch this in the future.

### DIFF
--- a/server/svix-server/src/v1/endpoints/attempt.rs
+++ b/server/svix-server/src/v1/endpoints/attempt.rs
@@ -276,9 +276,9 @@ fn list_attempts_by_endpoint_or_message_filters(
     if event_types.is_some() || channel.is_some() {
         query = query.join_rev(
             sea_orm::JoinType::InnerJoin,
-            messageattempt::Entity::belongs_to(message::Entity)
-                .from(messageattempt::Column::MsgId)
-                .to(message::Column::Id)
+            message::Entity::belongs_to(messageattempt::Entity)
+                .from(message::Column::Id)
+                .to(messageattempt::Column::MsgId)
                 .into(),
         );
 
@@ -287,7 +287,11 @@ fn list_attempts_by_endpoint_or_message_filters(
         }
 
         if let Some(channel) = channel {
-            query = query.filter(Expr::cust_with_values("channels ?? ?", vec![channel]));
+            // sea_orm evaluates the '$1' relative to the # of params in `Expr::cust_with_values`,
+            // NOT relative to the total number of params in the final query like you might expect.
+            // As such, this won't break if more $N params are added in earler/later
+            // `.filter` calls.
+            query = query.filter(Expr::cust_with_values("channels @> $1", [channel.jsonb()]));
         }
     }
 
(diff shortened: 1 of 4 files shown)


server/svix-server/src/v1/endpoints/attempt.rs
```diff
@@ -287,7 +287,7 @@ fn list_attempts_by_endpoint_or_message_filters(
         }
 
         if let Some(channel) = channel {
-            query = query.filter(Expr::cust_with_values("channels ?? ?", vec![channel]));
+            query = query.filter(Expr::cust_with_values("channels @> $1", [channel.jsonb()]));
```
svix-andor, line 290: This is a little error-prone. `$1` is dependent on this particular fragment staying the first placeholder. Is there' no way to give SeaORM some kind of position-independent placeholder which it can replace with the correctly indexed one when it builds the query?

svix-gabriel: It's a bit deceptive, but AFAICT the variable substitution happens "eagerly", so the $N placeholder only needs to match the number of values passed into `Expr::cust_with_values`.

I tested this real quick by just adding another variable and printing the resulting sql
```rust
        if let Some(channel) = channel {
            query = query.filter(Expr::cust_with_values(
                "channels @> $1",
                [channel.clone().jsonb()],
            ));
            query = query.filter(Expr::cust_with_values("channels @> $1", [channel.jsonb()]));
        }

        tracing::warn!(
            "XXX {}",
            sea_orm::QueryTrait::build(&query, sea_orm::DbBackend::Postgres).to_string()
        )
```

```
XXX SELECT "messageattempt"."id", "messageattempt"."created_at", "messageattempt"."msg_id", "messageattempt"."msg_dest_id", "messageattempt"."endp_id", "messageattempt"."url", "messageattempt"."status", "messageattempt"."response_status_code", "messageattempt"."response", "messageattempt"."ended_at", "messageattempt"."trigger_type" FROM "messageattempt" INNER JOIN "message" ON "message"."id" = "messageattempt"."msg_id" WHERE "messageattempt"."endp_id" = 'ep_2LMjmmJ10MYEbAu7rafZuMkmud8' AND channels @> E'\"obits\"' AND channels @> E'\"obits\"''
```

The last bit `AND channels @> E'\"obits\"' AND channels @> E'\"obits\"'` being what matters. 

On the flip side, if you try and match the `$N` to the number of params in the final query, the query builder actually panics.

So this
```rust
            query = query.filter(Expr::cust_with_values(
                "channels @> $1",
                [channel.clone().jsonb()],
            ));
            query = query.filter(Expr::cust_with_values("channels @> $2", [channel.jsonb()]));
```

panics with
```
panicked at 'index out of bounds: the len is 1 but the index is 1'
```

I'll add a comment explaining this since it caught me off guard too.

svix-andor: Ahh, neat!

svix-andor (approved):

PR #810, opened 2023-02-06
Fix event_type array query parsing

svix-gabriel:
This fixes how `?event_types` query parameters are parsed in `*/attempt/endpoint/*`, in addition to some adjacent refactors.

## Motivation

`axum::Query` doesn't support URL query array parameter parsing very well. For example,
- `?a=1&a=2` gets rejected
- `?a[]=1&a[]=2` also gets rejected
- `?a[0]=1&a[1]=2` _also_ gets rejected

This is largely because `serde::urlencoded` doesn't support any of these formats. See https://github.com/nox/serde_urlencoded/issues/75

There is `serde_qs`, which supports the last two. But not the first format. See https://github.com/samscott89/serde_qs/issues/16

Unfortunately, we _need_ to support the first format because that is what our internal API has supported up to today.

However, our OSS has been inconsistent. Specifically with `?event_types`.

Most endpoints that took in `?event_types` used `MessageListFetchOptions`, which manually parsed `?event_types` only using the first format. Ideally, we'd like to support all 3.

In addition, `*/attempts/endpoint/` didn't use `MessageListFetchOptions`. Instead it relied on `axum::Query` deserialization, so `?event_types` outright didn't work as expected with this endpoint.

## Solution

This PR does a couple small things to fix these issues

1) `MessageListFetchOptions` is axed in favor of a more explicit `EventTypesQuery`. `MessageListFetchOptions` had an extra timestamp parameter, `before`, that is already parsed just fine by `axum::Query`.
2) `EventTypesQuery` is _similar_ to `MessageListFetchOptions`, in respect to how `event_types` are parsed. With a few exceptions
    * `EventTypesQuery` validates the input.
    * `EventTypesQuery` handles all 3 formats, not just the first one
    * `EventTypesQuery` should be a bit more performant, since it uses `form_urlencode`, which parses pairs as `Cow<str>`, so fewer allocations. (Note that `form_urlencode` is what `serde_urlencoded` uses under the hood)
3) Updates `*/attempts/endpoint/` to use `EventTypesQuery`, so that `?event_types` are parsed correctly. Tests are added to demonstrate and validate this behavior.

### DIFF
--- a/server/svix-server/tests/e2e_attempt.rs
+++ b/server/svix-server/tests/e2e_attempt.rs
@@ -262,6 +262,33 @@ async fn test_list_attempts_by_endpoint() {
         .unwrap();
     assert_eq!(obits_attempts.data.len(), 1);
 
+    let exploded_attempts: ListResponse<MessageAttemptOut> = client
+        .get(
+            &format!("api/v1/app/{app_id}/attempt/endpoint/{endp_id_2}/?event_types=user.exploded"),
+            StatusCode::OK,
+        )
+        .await
+        .unwrap();
+    assert_eq!(exploded_attempts.data.len(), 1);
+
+    let regular_attempts: ListResponse<MessageAttemptOut> = client
+        .get(
+            &format!("api/v1/app/{app_id}/attempt/endpoint/{endp_id_2}/?event_types[]=event.type"),
+            StatusCode::OK,
+        )
+        .await
+        .unwrap();
+    assert_eq!(regular_attempts.data.len(), 2);
+
+    let all_attempts: ListResponse<MessageAttemptOut> = client
+    .get(
+        &format!("api/v1/app/{app_id}/attempt/endpoint/{endp_id_2}/?event_types[0]=event.type&event_types[1]=user.exploded"),
+        StatusCode::OK,
+    )
+    .await
+    .unwrap();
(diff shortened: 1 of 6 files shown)


server/svix-server/tests/e2e_attempt.rs
```diff
@@ -262,6 +262,33 @@ async fn test_list_attempts_by_endpoint() {
...
+
+    let all_attempts: ListResponse<MessageAttemptOut> = client
+    .get(
+        &format!("api/v1/app/{app_id}/attempt/endpoint/{endp_id_2}/?event_types[0]=event.type&event_types[1]=user.exploded"),
+        StatusCode::OK,
+    )
+    .await
+    .unwrap();
+    assert_eq!(all_attempts.data.len(), 3);
+
```
svix-andor, line 291: Slightly nitpicky, but since the way it's implemented now also permits mixing and matching the styles could you add a test for that, too?

svix-gabriel: You mean like `?a=1&a[]=2`?

I know this implementation _technically_ allows that to work, but that's more for simplicity. I think the chance of a client mixing and matching like that is remote enough that I don't think we should commit to supporting that format.

svix-andor (approved):

PR #812, opened 2023-02-07
Implement bidirectional pagination for endpoints

svix-andor:
With this change `Endpoint` also returns a `prev_iterator` in its list response, and accepts such an iterator.

### DIFF
--- a/server/svix-server/src/v1/endpoints/endpoint/crud.rs
+++ b/server/svix-server/src/v1/endpoints/endpoint/crud.rs
@@ -5,7 +5,7 @@ use axum::{
     Json,
 };
 use hyper::StatusCode;
-use sea_orm::{entity::prelude::*, ActiveValue::Set, QueryOrder, TransactionTrait};
+use sea_orm::{entity::prelude::*, ActiveValue::Set, TransactionTrait};
 use sea_orm::{ActiveModelTrait, DatabaseConnection, QuerySelect};
 use url::Url;
 
@@ -21,9 +21,10 @@ use crate::{
     db::models::{application, endpoint, endpointmetadata, eventtype},
     error::{HttpError, Result, ValidationErrorItem},
     v1::utils::{
+        apply_pagination_asc,
         patch::{patch_field_non_nullable, UnrequiredField, UnrequiredNullableField},
         ApplicationEndpointPath, EmptyResponse, ListResponse, ModelIn, ModelOut, Pagination,
-        PaginationLimit, ValidatedJson, ValidatedQuery,
+        PaginationLimit, ReversibleIterator, ValidatedJson, ValidatedQuery,
     },
     AppState,
 };
@@ -31,19 +32,19 @@ use hack::EventTypeNameResult;
 
 pub(super) async fn list_endpoints(
     State(AppState { ref db, .. }): State<AppState>,
-    pagination: ValidatedQuery<Pagination<EndpointId>>,
+    ValidatedQuery(pagination): ValidatedQuery<Pagination<ReversibleIterator<EndpointId>>>,
     permissions::Application { app }: permissions::Application,
 ) -> Result<Json<ListResponse<EndpointOut>>> {
     let PaginationLimit(limit) = pagination.limit;
-    let iterator = pagination.iterator.clone();
+    let iterator = pagination.iterator;
+    let is_prev = matches!(iterator, Some(ReversibleIterator::Prev(_)));
 
-    let mut query = endpoint::Entity::secure_find(app.id)
-        .order_by_asc(endpoint::Column::Id)
-        .limit(limit + 1);
-
-    if let Some(iterator) = iterator {
-        query = query.filter(endpoint::Column::Id.gt(iterator))
-    }
+    let query = apply_pagination_asc(
+        endpoint::Entity::secure_find(app.id),
+        endpoint::Column::Id,
+        limit,
+        iterator,
+    );
 
     let results = ctx!(
         query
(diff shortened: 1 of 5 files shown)


server/svix-server/tests/e2e_endpoint.rs
```diff
@@ -649,6 +649,68 @@ async fn test_list() {
     .unwrap();
 }
 
+#[tokio::test]
+async fn test_endpoint_list_ordering() {
+    let (client, _jh) = start_svix_server().await;
+
+    let app_id = create_test_app(&client, "App1").await.unwrap().id;
+
+    for i in 0..12 {
```
tasn, line 658: No biggie, but 12 sounds like a large number (so slower tests), 5-6 would probably suffice (it adds up, and our tests are already too slow)

server/svix-server/src/v1/endpoints/endpoint/crud.rs
```diff
@@ -21,29 +25,39 @@ use crate::{
...
         ApplicationEndpointPath, EmptyResponse, ListResponse, ModelIn, ModelOut, Pagination,
-        PaginationLimit, ValidatedJson, ValidatedQuery,
+        PaginationLimit, ReversibleIterator, ValidatedJson, ValidatedQuery,
     },
     AppState,
 };
 use hack::EventTypeNameResult;
 
+#[derive(Deserialize, Validate, JsonSchema)]
+pub struct ListEndpointsQueryParameters {
```
tasn, line 38: Why this? We don't want this.

svix-andor: Only because other endpoints with pagination offered it too, but it is not necessary, dropping.

server/svix-server/tests/e2e_endpoint.rs
svix-andor: Reduced to 5.

server/svix-server/src/v1/endpoints/endpoint/crud.rs
tasn: We have it in the other ones because they are time-based, it's not the case for endpoints/apps (time based doesn't matter).

server/svix-server/src/v1/endpoints/endpoint/crud.rs
```diff
@@ -58,9 +59,14 @@ pub(super) async fn list_endpoints(
     })
     .collect();
 
-    Ok(Json(EndpointOut::list_response_no_prev(
+    if is_prev {
```
svix-gabriel, line 62: I saw the slack thread, so I know this is how we typically correct the ordering. But as you saw, it is a bit brittle since it's _very_ easy to forget.

I wonder if we could add this logic to `::list_response`. Just `reverse` in `list_response_inner`, so we're less likely to forget in the future.

svix-gabriel (dismissed):

server/svix-server/src/v1/endpoints/endpoint/crud.rs
tasn: I fully agree with this, and in fact, we've done similar things elsewhere. This is extremely brittle (as you said), and extremely easy to miss.

server/svix-server/src/v1/utils/mod.rs
```diff
@@ -175,6 +175,34 @@ pub fn apply_pagination<
     }
 }
 
+/// Applies sorting and filtration to a query from its iterator, sort column, and limit,
+/// in an ascending ordering, i.e. one where `Normal` corresponds to starting at the lowest
+/// and counting up, and `Prev` corresponds to starting at the highest and counting down.
+pub fn apply_pagination_reverse<
```
tasn, line 181: I'm confused about the name `reverse` given that it's asceding order by default?

svix-andor: There's an existing `apply_pagination` which is the exact opposite of this. I could rename that to `apply_pagination_desc` and this one `apply_pagination_asc`. WDYT?

tasn: Much better, yeah!

server/svix-server/src/v1/endpoints/endpoint/crud.rs
svix-andor: This can't be generically moved into `list_response` as is because there are other listing endpoints that use it that do ordering the opposite way by default (highest to lowest). At the very least we'd need to add an extra parameter to it to tell what the data looks like and/or add another method like `list_response_ascending` or something like that.

tasn: Yeah, split to two functions then, list_response_asceding and descending. It sounds reasonable, and makes it less error prone.

svix-andor: How does https://github.com/svix/svix-webhooks/pull/812/files#diff-9cb85d646fc32039b2a4538aae5cabf2da93cdd2ad85f826479937ac88206f09L231-R269 look?

server/svix-server/src/v1/utils/mod.rs
```diff
@@ -199,6 +227,7 @@ fn list_response_inner<T: ModelOut>(
     limit: usize,
     is_prev_iter: bool,
```
svix-gabriel, line 228: Ooooof, three `bool` params in a row is really to misuse.

I guess it's fine since the function is private, and the callers are right below. I'm wondering if there's a way to make the params clearer.

svix-gabriel (approved):

svix-gabriel (approved):

PR #815, opened 2023-02-07
Query parameter to list endpoints newest-to-oldest

svix-andor:
The overall pagination is still done first-to-last, but within a page results are listed last-to-first.

Builds on top of #812

### DIFF
--- a/server/svix-server/src/v1/endpoints/endpoint/crud.rs
+++ b/server/svix-server/src/v1/endpoints/endpoint/crud.rs
@@ -21,10 +21,10 @@ use crate::{
     db::models::{application, endpoint, endpointmetadata, eventtype},
     error::{HttpError, Result, ValidationErrorItem},
     v1::utils::{
-        apply_pagination_asc,
+        apply_pagination,
         patch::{patch_field_non_nullable, UnrequiredField, UnrequiredNullableField},
-        ApplicationEndpointPath, EmptyResponse, ListResponse, ModelIn, ModelOut, Pagination,
-        PaginationLimit, ReversibleIterator, ValidatedJson, ValidatedQuery,
+        ApplicationEndpointPath, EmptyResponse, ListOrdering, ListResponse, ModelIn, ModelOut,
+        Ordering, Pagination, PaginationLimit, ReversibleIterator, ValidatedJson, ValidatedQuery,
     },
     AppState,
 };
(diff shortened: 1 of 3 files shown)


server/svix-server/src/v1/endpoints/attempt.rs
```diff
@@ -18,10 +18,10 @@ use crate::{
     v1::{
```
tasn, line 18: A lot of unrelated stuff

server/svix-server/src/v1/endpoints/endpoint/crud.rs
```diff
@@ -58,9 +60,11 @@ pub(super) async fn list_endpoints(
     })
     .collect();
 
-    Ok(Json(EndpointOut::list_response_no_prev(
+    Ok(Json(EndpointOut::list_response_asc(
         results,
         limit as usize,
+        is_prev,
+        newest_first.unwrap_or(false),
```
tasn, line 67: Why not set a default in the newest_first structure directly and have it non-null?

server/svix-server/src/v1/endpoints/endpoint/crud.rs
```diff
@@ -21,29 +21,31 @@ use crate::{
...
     },
     AppState,
 };
 use hack::EventTypeNameResult;
 
 pub(super) async fn list_endpoints(
     State(AppState { ref db, .. }): State<AppState>,
-    pagination: ValidatedQuery<Pagination<EndpointId>>,
+    ValidatedQuery(pagination): ValidatedQuery<Pagination<ReversibleIterator<EndpointId>>>,
+    ValidatedQuery(NewestFirst { newest_first }): ValidatedQuery<NewestFirst>,
```
tasn, line 36: Also, I think we can probably come up with a better name. It was just a terrible place-holder I suggested. :P

server/svix-server/src/v1/endpoints/attempt.rs
svix-andor: unrelated stuff gone after rebase :)

server/svix-server/src/v1/endpoints/endpoint/crud.rs
svix-andor: This way the struct remains more generic and can be reused elsewhere where default `true` is needed, and number of lines-wise it ends up being more or less the same. But I don't feel super strong about this, if you think having a default on the struct itself is significantly better I'm ok with changing to that.

server/svix-server/src/v1/utils/mod.rs
```diff
@@ -288,7 +311,14 @@ pub trait ModelOut: Clone {
...
             IteratorDirection::Next
         };
-        list_response_inner(data, limit, ListOrdering::Descending, direction, true)
+        list_response_inner(
+            data,
+            limit,
+            ListOrdering::Descending,
+            direction,
+            true,
+            false,
```
tasn, line 320: 1. This is going out of control with this true, false, and a million params.
2. Luckily it's not necessarily. `newest_first` == order descending. You don't need another param, you already have it. If it's newest first, you just do descending order...

server/svix-server/src/v1/endpoints/endpoint/crud.rs
svix-andor: Renamed it to `order=ascending`/`order=descending`.

server/svix-server/src/v1/utils/mod.rs
svix-andor: I made `apply_pagination` and `list_response` generic again, they simply take the ordering as a parameter.

server/svix-server/src/v1/endpoints/endpoint/crud.rs
svix-andor: Resolving this since it's not a boolean anymore, therefore not relevant.

tasn (approved):
I think it's fine. Just making sure: you have the tests to ensure this works + tests to verify you haven't broken the ordering for anything else?

svix-andor:
I believe so. The again-generic `apply_pagination` and `listing_response` is only used for Endpoint listing, the remainders still use the direction specific variant (which I intend to remove once all have bidirectional listing).

PR #817, opened 2023-02-07
Fire operational webhook on endpoint secret rotation

svix-andor:
As per title, rotating the secrets of an endpoint is an update to the endpoint, therefore it should fire the corresponding operational webhook.

### DIFF
--- a/server/svix-server/src/v1/endpoints/endpoint/mod.rs
+++ b/server/svix-server/src/v1/endpoints/endpoint/mod.rs
@@ -374,7 +374,7 @@ impl From<(endpoint::Model, Metadata)> for EndpointOut {
     }
 }
 
-#[derive(Clone, Debug, PartialEq, Eq, Validate, Deserialize, JsonSchema)]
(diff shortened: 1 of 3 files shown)


server/svix-server/src/v1/endpoints/endpoint/mod.rs
```diff
@@ -374,7 +374,7 @@ impl From<(endpoint::Model, Metadata)> for EndpointOut {
     }
 }
 
-#[derive(Clone, Debug, PartialEq, Eq, Validate, Deserialize, JsonSchema)]
+#[derive(Default, Clone, Debug, PartialEq, Eq, Validate, Serialize, Deserialize, JsonSchema)]
```
tasn, line 377: Why is this needed?

tasn: Ah, nvm, just for the tests. :)

svix-gabriel (approved):

PR #826, opened 2023-02-09
Add support for ordering and prev_iter in app list

svix-andor:
It also extends the `common_test_list` function to test for reverse ordering, where supported.

### DIFF
--- a/server/svix-server/src/v1/endpoints/application.rs
+++ b/server/svix-server/src/v1/endpoints/application.rs
@@ -11,14 +11,14 @@ use crate::{
     error::{http_error_on_conflict, HttpError, Result},
     transaction,
     v1::utils::{
-        openapi_desc, openapi_tag,
+        apply_pagination, openapi_desc, openapi_tag,
         patch::{
             patch_field_non_nullable, patch_field_nullable, UnrequiredField,
             UnrequiredNullableField,
         },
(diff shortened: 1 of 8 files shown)


server/svix-server/src/v1/endpoints/application.rs
```diff
@@ -193,20 +194,40 @@ const LIST_APPLICATIONS_DESCRIPTION: &str = "List all of the organization's appl
 
 async fn list_applications(
     State(AppState { ref db, .. }): State<AppState>,
-    pagination: ValidatedQuery<Pagination<ApplicationId>>,
+    ValidatedQuery(pagination): ValidatedQuery<Pagination<ReversibleIterator<ApplicationId>>>,
+    ValidatedQuery(Ordering { order }): ValidatedQuery<Ordering>,
```
svix-gabriel, line 198: Why isn't `order` just another field on `pagination`? It seems like they'll always go together, no?

svix-andor: 1. While a niche use-case I think ordering can have uses for non-paginated results too.
2. I can remove it if you think the above is too unlikely to happen. However, since `Ordering` isn't being added in this PR, it's only a usage of it, is it ok if we get this merged and I can merge it with Pagination in a separate one?

svix-gabriel (approved):

PR #833, opened 2023-02-09
Add new event type and endpoint params to libs

svix-andor:
These were missed when the OpenAPI spec was updated, so fixing that here.

### DIFF
--- a/csharp/Svix/Endpoint.cs
+++ b/csharp/Svix/Endpoint.cs
@@ -260,6 +260,7 @@ namespace Svix
                     appId,
                     options?.Iterator,
                     options?.Limit,
+                    options?.Order,
                     idempotencyKey);
 
                 return lEndpoints?.Data;
@@ -284,6 +285,7 @@ namespace Svix
                     appId,
                     options?.Iterator,
                     options?.Limit,
+                    options?.Order,
                     idempotencyKey,
                     cancellationToken);
 
--- a/csharp/Svix/EventType.cs
+++ b/csharp/Svix/EventType.cs
@@ -28,6 +28,7 @@ namespace Svix
             {
                 var lResponse = _eventTypeApi.DeleteEventTypeApiV1EventTypeEventTypeNameDeleteWithHttpInfo(
                     eventType,
+                    null,
                     idempotencyKey);
 
                 return lResponse.StatusCode == HttpStatusCode.NoContent;
@@ -49,6 +50,7 @@ namespace Svix
             {
                 var lResponse = await _eventTypeApi.DeleteEventTypeApiV1EventTypeEventTypeNameDeleteWithHttpInfoAsync(
                     eventType,
+                    null,
                     idempotencyKey,
                     cancellationToken);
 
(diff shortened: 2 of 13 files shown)


csharp/Svix/Endpoint.cs
```diff
@@ -252,14 +252,15 @@ public string GetSecret(string appId, string endpointId, string idempotencyKey =
             }
         }
 
-        public List<EndpointOut> List(string appId, ListOptions options = null, string idempotencyKey = default)
+        public List<EndpointOut> List(string appId, Ordering order = default, ListOptions options = null, string idempotencyKey = default)
```
tasn, line 255: 1. Doesn't this break things if you put them before other params?
2. Also, it should be part of ListOptions, not a different param.

csharp/Svix/EventType.cs
```diff
@@ -43,12 +44,13 @@ public bool Archive(string eventType, string idempotencyKey = default)
             }
         }
 
-        public async Task<bool> ArchiveAsync(string eventType, string idempotencyKey = default, CancellationToken cancellationToken = default)
+        public async Task<bool> ArchiveAsync(string eventType, bool? expunge = null, string idempotencyKey = default, CancellationToken cancellationToken = default)
```
tasn, line 47: Same with expunge, I'm thinking it should probably be part of "Options" rather than its own param?

java/lib/src/main/java/com/svix/Endpoint.java
```diff
@@ -25,8 +26,12 @@ public Endpoint() {
 	}
 
 	public ListResponseEndpointOut list(final String appId, final EndpointListOptions options) throws ApiException {
+		return list(appId, null, options);
+	}
+
+	public ListResponseEndpointOut list(final String appId, final Ordering order, final EndpointListOptions options) throws ApiException {
```
tasn, line 32: I'll stop commenting, but Ordering should be part of options.

csharp/Svix/Endpoint.cs
svix-andor: Updated, now ordering is part of `ListOptions` in all libs. Ruby does not appear in the list because it's all implicitly done in it.

csharp/Svix/EventType.cs
svix-andor: There's no `delete`/`archive` options yet. We could add one but it'd be changing the signature either way. Also, I feel like this is distinct enough where its own parameter is warranted.

csharp/Svix/Abstractions/IEventType.cs
```diff
@@ -8,9 +8,9 @@ namespace Svix.Abstractions
 {
     public interface IEventType
     {
-        bool Archive(string eventType, string idempotencyKey = default);
+        bool Archive(string eventType, bool? expunge = null, string idempotencyKey = default);
 
-        Task<bool> ArchiveAsync(string eventType, string idempotencyKey = default,
+        Task<bool> ArchiveAsync(string eventType, bool? expunge = null, string idempotencyKey = default,
```
tasn, line 13: This breaks API, let's not do this for row.

csharp/Svix/EventType.cs
tasn: Yeah, but then it means breaking the API twice!

Anyhow, it should probably just be its own function anyway, so you'd have `archive`, and you'd have `expunge` as a standalone function (I now remembered that we talked about it in the original PR).

svix-andor: Changed, just passing `null` everywhere for now.

tasn (approved):

PR #835, opened 2023-02-09
Alter DbErr Traceable impl to make conversions explicit

svix-daniel:
These changes alter the `Traceable` implementation for `DbErr` to only return `Error::database`s instead of having logic which implicitly converts these errors to `HttpError`s depending on the error type.

Instead the relevant conversion is performed explicitly via the new utility function `http_error_on_conflict`. Note that 404 responses did not need to be explicitly checked as every instance where a 404 should be returned has logic that handles `None`  responses.

### DIFF
--- a/server/svix-server/src/error.rs
+++ b/server/svix-server/src/error.rs
@@ -21,7 +21,6 @@ use sea_orm::RuntimeErr;
 use sea_orm::TransactionError;
 use serde::Serialize;
 use serde_json::json;
-use sqlx::Error as SqlxError;
 
 use crate::core::webhook_http_client;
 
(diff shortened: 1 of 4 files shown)


server/svix-server/src/error.rs
```diff
@@ -455,3 +427,22 @@ impl From<crate::core::webhook_http_client::Error> for Error {
...
+            if e.as_database_error()
+                .and_then(|e| e.code())
+                .filter(|code| code == "23505")
+                .is_some() =>
+        {
+            HttpError::conflict(None, None).into()
+        }
+        // This always inserts this line's locaton, but because it should be wrapped by a [`ctx`]
+        // call, you should still get a meaningful trace
+        e => Error::database(e, location!()),
```
svix-gabriel, line 446: Maybe it's better use a `""` instead of `location!`? I'm just thinking from a debugging perspective, for something like

```rust
ctx!(app.insert(txn).await.map_err(http_error_on_conflict))?
```

The line # in `error.rs` might just be noise.

svix-gabriel (approved):

svix-gabriel (approved):

PR #836, opened 2023-02-10
Add query flag to insert evt type into payload

svix-andor:
Some users use one endpoint to consume different kinds of event types. In these cases it can be useful to have a discriminant in the payload itself to determine the type of event being sent. This PR adds a flag which automates this.

The flag defaults to false (don't insert). If set, it checks if the payload is a JSON object. If not then there's nothing sensible to do. If it is, it checks if there's already a field with the name `"event"`. If yes, nothing is done. If no, it creates a new field with this name and sets its value to the event type name.

### DIFF
--- a/server/svix-server/src/v1/endpoints/message.rs
+++ b/server/svix-server/src/v1/endpoints/message.rs
@@ -225,6 +225,7 @@ async fn list_messages(
 pub struct CreateMessageQueryParams {
     #[serde(default = "default_true")]
     with_content: bool,
+    insert_event_type: Option<bool>,
 }
 
 const CREATE_MESSAGE_DESCRIPTION: &str = r#"
@@ -245,11 +246,12 @@ async fn create_message(
         cache,
         ..
     }): State<AppState>,
-    ValidatedQuery(CreateMessageQueryParams { with_content }): ValidatedQuery<
-        CreateMessageQueryParams,
-    >,
+    ValidatedQuery(CreateMessageQueryParams {
+        with_content,
+        insert_event_type,
+    }): ValidatedQuery<CreateMessageQueryParams>,
     permissions::OrganizationWithApplication { app }: permissions::OrganizationWithApplication,
-    ValidatedJson(data): ValidatedJson<MessageIn>,
+    ValidatedJson(mut data): ValidatedJson<MessageIn>,
 ) -> Result<(StatusCode, Json<MessageOut>)> {
     let create_message_app = CreateMessageApp::layered_fetch(
         &cache,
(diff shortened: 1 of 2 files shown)


server/svix-server/src/v1/endpoints/message.rs
```diff
@@ -225,6 +225,7 @@ async fn list_messages(
 pub struct CreateMessageQueryParams {
     #[serde(default = "default_true")]
     with_content: bool,
+    insert_event_type: Option<bool>,
```
tasn, line 228: We need a better name probably. Dunno what it is though. :)

server/svix-server/src/v1/endpoints/message.rs
```diff
@@ -263,6 +265,14 @@ async fn create_message(
     // Should never happen since you're giving it an existing Application, but just in case
     .ok_or_else(|| err_generic!("Application doesn't exist: {}", app.id))?;
 
+    if insert_event_type.unwrap_or(false) {
+        if let serde_json::Value::Object(ref mut obj) = data.payload {
+            let value = serde_json::to_value(data.event_type.clone())
```
tasn, line 270: This looks expensive, what doss it do?

svix-andor: Ahh, I'm silly, I can just construct the `serde_json::Value::String` directly, good catch.

server/svix-server/src/v1/endpoints/message.rs
svix-andor: No strong opinion on this one. `insert_event_type` is a bit of a mouthful but I think it's pretty clear what it does.

svix-gabriel (dismissed):

server/svix-server/src/v1/endpoints/message.rs
```diff
@@ -263,6 +265,13 @@ async fn create_message(
     // Should never happen since you're giving it an existing Application, but just in case
     .ok_or_else(|| err_generic!("Application doesn't exist: {}", app.id))?;
 
+    if insert_event_type.unwrap_or(false) {
+        if let serde_json::Value::Object(ref mut obj) = data.payload {
```
svix-andor, line 269: @tasn raised a good point on slack: the payload is enforced to be a JSON object elsewhere in code, but since rust doesn't allow using enum variants as types themselves it must be a `Value`. Question: should I just do an `as_object().unwrap()` here since it's guaranteed to be an object? Or is the current "silently do nothing if this invariant is violated" preferred? Or augment the latter by logging in case the invariant is broken?

svix-gabriel: I actually like what you did. There's no performance difference, but if we ever relaxed the requirements for `data.payload`, it's hard to audit where we made the assumption that it's always an object.

tasn: Yup, I'm with @svix-gabriel. Though maybe add an error log for the `else` clause? Just so we know if it's ever not the case?

svix-andor: Updated with a log message.

server/svix-server/src/v1/endpoints/message.rs
tasn: `inject_event_type`?

svix-andor: Renamed to `inject_event_type`.

PR #837, opened 2023-02-10
Add prev_iter support for event type list

svix-andor:
Two noteworthy changes:
* Introduce a new `IdIterator` trait to represent all identifier-like things we use for iteration in the API. These are generally the `BaseId` types but there are some exceptions, like `EventTypeName` in this change's case.
* Move `From<String>` impl from the `create_id_type` macro to the `string_wrapper_impl` macro since all string wrappers will want this, and it's needed for `EventTypeName` to function as a `ReversibleIterator`.

### DIFF
--- a/server/svix-server/src/v1/utils/mod.rs
+++ b/server/svix-server/src/v1/utils/mod.rs
@@ -179,10 +179,17 @@ pub fn apply_pagination_desc<
     }
 }
 
+/// Marker trait for any type that is used for iterating through results
+/// in the public API.
+pub trait IdIterator: Validate + Into<sea_orm::Value> {}
+
+impl<T: BaseId + Validate + Into<sea_orm::Value>> IdIterator for T {}
+impl IdIterator for EventTypeName {}
+
 pub fn apply_pagination<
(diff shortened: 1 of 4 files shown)


server/svix-server/src/v1/utils/mod.rs
```diff
@@ -179,10 +179,17 @@ pub fn apply_pagination_desc<
     }
 }
 
+/// Marker trait for any type that is used for iterating through results
+/// in the public API.
+pub trait IdIterator: Validate + Into<sea_orm::Value> {}
```
svix-gabriel, line 184: I know you copied this from what was here before, but does this actually need to implement `Validate`?

svix-andor: It does insofar as `ReversibleIterator` does. Because this is used as the type param to `ReversibleIterator` in `apply_pagination`. So if we're ok dropping it from that, then it can be dropped here.

svix-gabriel: Hmmm that's annoying.

svix-gabriel (approved):

Issue #819, opened 2023-02-08
ListAttemptsByMessageAsync throws error


Obl3n:
### Version
0.77.0

### Platform
Windows

### Description

Using C# Client 

Svix 0.77.0

var msg= await svix.MessageAttempt.ListAttemptsByMessageAsync("app_id", "msg_id");
throws
``````
System.InvalidOperationException: Nullable object must have a value.
   at System.Nullable`1.get_Value()
   at Svix.MessageAttempt.ListAttemptsByMessageAsync(String appId, String messageId, AttemptsByMessageListOptions options, String idempotencyKey, CancellationToken cancellationToken)
``````

``````Nullable object must have a value.``````

**Comment by Obl3n** [comment]
```svix``` is not null and other methods are working fully fine.

**Comment by tasn** [comment]
Can you find the exact line? I just looked at the code and it looks fine to me!

**Comment by Obl3n** [comment]
Hey, made some debugging and it seems like it only works if AttemptsByMessageListOptions passed with not null values Status,Code.

var msg= await svix.MessageAttempt.ListAttemptsByMessageAsync("app_id", "msg_id", new AttemptsByMessageListOptions
{
    Limit = 50,
    Status = 3,
    Code = 200
});

But how to make it that it fetchs all the attempts no matter status and code?

**Comment by Obl3n** [comment]
```
  public async Task<List<MessageAttemptOut>> ListAttemptsByMessageAsync(string appId, string messageId, AttemptsByMessageListOptions options = null,
            string idempotencyKey = default, CancellationToken cancellationToken = default)
        {
            try
            {
                var lResults = await _messageAttemptApi.ListAttemptsByMsgApiV1AppAppIdAttemptMsgMsgIdGetAsync(
                    appId,
                    messageId,
                    options?.EndpointId,
                    options?.Iterator,
                    options?.Limit,
                    (Svix.Model.MessageStatus)options?.Status,
                    (Svix.Model.StatusCodeClass)options?.Code,
                    options?.EventTypes,
                    options?.Channel,
                    options?.Before,
                    options?.After,
                    idempotencyKey,
                    cancellationToken);

                return lResults?.Data;
            }
            catch (ApiException e)
            {
                Logger?.LogError(e, $"{nameof(ListAttemptsByMessageAsync)} failed");

                if (Throw)
                    throw;

                return new List<MessageAttemptOut>();
            }
        }

```

Probably because it casts  like this
(Svix.Model.MessageStatus)options?.Status,
[...]

PR #841, opened 2023-02-11
Update openapi.json

Obl3n:
## Motivation

This should solve #819 problem
## Solution

Making nullable MessageStatuse and StatusCodeClass.

### DIFF
--- a/openapi.json
+++ b/openapi.json
@@ -2734,6 +2734,7 @@
                     2,
                     3
                 ],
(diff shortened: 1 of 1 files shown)


tasn:
The OpenAPI spec is generated automatically, so you can't just update it manually unfortunately.

Moreso, the OpenAPI spec is correct. The value is not nullable, it's just not required. If passed, it shouldn't be null.

PR #845, opened 2023-02-13
Add flag to ignore missing app on msg creation

svix-andor:
Some users use Svix in a stateless mode where they do not know if an app (consumer) exists or not when they send a message. These users just want to ignore the error when an app does not exist when a message is sent.

This change adds a new query parameter `ignore_missing_app`, which can be set to `true` to return 200 OK on `create_message` when the app does not exist.

### DIFF
--- a/server/svix-server/src/v1/endpoints/message.rs
+++ b/server/svix-server/src/v1/endpoints/message.rs
@@ -10,22 +10,28 @@ use crate::{
             MessageAttemptTriggerType, MessageId, MessageUid,
         },
     },
-    ctx, err_generic,
+    ctx,
+    db::models::application,
+    err_generic,
     error::{HttpError, Result},
     queue::MessageTaskBatch,
     v1::utils::{
-        apply_pagination_desc, iterator_from_before_or_after, openapi_tag, validation_error,
-        ApplicationMsgPath, EventTypesQuery, ListResponse, ModelIn, ModelOut, PaginationLimit,
-        ReversibleIterator, ValidatedJson, ValidatedQuery,
+        apply_pagination_desc, iterator_from_before_or_after, openapi_tag,
+        validation_error, ApplicationMsgPath, ApplicationPath, EventTypesQuery, ListResponse,
+        ModelIn, ModelOut, PaginationLimit, ReversibleIterator, ValidatedJson, ValidatedQuery,
     },
     AppState,
 };
-use aide::axum::{
-    routing::{delete_with, get_with, post_with},
-    ApiRouter,
+use aide::{
+    axum::{
+        routing::{delete_with, get_with, post_with},
+        ApiRouter,
+    },
+    transform::TransformOperation,
 };
 use axum::{
     extract::{Path, State},
+    response::IntoResponse,
     Json,
 };
 use chrono::{DateTime, Duration, Utc};
@@ -224,6 +230,9 @@ async fn list_messages(
 pub struct CreateMessageQueryParams {
     #[serde(default = "default_true")]
     with_content: bool,
+    /// If set to true a 200 OK is returned instead of 404 when the
+    /// application specified in the request does not exist.
+    ignore_missing_app: Option<bool>,
 }
 
 /// Creates a new message and dispatches it to all of the application's endpoints.
@@ -243,12 +252,28 @@ async fn create_message(
         cache,
         ..
     }): State<AppState>,
-    ValidatedQuery(CreateMessageQueryParams { with_content }): ValidatedQuery<
-        CreateMessageQueryParams,
-    >,
-    permissions::OrganizationWithApplication { app }: permissions::OrganizationWithApplication,
+    Path(ApplicationPath { app_id }): Path<ApplicationPath>,
+    ValidatedQuery(CreateMessageQueryParams {
+        with_content,
+        ignore_missing_app,
+    }): ValidatedQuery<CreateMessageQueryParams>,
+    permissions::Organization { org_id }: permissions::Organization,
     ValidatedJson(data): ValidatedJson<MessageIn>,
-) -> Result<(StatusCode, Json<MessageOut>)> {
+) -> Result<(StatusCode, axum::response::Response)> {
+    let app = match ctx!(
+        application::Entity::secure_find_by_id_or_uid(org_id, app_id.to_owned())
+            .one(db)
+            .await
+    )? {
+        Some(app) => app,
+        None if ignore_missing_app.unwrap_or(false) => {
+            return Ok((StatusCode::OK, ().into_response()));
+        }
+        None => {
+            return Err(HttpError::not_found(None, Some("Application not found".into())).into())
+        }
+    };
+
     let create_message_app = CreateMessageApp::layered_fetch(
         &cache,
         db,
@@ -291,7 +316,7 @@ async fn create_message(
         MessageOut::without_payload(msg)
     };
 
-    Ok((StatusCode::ACCEPTED, Json(msg_out)))
+    Ok((StatusCode::ACCEPTED, Json(msg_out).into_response()))
 }
 
 #[derive(Debug, Deserialize, Validate, JsonSchema)]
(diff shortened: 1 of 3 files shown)


server/svix-server/src/v1/endpoints/message.rs
```diff
@@ -225,6 +231,9 @@ async fn list_messages(
 pub struct CreateMessageQueryParams {
     #[serde(default = "default_true")]
     with_content: bool,
+    /// If set to true a 200 OK is returned instead of 404 when the
+    /// application specified in the request does not exist.
+    ignore_missing_app: Option<bool>,
```
tasn, line 236: Sorry to bikeshed, but this is public API. Do we have a better name in mind?

server/svix-server/src/v1/endpoints/message.rs
```diff
@@ -351,10 +376,17 @@ async fn expunge_message_content(
 
 pub fn router() -> ApiRouter<AppState> {
     let tag = openapi_tag("Message");
+
+    fn document_create_message(op: TransformOperation) -> TransformOperation {
+        op.response::<202, Json<MessageOut>>()
+            .response::<200, ()>()
```
tasn, line 382:
```suggestion
            .response::<200, Json<()>>()
```
No?

server/svix-server/src/v1/endpoints/message.rs
```diff
@@ -245,12 +254,28 @@ async fn create_message(
...
-    permissions::OrganizationWithApplication { app }: permissions::OrganizationWithApplication,
+    Path(ApplicationPath { app_id }): Path<ApplicationPath>,
+    ValidatedQuery(CreateMessageQueryParams {
+        with_content,
+        ignore_missing_app,
+    }): ValidatedQuery<CreateMessageQueryParams>,
+    permissions::Organization { org_id }: permissions::Organization,
     ValidatedJson(data): ValidatedJson<MessageIn>,
-) -> Result<(StatusCode, Json<MessageOut>)> {
+) -> Result<(StatusCode, axum::response::Response)> {
```
tasn, line 264: I think it would be cleaner to have it as an enum rather than either? (assuming enum works with axum)
So an enum for Ok and Accepted and then strict typing inside?

server/svix-server/src/v1/endpoints/message.rs
svix-andor: That depends on what you want. `()` is a response with no body at all and no `Content-Type` either. `Json<()>` will have `Content-Type: application/json`, but I'm not sure what `()` is in JSON, it doesn't really have a meaningful equivalent. I went with `()` because that's really _nothing at all_.

We can also do `Json<EmptyResponse>` which I think we currently serialize as `{}`.

server/svix-server/src/v1/endpoints/message.rs
svix-andor: Enum is an option, axum doesn't really care either way, as long as it's serializable JSON it's all the same. It does not make a difference to aide either, it can't infer that one status code is one variant of the enum and the other another, it has to be manually told with the operation transformation (the other thing you commented on).

Using `Response` + `into_response()` is their recommendation: https://docs.rs/axum/latest/axum/response/index.html#returning-different-response-types

Do you think the small added extra type safety is worth it for the enum?

server/svix-server/src/v1/endpoints/message.rs
svix-andor: Ugh, `()` is serialized as `null`: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a77a840624e3b8377952dbfc1982ff81

So, you probably don't want `Json<()>`

server/svix-server/src/v1/endpoints/message.rs
svix-andor: As usual, I went with the most in-your-face thing I could think of. I think it's fairly clear, if perhaps a bit long. If you want something shorter maybe `stateless=true`? I'm afraid that's a bit too opaque/vague though.

svix-gabriel: FWIW `ignore_missing_app` seems pretty intuitive to me as well.

server/svix-server/src/v1/endpoints/message.rs
```diff
@@ -245,12 +254,28 @@ async fn create_message(
...
+    let app = match ctx!(
```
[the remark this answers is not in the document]
svix-gabriel: Nevermind, ignore me. I read too quickly.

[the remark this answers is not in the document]
svix-gabriel: Just to clarify, I missed that we were already querying the app in `permissions::OrganizationWithApplication`. So I think what you did is fine.

It's still _probably_ better to rely on the lookup from `layered_fetch`, you'll just have to change it to accept an `ApplicationIdOrUid` instead of an `ApplicationId`. Maybe not too bad of a change but I also understand if you don't want to do that since it's out of scope.

server/svix-server/src/v1/endpoints/message.rs
tasn: Json<EmptyResponse> sounds right

server/svix-server/src/v1/endpoints/message.rs
tasn: I'd think so, yeah. It's not a lot of work, and we want to have type checking where we can.

server/svix-server/src/v1/endpoints/message.rs
tasn: It just feels a bit JAVA to me "IgnoreMissingAppAndReturn404IfNotExist" <-- I wonder if there's a more generic and short name for this.

server/svix-server/src/v1/endpoints/message.rs
svix-andor: I guess by that you mean `Json<EmptyResponse>` right?

svix-andor: Changed response to `Json<EmptyResponse>`

server/svix-server/src/v1/endpoints/message.rs
svix-andor: A day later, I still think `ignore_missing_app` is good, I can't think of better alternatives.

tasn: Still not a fan. "missing" doesn't really mean what we want, and "ignore" also doesn't fully capture what we are after.

* Maybe `skip_missing_app` feels like it's maybe better, but still feels lacking.
* `skip_no_app`? too vague/broken English
* `skip_nonexistent_app`? too long
* `skip_absent_app`? blah.

Maybe `ignore_missing_app` is the best we have. Any other thoughts? If no, I guess we can go ahead. :|

server/svix-server/src/v1/endpoints/message.rs
```diff
@@ -347,10 +386,20 @@ async fn expunge_message_content(
 
 pub fn router() -> ApiRouter<AppState> {
     let tag = openapi_tag("Message");
+
+    fn document_create_message(op: TransformOperation) -> TransformOperation {
```
tasn, line 390: This PR can go in, though there's one question to answer before we can (which I think I asked already? If so, sorry for repeating myself!): how do the auto-generated libs deal with this new multi-status point API? E.g. what does the generated Rust lib does? TypeScript? Python? Java?

svix-andor: Hmm, it's actually not great 😞 
I tried regenerating after adding a response code `200` response to this operation, with no response body. I checked two languages:
1. In Go, it removed the automatic deserialization from the response, so the generated method just returns the http response to you and it's up to you to deserialize the correct response.
2. In Rust it's even worse because it seems if there's a response with code 200 it prefers that and disregards other successful response codes. In this case it changed it from `Result<MessageOut, SomeErrorType>` to `Result<(), SomeErrorType>`.

svix-andor: I scoured the OpenAPI generator issue tracker a little bit and it seems that support for this varies language by language:
* https://github.com/OpenAPITools/openapi-generator/issues/7714 - Ruby is wrong
* https://github.com/OpenAPITools/openapi-generator/issues/7426 - Python might be correct, or at least the issue is marked resolved
* In the mean time I also checked Java locally, which is also wrong.

So yeah, we might not want to add another response code, rather add a different response type maybe? 😕

svix-andor:
Closing this as it was superseded by another solution.

PR #848, opened 2023-02-13
Fix 5xx responses during concurrent UID updates

svix-daniel:
Followup to #846 which identifies and fixes the same problem in other parts of the code. In brief, with unluckily timed concurrent PUT/PATCH requests to the same object, the checks for UID conflicts done during the start of the request would become stale as one request processed. This leads to a specific `DbErr` which the changes now manually map to a HTTP CONFLICT response.

### DIFF
--- a/server/svix-server/src/v1/endpoints/application.rs
+++ b/server/svix-server/src/v1/endpoints/application.rs
@@ -341,7 +341,7 @@ async fn patch_application(
     let (app, metadata) = model;
 
     let (app, metadata) = transaction!(db, |txn| async move {
-        let app = ctx!(app.update(txn).await)?;
+        let app = ctx!(app.update(txn).await.map_err(http_error_on_conflict))?;
         let metadata = ctx!(metadata.upsert_or_delete(txn).await)?;
         Ok((app, metadata))
     })?;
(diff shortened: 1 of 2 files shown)


svix-gabriel (approved):

server/svix-server/src/v1/endpoints/endpoint/crud.rs
```diff
@@ -87,7 +87,7 @@ async fn create_endp_from_data(
         let txn = ctx!(db.begin().await)?;
```
tasn, line 87: Why does app (above) uses the `transaction!` macro and this one doesn't?

Also, does it happen at the transaction commit, or just when you insert inside of the transaction?

Also also, if it's at the end of the transaction then I think the above (app) is still broken? If it's not, then this change below is superfluous?

svix-daniel: I'll check whether the error happens on insert or on committing the transaction

svix-daniel: It happens inside the transaction, not on commit. I removed what was unnecessary.

svix-gabriel (approved):

Issue #717, opened 2022-12-03
Server: add OpenAPI generation


tasn:
Alternatives:
- https://github.com/tamasfe/aide
- https://github.com/juhaku/utoipa

I think `aide` looks more promising, but haven't given either a deep enough look.

svix-andor:
Just a quick note on where this is: `aide` was integrated into svix-webhooks, so it generates a crude spec, it is missing the following still:
* [ ] Operation descriptions - #811 
* [ ] Operation summaries
* [ ] Operation IDs
* [ ] Code examples

svix-andor:
@tasn the Python API automatically adds all of the HTTP error responses used in the API to all operations. There's no easy way to do this with aide, only by manually iterating over all operations once the API docs have been constructed and then manually adding each error response to all operations. Is this something that must be retained?

tasn:
Can't it be part of the macro you added and just add all of them (or only a subset if defined)?

I know OpenAPI also supports a "default error", which essentially shows the structure of the error struct without specifying specific codes. To be honest, this is probably preferable if can be easily done.

tasn:
Done.

PR #849, opened 2023-02-15
Add attr macro to generate operation transform

svix-andor:
This macro can be used to annotate axum handler functions to automatically generate an operation transformation that adds an operation ID, summary, and description.

Here's how to use it:

```rust
/// Doc comment
#[aide_annotate]
async fn some_handler(/* some params */) -> impl IntoResponse {

}
```

This generates a new function `some_handler_operation` which has the signature `fn(TransformOperation) -> TransformOperation`, i.e. what's needed for aide. This function sets the following in OpenAPI:
1. Operation ID becomes the name of the function, or can be optionally overridden with an `#[aide_annotate(op_id = "foobar")]`.
2. Operation summary becomes the title cased name of the function, e.g. `create_message` becomes `Create Message`. This can also be overridden, with `op_summary = "summary here"`.
3. Operation description becomes the doc comment of the function. This macro makes setting a doc comment mandatory.

Part of #717

### DIFF
--- a/server/svix-server/src/v1/endpoints/application.rs
+++ b/server/svix-server/src/v1/endpoints/application.rs
@@ -11,7 +11,7 @@ use crate::{
     error::{http_error_on_conflict, HttpError, Result},
     transaction,
     v1::utils::{
-        apply_pagination, openapi_desc, openapi_tag,
+        apply_pagination, openapi_tag,
         patch::{
             patch_field_non_nullable, patch_field_nullable, UnrequiredField,
             UnrequiredNullableField,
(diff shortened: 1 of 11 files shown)


server/svix-server/src/v1/endpoints/application.rs
```diff
@@ -349,8 +349,8 @@ async fn patch_application(
     Ok(Json((app, metadata).into()))
 }
 
-const DELETE_APPLICATION_DESCRIPTION: &str = "Delete an application.";
-
+/// Delete an application.
+#[aide_annotate(op_id = "delete_application_api_v1_app__app_id___delete")]
```
tasn, line 353: So it takes the description it from the comment?

Is there a way to automatically generate the op_id? E.g. if you do it on the router itself?

I'd much rather have something that's automatically generated than something we need to manually update like this. Even at the expense of compatibility with what we have in the op_id.

svix-andor: There is a default, it picks the name of the function (and just that, nothing else) if you don't specify one. But I also put in an override as seen here, so we can use the existing operation IDs for operations that already exist. There are a few operations in rust which don't exist in the OpenAPI (some `PATCH` ones), you can see those don't set an `op_id` by hand.

tasn: OK, we can always worry about this later if we have other thoughts/concerns.

svix-gabriel (approved):
The actual proc_macro implementation is all greek to me. But the function annotation seems pretty intuitive to use.

PR #853, opened 2023-02-16
Install OpenSSL in Windows pipeline

svix-andor:
Let's see if this is enough to get Windows CI working again

### DIFF
--- a/.github/workflows/server-release.yml
+++ b/.github/workflows/server-release.yml
@@ -1,8 +1,9 @@
 name: Server Release
 
 on:
-  release:
-    types: [published]
+  push:
+    branches:
+      - main
 
 jobs:
   release:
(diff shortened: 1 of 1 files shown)


.github/workflows/server-release.yml
```diff
@@ -42,6 +42,10 @@ jobs:
         if: ${{ matrix.target == 'x86_64-unknown-linux-musl' }}
         run: sudo apt-get install -y musl-dev musl-tools
 
+      - name: Install OpenSSL for Windows
+        if: ${{ matrix.target == 'x86_64-pc-windows-gnu' }}
```
tasn, line 46: Are we using windows-gnu?

svix-andor: https://github.com/svix/svix-webhooks/blob/main/.github/workflows/server-release.yml#L15
yep

.github/workflows/server-release.yml
```diff
@@ -1,8 +1,9 @@
 name: Server Release
 
 on:
-  release:
-    types: [published]
+  push:
+    branches:
+      - main
```
tasn, line 6: 🤔

svix-andor: Accidentally pushed this here while debugging this on my forked copy.

tasn: Keep this comment open, so that we remember to fix it before merging.

PR #857, opened 2023-02-24
Add JSON response type wrapper with response code

svix-andor:
This way the response code can be correctly inferred by aide for endpoints where we were previously returning `(StatusCode, Json<T>)` style responses. With the tuple style aide had no way of knowing what was the response like that's being returned. This way the response code is known statically at compile time which we can use to inform aide.

There are two structs, `JsonStatus`, for endpoints that always return the same status code on success, and `JsonStatusUpsert`, which is appropriate for endpoints that do upsert-style operations, e.g. where you have two status codes to distinguish between what happened internally, e.g. create versus update, but the output is the same. There's a fair number of endpoints that follow this pattern so it seemed worthwhile to add this.

Unfortunately custom types, i.e. `http::StatusCode` can't be used as a const generic in Rust currently, so I used `u16` which is essentially what `StatusCode` is on the inside, which I just quasi-cast into `StatusCode` with `from_u16(v).unwrap()`, but this _can_ panic because `from_u16` validates that the value passed is `100 <= n < 1000`. Therefore, some vigilance must be exercised with this, because `JsonStatus<1234, Foo>` will compile but blow up at runtime/test time.

### DIFF
--- a/server/svix-server/src/v1/endpoints/application.rs
+++ b/server/svix-server/src/v1/endpoints/application.rs
@@ -17,8 +17,9 @@ use crate::{
             UnrequiredNullableField,
         },
         validate_no_control_characters, validate_no_control_characters_unrequired,
-        validation_error, ApplicationPath, EmptyResponse, ListOrdering, ListResponse, ModelIn,
-        ModelOut, Pagination, PaginationLimit, ReversibleIterator, ValidatedJson, ValidatedQuery,
+        validation_error, ApplicationPath, EmptyResponse, JsonStatus, JsonStatusUpsert,
+        ListOrdering, ListResponse, ModelIn, ModelOut, Pagination, PaginationLimit,
+        ReversibleIterator, ValidatedJson, ValidatedQuery,
     },
     AppState,
 };
@@ -31,7 +32,6 @@ use axum::{
     Json,
 };
 use chrono::{DateTime, Utc};
-use hyper::StatusCode;
 use schemars::JsonSchema;
 use sea_orm::ActiveModelTrait;
 use sea_orm::ActiveValue::Set;
@@ -246,13 +246,14 @@ async fn create_application(
     query: ValidatedQuery<CreateApplicationQuery>,
     permissions::Organization { org_id }: permissions::Organization,
     ValidatedJson(data): ValidatedJson<ApplicationIn>,
-) -> Result<(StatusCode, Json<ApplicationOut>)> {
+) -> Result<JsonStatusUpsert<ApplicationOut>> {
     if let Some(ref uid) = data.uid {
         if let Some((app, metadata)) = ctx!(
             application::Model::fetch_with_metadata(db, org_id.clone(), uid.clone().into()).await
         )? {
             if query.get_if_exists {
-                return Ok((StatusCode::OK, Json((app, metadata).into())));
+                // Technically not updated, but it fits.
+                return Ok(JsonStatusUpsert::Updated((app, metadata).into()));
             }
             return Err(HttpError::conflict(
                 None,
@@ -275,7 +276,7 @@ async fn create_application(
         Ok((app_result, metadata))
     })?;
 
-    Ok((StatusCode::CREATED, Json((app, metadata).into())))
+    Ok(JsonStatusUpsert::Created((app, metadata).into()))
 }
 
 /// Get an application.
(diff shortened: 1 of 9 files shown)


server/svix-server/src/v1/endpoints/application.rs
```diff
@@ -293,7 +294,7 @@ async fn update_application(
     Path(ApplicationPath { app_id }): Path<ApplicationPath>,
     permissions::Organization { org_id }: permissions::Organization,
     ValidatedJson(data): ValidatedJson<ApplicationIn>,
-) -> Result<(StatusCode, Json<ApplicationOut>)> {
+) -> Result<JsonStatusUpsert<200, 201, ApplicationOut>> {
```
tasn, line 297:
```suggestion
) -> Result<JsonStatusUpsert<ApplicationOut>> {
```
Nitpick

It already feels like a special case, and we'll have this exact behavior everywhere anyway, so may as well just assume the status codes? Seeing both here, especially since they are in reverse order (create is second, update is first, though I guess order doesn't matter here).

server/svix-server/src/v1/utils/mod.rs
```diff
@@ -563,6 +564,95 @@ pub struct EventTypeNamePath {
...
+        };
+        (
+            StatusCode::from_u16(status).unwrap(),
+            axum::extract::Json(body),
+        )
+            .into_response()
+    }
+}
+
+impl<const SUPDATE: u16, const SCREATE: u16, T: JsonSchema + Serialize> OperationOutput
```
tasn, line 630: Looking at this, it looks like we can use the StatusCode enum and we don't need to actually use the ints?

The status code enum has it faults, but I think it's still better than magic numbers everywhere that are easy to forget what they mean exactly everywhere.

svix-andor: https://github.com/hyperium/http/blob/master/src/status.rs#L45

`StatusCode` is not an enum, it's a newtype struct, and it has some `const`s defined for its various values (`StatusCode::OK`) which is why it might have looked like an enum.

Optionally, we can hand-roll our own status code enum which we map to `http::StatusCode`.

server/svix-server/src/v1/endpoints/application.rs
svix-andor: Yeah, on second thought I agree, `upsert` is a special case anyway. Hardcoded this one to 200-UPDATE/201-CREATE.

server/svix-server/src/v1/utils/mod.rs
tasn: And we can't StatusCode::OK.as_u16()? It's a bit ugly I guess, grr.

OK, let's keep it as is.

tasn: Ah, no, wait, we don't need the as_u16.

Can't we just accept that, and then inside, in the describe function just do `as_u16`?

svix-andor: `as_u16` is not `const fn` so that's not an option (yet). I looked at `http`s issue tracker but no one seems to have even requested it, even though it seems pretty natural.

svix-andor: > Ah, no, wait, we don't need the as_u16.
> 
> Can't we just accept that, and then inside, in the describe function just do `as_u16`?

No, const generics can only be integral primitives (iN, uN, bool, etc.) and enums. StatusCode isn't either of those.

tasn: I mean *inside* of the implementation, but OK, if it has to be a primitive then that won't work.

server/svix-server/src/v1/utils/mod.rs
```diff
@@ -563,6 +564,88 @@ pub struct EventTypeNamePath {
     pub event_type_name: EventTypeName,
 }
 
+/// JsonStatus is a wrapper over `axum::extract::Json` as a handler output.
+/// Setting the `STATUS` const parameter automatically sets the response
+/// status code, as well as inserting it into the aide documentation.
+pub struct JsonStatus<const STATUS: u16, T: JsonSchema + Serialize>(pub T);
```
svix-gabriel, line 570: I wonder if it's worth just calling this `Json` instead of `JsonStatus`. A bit overloady of `axum::Json`, but if we'll never use `axum::Json` after this maybe it's worth it.

svix-gabriel (dismissed):

server/svix-server/src/v1/utils/mod.rs
svix-andor: `axum::Json` will be used, because that's the extractor for JSON bodies, so the request side of things. Alternatively, we could choose to name this `Json` and always alias that to `JsonBody` for example, but I think the way it is at the moment is clearer.

svix-gabriel (approved):
const generics are so cool

PR #871, opened 2023-03-17
catchup ee from oss 

svix-onelson:
n.b. there were some conflicts in `server/svix-server/src/v1/endpoints/event_type.rs` which I've made a best effort to resolve. I'd appreciate extra attention on the changes for that file.


.version
```diff
@@ -1 +1 @@
-0.75.0
+0.81.0
```
svix-onelson, line 1: Not sure if `ee` has it's own versioning scheme, but this is what it is in `main`

server/svix-server/src/v1/endpoints/event_type.rs
svix-onelson, line 1: This was the only file with conflicts that had to be manually resolved.

svix-gabriel (approved):

PR #877, opened 2023-03-24
Run cargo update

svix-daniel:
These changes are simply the result of running `cargo update` in an attempt to avoid a security advisory on OpenSSL's Rust bindings.

### DIFF
--- a/server/Cargo.lock
+++ b/server/Cargo.lock
@@ -80,7 +80,7 @@ dependencies = [
  "darling",
  "proc-macro2",
  "quote",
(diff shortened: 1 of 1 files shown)


server/Cargo.lock
```diff
@@ -100,9 +100,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.69"
```
svix-gabriel, line 103: Nice

svix-gabriel (approved):

PR #878, opened 2023-03-28
Alias types from internal dir / openapi package

tkonya:
When these ListResponse* types are aliased in the svix package, the struct type they're returning in the Data array should also be aliased.

## Motivation

If the user would like to, say, make a helper function that cycles through all the iterators for this, and returns the structs on a channel, they cannot do that without the types being available (or defining their own struct types that mimic the openapi ones).

## Solution

Simply alias the struct types that are used in each of these ListReponse* type's Data array field.


svix-onelson (approved):
LGTM. Thanks for this :pray:

PR #885, opened 2023-04-07
Create message: change to RawValue instead of Value

tasn:
We were using the `serde_json::Value` for the message payload which resulted in a lot of allocations when the payload was complex, and was causing high memory usage and fragmentation. With this, we don't actually parse this part of the json, but rather just treat it as a string (serde verifies it's valid json though). We unfortunately turn it into a `Value` when persisting to PG, we should fix this extra thing that diminishes the point of this change a bit.

Includes a hack to workaround the serde_json flatten issue with RawValue: https://github.com/serde-rs/json/issues/599

### DIFF
--- a/server/svix-server/src/v1/endpoints/message.rs
+++ b/server/svix-server/src/v1/endpoints/message.rs
@@ -36,6 +36,7 @@ use sea_orm::{entity::prelude::*, IntoActiveModel};
 use sea_orm::{sea_query::Expr, ActiveValue::Set};
 use serde::{Deserialize, Serialize};
 
+use serde_json::value::RawValue;
 use svix_server_derive::{aide_annotate, ModelIn, ModelOut};
 use validator::{Validate, ValidationError};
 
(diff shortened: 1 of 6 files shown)


server/svix-server/src/v1/endpoints/message.rs
```diff
@@ -56,15 +57,48 @@ pub fn validate_channels_msg(
...
+
+pub fn validate_raw_payload_is_object(
+    payload: &RawPayload,
 ) -> std::result::Result<(), ValidationError> {
-    match payload {
-        serde_json::Value::Object(_) => Ok(()),
-        _ => Err(validation_error(
+    // Verify it's an object/map
+    if payload.0.get().starts_with('{') {
+        Ok(())
```
svix-onelson, line 96: I'm assuming leading whitespace is trimmed automatically with `RawValue`? I saw there were remarks in the docs about how formatting/whitespace is retained as-is, but unsure if that's only the interior of the value.

I'd expect the surrounding whitespace to be removed just based on how I expect the parser to work, but unsure.

Maybe add a small test case just to check the behavior with leading whitespace?

tasn: There's no concept of interior whitespace, because this is the token so it has to be the value (no whitespace). I also tested it manually.

svix-onelson (approved):

PR #888, opened 2023-04-12
Gabriel/acks on nacks on stacks

svix-gabriel:


server/svix-server/src/worker.rs
```diff
@@ -994,13 +994,13 @@ pub async fn queue_handler(
                             .await
                             .is_err()
                         {
-                            if let Err(err) = queue_tx.nack(delivery).await {
+                            if let Err(err) = delivery.nack().await {
```
svix-gabriel, line 997: Unfortunately, we can't add `ack`/`nack` methods to the `TaskQueueReceiver`. From the worker
```rust
match queue_rx.receive_all().await {
            Ok(batch) => {
                for delivery in batch {
                    // ...
                    tokio::spawn(async move {
                        let worker_context = WorkerContext { ... };
                        let queue_task = // ...
                        if process_queue_task(worker_context, queue_task)
                            .await
                            .is_err()
                        {
                            if let Err(err) = delivery.nack().await {
                                // ...
                            }
                        } else if let Err(err) = delivery.ack().await {
                            // ...
                        }
                    });
                }
            }
            // ...
        }
```
`queue_rx.receive_all()` takes `queue_rx` by a mutable reference, so we can't have a reference to it in the spawned task ☹️

server/svix-server/src/queue/memory.rs
```diff
@@ -55,16 +55,25 @@ pub struct MemoryQueueConsumer {
...
-                    Ok(vec![delivery])
+                    vec![delivery]
                 } else {
-                    Err(err_queue!("Failed to fetch from queue"))
+                    return Err(err_queue!("Failed to fetch from queue"))
                 }
             }
+        };
+
+        // possible errors are `Empty` or `Disconnected`. Either way,
```
svix-james, line 70: A bit confused what's going on here. We receive once, then do it again?

svix-gabriel: `try_recv()` just receives the next value if and only if it's immediately available in the queue (hence why there's no `await`).

So all the change does is load as many messages from the queue as possible, without blocking more than once.

server/svix-server/src/worker.rs
tasn: You can use `RefCell` to workaround that and have inner mutability. I think that makes more sense than changing the whole API. Especially since I think it's nicer having the operations on the queue rather than on the "delivery".

What do you think?

svix-gabriel: Okay so my explanation was somewhat incomplete before.

There are two issues at play. The 1st is the mutability of `receive_all()`, which prevents `ack()` and `nack()` being called inside the loop. The 2nd is that `tokio::spawn(async move { ... })` necessarily takes ownership of everything passed in. Even if `receive_all()` did take `&self` (instead of `&mut self`), we'd have to `clone()` the `queue_rx`.

`Arc<RefCell<...>>` won't work here, since `RefCell` isn't sync, we won't be able to pass it into the spawned task (we'd get an error like `future cannot be sent between threads safely
the trait Sync is not implemented for queue_rx`). We could use `Mutex`, but now there's an unnecessary runtime cost to `queue_rx.receive_all()`.

So I agree in theory, I'd rather have `queue_rx.ack(delivery)` than `delivery.ack()`. But I think the tradeoff isn't worth it

server/svix-server/src/queue/mod.rs
```diff
@@ -103,91 +109,109 @@ pub enum QueueTask {
...
-        )
-        .await
-    }
+pub enum TaskQueueConsumer {
+    Redis(RedisQueueConsumer),
+    Memory(MemoryQueueConsumer),
 }
 
-pub struct TaskQueueConsumer(Box<dyn TaskQueueReceive + Send + Sync>);
+//pub struct TaskQueueConsumer(Box<dyn TaskQueueReceive + Send + Sync>);
```
svix-andor, line 140: leftover comment

server/svix-server/src/queue/mod.rs
```diff
@@ -103,91 +109,109 @@ pub enum QueueTask {
...
-        self.0.receive_all().await
+        match self {
+            TaskQueueConsumer::Redis(q) => ctx!(q.receive_all().await),
+            TaskQueueConsumer::Memory(q) => ctx!(q.receive_all().await),
+        }
     }
 }
 
+/// Used by TaskQueueDeliveries to Ack/Nack itself
+pub(self) enum Acker {
```
svix-andor, line 152: `pub(self)` unneeded

server/svix-server/src/queue/mod.rs
```diff
@@ -103,91 +109,109 @@ pub enum QueueTask {
...
+    Memory(MemoryQueueProducer),
+    Redis(Arc<RedisQueueInner>),
+}
+
 pub struct TaskQueueDelivery {
     pub id: String,
     pub task: Arc<QueueTask>,
+    pub(self) acker: Acker,
```
svix-andor, line 160: `pub(self)` unneeded

svix-daniel (approved):

PR #893, opened 2023-04-18
Improve OpenAPI generation

svix-andor:
This PR contains a series of commits aimed at improving the OpenAPI generation in order to make it closely match our current specification, but now instead generated by Rust. It is best reviewed commit by commit.


svix-onelson (approved):
LGTM :pray:

PR #895, opened 2023-04-18
Add RabbitMQ as message queue solution

svix-gabriel:
## Motivation

This adds `RabbitMQ` as an alternative internal queueing system (instead of in-memory or redis based queueing).

That's pretty much it. The RabbitMQ integration has only been tested locally using E2E tests, so there may be issues that only show up on a larger scale that need addressing. **As such, this should be treated as an MVP. Be cautious about using the RabbitMQ solution in a production environment for the time being.** The README hasn't been updated yet for precisely this reason.

## Solution

All in all, this requires a few changes
* Adding the integration to the local docker-compose testing suite. There's an adjacent `rabbit` folder for including the delayed message plugin, for, well, sending delayed messages 😃 
* Adding the actual queue implementation
* The test that checks multiple concurrent workers was updated to ack messages.


server/svix-server/src/queue/rabbitmq.rs
```diff
@@ -0,0 +1,269 @@
...
+    let consumer_chan = ctx!(conn.create_channel().await)?;
+
+    let exchange_name = ctx!(declare_delayed_message_exchange(&producer_chan).await)?;
+    ctx!(declare_bound_queue(&queue_name, &exchange_name, &producer_chan).await)?;
+
+    // With prefetch_size, there's a tradeoff.
+    // More prefetched messages drains RabbitMQ Queues faster, but also consumes more worker memory.
+    // Additionally, too many prefetched messages can starve other workers, and hurt total throughput.
+    // Leaving as 1 for now, to prevent starvation.
+    let prefetch_size = 1;
```
svix-james, line 49: Does prefetch define a minimum as well as a maximum?

svix-james: ... either way what I'm getting at though is that I think this should be configurable or (gasp) we should modify `receive_all`  to accept an optional max batch-size (though not sure if it's possible to specify at receive-time w/ rabbit)

svix-onelson: I didn't think there was? Gabriel may have seen the docs on this more recently than I. 

If you had a minimum, you'd usually need some other heuristic to prevent _"N where N  < PREFETCH_MIN"_ messages sitting there for ages with nobody electing to pick them up during slower periods.

server/svix-server/src/queue/rabbitmq.rs
```diff
@@ -0,0 +1,269 @@
...
+
+async fn start_queue_consumer(
+    queue_name: &str,
+    channel: &lapin::Channel,
+    prefetch_size: u16,
+) -> Result<lapin::Consumer> {
+    // Ref https://www.rabbitmq.com/amqp-0-9-1-reference.html#basic.consume.consumer-tag
+    // NOTE from @svix-gabriel: it's not really clear why this needs to be set to anything,
+    // so letting rabbit choose a default for now
+    let consumer_tag = "";
```
svix-onelson, line 150: A common case is to encode the software version (think like a user agent), or "role" in this iirc. Useful if/when deploys go sour and maybe you've got a lingering consumer running old code. Otherwise, they all look the same in the list.

server/svix-server/src/queue/rabbitmq.rs
```diff
@@ -0,0 +1,269 @@
...
+    consumer: lapin::Consumer,
+}
+
+pub async fn new_pair(
+    dsn: &str,
+    queue_name: String,
+) -> Result<(TaskQueueProducer, TaskQueueConsumer)> {
+    let conn = ctx!(lapin::Connection::connect(dsn, ConnectionProperties::default()).await)?;
+    let producer_chan = ctx!(conn.create_channel().await)?;
+    let consumer_chan = ctx!(conn.create_channel().await)?;
```
svix-onelson, line 40: I'm not sure how this fits with the abstraction, but it's not uncommon to have the connection break, requiring some sort of reconnect loop logic. I wonder where this can happen as written. If the pipe breaks, what do? Does that require the process to restart?

server/svix-server/src/queue/rabbitmq.rs
svix-gabriel: It's a maximum. https://www.rabbitmq.com/amqp-0-9-1-reference.html#basic.qos.prefetch-size

It's not necessarily related to how many values `receive_all` returns, it's just a tuning variable (basically trading RabbitMQ queue memory for consumer client's memory, with the other caveats mentioned in the comment.

I think having this as a configurable variable makes sense though. I'll make that change

server/svix-server/src/queue/rabbitmq.rs
svix-gabriel: Updated, although injected some randomness based on https://www.rabbitmq.com/amqp-0-9-1-reference.html#basic.consume.consumer-tag

server/svix-server/src/queue/rabbitmq.rs
svix-gabriel: We discussed this a bit offline, so adding those notes here for clarity

Handling broken connections is definitely a prerequisite for using this in prod. The main concerns are
1. Since `lapin` doesn't yet handle this for us (maybe [one day](https://github.com/amqp-rs/lapin/issues/312)), there's some hefty synchronization we'd have to add to handle graceful reconnections _given how our work queue is currently set up_. i.e. This is very complicated
2. We still need more time to exercise RabbitMQ and discover all the kinks, so we wouldn't want to recommend folks jump into using RabbitMQ for production projects yet _anyway_

Therefore, we're gonna table the reconnection issue for another PR.

svix-gabriel: Added https://github.com/svix/svix-webhooks/issues/898

svix-onelson (approved):

PR #896, opened 2023-04-19
ref verify_timestamp

itmilos:
used match Instead of if else 
moved timestamp_diff to variable to improve readability


tasn:
Hey @itmilos, thanks for the PR!

What's the purpose behind the change? Because the code looks less clear now than it was before.

itmilos:
This solution is also faster as timestamp difference is calculated only once.

tasn:
> This solution is also faster as timestamp difference is calculated only once.

I don't think it's faster in any meaningful way if at all. Did you see a noticeable performance improvement?

itmilos:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=df5c48a5c95900036fbc34353d9eda4f

PR #899, opened 2023-04-20
Add additional validation to some inputs

svix-daniel:
It was discovered that some fields in various `ModelIn` implementors were not validated correctly. 

These changes ensure that 

- The `#[validate]` macro is no longer missing for the `EventTypeName` in `EventTypeIn` instances
- `EventTypeName`s are validated to have a length greater than one
- That the `BaseUid` default validation implementation also checks for a length greater than one

Additionally fixes a typo in the `test_id_validation` test which caused one case to be skipped.

### DIFF
--- a/server/svix-server/src/core/types/mod.rs
+++ b/server/svix-server/src/core/types/mod.rs
@@ -197,7 +197,15 @@ fn validate_limited_str(s: &str) -> std::result::Result<(), ValidationErrors> {
         static ref RE: Regex = Regex::new(r"^[a-zA-Z0-9\-_.]+$").unwrap();
     }
     let mut errors = ValidationErrors::new();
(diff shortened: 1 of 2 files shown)


server/svix-server/src/core/types/mod.rs
```diff
@@ -218,11 +218,28 @@ fn validate_limited_str(s: &str) -> std::result::Result<(), ValidationErrors> {
     }
 }
 
+fn validate_non_empty_limited_str(s: &str) -> std::result::Result<(), ValidationErrors> {
```
tasn, line 221: Maybe just fix `validate_limited_str` to not accept empty? (it really shouldn't)

tasn (approved):

PR #900, opened 2023-04-20
Make redis-dsn optional

### LINKED ISSUE #870: Unable to start svix-server without redis
I am trying to run svix server without redis, but not able to run server when I remove redis. In documentation it is given redis_dsn field can be left empty but even after changing redis_dsn to empty still server is not starting and giving this error.
```
backend_1    | thread 'main' panicked at 'Invalid Redis DSN', svix-server/src/wait_for.rs:9:55
backend_1    | note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
server_backend_1 exited with code 101
```

jaymell:
The redis DSN is not required, given that redis
is not required for either queuing or caching. Nonetheless 
our config defaults were always adding it, causing it to waited on, 
and causing the application to panic.

Fixes #870.

### DIFF
--- a/server/svix-server/config.default.toml
+++ b/server/svix-server/config.default.toml
@@ -49,10 +49,10 @@ db_dsn = "postgresql://postgres:postgres@pgbouncer/postgres"
 # Higher values can significantly increase performance if your database can handle it.
 db_pool_max_size = 20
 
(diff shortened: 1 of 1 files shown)


svix-onelson (approved):
tbh the diff is not what I expected to see, but :shipit:

svix-onelson (approved):

PR #902, opened 2023-04-24
Use new rust-generated OpenAPI

svix-andor:
Change library code generation to use the new rust-generated OpenAPI specification. The number of lines changed is immense, but a lot of it is simple copy-pasting. 

Two parts to the change that are worth mentioning:
1. Order of parameters in some resources (endpoints, messages, attempts, integrations). The new OpenAPI matches the ordering in the endpoint URL.
2. The Python templates needed a change to the import statements.


csharp/Svix/EventType.cs
```diff
@@ -182,12 +177,12 @@ public List<EventTypeOut> List(EventTypeListOptions options = null, string idemp
         {
             try
             {
-                var lResults = await _eventTypeApi.ListEventTypesApiV1EventTypeGetAsync(
-                    options?.Iterator,
+                var lResults = await _eventTypeApi.V1EventTypeListAsync(
                     options?.Limit,
-                    options?.WithContent,
+                    options?.Iterator,
+                    null,
```
tasn, line 183: Same null comment.

csharp/Svix/MessageAttempt.cs
```diff
@@ -362,18 +352,18 @@ public MessageAttemptOut GetAttempt(string appId, string attemptId, string messa
         {
             try
             {
-                var lResults = await _messageAttemptApi.ListAttemptsApiV1AppAppIdMsgMsgIdAttemptGetAsync(
+                var lResults = await _messageAttemptApi.V1MessageAttemptListByMsgDeprecatedAsync(
                     appId,
                     messageId,
-                    options?.Iterator,
                     options?.Limit,
+                    options?.Iterator,
```
tasn, line 359: Is there type-checking for the ordering, or do we just hope it's right?
Do I need to double check you there, or have you verified that?

java/lib/src/main/java/com/svix/MessageAttempt.java
```diff
@@ -78,24 +78,23 @@ public void resend(final String msgId, final String appId, final String endpoint
 
 	public void resend(final String msgId, final String appId, final String endpointId, final PostOptions options) throws ApiException {
```
tasn, line 79: Wow, our SDK is terrible here! It should be app id -> msg id -> endpoint id.

csharp/Svix/EventType.cs
```diff
@@ -157,12 +152,12 @@ public List<EventTypeOut> List(EventTypeListOptions options = null, string idemp
         {
             try
             {
-                var lResults = _eventTypeApi.ListEventTypesApiV1EventTypeGet(
-                    options?.Iterator,
+                var lResults = _eventTypeApi.V1EventTypeList(
                     options?.Limit,
-                    options?.WithContent,
+                    options?.Iterator,
+                    null,
```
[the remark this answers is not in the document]
svix-andor: Ordering.

csharp/Svix/EventType.cs
svix-andor: Same, it's ordering.

csharp/Svix/MessageAttempt.cs
svix-andor: There is type checking yes, so things like limit and iterator can't be mixed up because one's an integer and the other is a string, but there are some parameters where mistakes are possible, for example before/after. A double check can't hurt, although I think the chance for mistakes are low because for languages where things are passed by position I checked things one by one.

java/lib/src/main/java/com/svix/MessageAttempt.java
svix-andor: Let's do that in a followup PR?

javascript/src/index.ts
```diff
@@ -643,7 +643,7 @@ class MessageAttempt {
     endpointId: string,
     options?: MessageAttemptListOptions
   ): Promise<ListResponseMessageAttemptEndpointOut> {
-    return this.api.listAttemptsForEndpointApiV1AppAppIdMsgMsgIdEndpointEndpointIdAttemptGet(
+    return this.api.v1MessageAttemptListByEndpointDeprecated(
```
tasn, line 646: How come we have "deprecated" in the name now? Not a biggie, just curious.

svix-andor: Because the API call itself is deprecated, and since we renamed the operation IDs it seemed like a good idea to drive the point home by including "deprecated" in the name, not just in the description.

tasn: OK.

PR #917, opened 2023-05-10
add js transforms to bridge

svix-onelson:
Embeds a deno runtime in the bridge binary to allow ad hoc reshaping of messages read from queues before they are sent to Svix.

## Motivation

When consuming JSON from messaging systems to generate new webhooks, it's common to need to transform the payload before making the Create Message request to Svix.

## Solution

Embedding a JS runtime into the bridge process allows us to run user-defined functions on the payloads as they move from input to output.

With this diff, plugin instances provided by `svix-webhook-bridge-plugin-queue-consumer` can now evaluate js included in the plugin config:
```yaml
plugins:
- type: "rabbitmqconsumer"
  input:
    uri: "amqp://guest:guest@localhost:5672/%2f"
    queue_name: "local"
    requeue_on_nack: false
  transformation: | 
    function handler(input) {
      return { 
        app_id: input.key, 
        message: { 
          eventType: input.event_type, 
          payload: input.data 
        }
      };
    }
  output:
    token: "***************"
```

The `transformation` key can now be set on any of these consumers. The JS fragment should include a default export of a function that accepts an object and returns an object.

These functions can be used to reshape the payload as necessary.

### DIFF
--- a/webhook-bridge/svix-webhook-bridge-types/src/lib.rs
+++ b/webhook-bridge/svix-webhook-bridge-types/src/lib.rs
@@ -1,9 +1,58 @@
 pub use async_trait::async_trait;
+use tokio::sync::mpsc;
+use tokio::sync::oneshot;
+
+/// Plain old JSON objects are what the transformations expect to receive and produce.
+pub type JsObject = serde_json::Map<String, serde_json::Value>;
+/// A channel for plugins to send payloads/scripts to for execution.
+pub type TransformerTx = mpsc::UnboundedSender<TransformerJob>;
+/// The receiver side for transformations. The JS executor reads from this.
+pub type TransformerRx = mpsc::UnboundedReceiver<TransformerJob>;
+/// A oneshot channel for the JS executor to "publish" return values to once complete.
+// FIXME: better error type?
+pub type TransformerCallbackTx = oneshot::Sender<Result<JsReturn, ()>>;
+/// Used by the caller of the transformer to await the execution's output.
+// FIXME: better error type?
+pub type TransformerCallbackRx = oneshot::Receiver<Result<JsReturn, ()>>;
+
+/// A transformation job sent to the JS executor.
+/// Once the script has been run on the payload, the transformed payload is sent back through the
+/// callback channel.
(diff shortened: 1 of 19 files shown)


webhook-bridge/svix-webhook-bridge-types/src/lib.rs
```diff
@@ -1,9 +1,58 @@
...
+}
 
 /// Effectively a black box to the supervisor.
 /// Plugins should run until they are done, and likely they should not be "done" until the program
 /// exits.
 #[async_trait]
 pub trait Plugin: Send {
+    /// For plugins that want to run JS transformations on payloads.
+    /// Giving them a sender lets them pass messages to the JS executor.
+    fn set_transformer(&mut self, _tx: Option<TransformerTx>) {}
```
svix-onelson, line 56: This is a bit of a kludge. It lets me feed in a Sender after the Plugin has been constructed which lets me avoid refactoring how Plugins are constructed.

I think ideally we need to separate the config validation from instantiation. 

Right now they are lumped together via `TryInto` impls for `-Config` into `-Plugin`. If these were split, we could feed the `Sender` into the constructor directly (removing the runtime `.expect()` to see if it was set).

I figure this arrangement should be fixed when we finally unify the different backends with `enum_dispatch`.

svix-gabriel (approved):

webhook-bridge/svix-webhook-bridge/Cargo.toml
```diff
@@ -8,9 +8,41 @@ edition = "2021"
...
+#deno_core = "0.158.0"
+
+# tagged 1.26.2 (leaks)
+#deno_core = "0.155.0"
+
+# tagged 1.25.4
+#deno_core = "0.152.0"
+
+# tagged 1.24.3 (leaks, but not quite as badly as later releases - backtrace showed v8 GC frames)
+#deno_core = "0.147.0"
```
svix-onelson, line 41: I can cut all this out. I was keeping notes as I went, profiling the memory for each minor at its highest patch-level.

svix-gabriel (approved):

PR #922, opened 2023-05-12
Server: Support comma-separated array query params

### LINKED ISSUE #921: Rust client does not pass supported query param format for arrays
## Bug Report

Chatted with @tasn and it looks like the Rust client library is passing in array query params as comma separate (e.g. `?event_types=a,b,c`) which Svix doesn't support.

### Version

0.85.1

### Platform

Linux: 5.19.0-41-generic

### Description

Code reproduction:
```
use svix::api::{MessageListOptions, Svix};

// Create an async main function
#[tokio::main]
async fn main() -> Result<(), anyhow::Error> {
    let app_uid = "somei-id";

    // Create a new Svix client
    let svix = Svix::new("REDACTED".to_string(), None);

    let event_types = vec![
        "card_transaction.updated".to_string(),
        "dispute.updated".to_string(),
    ];

    println!("event_types: {:?}", event_types);

    let options = MessageListOptions {
        event_types: Some(event_types),
       ..Default::default()
    };

    let result = svix
        .message()
        .list(app_uid.to_string(), Some(options))
        .await?;

    println!("got value from the server; result={:?}", result);

    Ok(())
}
```

I get the error:
```
Error: Varidation error Some(HttpValidationError { detail: [ValidationError { loc: ["query", "__all__"], msg: "String must match the following pattern: [a-zA-Z0-9\\-_.].", _type: "value_error" }] })
```

jaymell:
Our Rust lib generates query params in comma-separated syntax.
This is technically not what our API expects, but frankly it's easy to
support it and seems reasonable to do so. Added and updated tests
accordingly.

Fixes #921

### DIFF
--- a/server/svix-server/src/v1/utils/mod.rs
+++ b/server/svix-server/src/v1/utils/mod.rs
@@ -563,13 +563,14 @@ where
         let pairs = form_urlencoded::parse(parts.uri.query().unwrap_or_default().as_bytes());
 
         let event_types: HashSet<EventTypeName> = pairs
-            .filter_map(|(key, value)| {
+            .filter(|(key, _)|
                 // want to handle both `?event_types=`, `?event_types[]=`, and `?event_types[1]=`
-                if key == "event_types" || (key.starts_with("event_types[") && key.ends_with(']')) {
-                    Some(EventTypeName(value.into_owned()))
-                } else {
-                    None
-                }
+                key == "event_types" || (key.starts_with("event_types[") && key.ends_with(']')))
+            .flat_map(|(_, value)| {
+                value
+                    .split(',')
+                    .map(|x| EventTypeName(x.to_owned()))
+                    .collect::<Vec<_>>()
             })
             .collect();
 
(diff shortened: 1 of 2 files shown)


svix-onelson (approved):
:+1:

svix-onelson (approved):

server/svix-server/src/v1/utils/mod.rs
```diff
@@ -563,13 +563,14 @@ where
...
-            .filter_map(|(key, value)| {
+            .filter(|(key, _)|
                 // want to handle both `?event_types=`, `?event_types[]=`, and `?event_types[1]=`
-                if key == "event_types" || (key.starts_with("event_types[") && key.ends_with(']')) {
-                    Some(EventTypeName(value.into_owned()))
-                } else {
-                    None
-                }
+                key == "event_types" || (key.starts_with("event_types[") && key.ends_with(']')))
+            .flat_map(|(_, value)| {
```
tasn, line 569: This looks like it has an issue, what happens if someone passes: `event_types=1,2,3&bloop=4` you'll get back 1,2,3,4, no?

PR #925, opened 2023-05-15
WIP: polish bridge config

svix-onelson:


svix-james:
LGTM

webhook-bridge/svix-webhook-bridge.example2.yaml
```diff
@@ -0,0 +1,171 @@
...
+          payload: input.data
+        }
+      };
+    }
+  output:
+    # Required (the Svix token to use when creating messages with this consumer)
+    token: "XYZ"
+
+# Webhook Receiver
+- type: "webhookreceiver"
```
svix-onelson, line 128: I'd probably kebab-case this to match the suffix of the package name.

svix-onelson:
I've got a (larger than expected) diff going locally which will rework the consumer-side config details such that:

```yaml
plugins:
# The top-level is named `queue-consumer` (to match `svix-webhook-bridge-plugin-queue-consumer`)
- type: "queue-consumer"
  input:
    # The `backend` is now a field under the `input`, along with the backend-specific config details.
    backend: "rabbitmq"
    uri: "amqp://guest:guest@localhost:5672/%2f"
    queue_name: "local"
    requeue_on_nack: true
  transformation: | 
    function handler(input) {
      return { 
        "app_id": input.key,
        "message": { 
          "eventType": input.event_type,
          "payload": input.data
        }
      };
    }
  output:
    token: "testsk_5dRcm4SN6Z6KqJsdgJLxnbW4uMo0-lMS.eu"
```

This was a much larger code change than I initially expected so I need to do a little cleanup before I post it.

tasn:
@svix-onelson, can you do a "diff" (maybe in a new PR) that's just the result files? As in, just have diffs where new files are added so that we can comment on the syntax easily (like a PR review) with the annoying diff lines. Does this make sense?

svix-onelson:
> @svix-onelson, can you do a "diff" (maybe in a new PR) that's just the result files? As in, just have diffs where new files are added so that we can comment on the syntax easily (like a PR review) with the annoying diff lines. Does this make sense?

Yeah, I think that's what I did to kick off the conversation. I'll copy the example file to a new file.

svix-onelson:
@tasn 5937888508a9dbc0fa122a354f37367170cdfe5a adds a full copy of the example config (`svix-webhook-bridge.example2.yaml`). Drop your notes on that one and I can try to incorporate the updates to make it that real.

webhook-bridge/svix-webhook-bridge.example2.yaml
```diff
@@ -0,0 +1,176 @@
+# Svix Webhook Bridge Example Configuration
+
+# Set the log level for the service. Supported: error, info, warn, debug, trace. Default: info
```
tasn, line 3: listen_address:

svix-onelson:
Discussed with @tasn - there are a couple of action items to clear before this moves forward. 
The plan is to first do some exploratory "what if" example configs to examine:

- what if the jargon was more specific, less generic? 
  - Ex: splitting the config into "producers" and consumers instead of just "plugins."
- what if we rebrand in terms of [webhook] "senders" and "receivers" instead of `queue-consumer` vs `webhook-receiver`
- what if the HTTP interface was a shared resource (like the JS runtime), "owned by bridge" and plugins optionally register routes mapping back to functions?
- (building on prev) what if `(input, transformation, output)` was the shape of all config items?
  - make it so `HTTP` or  `webhook` is an input and the outputs are the various queue backends.

So, I guess the idea will be to sketch out these alternate realities and see which feels the most accessible, then after, look at the code changes needed to make them real.

I'll post a series of new yaml files to explore these ideas as subsequent revs, then we'll discuss.

webhook-bridge/svix-webhook-bridge.example3.yaml
```diff
@@ -0,0 +1,74 @@
...
+log_format: "json"
+
+# Optionally configure these settings to export OTEL data
+opentelemetry:
+  address: "http://localhost:1234"
+  service_name: "my-bridge"
+  sample_ratio: 0.5
+
+# Host/port to listen on when receiving webhooks
+receiver_http_addr: "0.0.0.0:5000"
```
tasn, line 11:
```suggestion
listen_address: "0.0.0.0:5000"
```
1. I'd keep it generic because I think we'll maybe use it for sender too at some point (as mentioned before).
2. No need to abbreviate, doesn't add anything.

webhook-bridge/svix-webhook-bridge.example3.yaml
```diff
@@ -0,0 +1,74 @@
...
+      # the options.debug and options.server_url are inherited from the global
+      token: "YYYY"
+
+# Webhook Receivers each register an HTTP endpoint that maps to an output (ie. a message queue or what have you).
+receivers:
+  - path: "/from-svix"
+    verification:
+      # Use the standard svix signature verification scheme.
+      type: "svix"
+      endpoint_secret: "whsec_XXXXXXXXXXX="
```
tasn, line 55:
```suggestion
    input:
      # Use the standard svix signature verification scheme.
      type: "svix"
      endpoint_secret: "whsec_XXXXXXXXXXX="
```
I think sticking with the input/output is potentially clear regardless of the other thing.

webhook-bridge/svix-webhook-bridge.example3.yaml
```diff
@@ -0,0 +1,74 @@
...
+  token: "XXXXX"
+  options:
+    debug: true
+    server_url: http://example.com/
+
+# Webhook Senders configure an input and (optional) transformation.
+# Use transformations to conform the payload we get as input to the expected shape (app_id, message, etc).
+# When transformation is left unset, the payload is used as-is.
+senders:
+  - name: "rabbit-local"
```
tasn, line 26: I think name is probably a good idea, so maybe have it for the receivers too?

webhook-bridge/svix-webhook-bridge.example3.yaml
```diff
@@ -0,0 +1,74 @@
...
+  - name: "rabbit-local2"
+    input:
+      type: "rabbitmq"
+      queue: "local2"
+      uri: "amqp://example.com/%2f"
+      requeue_on_nack: true
+    transformation: |
+      handler = (x) => ({ app_id: "app_1234", message: { eventType: "foo.bar", payload: x }})
+    # Output can optionally override `global_svix_client`
+    svix_client:
```
tasn, line 45: I think "output" is good.

webhook-bridge/svix-webhook-bridge.example3.yaml
```diff
@@ -0,0 +1,74 @@
...
+  service_name: "my-bridge"
+  sample_ratio: 0.5
+
+# Host/port to listen on when receiving webhooks
+receiver_http_addr: "0.0.0.0:5000"
+
+# Global svix settings.
+# A svix client is needed by each sender.
+# Each sender can override any of these fields by defining a `svix_client` section.
+global_svix_client:
```
tasn, line 16: I'm on the fence about having a global here. I can definitely see the value but also maybe adds complexity given that most people will probably just have one receiver/sender so it doesn't add much.

webhook-bridge/svix-webhook-bridge.example4.yaml
```diff
@@ -0,0 +1,65 @@
+log_level: "debug"
```
tasn, line 1: I think we probably want something in the middle. So probably the other as base, but taking in the input/output things from this one.

webhook-bridge/svix-webhook-bridge.example3.yaml
```diff
@@ -0,0 +1,74 @@
...
+  token: "XXXXX"
+  options:
+    debug: true
+    server_url: http://example.com/
+
+# Webhook Senders configure an input and (optional) transformation.
+# Use transformations to conform the payload we get as input to the expected shape (app_id, message, etc).
+# When transformation is left unset, the payload is used as-is.
+senders:
```
tasn, line 25: Senders and receivers make it very clear I think, good names.

tasn, line 25: TBH, I think the example files should just have either senders or receivers. As in you'll have one file demonstrating one and another demonstrating the other. We probably don't want an example file that has both just for simplicity. We can mention in the docs that it's possible though.

At least for the start.

webhook-bridge/svix-webhook-bridge.example3.yaml
tasn: We can also have a "shortcut" potentially by having:
```yaml
    input:
      # Use the standard svix signature verification scheme.
      type: "svix-webhook"
      endpoint_secret: "whsec_XXXXXXXXXXX="
```
So we can support both the above and this one and this is jut a short version for the above. For example.

webhook-bridge/svix-webhook-bridge.example3.yaml
svix-onelson: Seemed like the path would be useful in that case, but I could also see the use for a name.

My main want for this tbh is to include it in spans/logging.

webhook-bridge/svix-webhook-bridge.example3.yaml
svix-onelson: Can definitely leave it out for now. Just wanted to float the idea.

webhook-bridge/svix-webhook-bridge.example3.yaml
```diff
@@ -0,0 +1,74 @@
...
+    transformation: |
+      handler = (x) => ({ app_id: "app_1234", message: { eventType: "foo.bar", payload: x }})
+    # Can optionally specify`svix_client` here, but can be omitted entirely if the global is set.
+
+  - name: "rabbit-local2"
+    input:
+      type: "rabbitmq"
+      queue: "local2"
+      uri: "amqp://example.com/%2f"
+      requeue_on_nack: true
```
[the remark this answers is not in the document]
svix-onelson: I was reluctant to default the requeue since I have no idea which is more likely... however:

- SQS only lets you requeue (there's no choice)
- This is controlled by the way the topic is defined (outside of bridge entirely) for GCP Pub/Sub

We could default to `true` to fit with SQS, but it also feels weird to bias based on a single backend :shrug:

webhook-bridge/svix-webhook-bridge.example5.senders.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+
+# Each defines an `input`, (optional) `transformation`, and `output`.
+senders:
+  - name: "from-rabbit-local-to-svix"
+    input:
+      type: "rabbitmq"
+      queue: "local"
+      uri: "amqp://example.com/%2f"
+    # Optional transformation. JSON bodies will be forwarded as-is when unset.
```
svix-onelson, line 22: Omitted the `requeue_on_nack` --  let's assume we can pick a default either way and document what it is, though this feels like a "stopped clock is correct twice a day" situation to me.

webhook-bridge/svix-webhook-bridge.example5.receivers.yaml
```diff
@@ -0,0 +1,50 @@
...
+http_listen_address: "0.0.0.0:5000"
+
+
+# Each defines an `input`, (optional) `transformation`, and `output`.
+receivers:
+  - name: "from-svix-to-queue"
+    input:
+      # The "webhook" input type uses the HTTP listener. POST bodies/headers are "handled" by forwarding to the
+      # configured "output."
+      type: "webhook"
```
svix-onelson, line 21: I prefer "webhook" to "http" since we're biased towards POST requests with JSON objects for bodies. I'd expect an "HTTP" input to have a lot more config options to let me go wild, but "webhook" sort of curbs that feeling for me.

webhook-bridge/svix-webhook-bridge.example3.yaml
svix-onelson: I split things up in `.example5.senders.yaml` and `.example5.receivers.yaml`.

webhook-bridge/svix-webhook-bridge.example3.yaml
svix-onelson: I included this idea in `.example5.receivers.yaml` but kept the `verification` terminology since I think `authentication` speaks more to tokens and HTTP Authorization headers and signature verification feels sort of different.

I could easily see a case for needing proper HTTP authorization checks in addition to or instead of signature verification. Better to not conflate.

webhook-bridge/svix-webhook-bridge.example4.yaml
svix-onelson: Mashed the two up in `.example5.senders.yaml` and `.exampler5.receivers.yaml`. 

5th iteration feels pretty viable to me :rofl:

svix-onelson:
So, I'm going to dig in on the supporting refactors to make `.example5.{sender,receiver}.yaml` a reality and assume that the larger effort changes required will apply for wherever we finally land.

The biggest change required seems to be 

- moving the HTTP server code into the bridge binary
- (probably) splitting `Plugin` trait into 2 separate types to represent sending vs receiving.

There might be some other things in the HTTP code that needs to change but I'm not sure what that might include right now.

The rewrangling of configs to match the examples is probably the smallest lift overall.

---

If there are extra changes in direction for the config part, I'm betting the refactoring of the HTTP stuff will still make sense. It feels right that the binary owns the HTTP server just like it owns the JS Runtime. 

The doubt I have is that we might want to listen on other protocols in the future, and making the listener "belong to the plugin" would make the most sense for that, but as they say: "you ain't gonna need it" (YAGNI). That's tomorrow's problem.

webhook-bridge/svix-webhook-bridge.example4.yaml
```diff
@@ -0,0 +1,65 @@
...
+log_format: "json"
+
+# Optionally configure these settings to export OTEL data
+opentelemetry:
+  address: "http://localhost:1234"
+  service_name: "my-bridge"
+  sample_ratio: 0.5
+
+# Host/port to listen on for HTTP inputs
+http_listen_addr: "0.0.0.0:5000"
```
tasn, line 11:
```suggestion
http_listen_address: "0.0.0.0:5000"
```

webhook-bridge/svix-webhook-bridge.example5.receivers.yaml
```diff
@@ -0,0 +1,65 @@
+log_level: "debug"
```
tasn, line 1: Everything looks excellent in this and the senders examples!

One thing I'd make sure that we support in this YAML is loading from env var. Does YAML by any chance have an extension that supports just loading from env vars so I can do? `key: $MY_KEY` in the config and then we'll load the key from env var or env var file?

svix-onelson: Nothing is provided for this in `serde_yaml` from what I saw. We could pre-process the config to expand vars before parsing the YAML using something like [shellexpand](https://crates.io/crates/shellexpand).

In the past I'd used [envsubst](https://www.gnu.org/software/gettext/manual/html_node/envsubst-Invocation.html), in a shell command but that's a bit clumsy for launching services. My use case was for deployments, etc where you're usually running a series of commands anyway.

PR #927, opened 2023-05-16
Add Since/Until for endpoint stats to libs

svix-gabriel:
## Motivation

We added this feature to our API a while ago (see [here](https://api.svix.com/docs#tag/Endpoint/operation/v1.endpoint.get-stats)), but never updated the libs.

This PR corrects that.

## Solution

Update the libs.

### DIFF
--- a/go/internal/openapi/api_endpoint.go
+++ b/go/internal/openapi/api_endpoint.go
@@ -1156,12 +1156,14 @@ func (r ApiV1EndpointGetStatsRequest) Execute() (EndpointStats, *_nethttp.Respon
  * @param endpointId The ep's ID or UID
  * @return ApiV1EndpointGetStatsRequest
  */
-func (a *EndpointApiService) V1EndpointGetStats(ctx _context.Context, appId string, endpointId string) ApiV1EndpointGetStatsRequest {
+func (a *EndpointApiService) V1EndpointGetStats(ctx _context.Context, appId string, endpointId string, since *time.Time, until *time.Time) ApiV1EndpointGetStatsRequest {
 	return ApiV1EndpointGetStatsRequest{
 		ApiService: a,
 		ctx: ctx,
 		appId: appId,
 		endpointId: endpointId,
+		since: since,
+		until: until,
 	}
 }
 
--- a/java/lib/src/main/java/com/svix/Endpoint.java
+++ b/java/lib/src/main/java/com/svix/Endpoint.java
@@ -38,7 +38,8 @@ public final class Endpoint {
 		return this.create(appId, endpointIn, new PostOptions());
 	}
 
-	public EndpointOut create(final String appId, final EndpointIn endpointIn, final PostOptions options) throws ApiException {
+	public EndpointOut create(final String appId, final EndpointIn endpointIn, final PostOptions options)
+			throws ApiException {
 		try {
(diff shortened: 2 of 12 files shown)


java/lib/src/main/java/com/svix/Endpoint.java
```diff
@@ -94,7 +98,8 @@ public void recover(final String appId, final String endpointId, final RecoverIn
 		this.recover(appId, endpointId, recoverIn, new PostOptions());
 	}
 
-	public void recover(final String appId, final String endpointId, final RecoverIn recoverIn, final PostOptions options) throws ApiException {
+	public void recover(final String appId, final String endpointId, final RecoverIn recoverIn,
```
tasn, line 101: Why did you add options to all of these? Also, isn't this a breaking change? Now these options are required?

rust/src/api.rs
```diff
@@ -432,14 +438,20 @@ impl<'a> Endpoint<'a> {
         .await?)
     }
 
-    pub async fn get_stats(&self, app_id: String, endpoint_id: String) -> Result<EndpointStats> {
+    pub async fn get_stats(
+        &self,
+        app_id: String,
+        endpoint_id: String,
+        options: Option<EndpointStatsOptions>,
```
tasn, line 445: This is a breaking change, we should have added a `get_stats_with_options`, though I'm actually fine with this subtle break in Rust land tbh, as the strong version pinning make it mostly a non-issue.

java/lib/src/main/java/com/svix/Endpoint.java
svix-gabriel: With the exception of the new overloaded `getStats` method, I think these other changes are just reformats my code editor did automatically.

I can undo them though, I didn't think they were a big deal.

go/internal/openapi/api_endpoint.go
```diff
@@ -1156,12 +1156,14 @@ func (r ApiV1EndpointGetStatsRequest) Execute() (EndpointStats, *_nethttp.Respon
  * @param endpointId The ep's ID or UID
  * @return ApiV1EndpointGetStatsRequest
  */
-func (a *EndpointApiService) V1EndpointGetStats(ctx _context.Context, appId string, endpointId string) ApiV1EndpointGetStatsRequest {
+func (a *EndpointApiService) V1EndpointGetStats(ctx _context.Context, appId string, endpointId string, since *time.Time, until *time.Time) ApiV1EndpointGetStatsRequest {
```
svix-james, line 1159: So I'm sure this is just my ignorance here, but we're ok w/ breaking changes to the "internal" files?

svix-gabriel: It's been a while since I've used go, but iirc everything in `/internal` is only visible to the package, not users of the package. So folks using our libs can't call this function anyway.

And then the functions they do call (in `go/endpoint.go`) haven't had their interface change at all

svix-james (approved):

go/internal/openapi/api_endpoint.go
svix-onelson: Yeah. The gist is, Go won't let you import any package under an `internal/` path segment if the import isn't coming from inside the module itself. Dependents can't access that stuff.

svix-onelson (approved):

PR #933, opened 2023-05-18
integrate bridge js transformations with webhook-receiver plugin

svix-onelson:
Updates the webhook-receiver plugin to accept a `TransformerTx` so it can leverage the JS Runtime managed by bridge.

This is not the best diff, sadly.

The original webhook-ingester made some design decisions that made this a little awkward. Specifically, the request would accept a forwarder on its `forward` method which took responsibility for payload handling.

When transformations are optionally happening, the payload either needs to transform as a part of the `forward` method, or (my preference) the payload handling needs to be moved up to the handler level.

I went for the latter - The "forward" method now hangs off the forwarder and receives a payload. This means the handler can attempt the transformation before forwarding, which keeps the go/no-go logic all within the handler body. It also avoids having to somehow pass app state down into the forwarder so it could perform the transformation itself.

Unfortunately, the go/no-go logic in the handler is also sort of a mess. In order to tame the additional mess contributed by the transformation, that chunk of the work has been extracted into a helper function.

---

Ideally, we'd have some actual unit tests around the HTTP handler code, and the bits invoked by it, but since we don't yet have anything setup it'd be a bigger lift than I care to tackle right now.

Instead, manual testing was done, noted in the comments below.

### DIFF
--- a/webhook-bridge/svix-webhook-bridge-plugin-webhook-receiver/src/lib.rs
+++ b/webhook-bridge/svix-webhook-bridge-plugin-webhook-receiver/src/lib.rs
@@ -4,9 +4,12 @@ use axum::{
     routing::post,
     Router,
 };
+use forwarding::ForwardingMethod;
 use serde::Deserialize;
 use std::net::SocketAddr;
-use svix_webhook_bridge_types::{async_trait, Plugin};
+use svix_webhook_bridge_types::{
+    async_trait, JsObject, JsReturn, Plugin, TransformerJob, TransformerTx,
+};
 use tracing::instrument;
 use types::{IntegrationId, IntegrationState, InternalState, SerializableRequest, Unvalidated};
 
@@ -24,14 +27,18 @@ pub struct WebhookReceiverPluginConfig {
     pub routes: Vec<IntegrationConfig>,
 }
 
-#[derive(Clone, Debug, Deserialize, PartialEq)]
+#[derive(Clone, Debug)]
 pub struct WebhookReceiverPlugin {
     cfg: WebhookReceiverPluginConfig,
+    transformer_tx: Option<TransformerTx>,
 }
 
 impl WebhookReceiverPlugin {
     pub fn new(cfg: WebhookReceiverPluginConfig) -> Self {
-        Self { cfg }
+        Self {
+            cfg,
+            transformer_tx: None,
+        }
     }
 }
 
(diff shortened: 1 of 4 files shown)


svix-onelson:
Did some manual testing. Config like:

```yaml
plugins:
  - type: "webhookreceiver"
    listen_addr: "0.0.0.0:5000"
    routes:
      - name: "hello"
        verification:
          type: none
        transformation: |
          function handler(x) {
            return { __TRANSFORMED__: true, ...x }
          }
        destination:
          type: "rabbitmq"
          uri: amqp://guest:guest@localhost:5672/%2f
          exchange: ""
          routing_key: "local"
      - name: "hello2"
        verification:
          type: none
        destination:
          type: "rabbitmq"
          uri: amqp://guest:guest@localhost:5672/%2f
          exchange: ""
          routing_key: "local"
```

I ran some curl:

```
$ curl http://localhost:5000/webhook/hello -H 'content-type: application/json' -d '{"abc": 123}'
$ curl http://localhost:5000/webhook/hello2 -H 'content-type: application/json' -d '{"abc": 123}'
```

Then checked the rabbitmq GUI to see that:

- 2 messages were in the queue
- one had the `__TRANSFORMED__` key
- the other didn't


image

Seems to work. :+1:

webhook-bridge/svix-webhook-bridge-plugin-webhook-receiver/src/lib.rs
```diff
@@ -104,3 +145,35 @@ async fn route(
...
     }
 }
+
+/// Attempts to run the payload through a js transformation.
+async fn transform(
+    payload: JsObject,
+    script: String,
+    tx: Option<TransformerTx>,
+) -> Result<JsObject, http::StatusCode> {
+    if let Some(transformer_tx) = tx {
```
svix-gabriel, line 155: This is definitely nitpicky, but how do you feel about restructuring this to something like

```
let transformer_tx = if let Some(tx) {
   tx
} else {
   tracing::error!("transformations are not available");
   return Err(http::StatusCode::INTERNAL_SERVER_ERROR)
}
// ... the rest
```

This way, the core business logic isn't as nested.

svix-onelson: It's a great note, but I did it a little different.

I think the flow can be improved with a refactor of the handler (to return Result instead of StatusCode). Ideally, this function would accept a `TransformerTx` instead of an `Option<_>`.

If the config polish PR ends up moving the HTTP server code into the bridge binary, I'll look to fix this up more better during that work.

svix-gabriel (approved):

PR #936, opened 2023-05-25
fixups for bridge readme, CI, etc

svix-onelson:
Mostly this just is last touches prior to merging back to main.
- Fleshing out some of the readme to be more aligned with the rest of the repo.
- CI workflow updates for how they need to be IRL.
- adds a LICENSE


.github/workflows/bridge-release.yml
```diff
@@ -1,11 +1,61 @@
-name: Webhook Bridge Release
```
tasn, line 1: Nitpick: I'd maybe call the directory `svix-bridge`, so it feels like it's a separate binary rather than a small chunk of the server.

bridge/svix-bridge/LICENSE
```diff
@@ -0,0 +1,20 @@
+Copyright (c) 2021-2023 Svix Inc.
```
tasn, line 1: No need to have a license file in every directory (I think?), just have one top-level for Svix-bridge.

svix-onelson: I started out with one top-level license, but the security linter called these crates errors without the license in each. I think we need them in here.

.github/workflows/bridge-release.yml
svix-onelson: Can do, though the `bridge/` name has an affinity with `server/` in that:
- `server/svix-server`
- `bridge/svix-bridge`

Fine by me either way. LMK what you want and I can find/replace my way to it.

svix-james (approved):

PR #951, opened 2023-06-05
Pass `WithContent` option as query parameter to `List Attempted Messages` EP

ema-basiq:
## Motivation

The motivation behind this Pull Request is to enhance the Svix Go SDK by expanding the functionality of the List Attempted Messages endpoint. Currently, the SDK lacks the capability to return messages with payload. This limitation poses a challenge for developers who require access to the payload data when retrieving attempted messages. Therefore, this PR aims to introduce an option within the SDK that enables the retrieval of messages along with their payloads. By addressing this limitation, developers will have a more comprehensive and convenient way to access the desired message data.

## Solution

`WithContent` field of type `*bool` has been added to the `MessageAttemptListOptions` type. That value is then passed as query parameter in the request to the List Attempted Messages endpoint.

### DIFF
--- a/go/messageattempt.go
+++ b/go/messageattempt.go
@@ -35,6 +35,7 @@ type MessageAttemptListOptions struct {
 	StatusCodeClass *StatusCodeClass
 	Channel         *string
 	EndpointId      *string
+	WithContent     *bool
 }
 
 // Deprecated: use `ListByMsg` or `ListByEndpoint` instead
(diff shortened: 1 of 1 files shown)


tasn:
Thanks a lot! @svix-onelson is taking a look.

onelson (approved):
LGTM. Thanks for looking at this 🙏🏻

svix-onelson (approved):

tasn:
We'll probably release a new version of the libs later this week with this fix, but please let us know in case we forget!

PR #952, opened 2023-06-12
curb incredibly slow leak in openssl connector

svix-onelson:
## Motivation

Workers that run for a long time uninterrupted can leak memory _very slowly_.

## Solution

Running svix-server through [bytehound](https://github.com/koute/bytehound), grouping allocations by the "largest leak by backtrace" gives us:

```
#00 [svix-server] _start
#01 [libc.so.6] __libc_start_main
#02 [libc.so.6] 7f15c8e4dd8f
#03 [svix-server] main
#15 [svix-server] svix_server::main
#16 [svix-server] tokio::runtime::scheduler::multi_thread::MultiThread::block_on
#17 [svix-server] tokio::runtime::context::BlockingRegionGuard::block_on
#18 [svix-server] svix_server::main::{{closure}}
#19 [svix-server] <tracing::instrument::Instrumented<T> as core::future::future::Future>::poll
#20 [svix-server] svix_server::run_with_prefix::{{closure}}
#21 [svix-server] <tokio::future::poll_fn::PollFn<F> as core::future::future::Future>::poll
#22 [svix-server] svix_server::run_with_prefix::{{closure}}::{{closure}}
#23 [svix-server] svix_server::core::webhook_http_client::WebhookClient::new
#24 [svix-server] openssl::ssl::connector::SslConnector::builder
#25 [svix-server] openssl::ssl::connector::ctx
#26 [libssl.so.3] SSL_CTX_new_ex
#27 [libssl.so.3] 7f15c95cd2fa
#28 [libcrypto.so.3] EVP_CIPHER_fetch
#29 [libcrypto.so.3] 7f15c93e95ad
#30 [libcrypto.so.3] 7f15c93e8f57
#31 [libcrypto.so.3] 7f15c93204da
#32 [libcrypto.so.3] 7f15c930116d
#33 [libcrypto.so.3] 7f15c9301268
#34 [libcrypto.so.3] 7f15c92e1659
#35 [libcrypto.so.3] 7f15c933d455
#36 [libcrypto.so.3] 7f15c9317394
#37 [libcrypto.so.3] CRYPTO_zalloc
#38 [libbytehound.so] malloc [api.rs:294]
```

It's possible there's a leak in OpenSSL, and a proper fix can come from updating the lib _however_ in this case, we're linking dynamically.
Since dynamic linking means we have less control over which specific versions are used, I tried to mitigate the growth on our side too.

This diff tries to avoid the unwanted memory growth by instantiating one `WebhookClient` when the worker starts, cloning it each time we need to hand it off to a future instead of letting each future instantiate one itself.

Running a low reqs/sec workload through svix-server locally seems to show a (closer to) flat memory profile over the course of an hour.

### DIFF
--- a/server/svix-server/src/worker.rs
+++ b/server/svix-server/src/worker.rs
@@ -27,7 +27,6 @@ use chrono::Utc;
 
 use futures::future;
 use http::{HeaderValue, StatusCode, Version};
-use ipnet::IpNet;
 use lazy_static::lazy_static;
 use rand::Rng;
 
@@ -249,6 +248,7 @@ struct WorkerContext<'a> {
     db: &'a DatabaseConnection,
     queue_tx: &'a TaskQueueProducer,
     op_webhook_sender: &'a OperationalWebhookSender,
+    webhook_client: &'a WebhookClient,
 }
 
 struct FailedDispatch(messageattempt::ActiveModel, Error);
@@ -330,12 +330,8 @@ async fn make_http_call(
         created_at,
     }: PendingDispatch,
     msg_dest: &messagedestination::Model,
-    whitelist_subnets: &Option<Arc<Vec<IpNet>>>,
+    client: &WebhookClient,
 ) -> Result<CompletedDispatch> {
-    let client = WebhookClient::new(
-        whitelist_subnets.clone(),
-        Some(Arc::new(vec!["backend".to_owned()])),
-    );
     let req = RequestBuilder::new()
         .method(method)
         .uri_str(&url)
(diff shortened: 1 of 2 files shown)


server/svix-server/src/worker.rs
```diff
@@ -959,6 +961,7 @@ pub async fn queue_handler(
                             cache: &cache,
                             op_webhook_sender: &op_webhook_sender,
                             queue_tx: &queue_tx,
+                            webhook_client: webhook_client.clone(),
```
svix-gabriel, line 964: I know `clone`ing is cheap for the webhook_client, but I think we can easily have the WorkerContext take a reference to the client instead, no? That way you'd only have to `clone()` it once on 954.

svix-gabriel (approved):
Left a small comment - but if my reference suggestion isn't as straightforward as a thought, then this LGTM

server/svix-server/src/worker.rs
svix-onelson: Yep. Makes total sense, especially since the `WorkerContext` is already lifetime'd up.
Reworking it to borrow as much as possible.

svix-gabriel (approved):

PR #955, opened 2023-06-14
Server: Centralize header sanitize logic

jaymell:
We will need to reuse the current header sanitizing logic in a future PR, so 
introduce `SanitizedHeaders` as a type, use it where appropriate, and 
rename the non-sanitized field to indicate clearly that it could contain 
sensitive information.


svix-onelson (approved):
looks legit

PR #958, opened 2023-06-15
Server: add  support for one time token auth

svix-onelson:
Support for one time tokens depends on cache.

A JWT and one-time token are generated in `app_portal_access`, and the
one-time token is used as a part of the cache key where the value is the
JWT itself.

The cache entry has a 7 day TTL.

### DIFF
--- a/server/svix-server/Cargo.toml
+++ b/server/svix-server/Cargo.toml
@@ -56,6 +56,7 @@ thiserror = "1.0.30"
 bytes = "1.1.0"
 blake2 = "0.10.4"
 chacha20poly1305 = "0.9.0"
+getrandom = "0.2.4"
 # sea orm
 sea-orm = { version = "0.10.2", features = [ "sqlx-postgres", "runtime-tokio-rustls", "macros", "with-chrono", "with-json" ], default-features = false }
 sqlx = { version = "0.6.2", features = [ "runtime-tokio-rustls", "postgres", "migrate" ] }
--- a/server/svix-server/src/v1/endpoints/auth.rs
+++ b/server/svix-server/src/v1/endpoints/auth.rs
@@ -8,11 +8,18 @@ use axum::{
 };
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
+use std::time::Duration;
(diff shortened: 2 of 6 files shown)


server/svix-server/src/v1/endpoints/auth.rs
```diff
@@ -62,21 +69,36 @@ impl From<DashboardAccessOut> for AppPortalAccessOut {
...
+                // FIXME: what duration?
+                Duration::from_secs(60 * 60 * 24 * 7)
+            )
+            .await
+    )?;
     let login_key = serde_json::to_vec(&serde_json::json!({
         "appId": app.id,
-        "token": token,
+        // FIXME: is it okay to change this key? Seems like it's a breaking change.
+        //   Do we need to support both one time and plain jwt?
```
tasn, line 100: This is what we output, and I think it's fine as we don't promise the URL structure to be part of the API. Especially not the base64 encoded part!

server/svix-server/Cargo.toml
```diff
@@ -56,6 +56,7 @@ thiserror = "1.0.30"
 bytes = "1.1.0"
 blake2 = "0.10.4"
 chacha20poly1305 = "0.9.0"
+getrandom = "0.2.4"
```
tasn, line 59: Open source review (even if it's already a dep, you can then just say "we already use it").

server/svix-server/src/v1/mod.rs
```diff
@@ -16,6 +16,9 @@ pub fn router() -> ApiRouter<AppState> {
     let ret: ApiRouter<AppState> = ApiRouter::new()
         .merge(endpoints::health::router())
         .merge(endpoints::auth::router())
+        // FIXME: normally "authless" router is needed to make sure it's not nested under middleware.
```
tasn, line 19: I'd probably still split it if it takes < 20 mins.

svix-onelson:
# Open Source Review: getrandom

https://crates.io/crates/getrandom

Selected on the basis of: it's what we're using for the one-time auth impl elsewhere.

I just matched the dep here, with the same version spec.

svix-gabriel (approved):

PR #959, opened 2023-06-15
Server: Support sending example event

jaymell:
Support sending example events to given endpoints.
This works out-of-the-box for `svix.ping` events and 
for user-defined events with an example defined.

This also makes slight modification to `MessageTaskBatch`
to support optional endpoint overrides.

### DIFF
--- a/server/svix-server/src/db/models/eventtype.rs
+++ b/server/svix-server/src/db/models/eventtype.rs
@@ -101,8 +101,25 @@ pub fn schema_example() -> serde_json::Value {
 }
 
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Default)]
-pub struct Schema(HashMap<String, Json>);
+pub struct Schema(pub HashMap<String, Json>);
 json_wrapper!(Schema);
+
+impl Schema {
+    // Return the first example found for a particular schema,
+    // if one exists:
+    pub fn example(&self) -> Option<&serde_json::Value> {
+        self.0
+            .get("1")
+            .and_then(|version| match version {
+                serde_json::Value::Object(obj) => obj.get("examples"),
+                _ => None,
+            })
+            .and_then(|examples| match examples {
+                serde_json::Value::Array(arr) => arr.iter().next(),
+                _ => None,
+            })
+    }
+}
 
 impl JsonSchema for Schema {
     fn schema_name() -> String {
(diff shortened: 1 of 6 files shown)


server/svix-server/src/v1/endpoints/message.rs
```diff
@@ -276,35 +278,18 @@ async fn list_messages(
...
-/// The `eventId` is an optional custom unique ID. It's verified to be unique only up to a day, after that no verification will be made.
-/// If a message with the same `eventId` already exists for any application in your environment, a 409 conflict error will be returned.
-///
-/// The `eventType` indicates the type and schema of the event. All messages of a certain `eventType` are expected to have the same schema. Endpoints can choose to only listen to specific event types.
-/// Messages can also have `channels`, which similar to event types let endpoints filter by them. Unlike event types, messages can have multiple channels, and channels don't imply a specific message content or schema.
-///
-/// The `payload` property is the webhook's body (the actual webhook message). Svix supports payload sizes of up to ~350kb, though it's generally a good idea to keep webhook payloads small, probably no larger than 40kb.
-#[aide_annotate(op_id = "v1.message.create")]
-async fn create_message(
+pub async fn create_message_inner(
```
jaymell, line 281: Diff looks awful but it's just moving most of the logic to `_inner` function.

server/svix-server/src/db/models/eventtype.rs
```diff
@@ -101,9 +101,26 @@ pub fn schema_example() -> serde_json::Value {
...
-pub struct Schema(HashMap<String, Json>);
+pub struct Schema(pub HashMap<String, Json>);
 json_wrapper!(Schema);
 
+impl Schema {
+    // Return the first example found for a particular schema,
+    // if one exists:
+    pub fn example(&self) -> Option<&serde_json::Value> {
+        self.0
+            .get("1")
```
svix-onelson, line 112: Huh. Is this one of those situations where we're decoding an array into a map so the keys are numbered?

jaymell: We have a very strange convention at the moment where a schema must be stored with a top-level key of "1". I don't claim to understand it, but our existing APIs don't work if this is not specified when creating schemas/event types. A problem for another day, though, I suppose.

svix-onelson: yeah, okay

svix-onelson (approved):
Looks okay to me.

PR #960, opened 2023-06-15
pin redis images to 7.0.10

svix-onelson:
## Motivation

CI is failing for rediscluster configurations as of the 7.0.11 release that just came out.

## Solution

Pin the redis containers we use in CI to 7.0.10 instead so the tests can actually run cleanly.

### DIFF
--- a/server/testing-docker-compose.yml
+++ b/server/testing-docker-compose.yml
@@ -10,13 +10,13 @@ services:
       - "5432:5432"
 
   redis:
-    image: "redis:7-alpine"
+    image: "redis:7.0.10-alpine"
     ports:
       - "6379:6379"
 
   # Redis cluster
   redis-cluster:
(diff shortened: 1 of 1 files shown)


svix-james (approved):

server/testing-docker-compose.yml
```diff
@@ -10,13 +10,13 @@ services:
       - "5432:5432"
 
   redis:
-    image: "redis:7-alpine"
+    image: "redis:7.0.10-alpine"
```
svix-james, line 13: I don't think you need to touch this one, since it appears to be a bitnami-specific change, but all good.

svix-onelson: Yeah, I hear you, but I feel better seeing the same version in lock-step down the file.

PR #963, opened 2023-06-16
Server: add app attempt stats

svix-onelson:
Implements the "endpoint stats by app" endpoint, porting over the
original tests along with it.

Spot-checked the output with curl as well.

```
$ curl "http://localhost:8071/api/v1/stats/app/app_2RFwwVIcJtUu5ncRmIkD7RPUtjP/attempt/?startDate=2023-06-15T19:45:00.000Z&endDate=2023-06-15T20:15:00.000Z" -sS | jq
{
  "startDate": "2023-06-15T19:45:00Z",
  "endDate": "2023-06-15T20:15:00Z",
  "period": "FiveMinutes",
  "data": {
    "successCount": [
      0,
      0,
      0,
      5,
      0,
      0
    ],
    "failureCount": [
      0,
      0,
      0,
      1,
      0,
      0
    ]
  }
}
```

### DIFF
--- a/server/svix-server/src/v1/endpoints/stats.rs
+++ b/server/svix-server/src/v1/endpoints/stats.rs
@@ -1,0 +1,299 @@
+use crate::core::types::ApplicationIdOrUid;
+use crate::error::{HttpError, ValidationErrorItem};
+use crate::v1::utils::openapi_tag;
+use crate::{ctx, err_generic, err_validation, AppState};
+use aide::axum::routing::get_with;
+use aide::axum::ApiRouter;
+use axum::extract::{Path, Query, State};
+use axum::Json;
+use chrono::{DateTime, Timelike, Utc};
+use schemars::JsonSchema;
+use sea_orm::{ConnectionTrait, FromQueryResult, Statement};
+use serde::{Deserialize, Serialize};
+use std::collections::BTreeMap;
+use svix_server_derive::aide_annotate;
+
+/// Period length for a statistics data point
+#[derive(Deserialize, Serialize, JsonSchema, Debug, Eq, PartialEq, Copy, Clone)]
+#[allow(unused)]
+pub enum StatisticsPeriod {
+    OneDay,
+    FiveMinutes,
+}
+
+#[derive(Deserialize, Serialize, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct BaseStatisticsResponse {
+    pub start_date: DateTime<Utc>,
+    pub end_date: DateTime<Utc>,
+    pub period: StatisticsPeriod,
+}
+
+#[derive(Deserialize, Serialize, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct AttemptStatisticsData {
+    pub success_count: Option<Vec<i64>>,
+    pub failure_count: Option<Vec<i64>>,
+}
+
+#[derive(Deserialize, Serialize, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct AttemptStatisticsResponse {
+    #[serde(flatten)]
+    pub common_: BaseStatisticsResponse,
+    pub data: AttemptStatisticsData,
+}
+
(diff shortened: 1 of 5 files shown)


svix-onelson:
Here's what I've got after a 2nd pass on the SQL query and some handler updates. Right now I'm assuming this is meant to be giving a fixed number of elements for each success and failure, with zero-filled elements for the empty buckets.

Porting the tests will tell me if I read that right or not.

```
$ curl "http://localhost:8071/api/v1/stats/app/app_2RFwwVIcJtUu5ncRmIkD7RPUtjP/attempt/?startDate=2023-06-15T19:30:00.000Z&endDate=2023-06-15T20:30:00.000Z" -sS | jq
{
  "data": {
    "successCount": [
      0,
      0,
      0,
      0,
      0,
      0,
      5,
      0,
      0,
      0,
      0,
      0
    ],
    "failureCount": [
      0,
      0,
      0,
      0,
      0,
      0,
      1,
      0,
      0,
      0,
      0,
      0
    ]
  },
  "period": "FiveMinutes",
  "startDate": "2023-06-15T19:30:00Z",
  "endDate": "2023-06-15T20:30:00Z"
}
```

Anyway, this is getting closer. Still not done though.

server/svix-server/src/v1/endpoints/stats.rs
```diff
@@ -0,0 +1,160 @@
...
+    let stmt = Statement::from_sql_and_values(
+        db.get_database_backend(),
+        r#"
+        SELECT
+            status,
+            count(1),
+            to_timestamp(
+                floor(EXTRACT(epoch FROM ended_at) / EXTRACT(epoch FROM interval '5 min'))
+                * EXTRACT(epoch FROM interval '5 min')
+            ) AS bucket
```
svix-onelson, line 94: This is pretty gnarly. Worth noting `date_bin` was added in postgres v14 and is the proper solution for this problem, but I don't know that we can force folks to upgrade their db (assuming they aren't already on v14).

server/svix-server/src/v1/endpoints/stats.rs
```diff
@@ -0,0 +1,297 @@
...
+    pub data: AttemptStatisticsData,
+}
+
+#[derive(Deserialize, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+#[allow(unused)]
+struct StatisticsQueryParams {
+    start_date: Option<DateTime<Utc>>,
+    end_date: Option<DateTime<Utc>>,
+}
```
svix-onelson, line 50: I saw a comment in the original python that mentioned "camelCase for external API, but snake_case for response." 

The tests seem to suggest this is the case, but I'm still not sure about the JSON encode/decode on the Python side to know if that was actually what was happening.

@svix-lucho if you've got callsites on the frontend expecting these responses, it'd help me to confirm if you see snake_case or camelCase field names for the response body.

svix-onelson: N.b. the stubs we have for this endpoint's doc generation look like they are advertising camelCased fields. I can rename the fields, I just wasn't sure if they should be or not 😰

svix-onelson: Looking at the inheritance tree, it looks like all fields for responses are camelCased by default so I'll go ahead and just do it. Tell me if you see different, Luco!

svix-lucho: I'm a bit late here, but checking the frontend it seems we expect camelCase everywhere. Just tested the branch and it seems to work just fine :)

Screenshot 2023-06-19 at 4 35 41 PM

svix-lucho: (without any changes in frontend) - so the format you have here is correct

svix-onelson: Great. I think the comments I saw along the way must have been outdated then. All good!

svix-james (approved):

PR #966, opened 2023-06-19
Server: Disable clustered-redis lib retries

jaymell:
Because we already have our own generic retry implementation that we use for queuing and caching, we should disable the clustered redis client's own default retry mechanism, which defaults to 16(!) retries if none are specified. For certain classes of errors, this limit includes exponental backoffs, so if we add our own layer of retries on top of this, we could potentially spend minutes in an endless retry loop. Let's ensure that doesn't happen!

### DIFF
--- a/server/svix-server/src/redis/cluster.rs
+++ b/server/svix-server/src/redis/cluster.rs
@@ -17,7 +17,7 @@ impl RedisClusterConnectionManager {
         info: T,
     ) -> Result<RedisClusterConnectionManager, RedisError> {
         Ok(RedisClusterConnectionManager {
-            client: ClusterClientBuilder::new(vec![info]).build()?,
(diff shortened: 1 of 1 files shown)


svix-onelson (approved):
:shipit:

server/svix-server/src/redis/cluster.rs
```diff
@@ -17,7 +17,7 @@ impl RedisClusterConnectionManager {
         info: T,
     ) -> Result<RedisClusterConnectionManager, RedisError> {
         Ok(RedisClusterConnectionManager {
-            client: ClusterClientBuilder::new(vec![info]).build()?,
+            client: ClusterClientBuilder::new(vec![info]).retries(0).build()?,
```
tasn, line 20: And non-cluster doesn't have retries?

svix-james: Correct.

PR #975, opened 2023-06-28
Bridge: support raw inputs for transformations

svix-onelson:
Adds a bunch of plumbing to allow the message payloads from consumers to
be passed to transformations as strings rather than pre-parsing as JSON
in rust.

One concrete use case for this is if the system publishing those
messages to the queue is sending XML or perhaps base64 encoded data.

When configured to specify the `format` of a transformation, you can set
it to `string` rather than the default `json`. Then, when the
transformation receives that string as an argument, it can do whatever
parsing is required in order to return a valid `Object`.

Tests have been added up and down the stack to show that strings are
handled correctly, and that the new config data is propagated.

Notable omissions:

- redis relies on JSON values internally and can't easily support "raw"
  inputs. We can revisit this when we switch to `omniqueue` since I
  believe it handles this better.

- [x] webhook receivers _can support_ raw inputs, I think, but _don't currently_.
  I'll try to get this going before the PR merges.

Update: got receivers going now. For both senders and receivers, we now check the configuration to catch cases where redis is paired with a string transformation and call that an Error. There's some tests in place to make sure those errors show up early in the process, rather than waiting for some message to come through.

### DIFF
--- a/bridge/README.md
+++ b/bridge/README.md
@@ -27,6 +27,75 @@ publish the payload on to a specified "output."
 
 Both **senders** and **receivers** are defined in terms of their input, and optional JavaScript transformation, and their output.
 
+---
+
+Transformations are configured as either a single string of JS source code:
+
+```yaml
+transformation: |
+    function handler(input) {
+      return {
+        app_id: input.key,
+        message: {
+          eventType: input.event_type,
+          payload: input.data
+        }
+      };
+    }
+```
+In this situation, the input value will be pre-parsed as JSON.
+
+You can also configure transformations using the verbose form, allowing control of the `format`:
+```yaml
+transformation:
+  format: "json"
+  src: |
+    function handler(input) {
+      return {
+        app_id: input.key,
+        message: {
+          eventType: input.event_type,
+          payload: input.data
+        }
+      };
+    }
+```
+The `format` field can be set to either `string` (ie. the _raw payload_) or `json` (the default behavior).
+
+For example, using the string `format` you can parse data out of the input using whatever method you like:
+
+```yaml
+transformation:
+  # Let's say the input is an XML message like:
+  # `<msg appId="app_123" eventType="a.b"><payload>{"x": 123}</payload></msg>`
+
+  format: "string"
+  src: |
+    function handler(input) {
+      let parser = new DOMParser();
+      let doc = parser.parseFromString(input, "text/xml");
+      let msg = doc.firstChild;
+      let payload = JSON.parse(msg.getElementsByTagName("payload")[0].textContent)
+
+      return {
+        app_id: msg.attributes.appId,
+        message: {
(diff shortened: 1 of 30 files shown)


bridge/README.md
```diff
@@ -27,6 +27,75 @@ publish the payload on to a specified "output."
...
+      return {
+        app_id: input.key,
+        message: {
+          eventType: input.event_type,
+          payload: input.data
+        }
+      };
+    }
+```
+The `format` field can be set to either `string` (ie. the _raw payload_) or `json` (the default behavior).
```
svix-gabriel, line 63: Not `object`? (Instead of `json`)? I feel like `json` implies it could be a `[...]`, or `1` too

bridge/generic-queue/src/gcp_pubsub.rs
```diff
@@ -186,6 +186,10 @@ impl<T: DeserializeOwned + Send + Serialize + Sync> Delivery<T> for GCPPubSubDel
         serde_json::from_slice(&self.message.message.data).map_err(Into::into)
     }
 
+    fn raw_payload(&self) -> Result<String, QueueError> {
```
svix-gabriel, line 189:
```suggestion
    fn raw_payload<'a>(&'a self) -> Result<&'a str, QueueError> { ... }
```

Could we get away with returning a `&str`, using something like https://doc.rust-lang.org/std/str/fn.from_utf8.html?

That way, we give the consumers the option of `.clone()`ing if and only if they need to.

svix-onelson: I think this might need to be a Cow, for cases where we only have raw bytes for the payload, but otherwise I think you're right.

bridge/README.md
svix-onelson: Yeah, this is tricky to explain (which is probably bad) but at the outer edge of the boundary, we only require that the thing parses as JSON. The transformation can be required to conform to a Map type.

This is to say, you can send valid json (of any shape) and so long as the transformation converts to `Object`, we're good.

In `#5762`, we'll be explicitly requiring a certain shape for the JSON regardless of whether or not it comes out of a queue or a transformation, but still the input and output aren't symmetrical.

Does that make any sense? 😰

bridge/generic-queue/src/gcp_pubsub.rs
svix-onelson: Derp. You are 100% correct. The sad part is we will need to have an owned value to talk to the JS Runtime, but we can borrow right up until a certain point.

I haven't refactored `TransformerInput` to be lifetime'd, but it's possible we could avoid the allocation there, deferring all the way up to the actual script execution.

I'll look at that in a separate PR if time allows.

svix-gabriel (approved):

PR #976, opened 2023-06-28
Implement send-example route

svix-daniel:
This change implements the `send-example` route which allows dispatching example events to a given endpoint.


svix-onelson (approved):
Looks familiar :shipit:

PR #982, opened 2023-06-29
Bridge: receivers now publish to queues with `payload` field

svix-onelson:
This diff is a little more broad than just "put the payload in a field."
It also tries to formalize the requirements of each output case, both
Senders and Receivers.

For Senders, we need a certain shape useful for building a call to
Create Message.
For Receivers, outputs expect to get a JSON object with a `payload`
field on it, representing the data to publish to the queue or what have
you.

These boundary types are now both in the `-types` crate, and used to
ensure both as-is and transformed data conforms where needed.

Existing tests caught the changes on the receiver side (which is
the main thing for this diff).

Additionally, for Senders, I camelCased the `appId` field to make the
create message JSON cased more consistently. This was a remark that came
up during review of an up-coming blog post.

### DIFF
--- a/bridge/svix-bridge/src/webhook_receiver/mod.rs
+++ b/bridge/svix-bridge/src/webhook_receiver/mod.rs
@@ -8,7 +8,7 @@ use axum::{
 };
 use std::net::SocketAddr;
 use svix_bridge_types::{
-    JsObject, TransformationConfig, TransformerInput, TransformerInputFormat, TransformerJob,
+    ForwardRequest, TransformationConfig, TransformerInput, TransformerInputFormat, TransformerJob,
     TransformerOutput, TransformerTx,
 };
 use tracing::instrument;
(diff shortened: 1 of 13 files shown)


bridge/svix-bridge/src/webhook_receiver/mod.rs
```diff
@@ -136,7 +135,8 @@ async fn parse_payload(
             transform(input, xform.source().clone(), transformer_tx).await
         }
         // Keep the original payload as-is if there's no transformation specified.
-        // The as_json() only gets us to `Value`, so we also need a `from_value` call to marshal into a Map type.
+        // The as_json() only gets us to `Value`, so we also need a `from_value` call to marshal
+        // into a [`ForwardRequest`] type.
         None => serde_json::from_value(payload.as_json().map_err(|_| {
             tracing::error!("Unable to parse request body as json");
             http::StatusCode::BAD_REQUEST
```
svix-onelson, line 142: An open question: should the default "as-is" behavior take the entire body and shove it into a `payload` field, or literally treat it as-is like it does now?

I suspect auto-wrapping is the only way we'd see folks use this without requiring a transformation.

svix-gabriel (approved):

bridge/svix-bridge/src/webhook_receiver/mod.rs
tasn: Yeah, take it as is and just put it in a payload field. The idea is to give people the ability to configure, but this configuration doesn't make sense in the case of webhooks. Probably only when reading from internal queues (the same way we let them control the app-id from a queue, but not from a webhook).

svix-onelson: Dang. Too sleepy -- I misunderstood your comment. Following-up with another PR to auto-wrap the body with `{"payload": ...}` when the transformation isn't used.

That makes sense. The body _is the payload_, after all.

svix-onelson: Fixed via https://github.com/svix/svix-webhooks/pull/985

PR #986, opened 2023-06-30
Bridge: variable expansion for config files

svix-onelson:
Config files can now use "dollar brace" notation to do variable
substitutions. See the readme for an example.

~~Since variable expansion might not be assumed, this behavior is gated by
a CLI flag/env var and is **off by default**.~~

Furthermore, the variables available for substitution are exposed via a
simple HashMap meaning we can filter the vars exposed in the future, if
requested. This also has the benefit of being nicer for testing ;)

Tests are added to capture the initial behavior of the new dep, but also
to lock in the contract if we decided to "roll our own" in the future.

---

Open source review required.

Adds a dependency on [envsubst-rs](https://github.com/coreos/envsubst-rs)
to provide the base substitution capability.

This was selected based on the following criteria:

- support for the absolute minimum features (no expression evaluation
  for fallbacks, etc).
- Tiny. 170LOC (60+ of which are tests), easy to understand and fork if
  needed.
- Vaguely looks maintained, owned by coreos (part of redhat).

The big idea with adding this dep is, _"it already exists, but rolling
our own based on this if we want to customize behavior is easy."_

### DIFF
--- a/bridge/README.md
+++ b/bridge/README.md
@@ -133,6 +133,27 @@ The config file itself does the heavy lifting.
 
 When unset, the current working directory is checked for a file named `svix-bridge.yaml`.
 
+## Variable Expansion
+
+`svix-bridge` supports environment variable expansion inside the config file.
+
+Using "dollar brace" notation, e.g. `${MY_VARIABLE}`, a substitution will be made from the environment.
+If the variable lookup fails, the raw variable text is left in the config as-is.
+
+As an example, here's a RabbitMQ sender configured with environment variables:
+
+```yaml
+senders:
+  - name: "send-webhooks-from-${QUEUE_NAME}"
+    input:
+      type: "rabbitmq"
+      uri: "${MQ_URI}"
+      queue_name: "${QUEUE_NAME}"
+    output:
+      type: "svix"
+      token: "${SVIX_TOKEN}"
+```
+
 Each sender and receiver can optionally specify a `transformation`.
 Transformations should define a function called `handler` that accepts an object and returns an object.
 
--- a/bridge/svix-bridge/src/config/tests.rs
+++ b/bridge/svix-bridge/src/config/tests.rs
@@ -1,5 +1,8 @@
 use super::Config;
 use crate::config::{LogFormat, LogLevel, SenderConfig};
+use std::collections::HashMap;
+use svix_bridge_plugin_queue::config::{QueueConsumerConfig, RabbitMqInputOpts, SenderInputOpts};
+use svix_bridge_types::{SenderOutputOpts, SvixSenderOutputOpts};
 
 /// This is meant to be a kitchen sink config, hitting as many possible
 /// configuration options as possible to ensure they parse correctly.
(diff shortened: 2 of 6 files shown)


bridge/svix-bridge/src/config/tests.rs
```diff
@@ -331,3 +334,174 @@ fn test_senders_example() {
...
+    let cfg = Config::from_src(src, Some(&vars)).unwrap();
+    // when lookups succeed, the token should be replaced.
+    let otel = cfg.opentelemetry.unwrap();
+    // Not the user-intended outcome, but it simplifies the parsing requirements.
+    assert_eq!(&otel.address, "https://127.0.0.1:$OTEL_PORT");
+}
+
+#[test]
+fn test_variable_substitution_missing_numeric_var_is_err() {
+    // FIXME: discuss in code review
```
svix-onelson, line 410: The nutshell here is: when a var is specified in the config but is not set in the env, the current behavior is to leave the original var as-is.

Example:
```yaml
queue_name: "${QUEUE_NAME}"
```

For string fields, this config will parse just fine because the literal string `"${QUEUE_NAME}"` works as a string, but will error if we try to connect to a bad queue name.

If we wanted to error immediately when a var is missing (could be better), I'd probably ditch the dep on `envsubst` and rewrite it in here with that behavior.

tasn: Nah, I think this is great, as it's the least likely to clash with other things.

Does this lib also support escaping?

bridge/README.md
```diff
@@ -133,6 +133,32 @@ The config file itself does the heavy lifting.
 
 When unset, the current working directory is checked for a file named `svix-bridge.yaml`.
 
+## Variable Expansion
+
+`svix-bridge` supports variable expansion inside the config file, when launched with `--expand-vars` or
```
tasn, line 138: Why make it a configuration value? Just have it always on, I can't think when we would even want it off.

bridge/svix-bridge/src/config/tests.rs
svix-onelson: Sadly, nope. If there's not really anything to purposefully prevent it from trying to do a replacement.

Some other systems use `$${...}` to retain the literal instead of doing a substitution. Not supported here 😓

bridge/README.md
svix-onelson: Can do.

There was an earlier concern about security, but since this implementation doesn't actually evaluate expressions of any kind, the worst thing that can happen is leaking your env vars to yourself (not a real problem).

An edge case: things that look like vars but aren't meant to be replaced. With the current implementation, the original token is retained (which would be good) unless there's actually a matching env var set.

Escaping would solve for this case, but we don't have that currently.

bridge/svix-bridge/src/config/tests.rs
tasn: I think it's fine anyway, as clashes would be unlikely?

bridge/README.md
tasn: I think there's no security concern and this is just a source for confusion (why isn't it working?!).

svix-onelson:
Marked draft again until I rip out the CLI flag and make var expansion the default.

tasn (approved):

PR #989, opened 2023-07-06
Server: deflake test_lock

svix-onelson:
## Motivation

This test repeatedly fails, but not always. A real frustration!

## Solution

Rephrasing the assertions may help. Instead of using fixed deadlines, assert based on relative durations between the first and second.

### DIFF
--- a/server/svix-server/src/core/idempotency.rs
+++ b/server/svix-server/src/core/idempotency.rs
@@ -526,10 +526,9 @@ mod tests {
 
     #[tokio::test]
     async fn test_lock() {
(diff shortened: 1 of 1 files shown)


svix-onelson:
Anecdotally, when this test failed I saw the first assertion failing because the resp_1_instant was ~355ms. I'm not sure where the deadlines came from, but since sleeps are only a minimum, I'd expect the deadline to be missed arbitrarily (just like we see in practice).

svix-gabriel (approved):

Issue #983, opened 2023-06-30
Support more JWT signature algorithms


tasn:
We only support HS256 at the moment, but we should probably make it configurable and support more. The list from https://docs.rs/jwt-simple/latest/jwt_simple/ (the lib that we use) seems reasonable, and it should be trivial to add.

Just adding a config variable to the config file to choose the type, and everything else should just work.

tasn:
Just need to make sure that asymmetric keys are handled correctly in the sense that we can have a "read only" key that can only verify but not generate (and if so, what does `jwt generate` do? Make sure it errors nicely).

PR #994, opened 2023-07-07
Server: JWT Signing Algorithm Expansion

svix-nick:
## Motivation

See #983. The JWT library that we use exposes a number of different signing algorithms, and it'd be neat to allow people to use some of the more common ones instead of the default HS256.

## Solution

This PR adds the `JwtSigningConfig` enum to the configuration which wraps the algorithms exposed by `jwt-simple`. It's a fun, double-layered enum to allow a fallback shape for backwards compatibility purposes.

The hard part is the deserialization. After that, the signing config delegates generation and verification calls to whichever kind of key we have loaded.

The signing config is wrapped in an `Arc` to avoid having many copies of the key in memory, and also because several of the algorithms are not `Clone`.

The non-`Default` variants are wrapped in a `Box` to avoid wasting a ton of space when the `Default` variant is used (which is probably going to be most of the time).

Finally, the signing config has a custom Debug impl that just spits out the name of the algorithm in play, to avoid accidentally leaking the key.


server/svix-server/config.default.toml
```diff
@@ -1,5 +1,5 @@
 # Default configuration file
-# Values here can also be set by setting the appropriate env var, e.g. SVIX_DB_DSN for db_dsn
+# Values here can also be set by setting the appropriate env var, e.g. SVIX_DB_DSN for db_dsnk
```
svix-onelson, line 2: Extra k by mistake?

svix-onelson (commented):
Looks good to me, by the broad strokes. I'm no cryptographer, so I can't speak to the veracity.

The existing tests passing post-refactor is a good sign, but I assume all the tests are using a standard (the old default) algorithm.

Would it be possible to get some tests going to exercise the new code? I think even just some small unit tests to show the Deserializers doing their thing might be nice to have.

svix-onelson (approved):
lgtm. Let's try it :shipit:

PR #995, opened 2023-07-08
Server: Avoid panics when using the wait-for option, make timeouts slightly less arcane

svix-nick:
## Motivation

Even if the application can't continue, we can still let the runtime wind down gracefully and get a formatted error message.

## Solution

This PR refactors `wait_for_dsn` in two ways:

1. It removes the use of `select!` in favor of simply putting the timeout loop inside of its own timeout
2. Instead of `panic!`ing on timeout or other errors, it bubbles the message back up in an unstructured dyn Error so that main can log it nicely then exit gracefully


server/svix-server/src/wait_for.rs
```diff
@@ -1,43 +1,38 @@
 use std::time::Duration;
 
+const ATTEMPT_TIMEOUT: Duration = Duration::from_secs(3);
```
svix-gabriel, line 3: Small nit - but I actually prefer where the constants were before, since they're only used in a single function, I think it makes sense to keep them restricted to that scope.

svix-gabriel (commented):
More curiosity, but do you know what the bubbled up error message ends up looking like (now that it doesn't panic)?

svix-gabriel (approved):

PR #998, opened 2023-07-10
Update Rust to latest stable version

svix-gabriel:
This updates Rust to the latest stable version for both server and bridge.

That's it.

Folks are recommended to update Rust locally as well.
```
rustup update
```


svix-onelson (approved):
🙏🏻

PR #1005, opened 2023-07-14
Deprecate Endpoint version field and make it optional

svix-lucho:
## Motivation
The endpoint `version` field is not being used for anything and is confusing for API users. 

## Solution
Mark `version` as deprecated and remove it from the required fields to create an endpoint. For backward compatibility, rather than completely removing it, default to `1` when no version is specified.


server/svix-server/src/v1/endpoints/endpoint/mod.rs
```diff
@@ -209,6 +217,7 @@ impl EndpointIn {
 impl ModelIn for EndpointIn {
     type ActiveModel = endpoint::ActiveModel;
 
+    #[allow(deprecated)]
```
svix-gabriel, line 220: I don't think I've ever seen this before. Not saying it's wrong, but what's the error message if this isn't there?

svix-lucho: Screenshot 2023-07-18 at 2 50 27 PM

svix-gabriel (approved):

server/svix-server/src/v1/endpoints/endpoint/mod.rs
```diff
@@ -162,9 +166,13 @@ pub struct EndpointIn {
...
     #[schemars(url, length(min = 1, max = 65_536), example = "example_endpoint_url")]
     pub url: Url,
-    #[validate(range(min = 1, message = "Endpoint versions must be at least one"))]
+
+    #[deprecated]
+    #[serde(default = "default_endpoint_version")]
+    #[validate(range(min = 1, message = "Endpoint versions must be at least one if set"))]
     #[schemars(range(min = 1), example = "example_endpoint_version")]
-    pub version: u16,
+    pub version: Option<u16>,
```
tasn, line 174: This makes the field nullable, I think we just want optional (so just set default, don't allow null). I think this is fine in general for this change, but it's a subtle difference worth being aware of.

PR #1007, opened 2023-07-18
Bridge: better js transformations

svix-onelson:
This PR serves a few aims:

1. improve diagnostics around configuring transformations.
2. improve overall performance of transformations.
3. upgrade deno
3a. without the penalty of leaking memory.
3b. open the door to using async/await and JS module code (needed for
the "poller input" prototyped during our recent hack week.
3c. avoid that CVE we saw for all but recent deno versions.

This diff upgrades us from a very old `deno_core` to the current release
`0.195.0` via the latest `deno_runtime`. The latter is what provides
expanded JS features such as async/await and typical DOM APIs you might
expect to have, such as `fetch`.

In `deno_core` versions beyond `0.142.0`, we see memory leaked each time
we initialized a `JsRuntime` and drop it.

This was the old method for executing JS code in Bridge. With the upgrade, we
now mitigate the leak by retaining the `JsRuntime`s we make. A fixed
pool of `JsRuntime`s, each in a dedicated thread,  are initialized on
startup.
Permit-based accounting (via `deadpool`) is used to distribute work across
the pool.

With this new approach, casual load testing shows dramatically improved
throughput. Probably this is due to amortization of interpreter startup
time, which previously we paid for each and every message handled.

Casual load testing also showed the memory profile to be essentially
flat (~100Mb with 4 JS runtimes), but there's still the possibility of memory
leaks stemming from the JS code itself since the interpreters are now
long-living.

For the goal of better diagnostics, the config parser now passes each
transformation source string to `deno_ast` to check that the syntax is
at least valid. Since the config parser happens early on, this will mean
an error on startup instead of sometime later when an event tries to run
through the broken script.

### DIFF
--- a/bridge/svix-bridge/Cargo.toml
+++ b/bridge/svix-bridge/Cargo.toml
@@ -23,16 +23,16 @@ serde_yaml = "0.9"
 svix-ksuid = "0.7.0"
 svix-bridge-plugin-queue = { optional=true, path = "../svix-bridge-plugin-queue" }
 svix-bridge-types = { path = "../svix-bridge-types" }
-threadpool = "1"
 tokio = { version = "1", features = ["full"] }
 tower-http = { version = "0.3", features = ["trace"] }
 tracing = "0.1"
 tracing-opentelemetry = "0.18.0"
 tracing-subscriber = { version="0.3", features=["env-filter", "fmt", "json"] }
-
-[dependencies.deno_core]
-# tagged deno@1.23.3 (does not seem to leak) though quite old.
-version = "0.142.0"
+# N.b. for newer deno versions (like this) the runtimes must be retained and reused since they will leak memory if you
+# create/drop them.
+deno_runtime = "0.119.0"
+deno_ast = "0.27.2"
+deadpool = { version = "0.9.5", features = ["unmanaged", "rt_tokio_1"] }
 
 [dev-dependencies]
 chrono = "0.4"
--- a/bridge/svix-bridge/src/config/mod.rs
+++ b/bridge/svix-bridge/src/config/mod.rs
@@ -3,6 +3,7 @@ use std::borrow::Cow;
 use std::collections::HashMap;
 use std::io::{Error, ErrorKind};
 use std::net::SocketAddr;
+use std::num::NonZeroUsize;
 use svix_bridge_plugin_queue::config::{
     into_receiver_output, QueueConsumerConfig, ReceiverOutputOpts as QueueOutOpts,
 };
@@ -27,6 +28,8 @@ pub struct Config {
     pub opentelemetry: Option<OtelExporterConfig>,
     #[serde(default = "default_http_listen_address")]
     pub http_listen_address: SocketAddr,
+    #[serde(default = "default_transformation_worker_count")]
+    pub transformation_worker_count: NonZeroUsize,
 }
 
 impl Config {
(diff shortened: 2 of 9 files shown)


bridge/generic-queue/src/redis.rs
```diff
@@ -119,7 +119,7 @@ impl<T: DeserializeOwned> RedisStreamDeserialize<RedisStreamJsonSerde> for T {
         let Some(payload) = map.get("payload") else {
             return Err(QueueError::Serde(serde_json::Error::missing_field(
                 "payload",
-            )))
+            )));
```
svix-onelson, line 122: rustfmt (in beta) complained about this 🤷🏻

bridge/svix-bridge/src/runtime/mod.rs
```diff
@@ -1,45 +1,128 @@
...
+            // carrying over from call to call.
+            // Refs:
+            // - <https://github.com/denoland/deno/issues/17861>
+            // - <https://github.com/denoland/deno/issues/17199>
+            let realm = runtime.create_realm().expect("create realm");
+            for Job { input, script, cb } in rx {
+                let isolate = runtime.v8_isolate();
+                let ret = run_script_inner(isolate, &realm, input, script);
+                cb.send(ret).expect("callback");
+            }
```
svix-onelson, line 37: It's looking more and more like `JsRealm` won't help much for anything. The isolate fed in there from the runtime is always the same (belonging to the runtime), not a fresh one.

Constructing a new Isolate using the v8 APIs causes a panic here. May as well skip the realm entirely and just try to mitigate the global namespace from being polluted some other way.

bridge/svix-bridge/src/config/mod.rs
```diff
@@ -46,8 +46,39 @@ impl Config {
...
+        let cfg: Self = serde_yaml::from_str(&src)
+            .map_err(|e| Error::new(ErrorKind::Other, format!("Failed to parse config: {}", e)))?;
+
+        for sc in &cfg.senders {
+            if let Some(tc) = sc.transformation() {
+                crate::runtime::validate_script(tc.source().as_str()).map_err(|e| {
+                    Error::new(
+                        ErrorKind::Other,
+                        format!(
+                            "failed to parse transformation for sender `{}`: {:?}",
```
svix-onelson, line 58: Commonly we validate stuff via proc macros that run during the serde parsing.

In this case, I deferred the work since the underlying types that hold the source code to validate are split out across a few different crates in the workspace and I didn't want to start pushing deno things into more crates.

svix-onelson:
# Open Source Review(s)

## `deno_runtime`

Previously we depended on `deno_core` for bare-bones JS eval. `deno_runtime` adds more features/libraries on top, needed for future work (e.g. as https://github.com/svix-onelson/poller-input-poc).

## `deno_ast`

Related to the previous, this crate provides APIs for parsing of JS scripts/modules.
This helps validate JS transformations.

## `deadpool`

Selected to help manage our fixed pool of `JsRuntime`s.

Initially, I selected [workerpool](https://github.com/lorepozo/workerpool) for this, which seemed like a great fit until `cargo-deny` showed many unsoundness/security problems with the implementation.

After seeing the poor audit for `workerpool`, I landed on `deadpool` because of:
- active maintainership
- small relevant API surface
- 700+ github stars
- "word of mouth"

I also looked at [lifeguard](https://github.com/zslayton/lifeguard) and [object-pool](https://github.com/CJP10/object-pool), but the semantics/contracts or APIs were not quite right.

deny.toml
```diff
@@ -32,6 +32,8 @@ allow = [
     "MPL-2.0",
     "OpenSSL",
     "Unicode-DFS-2016",
+    "CC0-1.0",
+    "Artistic-2.0",
```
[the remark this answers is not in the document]
tasn: I think we need a better explanation/review of these licenses in the top-level description/comment.

bridge/svix-bridge/Cargo.toml
```diff
@@ -23,16 +23,16 @@ serde_yaml = "0.9"
...
 tracing-subscriber = { version="0.3", features=["env-filter", "fmt", "json"] }
-
-[dependencies.deno_core]
-# tagged deno@1.23.3 (does not seem to leak) though quite old.
-version = "0.142.0"
+# N.b. for newer deno versions (like this) the runtimes must be retained and reused since they will leak memory if you
+# create/drop them.
+deno_runtime = "0.119.0"
+deno_ast = "0.27.2"
+deadpool = { version = "0.9.5", features = ["unmanaged", "rt_tokio_1"] }
```
tasn, line 35: Need to explain these deps in the top-level as an OSS review.

bridge/generic-queue/src/redis.rs
tasn: Already fixed.

deny.toml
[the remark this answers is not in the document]
svix-onelson: Done.

tasn:
OSS licenses and dependencies look good to me!

svix-james (dismissed):

bridge/svix-bridge/src/runtime/mod.rs
```diff
@@ -1,45 +1,137 @@
...
+
+#[derive(Clone)]
+pub struct JsPooler {
+    executors: Pool<Executor>,
+}
+
+impl JsPooler {
     pub fn new() -> Self {
-        Self(Arc::new(Mutex::new(ThreadPool::default())))
+        const POOL_SIZE: usize = 4;
```
svix-james, line 65: Is there any merit in making this configurable or dynamic based on number of cores or something?

svix-onelson: Yeah... I guess that'd be smart.

svix-onelson: This is now exposed as `transformation_worker_count` in the config.

PR #1014, opened 2023-07-20
Explicitly allow "CC0-1.0" license in deny.toml

svix-aaron1011:
This is an explicit public domain license:
https://creativecommons.org/publicdomain/zero/1.0/

This is needed by a transitive dependency I'm adding.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -27,6 +27,7 @@ allow = [
     "Apache-2.0",
     "BSD-2-Clause",
     "BSD-3-Clause",
+    "CC0-1.0",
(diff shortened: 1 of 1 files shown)


svix-onelson:
Ran into the same thing in OSS for https://github.com/svix/svix-webhooks/pull/1007.

It's been added there and should move through with the next OSS merge, but the merge looks like it might take a minute (due to several conflicting modules).

svix-aaron1011:
This was already added in https://github.com/svix/svix-webhooks/pull/1007

PR #1015, opened 2023-07-20
Fix API version in OpenAPI spec

svix-nick:
## Motivation

The OpenAPI specification generation is not updating the API version number as we make new releases.

## Solution

This pulls in the crate version number when generating the OpenAPI specification.


server/svix-server/src/openapi.rs
```diff
@@ -41,7 +43,7 @@ pub fn initialize_openapi() -> OpenApi {
     OpenApi {
         info: openapi::Info {
             title: "Svix API".to_owned(),
-            version: "1.4".to_owned(),
+            version: VERSION.to_owned(),
```
svix-onelson, line 46: Huh, interesting.

svix-onelson (approved):

server/svix-server/src/openapi.rs
svix-onelson: And, by "interesting" I mean I wonder if the version bump script previously used to increment this, but no more, or what the breakage might have been.

Issue #1019, opened 2023-07-23
(golang) Consider `svix.NullableString()` to accept a pointer.


rkuprov:
### Version
1.7.0

### Platform
Mac

### Description
Not really a bug, but an potential oversight. Passing a pointer to string would allow distinguishing between `nil` and `""`.

tasn:
Fixed in the linked PR.

PR #1020, opened 2023-07-23
`NullableString()` now takes a pointer argument

rkuprov:
why:
- to distinguish between `""` and `nil`

## Solution

solves #1019

### DIFF
--- a/go/svix.go
+++ b/go/svix.go
@@ -37,8 +37,8 @@ var defaultHTTPClient = &http.Client{
 func String(s string) *string {
 	return &s
 }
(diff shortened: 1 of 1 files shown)


tasn (approved):

tasn:
Thanks a lot! As discussed on Slack:
> While I'd really prefer not to break library API, if it's broken currently, an easy change for our customers to make, and will be caught at compile time (so people know they'll need to change the code), I think it's potentially OK.

https://svixcommunity.slack.com/archives/C022BJ2EBHV/p1690057398340549

PR #1022, opened 2023-07-24
Update `chrono` and run Cargo update

svix-gabriel:
## Motivation

Keep things up to date.

I originally hoped these would fix a couple of SVEs, but I don't think it does.

## Solution

Run cargo-update.

### DIFF
--- a/server/svix-server/Cargo.toml
+++ b/server/svix-server/Cargo.toml
@@ -46,7 +46,7 @@ opentelemetry-otlp = { version = "0.10" }
 validator = { version = "0.14.0", features = ["derive"] }
 jwt-simple = "0.10.8"
 ed25519-compact = "1.0.11"
-chrono = { version="0.4.19", features = ["serde"] }
+chrono = { version="0.4.26", features = ["serde"] }
 reqwest = { version = "0.11.9", features = ["json", "rustls-tls", "trust-dns"], default-features = false }
 bb8 = "0.8"
 bb8-redis = "0.13.1"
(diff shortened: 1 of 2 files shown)


server/svix-server/Cargo.toml
```diff
@@ -46,7 +46,7 @@ opentelemetry-otlp = { version = "0.10" }
 validator = { version = "0.14.0", features = ["derive"] }
 jwt-simple = "0.10.8"
 ed25519-compact = "1.0.11"
-chrono = { version="0.4.19", features = ["serde"] }
+chrono = { version="0.4.26", features = ["serde"] }
```
svix-onelson, line 49: The CVE is against certain versions of `time`, and chrono have been _taking a lot of time_ to decide if they want to make a breaking change in `0.4.x` to fix it, or wait for the `0.5.x` release line to start.

https://github.com/chronotope/chrono/issues/1073

svix-onelson (approved):
lgtm regardless

server/svix-server/Cargo.toml
svix-nick: If you disable the `oldtime` feature (which is on by default for niche backwards compat cases) then it will pull in a newer version of the `time` crate. See: https://docs.rs/chrono/latest/chrono/#duration

svix-onelson: Yeah, I tried that but... I think there's something funky in cargo's features assessment where the offending `time` crate still winds up in the lockfile, and we still get dinged by dependabot.

I think it might have been because of other transitive deps on chrono, from things like seaorm.

svix-nick: Ah, good point. It looks like our ksuid library might be the culprit:
image

svix-onelson: If the recent sea_orm update removed the issue on that side, we can probably fixup ksuid (assuming it's the last dependant in the graph) 🤔

PR #1023, opened 2023-07-24
Upgrade openapi-to-postmanv2 to latest version

svix-gabriel:
## Motivation

Originally to try and fix https://github.com/svix/svix-webhooks/security/dependabot/57, but looking at the diff I'm not sure it actually fixes it.

## Solution

`yarn upgrade openapi-to-postmanv2`


svix-onelson (approved):
:shipit:

PR #1024, opened 2023-07-25
Server: Replace lazy static with once cell everywhere

svix-nick:
## Motivation

The `lazy_static` crate has not received maintenance in about four years, and has generally been deprecated in favor of `once_cell`. The latter's interface is being incorporated into the standard library.

## Solution

This PR simply swaps out `lazy_static` macro invocations for `once_cell`'s `Lazy` interface.


server/svix-server/src/lib.rs
```diff
@@ -47,9 +47,7 @@ pub mod worker;
 
 const CRATE_NAME: &str = env!("CARGO_CRATE_NAME");
 
-lazy_static! {
-    pub static ref SHUTTING_DOWN: AtomicBool = AtomicBool::new(false);
-}
+pub static SHUTTING_DOWN: Lazy<AtomicBool> = Lazy::new(|| AtomicBool::new(false));
```
svix-aaron1011, line 50: This can just be `pub static SHUTTING_DOWN: AtomicBool = AtomicBool::new(false)`, since that become a `const fn` since this code was originally written.

svix-nick: Aha! Good catch

svix-james (approved):

svix-nick:
Open source review:
`once_cell` is the new standard for lazily-initialized static values. The intent is for its API to be incorporated into the standard library. Parts of it already have been as of Rust 1.70.0.

Our existing solution `lazy_static` is not maintained, and [urges users to move to once_cell](https://github.com/rust-lang-nursery/lazy-static.rs/issues/214).

`once_cell` is dual licensed Apache 2.0 and MIT, as are all of its dependencies.

tasn:
Looks good to me!

PR #1027, opened 2023-08-02
Update openapi schema and regenerate libraries

svix-lucho:
Update the openApi spec with the latest updates, including new/updated events in `x-webhooks` and the deprecation of the endpoint `version` field.


openapi.json
```diff
@@ -1006,8 +1010,10 @@
                         "type": "string"
                     },
                     "version": {
+                        "deprecated": true,
+                        "example": 1,
                         "format": "uint16",
-                        "minimum": 0,
+                        "minimum": 1,
```
svix-onelson, line 1016: This is the `minimum: 0` mentioned by Tom

go/internal/openapi/model_endpoint_in.go
```diff
@@ -370,28 +373,46 @@ func (o *EndpointIn) SetUrl(v string) {
 	o.Url = v
 }
 
-// GetVersion returns the Version field value
+// GetVersion returns the Version field value if set, zero value otherwise (both if not set or set to explicit null).
 func (o *EndpointIn) GetVersion() int32 {
-	if o == nil {
+	if o == nil || o.Version.Get() == nil {
 		var ret int32
 		return ret
```
svix-onelson, line 380: 🥴 this Go-ism for "return zero"

svix-onelson (dismissed):
Mentioned in slack -- I wasn't sure if either of the new parameters (with content, event types) needed extra updates in the public API for Go (or any other lib).

Might want to check the wrapper code just to be sure the params will get passed through to the codegen'd stuff for Go. I could be misremembering.

svix-lucho:
> Mentioned in slack -- I wasn't sure if either of the new parameters (with content, event types) needed extra updates in the public API for Go (or any other lib).
> 
> Might want to check the wrapper code just to be sure the params will get passed through to the codegen'd stuff for Go. I could be misremembering.

Yes, you were right, there were some methods in the public API that needed updates. Updated them.

PR #1029, opened 2023-08-09
Rearrange main to allow graceful trace provider shutdown in all paths

svix-nick:
Moving the main `run()` invocation into the command match block allows us to gracefully shutdown the tracer provider no matter which path is taken. It also makes it clearer that no fall-through is intended; either a command is run or the server starts.


svix-onelson (approved):
Makes good sense to me.

Removing the `std::process::exit` is also nice because I think it allows for things like `Drop` impls to kick in where they might be skipped otherwise.

PR #1034, opened 2023-08-11
Bridge: omniqueue adoption

svix-onelson:
Replaces the initial (sort of proof-of-concept) `generic-queue` crate with `omniqueue`.

Since omniqueue offers an _opaque_ representation of the producers and consumers via the `make_dynaic()` builder option, further refactors can be made to de-bulk the consumer code.

Currently there's a distinct type for each backend, but at this point we should be able to reduce this to a single type. This is planned as a follow-up PR (time allowing).

Since this diff will be an upheaval to the dependency graph no matter what, I've included a `cargo update` to help keep things current.

> Initial work towards de-bulking: https://github.com/svix/svix-webhooks/pull/1035


bridge/svix-bridge-plugin-queue/Cargo.toml
```diff
@@ -7,7 +7,7 @@ edition = "2021"
 
 [dependencies]
 futures-lite = "1.12.0"
-generic_queue = { path = "../generic-queue" }
+omniqueue = { path = "/home/onelson/Projects/omniqueue-rs/omniqueue" }
```
svix-onelson, line 10: I'm working through some stuff 😓 
Once I know what I need, I'll put up a PR on the omniqueue side.

bridge/svix-bridge-plugin-queue/tests/sqs_consumer.rs
```diff
@@ -74,7 +84,7 @@ async fn publish(client: &Client, url: &str, payload: &str) {
 }
 
 /// General "pause while we wait for messages to travel" beat. If you're seeing flakes, bump this up.
-const WAIT_MS: u64 = 100;
+const WAIT_MS: u64 = 300;
```
svix-onelson, line 87: Not sure what accounts for this, but SQS got a lot more flaky with omniqueue. Might be able to reduce this again once "de-bulking" has happened. Fewer trait objects in the call graph might make a difference.

tasn: This is very concerning, no? As in, it feels like something is bad with omniqueue and we shouldn't release bridge with this?

bridge/svix-bridge-plugin-queue/Cargo.toml
```diff
@@ -7,7 +7,7 @@ edition = "2021"
 
 [dependencies]
 futures-lite = "1.12.0"
-generic_queue = { path = "../generic-queue" }
+omniqueue = { git = "https://github.com/svix/omniqueue-rs", rev = "247904053bcf90cf693df4429092923bf97770eb" }
```
tasn, line 10: Probably need to make a release soon?

bridge/svix-bridge-plugin-queue/tests/sqs_consumer.rs
svix-onelson: I'm not quite so concerned right now. I've been more worried about correctness during this refactor process 😅 

I'll try and step it back down now that the diff has settled.

There's a range in use across the different backends:
- gcp, 100
- rabbit, 200
- redis, 250

tbh, it's hard to say if this was related to resource starvation on my machine while building/running. SQS was the only one I saw flake though.

tasn: OK, just feels like it's worth investigating.

svix-onelson: Managed to pull it down to 250 so at least it's in line with redis 🤣 

I think it's probably okay. We can definitely do a separate pass to tighten things up on both sides, here and omniqueue.

svix-onelson: Filed https://github.com/svix/monorepo-private/issues/6140 to follow-up on this.

svix-onelson: Filed an issue to look into this further.

svix-gabriel (approved):

PR #1039, opened 2023-08-16
Add short blurb about setting jwt_algorithm to README

svix-nick:
We added some new options for the jwt signing algorithm a while back. It would be neat if they were mentioned in the readme.


svix-onelson (dismissed):

README.md
```diff
@@ -282,6 +282,10 @@ Example valid JWT for the secret `x` (so you can see the structure):
 }
 ```
 
+### Using a different hashing algorithm
```
tasn, line 285:
```suggestion
### Using a different signing algorithm
```

README.md
```diff
@@ -282,6 +282,10 @@ Example valid JWT for the secret `x` (so you can see the structure):
...
+
+As mentioned above, the default hashing algorithm for signing JWTs is `HS256`. You can select a different algorithm by setting the `jwt_algorithm` config to one of these supported values: `HS384`, `HS512`, `RS256`, `RS384`, `RS512`, or `EdDSA`.
```
tasn, line 287:
```suggestion
As mentioned above, the default algorithm for signing JWTs is `HS256`. You can select a different algorithm by setting the `jwt_algorithm` config to one of these supported values: `HS384`, `HS512`, `RS256`, `RS384`, `RS512`, or `EdDSA`. 
```

README.md
tasn: Some of them are hashing, but some aren't. Collectively they are all signing.

svix-nick: Good point, updated.

README.md
svix-nick: Removed

svix-onelson (approved):

PR #1040, opened 2023-08-16
Server: update ouroboros via sea-orm

svix-onelson:
Ouroboros is a transitive dep of SeaORM, and SeaORM bump their dependency on it in the `0.12` release.

Refs: https://github.com/svix/svix-webhooks-private/security/dependabot/31

In truth there are a handful of crates that had to be updated in concert to get the build green after bumping `sea-orm`.

- `sea-orm`, `0.11.3` -> `0.12.2`
- `sqlx`, `0.6.2` -> `0.7.1`
- `jwt-simple`, `0.10.8` -> `0.11.6`

### DIFF
--- a/server/Cargo.lock
+++ b/server/Cargo.lock
@@ -3,12 +3,6 @@
 version = 3
 
 [[package]]
-name = "Inflector"
-version = "0.11.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fe438c63458706e03479442743baae6c88256498e6431708f6dfc520a26515d3"
-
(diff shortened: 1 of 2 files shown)


tasn:
Any scary items in the changelog of these libs? 
And for SeaORM: any cool stuff we could use?

svix-onelson:
@svix-gabriel maybe you could skim the SeaORM/sqlx changelogs linked above and note if there are concerns? You've got a more well-trained eye when it comes to these.

svix-james (approved):

PR #1044, opened 2023-08-21
Allow disabling TLS verification in configuration

### LINKED ISSUE #1030: Allowing ignoring certificate authenticity
## Feature Request
Allow the use of webhook endpoints which are protected using self-signed certificates.

svix-daniel:
Fixes #1030 by allowing the `WebhookClient` to bypass certificated validation using the a new member of the configuration. A warning log is emitted upon `WebhookClient` creation when this environment variable is set to help dissuade it's use.


svix-james:
Can we test this?

Also, is there any other place we only allow config change through an environment variable? I guess it's fine, but I wonder why a "scary-looking" configuration parameter is also not ok? The users of our software are adults, and I personally don't see a reason why this can't be a proper config param like everything else.

svix-onelson:
I guess writing tests for this would be easier if there was proper config (though I'm not sure that's what prompted James' two separate points).

server/svix-server/src/core/webhook_http_client.rs
```diff
@@ -764,4 +777,61 @@ mod tests {
             "127.0.0.1:8000".as_bytes()
         );
     }
+
+    #[tokio::test]
+    async fn test_tls_verification_disable() {
```
tasn, line 782: If we are already testing this dangerous flag, maybe also have a test to test the happy path? So if it's false, that we actually fail connecting to it?

svix-daniel: That is one of the assertions in this test case already.

tasn: Ah, missed that, awesome.

svix-daniel:
A test has been added with the caveat that it requires a new dev-dependency, `axum-server`.

`axum-server` provides an alternative to the `hyper::Server`/`axum::Server` which allows for HTTPS traffic via either `openssl` or `rustls` (of which it was decided to use the `tls-openssl` feature over the `rustls` equivalent to save on compilation time and to avoid second version of `rustls` from being required).

The crate appears reputable for the following reasons:

- `cargo deny` and the `security_audit` workflow report no known vulnerabilities in this crate or its dependencies
- The crate has been recommended by a maintainer of `axum` (https://github.com/tokio-rs/axum/discussions/1032#discussioncomment-2754752) and is used in official `axum` examples (https://github.com/tokio-rs/axum/tree/main/examples/tls-rustls). `axum` is a crate that we already use and trust.
- It seems to be actively maintained with the latest commit about nine days ago as of the time of writing this comment.

In searching for a solution to allow an HTTPS-compatible server to be used for testing, no other possibilities came up. At least any that still use `axum` at their core.

With a cursory review of the source code, and a dive into the documentation, the crate seems to be of a high enough quality for our use. 

It is licensed under MIT which is compatible for all our use cases.

Finally, there are very few risks associated with the use of this dependency as it is constrained to a dev-dependency.

svix-james (approved):

tasn (approved):

Issue #1042, opened 2023-08-19
Change serde/serde_derive to >= 1.0.184 to avoid precompiled binaries


tasn:
Serde started shipping pre-compiled binaries, which is obviously not a good thing... So we'll have to pin to an old version until they undo this.

Also need to pin serde derive.

Ref: https://github.com/serde-rs/serde/issues/2538

svix-daniel:
Version 1.0.184 removes this binary blob. See https://github.com/serde-rs/serde/releases/tag/v1.0.184

tasn:
OK, make  v1.0.184 a minimum req instead.

tasn:
Add to cargo-deny if not too hard.

PR #1046, opened 2023-08-21
Avoid serde_derive blob and ban affected versions via cargo-deny

svix-daniel:
Solves #1042 by updating the version of `serde` specified in the server's `Cargo.toml` to ensure `serde_derive`'s version is not affected by the binary blob.

Similarly ensures that all versions of the crate with the blob are banned via `cargo-deny`.

### DIFF
--- a/server/Cargo.lock
+++ b/server/Cargo.lock
@@ -321,7 +321,7 @@ checksum = "16e62a023e7c117e27523144c5d2459f4397fcc3cab0085af8e2224f643a0193"
 dependencies = [
  "proc-macro2",
  "quote",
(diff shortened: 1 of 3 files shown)


server/Cargo.lock
```diff
@@ -2956,9 +2968,9 @@ dependencies = [
 
 [[package]]
 name = "redis"
-version = "0.23.0"
+version = "0.23.2"
```
tasn, line 2971: Don't update redis, this was part of the issue we had today.

Why are you updating things anyway? Just do the cargo deny to block the serde version.

PR #1048, opened 2023-08-21
Server: Add Sentry support

svix-aaron1011:
## Motivation

Sentry is a useful error-reporting service

## Solution

This adds two new config values - `sentry_dsn` and `sentry_environment`. When `sentry_dsn` is provided, we enable Sentry reporting at the provided DSN with the supplied environment name.

Currently, we report both panics (handled by default by Sentry), and all warnings/errors logged through `tracing`

### DIFF
--- a/server/svix-server/Cargo.toml
+++ b/server/svix-server/Cargo.toml
@@ -75,6 +75,8 @@ strum_macros = "0.24"
 strum = { version = "0.24", features = ["derive"] }
 form_urlencoded = "1.1.0"
 lapin = "2.1.1"
+sentry = { version = "0.31.5", features = ["tracing"] }
+sentry-tracing = "0.31.5"
 
 [dev-dependencies]
 anyhow = "1.0.56"
--- a/server/svix-server/Cargo.toml
+++ b/server/svix-server/Cargo.toml
@@ -76,7 +76,6 @@ strum = { version = "0.24", features = ["derive"] }
 form_urlencoded = "1.1.0"
 lapin = "2.1.1"
 sentry = { version = "0.31.5", features = ["tracing"] }
-sentry-tracing = "0.31.5"
 
 [dev-dependencies]
 anyhow = "1.0.56"
--- a/server/svix-server/config.default.toml
+++ b/server/svix-server/config.default.toml
@@ -39,6 +39,13 @@ log_format = "default"
 # always sending. If the OpenTelemetry address is not set, this will do nothing.
 # opentelemetry_sample_ratio = 1.0
 
+# The Sentry DSN to use for error reporting. Disabled when omitted/null
+# sentry_dsn = "https://somedsn.ingest.sentry.io/12345"
+
+# The environment that the server is running in.
+# Supported: "dev", "staging", "prod"
+environment = "dev"
+
 # Whether to enable the logging of the databases at the configured log level. This may be useful for
 # analyzing their response times.
 db_tracing = false
--- a/server/svix-server/src/cfg.rs
+++ b/server/svix-server/src/cfg.rs
@@ -342,10 +342,9 @@ pub enum DefaultSignatureType {
     Ed25519,
 }
 
-#[derive(Clone, Debug, Deserialize, Default)]
+#[derive(Clone, Debug, Deserialize)]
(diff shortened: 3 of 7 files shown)


server/svix-server/config.default.toml
```diff
@@ -39,6 +39,12 @@ log_format = "default"
 # always sending. If the OpenTelemetry address is not set, this will do nothing.
 # opentelemetry_sample_ratio = 1.0
 
+# The Sentry DSN to use for error reporting. Disabled when omitted/null
+# sentry_dsn = "https://somedsn.ingest.sentry.io/12345"
+
+# The environment name to report to Sentry (when Sentry is enabled). Uses the Sentry default when omitted/null
+# sentry_environment = "development"
```
tasn, line 46: I think we already have an `env` parameter that you can use? Though even if not let's make it a general config.

tasn: We have it in Python for sure (and it should be an enum btw, unlike python)

svix-aaron1011: I renamed this to `environment`, and made it a `dev`/`staging`/`prod` enum

server/svix-server/config.default.toml
```diff
@@ -39,6 +39,13 @@ log_format = "default"
 # always sending. If the OpenTelemetry address is not set, this will do nothing.
 # opentelemetry_sample_ratio = 1.0
 
+# The Sentry DSN to use for error reporting. Disabled when omitted/null
+# sentry_dsn = "https://somedsn.ingest.sentry.io/12345"
+
+# The environment that the server is running in.
+# Supported: "dev", "staging", "prod"
+environment = "dev"
```
tasn, line 47: Let's keep it unset here so that we are sure we set it correctly in terraform (and not forget by accident).

server/svix-server/src/lib.rs
```diff
@@ -268,6 +270,18 @@ pub fn setup_tracing(cfg: &ConfigurationInner) {
         tracing_opentelemetry::layer().with_tracer(tracer)
     });
 
+    let sentry_guard = sentry::init(sentry::ClientOptions {
```
tasn, line 273: Local dev probably shouldn't hit sentry, only deployed dev, staging, and prod. Right?

server/svix-server/Cargo.toml
```diff
@@ -75,6 +75,8 @@ strum_macros = "0.24"
 strum = { version = "0.24", features = ["derive"] }
 form_urlencoded = "1.1.0"
 lapin = "2.1.1"
+sentry = { version = "0.31.5", features = ["tracing"] }
+sentry-tracing = "0.31.5"
```
svix-onelson, line 79: The tracing crate is re-exported, looks like.
https://github.com/getsentry/sentry-rust/blob/master/sentry/src/lib.rs#L221-L224

Should be available as `sentry::integrations::tracing` when the feature is enabled.

server/svix-server/src/lib.rs
svix-onelson: Could be useful to do this when iterating on adding context data (to see how it looks on the Sentry side).

In that situation, we'd probably want to use a DSN for a garbage/sandbox project, so probably the environment doesn't matter 🤣

server/svix-server/src/cfg.rs
```diff
@@ -337,6 +342,29 @@ pub enum DefaultSignatureType {
     Ed25519,
 }
 
+#[derive(Clone, Debug, Deserialize, Default)]
+#[serde(rename_all = "lowercase")]
+pub enum Environment {
+    #[default]
```
tasn, line 348:
```suggestion
```

server/svix-server/src/lib.rs
svix-aaron1011: Sentry gets disabled when the DSN is `None`, so I don't think we need an explicit disable for dev.

tasn: Anyhow, cool, sounds good. CI is failing, but good otherwise.

svix-aaron1011:
Open source review:

`sentry` is the official Rust SDK for sentry (https://github.com/getsentry/sentry-rust). It has an Apache 2.0 license

PR #1051, opened 2023-08-22
Bridge: Use jemalloc as the global allocator, except for windows

svix-nick:
This PR changes the global allocator for svix-bridge to jemalloc. Said allocator is only enabled on non-Windows targets.

This also includes some runtime memory metrics exposed via opentelemetry.

### DIFF
--- a/bridge/svix-bridge/src/allocator.rs
+++ b/bridge/svix-bridge/src/allocator.rs
@@ -50,6 +50,7 @@ mod supported {
 
 #[cfg(any(target_env = "msvc", not(feature = "jemalloc")))]
 mod unsupported {
+    use anyhow::anyhow;
     use std::sync::Arc;
     pub struct AllocatorStatMibs;
 
@@ -61,6 +62,6 @@ mod unsupported {
     }
 
     pub fn get_allocator_stat_mibs() -> anyhow::Result<Arc<AllocatorStatMibs>> {
-        Arc::new(AllocatorStatMibs)
+        Err(anyhow!("metric collection is not supported"))
     }
 }
--- a/bridge/svix-bridge/src/allocator.rs
+++ b/bridge/svix-bridge/src/allocator.rs
@@ -35,16 +35,16 @@ mod supported {
         Ok(Some((allocated, resident)))
     }
 
-    pub fn get_allocator_stat_mibs() -> anyhow::Result<Option<Arc<AllocatorStatMibs>>> {
+    pub fn get_allocator_stat_mibs() -> anyhow::Result<Arc<AllocatorStatMibs>> {
         let e = epoch::mib()?;
         let allocated = stats::allocated::mib()?;
         let resident = stats::resident::mib()?;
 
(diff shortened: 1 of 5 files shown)


bridge/svix-bridge/src/allocator.rs
```diff
@@ -0,0 +1,66 @@
...
+        _bust_cache: bool,
+        _mibs: Arc<AllocatorStatMibs>,
+    ) -> anyhow::Result<Option<(usize, usize)>> {
+        Ok(None)
+    }
+
+    pub fn get_allocator_stat_mibs() -> anyhow::Result<Option<Arc<AllocatorStatMibs>>> {
+        Ok(None)
+    }
+}
```
svix-onelson, line 66: I think we can probably get rid of the `Option` from some of these signatures, which will simplify the callsites further.

Since the `unsupported` impl can return a unit-struct for `AllocatorStatMibs` here, and in `get_allocator_stats` we can keep returning `Ok(None)`.

It might just remove one `if let` each time. What do you think?

bridge/svix-bridge/src/main.rs
```diff
@@ -2,18 +2,36 @@ use self::config::Config;
...
 
+use crate::allocator::{get_allocator_stat_mibs, get_allocator_stats};
+use crate::metrics::CommonMetrics;
+
+#[cfg(all(not(target_env = "msvc"), feature = "jemalloc"))]
+#[global_allocator]
+static GLOBAL: Jemalloc = Jemalloc;
+
+#[cfg(all(target_env = "msvc", feature = "jemalloc"))]
+compile_error!("jemalloc cannot be enabled on msvc");
```
svix-onelson, line 33: I don't remember if the line ordering matters (I think it does?) but if we ensure this `compiler_error!()` invocation happens as early as possible, all the rest of the `cfg` attributes can simplify down to _just the feature_, I think.

Does that make sense?
Maybe this sort of trick only makes sense from a `build.rs` -- something that happens before the main sources are processed. 🤔

svix-onelson (approved):
Looks great! Some notes for _possible changes_ but nothing blocking.

bridge/svix-bridge/src/allocator.rs
svix-nick: I like it, but the functions are still fallible, so there has to be a pattern match in there somewhere.

I got a little weirder with it. Let me know what you think.

bridge/svix-bridge/src/main.rs
svix-nick: I flipped the target_env to "gnu" and simplified the other conditions to try this out. Unfortunately it results in a variety of unresolved import errors.

Putting it in a build script works, but is still kind of a weird IDE experience. Compilation does exit with one nice clean error, but the unresolved import errors remain and _look_ like they're being used in the code.

The more complicated conditionals appear to be the best solution at the moment.

svix-onelson: Boo.

svix-onelson (approved):

PR #1052, opened 2023-08-22
Update cargo deny config to temporarily ignore webpki advisory

svix-nick:
To be removed once webpki has a safe release to upgrade to

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -19,6 +19,10 @@ ignore = [
 
 	# TODO: Eliminate `chrono` crate when `time` update is possible.
     "RUSTSEC-2020-0159",
+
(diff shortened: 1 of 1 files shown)


svix-james (approved):

tasn:
There's a dependabot thing to update it, why not just update it? https://github.com/svix/svix-webhooks/pull/1050

svix-nick:
I can absolutely do that, but it'll only solve one of the advisories. A fix for `webpki` is still outstanding.

tasn:
Let's merge that one and remove the deny for that one, and sure, let's keep the pki one for now. Just better to address the updates than to silence them.

PR #1053, opened 2023-08-22
Update rustls-webpki

svix-nick:
This PR updates the rustls-webpki crate in both bridge and server to avoid [RUSTSEC-2023-0053](https://rustsec.org/advisories/RUSTSEC-2023-0053.html)


svix-onelson (approved):
:shipit:

tasn:
There was already an automatically created dependabot PR, but cool, we can start with this. :)

PR #1054, opened 2023-08-23
Remove ctx family of macros in favor of track_caller

svix-nick:
This PR alters the way we're doing error propagation and backtrace handling.

Currently we use a family of macros that create and modify errors using the `file!` and `line!` std macros. Our primary `ctx!` macros embeds the location information into the call sites at compile time.

This PR instead piggybacks off of the `#[track_caller]` proc macro used by rust's panic system. This embeds the caller information into the binary in a similar way, but allows us to annotate a chain of functions directly instead of a chain of macros. That means that we can do the backtrace assembly automatically as part of the From impl, which means a lot fewer manual tracing invocations.

I couldn't think of a way to automatically provide another trace line when we're already propagating one of our own errors. This means that a manual call to `trace()` is still required in a few places unfortunately.

If there's a downside to this, I would say that reducing the tracing calls (formerly ctx! and now trace()) from mandatory to sometimes means that it's not always clear if it's necessary or desired to add one.


tasn:
> If there's a downside to this, I would say that reducing the tracing calls (formerly ctx! and now trace()) from mandatory to sometimes means that it's not always clear if it's necessary or desired to add one.

Yeah, that's actually a concern I'd say. I wonder if there's a way to hack a fix for that too?

svix-nick:
About the closest I was able to get was the unstable `Try` trait, which might let us do funky things with the `?` operator in the future. Not an option at the moment, alas.

Primarily I think it's a problem of exposure. The ergonomics of the tracing haven't changed much; developers should still attempt to call `trace()` on any Results they produce, like they should with `ctx!`. It's just that most of those calls are automated now, so people might not remember due to not seeing the macro all over the place.

tasn:
@svix-nick, yeah, that's my concern. I'd almost say it's a concerning regression. Though I guess the benefits everywhere else outweigh the potential gap in tracing. So it's probably fine.

Do we still have full backtraces with errors?

svix-nick:
We've got the same sort of traces as we had before. Not full traces, just from the error's origin to wherever it was logged.

server/svix-server/src/error.rs
```diff
@@ -31,57 +28,71 @@ pub type Result<T> = std::result::Result<T, Error>;
...
     // the file name and line number of the error. Used for debugging non Http errors
-    pub trace: Vec<&'static str>,
+    pub trace: Vec<&'static Location<'static>>,
     pub typ: ErrorType,
 }
 
 impl Error {
-    pub fn generic(s: impl fmt::Display, location: &'static str) -> Self {
+    #[track_caller]
+    pub fn generic(s: impl fmt::Display) -> Self {
```
svix-gabriel, line 37: In general, I like this change. But there are a few places that have gotten less ergonomic, not more.

E.g.

```rust
.map_err(|e| err_generic!("Error parsing message timestamp {e:?}"))?;
```

becoming

```rust
.map_err(|e| Error::generic(format!("Error parsing message timestamp: {e:?}")))?;
```

Since `Error::generic` is now capturing the callsite automatically, what do you think about also changing these functions (`Error::generic`, `Error::database`, etc) to instead take `s: impl fmt::Debug`

And then
```rust
.map_err(|e| err_generic!("Error parsing message timestamp {e:?}"))?;
```

just becomes
```rust
.map_err(Err::generic)?;
```

Which is a helluva lot more ergonomic, and essentially doesn't lose any information (since we have the Location)

server/svix-server/src/error.rs
```diff
@@ -34,34 +34,34 @@ pub struct Error {
 
 impl Error {
     #[track_caller]
-    pub fn generic(s: impl fmt::Display) -> Self {
+    pub fn generic(s: impl fmt::Debug) -> Self {
```
tasn, line 37: @svix-aaron1011, is literally doing the opposite at the moment internally. 😂

svix-nick: I had noticed a trend towards logging Debug representations, so I went with @svix-gabriel 's suggestion. I'm not sure it gains us a lot of information, but I also don't think it hurts us.

Before I go changing a bunch of call sites, do we want to go with Debug or Display formatting by default for these messages?

My (obviously pretty weak) opinion would be to go with Display and leave more complicated formatting up to the caller, but that does make some places kinda chunky.

svix-nick: Looking at @svix-aaron1011's PR, it seems like Debug-by-default does hurt us in some contexts. I'm going to undo that commit unless there are dissenting opinions.

svix-gabriel: 😂 - I like the detail of `Debug` representations (and occasionally I find them more useful), but I understand the argument for `Display`.

So I won't push back either way

server/svix-server/src/error.rs
```diff
@@ -31,57 +28,71 @@ pub type Result<T> = std::result::Result<T, Error>;
...
+            trace: vec![Location::caller()],
             typ: ErrorType::Cache(s.to_string()),
         }
     }
 
-    fn init_trace(location: &'static str) -> Vec<&'static str> {
-        let mut trace = Vec::with_capacity(10); // somewhat arbitrary capacity, but avoids reallocation when building an error trace later on
-        trace.push(location);
-        trace
+    #[track_caller]
```
svix-gabriel, line 84: Sorry, I just realized this, but we don't need `#[inline(always)]` here as well?

svix-nick: Nope, that's not necessary for `track_caller` to work

svix-gabriel (approved):

PR #1055, opened 2023-08-24
Bridge: Replace lazy_static macro with once_cell

svix-nick:
See also #1024 

lazy_static is unmaintained, once_cell is the new hotness. This PR swaps out the one lazy_static invocation for the equivalent lazy construct from once_cell.

### DIFF
--- a/bridge/Cargo.lock
+++ b/bridge/Cargo.lock
@@ -5754,7 +5754,7 @@ dependencies = [
  "envsubst",
  "http",
  "hyper 0.14.27",
- "lazy_static",
+ "once_cell",
  "opentelemetry",
  "opentelemetry-http",
  "opentelemetry-otlp",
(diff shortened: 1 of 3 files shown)


svix-nick:
:star: Open source review :star:

Might not be necessary, but given that this is technically a different project, here's what I said last time:

`once_cell` is the new standard for lazily-initialized static values. The intent is for its API to be incorporated into the standard library. Parts of it already have been as of Rust 1.70.0.

Our existing solution `lazy_static` is not maintained, and [urges users to move to once_cell](https://github.com/rust-lang-nursery/lazy-static.rs/issues/214).

`once_cell` is dual licensed Apache 2.0 and MIT, as are all of its dependencies.

svix-onelson:
OSS review approved!

PR #1056, opened 2023-08-25
Update Docker image

jaymell:
Update Rust/Debian versions.


svix-gabriel (dismissed):
I think we can do bridge too, no?

svix-gabriel (approved):

PR #1060, opened 2023-08-30
Server/Bridge: enable arm64 docker builds.

tasn:
We want to also release arm64 images for both server and bridge. These are popular targets and we should be building for them.

The builds take forever, but at least it's arm64 + amd64, and it only runs on release anyway.


.github/workflows/bridge-release.yml
```diff
@@ -59,6 +59,12 @@ jobs:
   docker:
     name: release docker
     runs-on: ubuntu-latest
+    strategy:
```
tasn, line 62: I think maybe the matrix makes it that the later one overrides the first.

PR #1063, opened 2023-08-30
Fix Docker Images

### LINKED ISSUE #1061: Latest svix docker image is broken
After pulling the latest svix docker image, the svix server no longer starts due to a missing shared object dependency, see below:

```shell
$ cd ~/dev/svix-webhooks/server

$ docker pull svix/svix-server
Using default tag: latest
latest: Pulling from svix/svix-server
Digest: sha256:c0ff61a897907074395f34dd7e575e4c03d7904e30963465843a0ef497686abf
Status: Image is up to date for svix/svix-server:latest
docker.io/svix/svix-server:latest

$ docker compose up -d
[+] Running 7/7
 ⠿ Network server_default         Created                                                                                                                                                                                                                                           0.1s
 ⠿ Volume "server_postgres-data"  Created                                                                                                                                                                                                                                           0.0s
 ⠿ Volume "server_redis-data"     Created                                                                                                                                                                                                                                           0.0s
 ⠿ Container server-redis-1       Started                                                                                                                                                                                                                                           0.9s
 ⠿ Container server-postgres-1    Started                                                                                                                                                                                                                                           0.9s
 ⠿ Container server-pgbouncer-1   Started                                                                                                                                                                                                                                           1.1s
 ⠿ Container server-backend-1     Started

$ docker compose ps --all
NAME                 IMAGE                      COMMAND                  SERVICE             CREATED              STATUS                        PORTS
server-backend-1     svix/svix-server           "/bin/sh -c 'set -ex…"   backend             About a minute ago   Exited (127) 59 seconds ago   
server-pgbouncer-1   edoburu/pgbouncer:1.15.0   "/entrypoint.sh /usr…"   pgbouncer           About a minute ago   Up About a minute (healthy)   0.0.0.0:8079->5432/tcp, :::8079->5432/tcp
server-postgres-1    postgres:13.4              "docker-entrypoint.s…"   postgres            About a minute ago   Up About a minute             5432/tcp
server-redis-1       redis:7-alpine             "docker-entrypoint.s…"   redis               About a minute ago   Up About a minute             0.0.0.0:8078->6379/tcp, :::8078->6379/tcp

$ docker compose logs backend
server-backend-1  | + [ ! -z true ]
server-backend-1  | + WAIT_FOR_ARG=--wait-for 15
server-backend-1  | + exec svix-server --run-migrations --wait-for 15
server-backend-1  | svix-server: error while loading shared libraries: libssl.so.1.1: cannot open shared object file: No such file or directory
```

jaymell:
Images built on Debian 12 were failing due to openssl lib issue. Upgrading to Debian 12 will require more significant changes to our dependencies, so for now use `bullseye`-based images, which will keep our images on latest 11.x release.

Fixes #1061

### DIFF
--- a/bridge/Dockerfile
+++ b/bridge/Dockerfile
@@ -49,7 +49,7 @@ RUN touch */src/lib.rs && touch */src/main.rs
 RUN cargo build --release --frozen
 
 # Production
-FROM debian:12.1-slim AS prod
+FROM debian:bullseye-slim AS prod
 
 RUN set -ex ; \
         mkdir -p /app ;\
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -27,7 +27,7 @@ COPY . .
 RUN cargo build --release --frozen
 
 # Production
-FROM debian:12.1-slim AS prod
+FROM debian:bullseye-slim AS prod
 
 RUN set -ex ; \
         mkdir -p /app ;\


bridge/Dockerfile
```diff
@@ -49,7 +49,7 @@ RUN touch */src/lib.rs && touch */src/main.rs
 RUN cargo build --release --frozen
 
 # Production
-FROM debian:12.1-slim AS prod
+FROM debian:bullseye-slim AS prod
```
tasn, line 52:
```suggestion
FROM debian:11.7-slim AS prod
```
1. I think also being generic is maybe not too awesome, as we can't know what's actually built because of docker cache. I think being explicit is better.
2. Also, numbers are clearer than debian names.

tasn:
Can you also link to the ticket it fixes please?

bridge/Dockerfile
jaymell: Except in this case we are already using debian names in the Rust image, so making the debian image match that IMO *is* clearer.

tasn: Fair. It doesn't solve the issue about not knowing which version, but that's god enough.

PR #1066, opened 2023-08-31
Libs: bump spec, regenerate, add methods for `EventType.importOpenApi()`

svix-onelson:
This work has been split into 2 revs to assist the reviewer 🤓 

- The first commit updates the spec and regenerates the OpenAPI Go sources
- The second commit adds new methods and functions to the public APIs in each lib for `EventType.importOpenApi()`.


go/internal/openapi/api_environment.go
```diff
@@ -1,9 +1,9 @@
...
+ * No description provided (generated by Openapi Generator https://github.com/openapitools/openapi-generator)
```
svix-james, line 4: Why are all these comments changing? I have no idea of the practical use of them in the first place, but is "No description provided" what we want now?

svix-james (approved):

go/internal/openapi/api_environment.go
svix-james: Fair enough.

PR #1071, opened 2023-09-05

svix-lucho:


openapi.json
```diff
@@ -3295,6 +3302,13 @@
                     "description": {
                         "type": "string"
                     },
+                    "featureFlag": {
```
tasn, line 3305: This is wrong, I explained more over chat.

svix-james (approved):

PR #1073, opened 2023-09-08
README: improve documentation for using operational webhooks.

tasn:


README.md
```diff
@@ -284,17 +284,58 @@ Example valid JWT for the secret `x` (so you can see the structure):
...
-To turn operational webhooks on, set the `operational_webhook_address` config to point to your Svix server, and create a JWT for the special account.
-Once those are set, create an `Application` with the `uid` set to the `org_id` you're interested in, and add `Endpoint`s for all of the events you'd like to subscribe to.
+The operational webhooks utilize Svix, and are controlled by a special account service account with the following ID: `org_00000000000SvixManagement00`.
 
-For example, for the default account, just create an app with the `uid` set to `org_23rb8YdGqMT0qIzpgGwdXfHirMu`.
+The first step is to turn it on by setting the `operational_webhook_address` config to point to your Svix server. The most common value for this setting is `http://127.0.0.1:8071/`, though it may be different based on your specific setup.
+
+The above step enables operational webhooks on this instance, and the next step is to enable it for your specific organization. As mentioned above, operational webhooks use a normal Svix account behind the scenes, so we'll first need to get the authentication token for this account, to do this you should run:
```
svix-gabriel, line 297:
```suggestion
The above step enables operational webhooks on this instance, and the next step is to enable it for your specific organization. As mentioned above, operational webhooks use a normal Svix account behind the scenes, so we'll first need to get the authentication token for this account. To do this you should run:
```

README.md
```diff
@@ -284,17 +284,58 @@ Example valid JWT for the secret `x` (so you can see the structure):
...
+  -H 'Authorization: Bearer op_webhook_token_123' \
+  -H 'Accept: application/json' \
+  -H 'Content-Type: application/json' \
+  -d '{
+        "name": "Operational webhook for default org",
+        "uid": "org_23rb8YdGqMT0qIzpgGwdXfHirMu"
+    }'
+```
+
+This is it, we now have operational webhooks enable for the default account. The only thing left is adding an endpoint where the operational webhooks are going to be sent to. For example:
```
svix-gabriel, line 319:
```suggestion
This is it, we now have operational webhooks enabled for the default account. The only thing left is adding an endpoint where the operational webhooks are going to be sent to. For example:
```

README.md
```diff
@@ -284,17 +284,58 @@ Example valid JWT for the secret `x` (so you can see the structure):
...
+  -d '{
+        "url": "https://operational-webhook-destination.com/webhook/",
+        "filterTypes": [
+          “endpoint.updated”,
+          “endpoint.deleted”
+        ],
+    }'
+```
+
+Please note, how we again use the org ID of the default account as the `app_id` (or rather `uid` in this case), when calling this function.
```
svix-gabriel, line 336:
```suggestion
Note how we use the org ID of the default account as the `app_id` (or rather `uid` in this case), when creating an endpoint.
```

Not a required change, but the original sentence read a bit weird to me.

README.md
```diff
@@ -284,17 +284,58 @@ Example valid JWT for the secret `x` (so you can see the structure):
...
+
+That's it. You should now have working operational webhooks. If you ever want to create a new endpoint, modify this existing one, or anything else, you just need to again get a JWT for the service account, and then use it like you would use any other Svix account.
```
svix-gabriel, line 338:
```suggestion
That's it. You should now have working operational webhooks. If you ever want to create a new endpoint, or modify an existing endpoint, you just need to generate a JWT for the service account, and then use the JWT like you would use any other Svix account.
```

Also not a required change. Some of the language just read a bit weird to me.

tasn:
Accepted all of the suggestions.

svix-gabriel (approved):

Issue #964, opened 2023-06-19
(Ruby) Svix::ApiError Error in the HTTP2 framing layer


rolznz:
### Version

0.85.1 (Ruby)

### Platform

Debian 5.10.136-1~deb10u3 (2022-09-06)(5.10.0-0.deb10.17-amd64)

### Description

~~After updating Svix to 0.85.1 from 0.82.1 we sometimes see an error being reported when `SVIX_CLIENT.message.create` is called.~~ Edit: the error seems more about concurrent usage or increased traffic on the server rather than a specific Svix version.

```
Svix::ApiError
Error in the HTTP2 framing layer
HTTP status code: 0
```

tasn:
Whoops, interesting. And it doesn't happen with 0.82.1 at all? Even if you downgrade back now?

I wonder if we also updated a dependency that made it fail.

rolznz:
@tasn we will downgrade and let you know if the issue persists or not.

tasn:
Thanks! Let me know and we'll start investigating. The reason why we need this verification is that it only happens sometimes which means it'll be hard to reproduce.

Btw, please *only* downgrade Svix (and whatever dependencies it'll autodowngrade), so that we can make sure it's just the Svix version change that caused it.

rolznz:
@tasn Sorry for the delay. I have downgraded only Svix but the change hasn't been deployed yet. I will keep you updated.

rolznz:
@tasn unfortunately the downgrade didn't fix it. It must have been some of the other changes that also went in the same release.

Thanks for the support!

tasn:
With pleasure. Let me know if there's anything we can do to help!

rolznz:
@tasn we are wondering if it is because we create a static client rather than one client per request. Do you recommend creating a new client for each request, or is it able to handle multiple requests simultaneously?

tasn:
Let me reopen this then.

Some people do a client per request, some a long-lived, it varies. I don't know how many do either in Ruby. Are you able to reproduce it with reusing a client but not when creating one per request?

rolznz:
@tasn thanks. I will let you know once we have deployed the change and confirmed if the issue improves or not

tasn:
Thanks!

tasn:
OK, thanks, and weird! What version of Ruby are you using? Not the svix lib, actual Ruby.

rolznz:
[...]

PR #1077, opened 2023-09-18
fix: retry if response code is 0

rolznz:
## Motivation

Currently if there is a general network error, there is no retry to send the message. This fix is suggested here: https://github.com/svix/svix-webhooks/issues/964#issuecomment-1722595011

## Solution

if there is a general network error the response code is 0. In this case we should still retry.

Note: there is no guarantee the second retry will work. Maybe it would be nice to be able to configure the number of retries, if this fix does indeed improve the original issue.

There seem to be no specs for this part of the code at all, please could you confirm whether a test needs to be added here.

### DIFF
--- a/ruby/templates/api_client_typhoeus_partial.mustache
+++ b/ruby/templates/api_client_typhoeus_partial.mustache
@@ -9,7 +9,7 @@
       request = build_request(http_method, path, opts)
       response = request.run
 
-      # retry 500s
+      # retry on 5XX and general network errors
       sleep_time = 0.05
       retry_count = 0
       (0..1).each do |n|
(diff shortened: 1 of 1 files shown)


ruby/templates/api_client_typhoeus_partial.mustache
```diff
@@ -13,7 +13,7 @@
       sleep_time = 0.05
```
svix-lucho, line 13: can you update this comment to `# retry 500s and network errors` ?

svix-lucho:
LGTM. Just had a small comment.

svix-lucho (dismissed):

ruby/templates/api_client_typhoeus_partial.mustache
rolznz: @svix-lucho updated, thanks!

PR #1087, opened 2023-10-02
Bump version and update changelog.

svix-onelson:
## Motivation

Server CI was failing because of "no space left on device" while building test binaries via `run-tests.sh`.

## Solution

Bump to the latest version of the rust cache action.

Updating to the latest rust-cache seems to reset the starting disk to `70Gb` which leaves around `14Gb` for the rest of the job to run.

This may just be a bandaid, perhaps requiring us to "bump the cache prefix" in the future to shed accumulated artifacts. Unsure.

Still, bumping the action seems to solve the immediate need.

### DIFF
--- a/.github/workflows/server-ci.yml
+++ b/.github/workflows/server-ci.yml
@@ -27,9 +27,9 @@ jobs:
         override: true
         profile: minimal
         components: clippy, rustfmt
-    - uses: Swatinem/rust-cache@v1
+    - uses: Swatinem/rust-cache@v2
       with:
-        working-directory: server
+        workspaces: "server -> target"
 
     - name: rustfmt
       uses: actions-rs/cargo@v1


svix-onelson:
I added a `df -h` to the `run-tests.sh` script.

The stable job is showing
```
Filesystem      Size  Used Avail Use% Mounted on
/dev/root        84G   77G  6.3G  93% /
...
```

while the beta job shows:

```
Filesystem      Size  Used Avail Use% Mounted on
/dev/root        84G   71G   13G  85% /
...
```

Running again, the disk usage started larger on the beta run rather than the stable run as before.

Running `du -hs target/*` shows identical output for both stable and beta:

```
4.0K	target/CACHEDIR.TAG
4.4G	target/debug
4.0K	target/tmp
```


I'm not sure what accounts for the difference, but I wonder if it could be data held by `Swatinem/rust-cache@v1`. I see there's a v2 for this action, and I'll try bumping it since it should clean up some deprecation warnings from the GH side.

tasn (approved):

PR #1094, opened 2023-10-05
Libs(Ruby) Accept options on message get

amerritt14:
## Motivation

This PR adds the ability to pass options in the `message.get` calls. [The Documentation](https://api.svix.com/docs#tag/Message/operation/v1.message.get) shows that the "with_content" option is valid, but the ruby wrapper made no accommodation to pass the option.


tasn (approved):

tasn:
Thanks a lot!

PR #1095, opened 2023-10-09
Improve server onboarding

abooij:
## Motivation

I wanted to play around with the server codebase, but got stuck while following the instructions.

## Solution

I tried to fix any problems I found, such as the port numbers in `development.env` not matching the port numbers of the docker compose configs.

I also reorganized the docker compose files, deduplicating configuration using `include` syntax. It now exposes PG and Redis ports by default, i.e. the `docker-compose.override.yml` file is no longer needed. The auth token is now set via `docker-compose.docker-image.yml`. The instructions now use the new [v2 `docker compose` syntax](https://docs.docker.com/compose/migrate/).

I combined some steps in the `README` that belong together, so that there is a clear sequence of steps:
1. install rust toolchain and dependencies
2. build project
3. start ancillary services using `docker compose`
4. start the server

### DIFF
--- a/server/docker-compose.docker-image.yml
+++ b/server/docker-compose.docker-image.yml
@@ -1,3 +1,4 @@
+version: "3.7"
 include:
   - docker-compose.base.yml
 services:
--- a/server/docker-compose.testing.yml
+++ b/server/docker-compose.testing.yml
@@ -1,3 +1,4 @@
+version: "3.7"
 include:
   - docker-compose.base.yml
 services:


server/README.md
```diff
@@ -23,92 +23,82 @@ For information on how to use this server please refer to the [running the serve
...
 
 Now generate an auth token, you can do it by running:
-```
+```sh
 cargo run jwt generate
 ```
+Copy-paste this value into the value of `SVIX_JWT_SECRET` in `.env`.
 
-See [the main README](../README.md) for instructions on how to generate it in production.
+See [the main README](../README.md) for instructions on how to generate the auth token in production.
```
abooij, line 72: TODO fixme these instructions aren't right

tasn:
Thanks a lot! CI is failing, can you please take a look?

abooij:
Can't reproduce this locally, so I'm counting on the failure being a Docker Compose file version thing. (Newer versions of Docker Compose largely ignore this file version, which would explain why it Works For Me (tm).)

tasn:
Thanks!

Btw, I see you changed some of the ports and how hey are exposed. This is a thing we do on purpose as we have many services locally and we want to make sure we expose the ports in a way that don't clash. I think it's worth bringing this back.

abooij:
Hmm, I've thought about the ports issue for a bit. To be honest, I don't see a canonical way out there. Ideally, the Docker Compose `include` feature would support amending definitions in the `services` section, but that simply raises an error:

```
imported compose file [/home/auke/svix-webhooks/server/docker-compose.base.yml] defines conflicting service postgres
```

I have a branch that fixes the CI issues, but we should first figure out how to orchestrate port opening.

PR #1098, opened 2023-10-10
Libs(Go): corrected name + req assign

plarsson:
## Motivation

The endpoint transformation update didn't assign the request (payload) correctly, resulting in an error.

## Solution

Assign the request

## Misc

Corrected spelling of func name


tasn (approved):
Whoops, nice catch!

PR #1100, opened 2023-10-11
Add `messagecontent` model

jaymell:
Move message content from `message.payload` to its own table. Data is stored as `bytea` for future flexibility. Payloads will be queried from the new table but will fall back to the existing column, which has been renamed in code for clarity. The `expired_message_cleaner` process has been updated to expire records from both tables.

### DIFF
--- a/server/svix-server/migrations/20231010170520_messagecontent_table.up.sql
+++ b/server/svix-server/migrations/20231010170520_messagecontent_table.up.sql
@@ -1,0 +1,11 @@
+CREATE TABLE messagecontent (
+    id character varying NOT NULL COLLATE pg_catalog."C",
+    created_at timestamp with time zone NOT NULL,
+    payload bytea NOT NULL,
+    expiration timestamp with time zone NOT NULL DEFAULT now() + INTERVAL '90 day' 
+);
+
+ALTER TABLE ONLY messagecontent
+    ADD CONSTRAINT pk_messagecontent PRIMARY KEY (id);
+
+CREATE INDEX messagecontent_expiry ON messagecontent (expiration);
--- a/server/svix-server/src/db/models/messagecontent.rs
+++ b/server/svix-server/src/db/models/messagecontent.rs
@@ -21,9 +21,7 @@ pub enum Relation {
     #[sea_orm(
         belongs_to = "super::message::Entity",
         from = "Column::Id",
-        to = "super::message::Column::Id",
-        on_update = "NoAction",
-        on_delete = "Cascade"
+        to = "super::message::Column::Id"
     )]
     Message,
 }
--- a/server/svix-server/src/db/models/messagecontent.rs
+++ b/server/svix-server/src/db/models/messagecontent.rs
@@ -1,0 +1,55 @@
+// SPDX-FileCopyrightText: © 2022 Svix Authors
+// SPDX-License-Identifier: MIT
+
+use crate::core::types::MessageId;
+use chrono::Utc;
+use sea_orm::entity::prelude::*;
+use sea_orm::ActiveValue::Set;
+
+#[derive(Clone, Debug, PartialEq, Eq, DeriveEntityModel)]
+#[sea_orm(table_name = "messagecontent")]
+pub struct Model {
+    #[sea_orm(primary_key, auto_increment = false)]
+    pub id: MessageId,
+    pub created_at: DateTimeWithTimeZone,
+    pub payload: Vec<u8>,
+    pub expiration: DateTimeWithTimeZone,
+}
+
+#[derive(Copy, Clone, Debug, EnumIter, DeriveRelation)]
+pub enum Relation {
+    #[sea_orm(
+        belongs_to = "super::message::Entity",
+        from = "Column::Id",
+        to = "super::message::Column::Id",
+        on_update = "NoAction",
+        on_delete = "Cascade"
+    )]
+    Message,
+}
+
+impl Related<super::message::Entity> for Entity {
+    fn to() -> RelationDef {
+        Relation::Message.def()
+    }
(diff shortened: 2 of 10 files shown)


server/svix-server/migrations/20231010170520_messagecontent_table.up.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE messagecontent (
+    id character varying NOT NULL COLLATE pg_catalog."C",
+    created_at timestamp with time zone NOT NULL,
+    payload bytea NOT NULL,
```
svix-onelson, line 4: You mentioned in the PR description you chose `bytea` for future flexibility. Is there some advantage over just using a text field since JSON is utf-8 (I thought)?

server/svix-server/migrations/20231010170520_messagecontent_table.up.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE messagecontent (
+    id character varying NOT NULL COLLATE pg_catalog."C",
```
svix-onelson, line 2: Does this share an id with the corresponding message? I sort of expected a nullable fk added on the message side, but I suppose that's not needed if the ids just match.

server/svix-server/src/db/models/messagecontent.rs
```diff
@@ -0,0 +1,55 @@
...
+}
+
+#[derive(Copy, Clone, Debug, EnumIter, DeriveRelation)]
+pub enum Relation {
+    #[sea_orm(
+        belongs_to = "super::message::Entity",
+        from = "Column::Id",
+        to = "super::message::Column::Id",
+        on_update = "NoAction",
+        on_delete = "Cascade"
```
svix-onelson, line 26: Wonder how this works without the FK to connect the two tables.
I _thought_ cascade deletes required a trigger if you didn't have a constraint in place.

server/svix-server/src/v1/endpoints/attempt.rs
```diff
@@ -246,19 +247,37 @@ async fn list_attempted_messages(
...
+    let msg_ids = dests_and_msgs
+        .iter()
+        .filter_map(|(_, msg)| msg.as_ref())
+        .map(|msg| msg.id.clone())
+        .collect::<Vec<MessageId>>();
+
+    // `find_also_related` can't be used multiple times in a query, so rather
+    // than build a complicated custom query, just query all the content
+    // data separately. Hopefully this isn't too painful:
+    let mut msg_content_map = messagecontent::Entity::secure_find_by_id_in(msg_ids)
```
svix-onelson, line 261: Seems right to me. 👍🏻

server/svix-server/src/v1/endpoints/message.rs
```diff
@@ -182,25 +186,32 @@ pub struct MessageOut {
...
+    pub fn from_msg_and_payload(model: message::Model, content: Option<Vec<u8>>) -> Self {
+        // Thankfully it can never fail :)
+        let payload = content
+            .map(|p| serde_json::from_slice(&p).expect("Can never fail"))
+            .or(model.legacy_payload);
+        let payload = RawPayload::from_string(match payload {
+            Some(payload) => serde_json::to_string(&payload).expect("Can never fail"),
+            None => r#"{"expired":true}"#.to_string(),
+        })
+        .expect("Can never fail");
```
svix-onelson, line 198: This is `MessageOut` so it's probably true, but last time I checked we had similar comments/remarks in the ingest handler when consuming `MessageIn` about "can never fail" even while json parsing _can fail_ when the structure is very deep, for example.

svix-onelson (commented):
What's the test plan for this?

server/svix-server/src/db/models/messagecontent.rs
jaymell: I'm sure you're right. This was mostly a copy-paste of other models which generally have the same setting, and IIRC, we don't have _any_ FK constraints in the DB.

server/svix-server/migrations/20231010170520_messagecontent_table.up.sql
jaymell: Yeah, the IDs match. Seemed simple to me.

server/svix-server/migrations/20231010170520_messagecontent_table.up.sql
jaymell: The possibility is that we could actually store binary data in the future, hence the binary type.

server/svix-server/src/db/models/messagecontent.rs
jaymell: Removed entirely, since it's pointless anyway.

svix-james:
> What's the test plan for this?

Updated related test to check the new table, not the old column.

svix-onelson (dismissed):
:shipit:

server/svix-server/src/v1/endpoints/message.rs
jaymell: Yeah, I honestly didn't feel comfortable w/ this and was hoping someone would call me out on it :stuck_out_tongue:. I still don't feel great about the `expect`s, but I went ahead and removed the most egregious one.

svix-onelson (approved):

server/svix-server/src/v1/endpoints/message.rs
svix-onelson: More conservative is better here, yeah. I like your change. I'd normally only resort to the `expect()` if the containing scope forced me to make the body infallible, but that isn't the case here. 👍🏻

svix-aaron1011 (approved):

PR #1113, opened 2023-11-01
Allow configurable User-Agent Header

igavriil:
## Motivation

The User-Agent header in HTTP webhooks serves a pivotal role in identifying
the source of the request. 
For any platform sending out webhooks, it's crucial for the recipients to 
recognize and authenticate the origin of the requests they are processing.

## Solution

Add optional `user_agent_header` in server configuration

### DIFF
--- a/server/svix-server/config.default.toml
+++ b/server/svix-server/config.default.toml
@@ -90,6 +90,9 @@ cache_type = "memory"
 # If true, headers are prefixed with `Webhook-`, otherwise with `Svix-` (default).
 whitelabel_headers = false
 
+# Custom User-Agent header
+# user_agent_header = "User Agent"
+
 # If true, only allow https endpoints, otherwise also allow http.
 endpoint_https_only = false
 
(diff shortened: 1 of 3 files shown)


tasn:
Hey @igavriil, I'd love to hear more about your exact use-case. Where are you using Svix where this has been needed?

igavriil:
When someone receives a webhook from a service they expect all the headers to be somehow related to this service.  User-Agent falls in this category and having this header set to svix can create trust issues. This is not a specific usecase but more a general trust issue

tasn:
@igavriil, thanks for sharing more context. I don't think it creates trust issues, `nginx`, caddy, and every web server send their info from the `server` header. The same goes with other types of software, such as: web clients, web libraries, email clients, servers, libraries, and more. This is useful information for debugging (getting the software name and version), and also a tiny shoutout to the Svix open source project which is powering the webhooks. So I think that unless we see a clear reason driven by a real use-case, rather than a hypothetical one, this is probably not something we would like to add. I'm happy to be convinced otherwise, but we generally only make changes if they are driven by real needs, and this doesn't seem to be the case here.

PR #1119, opened 2023-11-14
Update libs with `Statistics` endpoints

jaymell:
Add support for two recently added `Statistics` endpoints:
* Aggregate App Stats 
* Aggregate Event Types


jaymell:
Note that `Aggregate Event Types` is technically still hidden in our API documentation. Should we fix that?

javascript/src/index.ts
```diff
@@ -875,3 +881,25 @@ export class Webhook {
...
+
+  public constructor(config: Configuration) {
+    this.api = new StatisticsApi(config);
+  }
+
+  public aggregateEventTypes(): Promise<AggregateEventTypesOut> {
+    return this.api.aggregateEventTypes({});
+  }
+
+  public calculateAggregateAppStats(
```
tasn, line 896:
```suggestion
  public aggregateApplicationMessages(
```
Stats and calculate are redundant. Need to make sure the event type name is also correct for this, but this should be the name of the function all around.

jaymell: Shall I change the name in our API docs then as well?

tasn: Yup, but that's a one-line change, no?

go/statistics.go
```diff
@@ -0,0 +1,46 @@
...
+func (s *Statistics) CalculateAggregateAppStats(ctx context.Context, appUsageStatsIn *AppUsageStatsIn, options *PostOptions) (*AppUsageStatsOut, error) {
+	req := s.api.StatisticsApi.CalculateAggregateAppStats(ctx)
+	if appUsageStatsIn != nil {
+		req = req.AppUsageStatsIn(*appUsageStatsIn)
+	}
+	if options != nil {
+		if options.IdempotencyKey != nil {
+			req = req.IdempotencyKey(*options.IdempotencyKey)
+		}
+	}
```
svix-onelson, line 28: The double nil check here always bites me. Good job.

go/statistics.go
```diff
@@ -0,0 +1,46 @@
...
+	api *openapi.APIClient
+}
+
+type (
+	AppUsageStatsIn        = openapi.AppUsageStatsIn
+	AppUsageStatsOut       = openapi.AppUsageStatsOut
+	AggregateEventTypesOut = openapi.AggregateEventTypesOut
+)
+
+func (s *Statistics) CalculateAggregateAppStats(ctx context.Context, appUsageStatsIn *AppUsageStatsIn, options *PostOptions) (*AppUsageStatsOut, error) {
```
svix-onelson:
```suggestion
func (s *Statistics) AggregateAppStats(ctx context.Context, appUsageStatsIn *AppUsageStatsIn, options *PostOptions) (*AppUsageStatsOut, error) {
```

Should these drop the "calculate" prefix too? Maybe we only did it for event types? I think I see the same in Java and probably others further down in the diff.

svix-onelson (commented):
I see Tom made a similar comment to mine, so maybe the updates are still pending.

go/statistics.go
jaymell: I didn't squash the commits. Is it possible you're looking at changes from the first commit only?

tasn (approved):

tasn (approved):

PR #1129, opened 2023-11-20
Server: adopt omniqueue as a queue backend

svix-onelson:
Nowhere near complete. Not ready to merge.

### DIFF
--- a/server/Cargo.toml
+++ b/server/Cargo.toml
@@ -1,5 +1,7 @@
 [workspace]
 members = ["svix-server", "svix-server_derive"]
+resolver = "2"
 
 [patch.crates-io]
-hyper = { git = "https://github.com/svix/hyper/", rev = "b901ca7c" }
+# FIXME(onelson): this rev points to a draft PR which should be merged before this PR gets merged.
+hyper = { git = "https://github.com/svix/hyper/", rev = "19c6acd14767b769303b8ae274cc30e3f9602d61" }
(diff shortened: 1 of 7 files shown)


server/Cargo.toml
```diff
@@ -1,5 +1,7 @@
 [workspace]
 members = ["svix-server", "svix-server_derive"]
+resolver = "2"
 
 [patch.crates-io]
-hyper = { git = "https://github.com/svix/hyper/", rev = "b901ca7c" }
+# FIXME(onelson): this rev points to a draft PR which should be merged before this PR gets merged.
+hyper = { git = "https://github.com/svix/hyper/", rev = "19c6acd14767b769303b8ae274cc30e3f9602d61" }
```
svix-onelson, line 7: There's a new 0.14.x branch in our fork of hyper which `main` now points to.

server/svix-server/src/queue/mod.rs
```diff
@@ -217,11 +247,9 @@ impl TaskQueueDelivery {
         tracing::trace!("nack {}", self.id);
         run_with_retries(
             || async {
+                // FIXME(onelson): borrowing &self in the match allows redis arm to pass it to `nack`
+                //   Omni arm needs `mut self` for `delivery.nack()`. How to satisfy both?
```
svix-onelson, line 251: I think this was the active concern when I was looking at this last.

The plan was to redo omniqueue to use interior mutability so the Acker in there can be `&self` (non-mut) like the rest.

The mutability in omni was needed for a "ack or nack once" check, and it looks like we could use `AtomicBool` for this and sidestep the issue.

svix-jplatte:
Superseded by #1188.

PR #1152, opened 2023-12-26

svix-lucho:


go/internal/openapi/client.go
```diff
@@ -64,12 +64,12 @@ type APIClient struct {
 
 	EnvironmentApi *EnvironmentApiService
 
-	EnvironmentSettingsApi *EnvironmentSettingsApiService
```
svix-onelson, line 67: I wonder if this was intentional, either to have it be in the lib at all in the first place, or the removal now.

Does anyone know the story here?

openapi.json
```diff
@@ -15198,16 +16049,76 @@
...
+                    }
                 ]
-            }
-        },
-        "/api/v1/environment/export/": {
+            },
             "post": {
                 "description": "Download a JSON file containing all org-settings and event types",
-                "operationId": "v1.environment.export",
+                "operationId": "v1.environment.export.post",
```
svix-onelson, line 16121: The last segment in an operation id is usually something CRUD-ish, right? create/delete/list... that sort of thing. I wonder why `post` is the thing here.

go/internal/openapi/client.go
svix-lucho: Hmm, no idea. From looking at the code, it seems the same methods are available in EnvironmentApi. It doesn't look like we are loosing any functionality

openapi.json
svix-lucho: My bad. I manually changed this op id here because it was failing to generate the libs due to a repeated op id with the one in line 15964. I changed the other one to be `v1.environment.export.get` so we can leave this one as is; no need to add `.post.` This endpoint is not making a CRUD operation AFAICT, so `create` doesn't look appropriate either.

I'll fix the repeated op id so we don't have to edit this again manually, but I wanted to get this PR unblocked.

cc @svix-aaron1011

svix-onelson (commented):
Yeah, I don't know what's going on with the spec changes. Some of these changes seem odd to me. I started to look at the history on the spec, but I still haven't gotten to an understanding. I'll have to revisit tomorrow 😭

go/internal/openapi/client.go
svix-onelson: Sure. It's still a breaking change I think. I just want to be sure it's intended breakage, but I don't have the context around this change.

svix-lucho: No longer happening after Aaron's change

openapi.json
```diff
@@ -18767,6 +19668,186 @@
                 ]
             }
         },
+        "/api/v1/transformation-template/generate": {
```
svix-onelson, line 19671: I wonder if this route should have a trailing `/` to match the rest?

Sorry to only notice this now 😭

svix-onelson (approved):
This looks much better to me. h/t to @svix-aaron1011 for the spec fix to make it look more how I'd expect!

openapi.json
svix-lucho: 😮 you're right. I think I missed that one in my PR. I'll merge this one and fix that in a follow up.

PR #1153, opened 2023-12-27
csharp: Remove tests of generated code

jaymell:
These tests are broken due to changes in underlying generated code. We probably don't need tests for the generated code, anyway, so removing.


svix-onelson (dismissed):

jaymell:
Added back in a couple application tests that weren't asserting calls to generated api.

svix-onelson (approved):

PR #1156, opened 2024-01-06
Server: backport expired message cleaner updates

svix-onelson:
Widens the period of the cleaner to be 12h when row counts cleaned dips
below the batch size.

The legacy cleaner query also has its timeout extended as was done when
troubleshooting the cleaner for one deployment. We may need to do the
same for the new `messagecontent` table, but for now I left it as-is.

Some additional refactors were needed with the 12h period to ensure the
process would not block (for hours) when the shutdown atomic gets set.

Additionally, since the "worst case" search for the `message` table
cleaner is the inevitable conclusion when a deployment starts writing to
`messagecontent` instead, a check is done when the cleaner first starts
up to see if there are any pending payloads at all in the `message` table.

This allows us to skip repeatedly scanning a table that will never have new
message payloads added to it.

### DIFF
--- a/server/svix-server/src/expired_message_cleaner.rs
+++ b/server/svix-server/src/expired_message_cleaner.rs
@@ -3,20 +3,54 @@
 
 use std::sync::atomic::Ordering;
 
-use crate::error::Result;
-use sea_orm::{ConnectionTrait, DatabaseConnection, DbErr, Statement, UpdateResult};
-use std::time::Duration;
-use tokio::time::sleep;
+use crate::error::{Error, Result};
+use sea_orm::{
+    ConnectionTrait, DatabaseConnection, DbErr, ExecResult, QueryResult, Statement,
+    TransactionTrait, UpdateResult,
+};
+use std::time::{Duration, Instant};
+
+type DbResult<T> = std::result::Result<T, DbErr>;
+
+async fn exec_without_timeout(pool: &DatabaseConnection, stmt: Statement) -> DbResult<ExecResult> {
+    let increase_timeout = Statement::from_string(
+        pool.get_database_backend(),
+        "SET LOCAL statement_timeout=0;",
+    );
+    let tx = pool.begin().await?;
+    let _ = tx.execute(increase_timeout).await?;
+    let res = tx.execute(stmt).await?;
+    tx.commit().await?;
+    Ok(res)
+}
+async fn query_one_without_timeout(
+    pool: &DatabaseConnection,
+    stmt: Statement,
+) -> DbResult<Option<QueryResult>> {
+    let increase_timeout = Statement::from_string(
+        pool.get_database_backend(),
+        "SET LOCAL statement_timeout=0;",
+    );
+    let tx = pool.begin().await?;
+    let _ = tx.execute(increase_timeout).await?;
+    let res = tx.query_one(stmt).await?;
+    tx.commit().await?;
+    Ok(res)
+}
 
 /// Nullifies the payload column for expired messages,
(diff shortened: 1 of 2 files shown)


server/svix-server/src/expired_message_cleaner.rs
```diff
@@ -14,9 +16,18 @@ pub async fn clean_expired_messages(
     pool: &DatabaseConnection,
     limit: u32,
 ) -> std::result::Result<UpdateResult, DbErr> {
-    let legacy_stmt = Statement::from_sql_and_values(
-        pool.get_database_backend(),
-        r#"
+    //
```
svix-gabriel, line 19:
```suggestion
```

server/svix-server/src/expired_message_cleaner.rs
```diff
@@ -14,9 +16,18 @@ pub async fn clean_expired_messages(
...
+    let legacy_res = {
```
svix-gabriel, line 20: I know why we're calling this legacy, but maybe it'd be better just to name these after the models they are destroying?

Or fill in the comment above about why this is "legacy"?

server/svix-server/src/expired_message_cleaner.rs
```diff
@@ -57,23 +71,27 @@ pub async fn clean_expired_messages(
...
+    const ON_ERROR: Duration = Duration::from_secs(10);
     const BATCH_SIZE: u32 = 5_000;
-    let mut sleep_time = Some(IDLE);
+    let mut sleep_time = None;
     loop {
         if let Some(duration) = sleep_time {
             sleep(duration).await;
         }
-        match clean_expired_messages(pool, BATCH_SIZE).await {
+        let pool = pool.clone();
```
svix-gabriel, line 85: Why the cloning?

server/svix-server/src/expired_message_cleaner.rs
svix-onelson: Can do.


Separately, a question I have for you is: should we add cfg to enable/disable this bit specifically since the worst-case search is inevitable after a certain point for anyone already switched over to messagecontent?

The most cost-effective thing to do is _not run this query_ after it's no longer needed, but I'm not sure how to know this programmatically.

I suppose if there are no rows with future expiration dates we can just skip this entirely since messagecontent would get all the new writes, right?

svix-onelson: I have done some refactoring and included some docs to try and explain "legacy."

I also did the thing where I check for pending non-null payloads in the `message` table. This seems important to have since otherwise installations will continue to scan that old table forever while never writing new payloads to it. That's a lose-lose situation.

server/svix-server/src/expired_message_cleaner.rs
svix-onelson: Yeah, dunno. Removed.

svix-james (dismissed):

server/svix-server/src/expired_message_cleaner.rs
```diff
@@ -48,32 +88,79 @@ pub async fn clean_expired_messages(
...
-        rows_affected: legacy_res.rows_affected() + res.rows_affected(),
+        rows_affected: legacy_row_count + res.rows_affected(),
     })
 }
 
+/// Checks to see if the message table has any non-null payloads requiring expiry.
+///
+/// ## Background
+///
+/// Initially payloads where modeled as a field in `message`, but later migrated to a separate
```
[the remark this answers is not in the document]
svix-onelson: 🙏🏻

svix-james (approved):

PR #1160, opened 2024-01-15
[javascript] Fix thrown error message when API status code is unknown

svix-lucho:
## Motivation

The response body was not read correctly, causing the SDK to throw errors like `Message: "Unknown API Status Code! Body: "[object Object]"`. This happens when the API returns an unexpected status code (like 500) or when the content type is not defined.

## Solution
Using `await response.body.text()` should return the actual response body.


svix-onelson (approved):
👨🏻‍🍳 🤌

PR #1162, opened 2024-01-17
[javascript] Remove fetch polyfill

arjunyel:
## Motivation

`svix-fetch` uses Node specific APIs which causes issues when using non-node environments, like Cloudflare Workers, and was imported as a side effect which means that bundlers cannot tree shake `svix`. `fetch` is now [supported](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API#browser_compatibility) on [all maintained Nodejs versions](https://endoflife.date/nodejs). Also as a general rule users should be responsible for polyfilling fetch, [not libraries](https://www.builder.io/blog/stop-polyfilling-fetch-in-your-npm-package).

## Solution

Use `fetch` from [globalThis](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis) and mark the library as side effect free.

### DIFF
--- a/javascript/package.json
+++ b/javascript/package.json
@@ -17,6 +17,7 @@
   "files": [
     "dist/**/*"
   ],
+  "sideEffects": false,
   "scripts": {
     "build": "tsc",
     "prepare": "yarn run build",
@@ -31,7 +32,6 @@
     "@stablelib/base64": "^1.0.0",
     "es6-promise": "^4.2.4",
     "fast-sha256": "^1.3.0",
-    "svix-fetch": "^3.0.0",
     "url-parse": "^1.4.3"
   },
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


svix-lucho:
@arjunyel The reason why we use `svix-fetch` is not to polyfill `fetch`, it's to enable keepalives when running in a node context. I agree with the issues with non-node environments and tree-shaking, but we need to make sure that we are not losing the keepalive behavior when needed.

svix-lucho (dismissed):

arjunyel:
The end user can set keep alive on their fetch for Node 18 https://stackoverflow.com/questions/62500011/reuse-tcp-connection-with-node-fetch-in-node-js

PR #1163, opened 2024-01-19
[javascript] Add es module build

arjunyel:
## Motivation

Creating an ES Module build allows ESM environments to treeshake the `svix` library

## Solution

Created both a CJS and ESM build using [tsup](https://github.com/egoist/tsup). Also updated all dependencies, upgraded to the [Eslint v9 config](https://eslint.org/blog/2022/08/new-config-system-part-2/), and switched from Jest to [Vitest](https://vitest.dev/).

Based on top of #1162


tasn:
Thanks for the contribution, this is a great addition!

Were the changes to eslint and prettier (moving to a different format) necessary? Also, I'm less familiar with tsup, is this what people use nowadays?

arjunyel:
Happy to help, thanks for making Svix :D

The eslint changes were necessary as the old configuration format is being [removed](https://eslint.org/blog/2023/10/flat-config-rollout-plans/).

To make npm packages I personally use [dnt](https://github.com/denoland/dnt) which lets you create both a Deno native and npm native packages. An example is I used it to make [Lago's SDK](https://github.com/getlago/lago-javascript-client). The downside it requires Deno while this repo already uses node.

For node projects `tsup` is the best tool I've found, here's a video of TypeScript expert Matt Pocock [talking about it](https://youtu.be/eh89VE3Mk5g).

PR #1167, opened 2024-01-20
Server: expired message cleaner cleanup

svix-onelson:
Simplifies some of the timing code on the expired message cleaner. No more 2 level check for slow vs idle. Just sleep big when the expired row count is less than the batch size.


svix-james:
Seems ok, but why do we need to simplify here?

svix-james (approved):

svix-onelson:
> Seems ok, but why do we need to simplify here?

There was just a lot of added line noise for what is today a much more simple check. If the difference in behavior is zero between an empty poll and a below-batch size poll, then there's no point in having independent branches for them.

Issue #1161, opened 2024-01-16
Why use threetenbp instead of java time for time


Burane:
### Motivation

When using threetenbp instead of java time it force to have all payload objects you want to send in a webhook message to use threetenbp.

### Proposal
In [java/openapi-generator-config.json](https://github.com/svix/svix-webhooks/blob/main/java/openapi-generator-config.json) add ` "dateLibrary": "java8"`
So it will use java time instead of  threetenbp

tasn:
More from Slack:
> I dont think there is an other way than support 2 version of the lib :confused:
> Do you have data for how many use pre-java 8 version (who will be 10 years old in a few months)

I think this is a great change and we should do it.

PR #1169, opened 2024-01-23
libs: Use Java time instead of threetenbp

svix-gabriel:
## Motivation

This closes https://github.com/svix/svix-webhooks/issues/1161.

## Solution

Once the openapi-generator-config.json was changed, I also had to update a bunch of our template and generated code to use the types.


java/templates/libraries/okhttp-gson/ApiClient.mustache
```diff
@@ -22,11 +22,6 @@ import org.joda.time.DateTime;
 import org.joda.time.LocalDate;
 import org.joda.time.format.DateTimeFormatter;
 {{/joda}}
-{{#threetenbp}}
```
tasn, line 25: Do we need this change? Doesn't it automatically do it because of the conditional? Or is this template file old for some reason?

svix-gabriel: Yeah, I think in theory it's not strictly necessary. I didn't see much reason to keep it around though either.

svix-james (approved):

PR #1171, opened 2024-01-23
[javascript] Remove Buffer usage

arjunyel:
## Motivation

`Buffer` is Node only and gets converted to a string anyway so user should be responsible to convert based on their platform

## Solution

Remove `Buffer`


tasn:
The problem is that this breaks API unfortunately. :(

Is there a real problem with having it be there?

arjunyel:
Buffer is a Node only thing, not standard JavaScript so it won't be acceptable in all runtimes.

@tasn would you prefer I close all my PRs and open them against the standard-webhooks library instead? That's much more new and I could do a 2.0 version there with breaking changes that won't effect as many people. In general though the Svix library has node only things which affects people like me who want to use it in other environments like Cloudflare workers, we have to use polyfills

arjunyel:
Why not have the user be responsible for turning the Buffer into a string? That way it works for every platform. That's the core of the issue, Buffer is a Node API, not a JavaScript one

tasn:
Yeah, for sure. The problem is breaking API though. :|

PR #1185, opened 2024-02-05
server: Some redis queue refactorings

svix-jplatte:
## Motivation

Preparation for omniqueue integration, in particular solves the problem discussed [here](https://github.com/svix/svix-webhooks/pull/1129/files#r1463610989).

## Solution

Refactor redis queue implementation to not take `TaskQueueDelivery` parameter in `ack`, `nack` methods, instead only taking references to the fields it is interested in, specifically *ex*cluding the `acker` field (such that it can be borrowed mutably when calling one of these methods).

Additionally, I improved one of the methods to use implicit named format arguments.

### DIFF
--- a/server/svix-server/src/queue/mod.rs
+++ b/server/svix-server/src/queue/mod.rs
@@ -196,7 +196,7 @@ impl TaskQueueDelivery {
             || async {
                 match &self.acker {
                     Acker::Memory(_) => Ok(()), // nothing to do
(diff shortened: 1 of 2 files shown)


svix-jplatte:
Unsure whether the test failure is related. Couldn't run tests locally, #1186 is related but not the full solution (I also got a timeout from the webhooks server after that); but also there's a clearly-unrelated CI failure there so maybe this is also flakiness? Any ideas appreciated, though at first I'll also try re-running CI for this.

PR #1186, opened 2024-02-05
Use fully-qualified image names in docker-compose files

svix-jplatte:
## Motivation

Compatibility with podman-compose w/ default configuration.

## Solution

Use `docker.io/` prefix for image names from DockerHub.


svix-aaron1011 (approved):

svix-jplatte:
Clippy error will be fixed by  #1185.

svix-aaron1011 (approved):

PR #1188, opened 2024-02-06
Adopt omniqueue as a queue backend and remove our own in-memory queue implementation

svix-jplatte:
~~Depends on #1185 (merged), https://github.com/svix/omniqueue-rs/pull/24.~~

I have deleted low-level tests for the in-memory queue since its basic functionality should be tested inside omniqueue, not here. Do we have tests for higher-level functionality that uses the in-memory queue backend in some scenarios?

The redis and SQS implementation should probably be replaced by the implementations found in Omniqueue in a follow-up PR, to keep this one reasonably small.


server/svix-server/src/queue/mod.rs
```diff
@@ -190,56 +218,80 @@ impl TaskQueueDelivery {
...
-            should_retry,
-            RETRY_SCHEDULE,
-        )
-        .await
+                })
+                .await
+            {
+                return result;
+            }
+        }
```
svix-jplatte, line 245: This weird formulation of the retry loop is unfortunately required for the `&mut ack` design to pass the borrow checker. It's a bit hard to explain but I can attempt to write up the details if somebody is interested. Or we just change omniqueue's `ack` and `nack` methods to take `&self` ^^

server/svix-server/src/queue/mod.rs
```diff
@@ -150,26 +170,34 @@ impl TaskQueueProducer {
...
+    Omni(DynConsumer),
 }
 
 impl TaskQueueConsumer {
     pub async fn receive_all(&mut self) -> Result<Vec<TaskQueueDelivery>> {
         match self {
             TaskQueueConsumer::Redis(q) => q.receive_all().await.trace(),
-            TaskQueueConsumer::Memory(q) => q.receive_all().await.trace(),
             TaskQueueConsumer::RabbitMq(q) => q.receive_all().await.trace(),
+            // FIXME(onelson): need to figure out what deadline/duration to use here
```
svix-jplatte, line 182: This looks like it also needs resolving before this can be merged.

svix-james:
LGTM, but somebody who was a bit more invested in the development of omniqueue than me should probably chime in.

server/svix-server/src/queue/mod.rs
svix-jplatte: @svix-onelson any ideas on how to find a suitable duration?

svix-jplatte:
Hm, even with the unbounded mpsc channel that the builtin in-memory queue used now adopted in omniqueue, some tests are still failing consistently, and with a *super weird* error message:

```
thread 'test_uid_across_users' panicked at library/alloc/src/raw_vec.rs:570:5:
capacity overflow
```

Plus a connection reset that I assume comes from the panic propagating up to hyper, which subsequently terminates the connection that the panicked task was serving.

server/svix-server/src/queue/mod.rs
svix-onelson: Yeah, this is a struggle. For the old memory backend it looks like there was a 30 sec window used. The rabbit impl we have sort of cheats and will wait as long as it takes for 1 message to come through, then yields that 1 immediately.

My concern is the polling semantics for each backend furnished by omniqueue might suggest a different value for the deadline. It might need to therefore be something we pull from the service config.

My suggestion is to keep it 30s to start with a mind to revisit and possibly add a config value to drive it.

svix-onelson:
> Hm, even with the unbounded mpsc channel that the builtin in-memory queue used now adopted in omniqueue, some tests are still failing consistently, and with a _super weird_ error message:
> 
> ```
> thread 'test_uid_across_users' panicked at library/alloc/src/raw_vec.rs:570:5:
> capacity overflow
> ```
> 
> Plus a connection reset that I assume comes from the panic propagating up to hyper, which subsequently terminates the connection that the panicked task was serving.

@svix-jplatte did this get solved somehow?

svix-onelson (approved):
LGTM

svix-jplatte:
Yes, there was one place where `usize::MAX` was passed as a buffer / channel capacity. I just put in a 128 instead, I doubt the exact number matters much since the in-memory queue isn't supposed to be optimized right?

svix-onelson:
> in-memory queue isn't supposed to be optimized right?

Correct. It's more of a stub than anything else. Like a dummy (no-op) cache implementation.

PR #1190, opened 2024-02-06
Rust verification library: fix comparison of different length signatures

tasn:
There was a bug in the code which meant that the signatures are only compared up to the length of the shorter signature, which means that an attacker can just pass `v1,` as the signature and that will always pass verification.

This change fixes it so that the length of the signature is also taken into account when comparing, to make sure that it's always the same length before comparing.

Manually verified all of the other libraries are correct, and added tests to JavaScript and Go (even though they were also not affected).

Many thanks to Fredrik Meringdal (@fmeringdal) for the report.

### DIFF
--- a/rust/src/webhooks.rs
+++ b/rust/src/webhooks.rs
@@ -86,10 +86,13 @@ impl Webhook {
             .filter_map(|x| x.split_once(','))
             .filter(|x| x.0 == SIGNATURE_VERSION)
             .any(|x| {
-                x.1.bytes()
-                    .zip(expected_signature.bytes())
-                    .fold(0, |acc, (a, b)| acc | (a ^ b))
-                    == 0
+                (x.1.len() == expected_signature.len())
+                    && (x
+                        .1
+                        .bytes()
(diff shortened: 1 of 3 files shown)


svix-aaron1011 (approved):

rust/src/webhooks.rs
```diff
@@ -86,10 +86,13 @@ impl Webhook {
...
-                    .zip(expected_signature.bytes())
-                    .fold(0, |acc, (a, b)| acc | (a ^ b))
-                    == 0
+                (x.1.len() == expected_signature.len())
+                    && (x
+                        .1
+                        .bytes()
+                        .zip(expected_signature.bytes())
+                        .fold(0, |acc, (a, b)| acc | (a ^ b))
+                        == 0)
```
svix-jplatte, line 95: I think we can use [`Iterator::eq`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.eq) instead?

tasn: No, it has to be constant time comparison, that's why it's done this way.

PR #1196, opened 2024-02-08
server: Extract Retry state machine out of run_with_retries

svix-jplatte:
For now the state machine is only used for the implementation of run_with_retries, but it will be used directly in the future, to get around control flow analysis issues with run_with_retries.

Extracted out of #1188 because CI is consistently failing there and I'm wondering whether I botched the retry state machine.


svix-james (approved):

svix-jplatte:
Failures in #1188 figured out. It's probably best to merge the commit from the PR as part of that one, not on its own.

PR #1197, opened 2024-02-08
rust: Switch from reqwest to hyper 1.0

### LINKED ISSUE #1146: Bump rust lib to Hyper 1.0
It's actually about the libs

## Feature Request

Upgrade axum to 0.7

### Motivation

I have an axum server that uses svix, both as an emitter and a consumer. With the recent changed to 1.0 of the hyper and http crates, a bunch of breaking changes were introduced in the ecosystem. Axum's 0.7 version now supports these 1.0 releases. However, since it uses a different http create version, the validation of webhooks fails due to the HeaderMap's version difference in the function signature.

I have put a lot of effort into migrating our codebase to axum 0.7 and this is the last piece of the puzzle.

### Proposal

Having skimmed through the lib's code, this should not introduce breaking changes, but will probably require custom feature flags like `axum06` and `axum07`.

Since I'm blocked, and you guys are probably busy, I can do these changes and submit a PR; but I might need some guidance into any standards or ways of doing things you guys have.

### Alternatives

Te alternative is to go back to 0.6, which will work for the time being, but I'm assuming this will eventually be needed

svix-jplatte:
~~Depends on #1193.~~
Closes #1146.

Probably best to review commit by commit, though most of the changes are in a single commit.


rust/src/lib.rs
```diff
@@ -1,21 +1,73 @@
...
+#[cfg(not(any(feature = "native-tls", feature = "rustls-tls")))]
+type Connector = HttpConnector;
+
+// If only native TLS is enabled, use that.
+#[cfg(all(feature = "native-tls", not(feature = "rustls-tls")))]
+type Connector = hyper_tls::HttpsConnector<HttpConnector>;
+
+// If rustls is enabled, use that.
+#[cfg(feature = "rustls-tls")]
+type Connector = hyper_rustls::HttpsConnector<HttpConnector>;
```
svix-jplatte, line 49: I wonder

- Should we support setups where TLS is completely disabled (at compile time)?
- If we were to not support the compile-time configuration, should we also enforce https at runtime (e.g. using `.https_only()` on the rustls connector builder)?
- If for some reason both native-tls and rustls are enabled at compile time, which one should we use? Currently this is rustls which is probably more secure and potentially avoids some compilation issues in particular around cross-compilation. It's less compatible with old TLS server implementations, but shouldn't matter unless somebody runs the API server behind an old reverse proxy.

svix-james: * I don't see any reason to support compile-time disabling of TLS -- that's not currently something we support, right?
* I don't think `https_only` is necessary -- it sounds prohibitive to anyone using this for local development. If we think it's an important option, that could potentially be a feature as well?
* I would default to rustls

svix-james (approved):

rust/src/lib.rs
svix-jplatte: Yeah, compile-time disabling of TLS is definitely not a high-profile feature. It almost comes about naturally by both TLS backends being optional. If we wanted to have a proper error message for if both are disabled, that would take an approximate 5 lines of code and a small compile time hit from having a `build.rs` file, whereas the current solution of supporting it is 4 lines of code :smile:

svix-jplatte:
Do we have examples or tests for the client libraries? I haven't seen anything and am wondering if compile checks are sufficient for merging this.

rust/src/lib.rs
svix-james: Ah, ok, I guess that makes sense. It's probably not a terrible idea to allow it then. But if we support one or the other by default, a user would have to go out of their way to do so though, right, i.e., by disabling default-features?

svix-jplatte: Yeah, exactly. I've set `native-tls` as the default now, same as reqwest does but we could switch the default to rustls too.

svix-jplatte:
Force-pushed for rebase and to remove three dependencies that were added but not actually used in the final version of the code.

Only the last three commits are actually new compared to yesterday, and the previous commit that bumped the version is removed since serious breaking changes are now avoided.

svix-james (approved):

PR #1203, opened 2024-02-12
[lib/c#] Return iterator information in list endpoints

### LINKED ISSUE #1200: csharp: List endpoint responses missing pagination data
Our list endpoints, e.g., `Endpoint.List`, `Endpoint.ListAsync`, do not return pagination data, which makes the library unusable if results are paginated. Need to return the full response, and not just the `data` portion of it.

svix-lucho:
## Motivation

Our list endpoints, e.g., Endpoint.List, Endpoint.ListAsync, do not return pagination data, which makes the library unusable if results are paginated. We need to return the full response, and not just the data portion of it.

## Solution

Return the full response (`data`, `iterator` and `done`) from the listing methods.

**This is a breaking change.** Users upgrading to the latest version of the C# lib will need to update their usage of these endpoints. `var res = Application.List(...)` becomes `var res = Application.List(...).Data`.

Fixes #1200

### DIFF
--- a/csharp/Svix/Abstractions/IApplication.cs
+++ b/csharp/Svix/Abstractions/IApplication.cs
@@ -24,9 +24,9 @@ namespace Svix.Abstractions
         Task<ApplicationOut> GetAsync(string appId, string idempotencyKey = default,
             CancellationToken cancellationToken = default);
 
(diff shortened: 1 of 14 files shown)


svix-onelson (approved):
Looks like it does what you mention in the description. LGTM.
:shipit:

PR #1204, opened 2024-02-13
[libs/Java]: don't serialize nulls in PATCH endpoint methods

### LINKED ISSUE #1141: [JAVA] Patch requests serialize null values
### Version

1.13.0
### Platform

Java svix library
### Description

When calling Patch endpoints from java svix library, we want to partialy update an object, so in most of case we dont precises all field of the object but just the fields we want to update.

But the GSON in the svix java lib is configured to serialize null values to null instead of ommiting them for Patch endpoints

svix-lucho:
## Motivation

When calling `PATCH` endpoints from the Java svix library, we want to update an object partialy, but GSON is configured to serialize null values to null instead of omitting them for Patch endpoints.

## Solution

Use a different serializer when the request method is `PATCH`, to ignore nulls during serialization in that case.

Fixes #1141


tasn:
Please add `Fixes #123` (with the right ticket number) to the description.

svix-james (approved):

PR #1206, opened 2024-02-14
python: Make request timeout and retries configurable

svix-jplatte:


python/svix/api.py
```diff
@@ -145,6 +145,19 @@
 class SvixOptions:
     debug: bool = False
     server_url: t.Optional[str] = None
+    """
+    The number of times to retry HTTP requests.
+
+    Defaults to 3, maximum 5.
+    """
+    num_retries: int = 3
```
svix-gabriel, line 153: I think the way we've typically handled retry scheduling is to actually encode it as an array of durations.

So the retry schedule could be something like
```
[1, 3, 10]
```
meaning "retry 3 times. Wait 1 second before the first retry, 3 seconds before the second retry, and 10 seconds before last retry". This representation is a bit more flexible, and matches how we let customers configure retry schedules for their webhooks too.

svix-jplatte:
Implemented retry schedule and also fixed a small bug I found with sync sleeping being used in an async function.

Also:

- Previously we had `NUM_RETRIES = 3` but since the first request was part of that loop, it actually only retried twice after initial failure; I implemented three retries now but can change that
- Previously a 5xx status code would always be followed up with a sleep, even if no further retry would happen afterwards, this is fixed

PR #1209, opened 2024-02-14
rust: Add timeout to SvixOptions

svix-jplatte:


rust/Cargo.toml
```diff
@@ -36,6 +36,7 @@ serde_repr = "0.1"
 thiserror = "1.0.30"
 time = "0.3"
 url = "2.2"
+tokio = { version = "1.36.0", features = ["time"] }
```
svix-gabriel, line 39:
```suggestion
tokio = { version = "1.36.0", default-features = false, features = ["time"] }
```
Just to avoid bloating clients?

rust/src/api.rs
```diff
@@ -15,6 +15,13 @@ const CRATE_VERSION: &str = env!("CARGO_PKG_VERSION");
 pub struct SvixOptions {
     pub debug: bool,
     pub server_url: Option<String>,
+    /// Timeout for HTTP requests.
+    ///
+    /// The timeout is applied from when the request starts connecting until
+    /// the response body has finished.
+    ///
+    /// By default, requests never time out.
```
svix-gabriel, line 23: FWIW, I think we should have some sort of opinionated default timeout, even if it's absurdly long (like 60 seconds or something). It's an easy thing to forget about when making requests to an external service.

rust/Cargo.toml
svix-jplatte: Tokio doesn't have any default features.

rust/src/api.rs
svix-jplatte: Okay, wanted to not change behavior at the same time but can do. Python seems to use 15s by default.

svix-gabriel (approved):
Part of the reason I suggested a default timeout is to avoid having to handle a `None` timeout.

But I don't think it matters much.

PR #1210, opened 2024-02-14
[javascript] Update timingSafeEqual

arjunyel:
## Motivation

Update `timingSafeEqual` to match Deno standard library

## Solution

Changed code to match [https://github.com/denoland/deno_std/blob/main/crypto/timing_safe_equal.ts](https://github.com/denoland/deno_std/blob/main/crypto/timing_safe_equal.ts)


svix-lucho:
The code doesn't match the Deno file you linked?

PR #1212, opened 2024-02-19
bridge: Upgrade dependencies

svix-jplatte:
The main interesting change here is the omniqueue upgrade. Changes for that are: https://github.com/svix/omniqueue-rs/compare/247904053bcf90cf693df4429092923bf97770eb...b6383db507ca2d61da1b31dfb4ea0981cf7c6881

I don't really know if the earlier commits in that range contain any runtime behavior changes (commit messages suggest no). Also the config struct updates I did were my best guesses as to what we might want to do about the new omniqueue redis backend features. Happy to iterate on all of this.

### DIFF
--- a/bridge/Cargo.lock
+++ b/bridge/Cargo.lock
@@ -14,9 +14,9 @@ dependencies = [
 
 [[package]]
 name = "addr2line"
-version = "0.20.0"
+version = "0.21.0"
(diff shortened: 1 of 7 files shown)


svix-gabriel (dismissed):

svix-onelson (approved):
LGTM

svix-onelson (approved):

PR #1217, opened 2024-02-20
ci: Build server tests earlier

svix-jplatte:
- Counts test build time against the build step
- Aborts earlier in case tests fail to build

### DIFF
--- a/.github/workflows/server-ci.yml
+++ b/.github/workflows/server-ci.yml
@@ -48,7 +48,7 @@ jobs:
       working-directory: server
 
     - name: Build
-      run: cargo build --locked
+      run: cargo test --all-features --no-run --locked
       working-directory: server
 
     - name: Install dependencies


svix-jplatte:
I was kind of expecting it. Looks like bridge CI failure is not because of any changes I made. :unamused:

svix-jplatte:
Seems like this consistently breaks tests, despite being such a trivial change :thinking:

svix-onelson:
> Seems like this consistently breaks tests, despite being such a trivial change 🤔
> But this PR is sufficiently un-important that this doesn't need to be figured out now.

I'm not sure it would behave differently, but `cargo test --no-run` should produce all the test binaries without running the tests. I just don't know if a subsequent `cargo test` will reuse those pre-built binaries in this situation, or if we'd have to invoke them each ourselves.

svix-jplatte:
Interesting idea, unfortunately the outcome is the same. I've backed out the bridge change, should still be worth to merge for server.

svix-onelson (approved):

PR #1228, opened 2024-02-22
Replace redis queue impl with omniqueue

svix-jplatte:


svix-jplatte:
Now with tests passing (locally)! Still want to do extra manual testing tomorrow before merging this, but otherwise it should finally be ready.

svix-jplatte:
And now the delay test is failing again... :sob:

server/svix-server/src/queue/redis.rs
```diff
@@ -287,16 +119,25 @@ async fn run_migration_schedule(delays: &[Duration], pool: RedisPool) {
...
     queue_prefix: &str,
     main_queue_name: &'static str,
     delayed_queue_name: &'static str,
-    delayed_lock: &'static str,
+    delayed_lock_name: &'static str,
 ) -> (TaskQueueProducer, TaskQueueConsumer) {
     let main_queue_name = format!("{queue_prefix}{main_queue_name}");
     let delayed_queue_name = format!("{queue_prefix}{delayed_queue_name}");
-    let delayed_lock = format!("{queue_prefix}{delayed_lock}");
+    let delayed_lock = format!("{queue_prefix}{delayed_lock_name}");
```
svix-james, line 131: Minor but this should probably be renamed as well to `delayed_lock_name`

svix-james (dismissed):

svix-jplatte:
So the next sometimes-falling test again assumed two `receive_all` calls would each return with a single item only. In this case it actually seems somewhat reasonable that this would be the case, but there's a good chance that other than the `receive_all` behavior (which is now more reasonable than before), nothing changed.

Specifically, the test sends two messages, first one with a delay of 2s, then one without delay, and expected `receive_all` to first only return the first item, but sometimes it actually returns both at the same time (but in the right order, second message first). I'm willing to call this correct. Worth nothing that the test always passed locally, and it would make sense that it only fails on slower machines.

svix-jplatte:
Alright, so CI succeeded after two retries. First was another flakiness, second was the messages coming back in the wrong order in `test_delay`. Since I could never reproduce locally but the test felt weirdly slow (slower than could be explained by sleeps / delay), I suspect that one of our dependencies is just horribly slow when build without optimizations, and that's why the message that's supposed to arrive ~2s later actually comes back first.

I also looked into how messages are read from the delayed queue in the omniqueue implementation but failed to figure it out :smiling_face_with_tear: 

I'll look into my horribly-slow-dependency theory since I've seen that before for stuff like sha2 (IIRC).

svix-jplatte:
Daamn, tests passed the first time after rebase. That's a good sign. So yeah, the latest problem that made tests flake was fixed in #1249! And finally, I did the local testing with old worker / new API plus the other way around, and that worked too :tada:

svix-james (approved):

PR #1235, opened 2024-02-27
server: Add defaulted generic error parameter to error::Result

svix-jplatte:
## Motivation

Removes the need to fully qualify `std::result::Result` anywhere except the one place the alias is introduced.

## Solution

Allow our `Result` alias to be a proper drop-in replacement for std's result by having an optional second type parameter.


svix-onelson (approved):
LGTM

PR #1236, opened 2024-02-27
server: Fix redis tests relying on dev configuration

svix-jplatte:
## Motivation

Previously, running tests would only work if there was a `config.toml` with a useable configuration, or the `SVIX_JWT_SECRET` environment variable was set (one test would also pick this up from `.env`, others not).

## Solution

Make tests work independently of dev configuration by hardcoding config values that work for testing.


svix-jplatte:
`run_tests.sh` was actually setting the env vars. There's probably some important improvements to be had, but this doesn't seem right to me anymore.

PR #1244, opened 2024-03-01
Tracing initialization improvements

svix-jplatte:
Previously, it was possible that tests would overwrite each other's intended logging configuration.


svix-james (approved):

svix-james:
This is a great change. I don't know whether it will actually fix our problems related to testing, though. For example, when I've tried to instrument tests to reconfigure the tracer per test, I've had no luck. If I make a direct call to `opentelemetry::global::shutdown_tracer_provider()` at the beginning of a test, the call will just hang indefinitely, which I believe is related to this issue: https://github.com/open-telemetry/opentelemetry-rust/issues/868

PR #1249, opened 2024-03-05
Update Cargo lockfiles

svix-jplatte:
## Motivation

bb8 0.8.2 fixes an important bug affecting us where acquiring a connection from a pool could take super long if a long-running request was blocking another connection from being returned to the pool (or something along those lines). Also CI failed in #1228 due to a broken dependency for some fringe OSes (and it's easier to bump than to add an exception to `deny.toml`).

## Solution

Upgrade all the dependencies! … except the `svix` library, to avoid pulling in hyper 1.0 for now.

### DIFF
--- a/bridge/Cargo.lock
+++ b/bridge/Cargo.lock
@@ -73,9 +73,9 @@ dependencies = [
 
 [[package]]
 name = "ahash"
-version = "0.8.10"
+version = "0.8.11"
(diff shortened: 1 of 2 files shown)


svix-aaron1011 (approved):

svix-jplatte:
Bad timing with the second commit :sweat_smile:

svix-aaron1011 (approved):

PR #1250, opened 2024-03-05
Make procedural macros faster by optimizing quote

svix-jplatte:
This takes extra compile time on first build of quote, but the cost is recouped by faster runtime of proc-macros, even for clean `cargo check`. (tested for server, but bridge has just as many proc-macro dependencies)


svix-jplatte:
Server CI likely a flake, security failure addressed by #1249.

svix-onelson (approved):

PR #1256, opened 2024-03-06
Upgrade omniqueue

svix-jplatte:
… and one related dependency.

Also fix a regression in the rabbitmq queue backend that would result in tasks from a new API process not being picked up by old worker processes due to missing message IDs.

### DIFF
--- a/server/svix-server/Cargo.toml
+++ b/server/svix-server/Cargo.toml
@@ -72,7 +72,7 @@ urlencoding = "2.1.2"
 form_urlencoded = "1.1.0"
 lapin = "2.1.1"
 sentry = { version = "0.32.2", features = ["tracing"] }
-omniqueue = { git = "https://github.com/svix/omniqueue-rs.git", rev = "044fc688670759859f193bb646fac043bbfd08ba", default-features = false, features = ["in_memory", "rabbitmq", "redis_cluster"] }
+omniqueue = { version = "0.2.0", default-features = false, features = ["in_memory", "rabbitmq", "redis_cluster"] }
 
 [target.'cfg(not(target_env = "msvc"))'.dependencies]
 tikv-jemallocator = { version = "0.5", optional = true }
--- a/server/svix-server/Cargo.toml
+++ b/server/svix-server/Cargo.toml
@@ -72,7 +72,7 @@ urlencoding = "2.1.2"
 form_urlencoded = "1.1.0"
 lapin = "2.1.1"
 sentry = { version = "0.32.2", features = ["tracing"] }
-omniqueue = { version = "0.2.0", default-features = false, features = ["in_memory", "rabbitmq", "redis_cluster"] }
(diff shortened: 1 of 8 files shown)


server/svix-server/src/queue/rabbitmq.rs
```diff
@@ -160,3 +160,69 @@ async fn declare_bound_queue(
...
+    use futures::StreamExt as _;
+    use lapin::options::{BasicConsumeOptions, BasicQosOptions};
+    use svix_ksuid::{KsuidLike as _, KsuidMs};
+
+    use crate::{cfg, queue::QueueTask};
+
+    #[tokio::test]
+    // run with `cargo test -- --ignored rabbitmq` only when rabbitmq is up and configured
+    #[ignore]
+    async fn test_messages_have_ids() {
```
svix-onelson, line 175: I think the prior art for this sort of thing is to return immediately (i.e. pass the test) for configs that don't support the thing.

iirc we see this in cache-related tests, specifically checking for the "none" cache.

https://github.com/svix/svix-webhooks/blob/910f29b1667c2fef973f22ef7c8b779413642b39/server/svix-server/tests/it/worker.rs#L138

Not a huge difference but it'll save us from having to run the test binary twice for this one case.

svix-jplatte: I quite dislike that, it makes it hard / impossible to tell whether the test actually ran and passed, or it saw a configuration that didn't match its expectations. I think we should change those tests to match this one ;)

Regarding the extra run of the test binary, note that `--ignored` only runs ignored tests, `--include-ignored` would also (re)run others.

server/svix-server/Cargo.toml
```diff
@@ -72,7 +72,7 @@ urlencoding = "2.1.2"
 form_urlencoded = "1.1.0"
 lapin = "2.1.1"
 sentry = { version = "0.32.2", features = ["tracing"] }
-omniqueue = { git = "https://github.com/svix/omniqueue-rs.git", rev = "044fc688670759859f193bb646fac043bbfd08ba", default-features = false, features = ["in_memory", "rabbitmq", "redis_cluster"] }
+omniqueue = { version = "0.2.0", default-features = false, features = ["in_memory", "rabbitmq-with-message-ids", "redis_cluster"] }
```
tasn, line 75: @svix-jplatte, why is this feature not just called `rabbitmq`?

tasn: We broke compatibility in a few ways in this version anyway, no? So why not break (read: fix) this too?

svix-jplatte: We broke runtime compatibility with omniqueue 0.1 for redis to preserve compatibility between svix'es redis backend and the omniqueue redis backend (the omniqueue 0.1 behavior didn't really make sense). This feature is also for compatibility with the pre-omniqueue queue in svix, but it's not useful for other omniqueue users (unlike the redis change).

svix-onelson (approved):

PR #1264, opened 2024-03-10
Libs(ruby): expose patch endpoints

maful:
## Motivation

I've noticed that the `patch` endpoints, which were introduced back in [v1.6.0](https://github.com/svix/svix-webhooks/releases/tag/v1.6.0), aren't currently exposed in the way they should be. This omission can lead to errors when users attempt to utilize these `patch` endpoints, particularly when building request bodies for endpoint resources. A case in point is the `Svix::EndpointPatch` error, which serves as an illustrative example of the underlying issue:

```
irb(main):001:0> Svix::EndpointPatch.new({})
(irb):1:in `<main>': uninitialized constant Svix::EndpointPatch (NameError)

Svix::EndpointPatch.new({})
    ^^^^^^^^^^^^^^^
Did you mean?  Svix::EndpointUpdate
               Svix::EndpointOut
               Svix::EndpointAPI
```

## Solution

To address this issue, I propose to expose the `patch` endpoints within the `ruby/lib/svix.rb` file.


svix-james (approved):
LGTM -- thanks!

PR #1267, opened 2024-03-11
server: Reduce redundant test runs

svix-jplatte:
The redis tests are now run once for the redis backend, and once for the redis cluster backend in run-tests.sh. Previously, all six runs would execute these tests, most of the time with the redis backend. Since these tests are relatively low-level, they are not affected by other configuration changes. It really only makes sense to run e2e tests with all the different configuration combinations.

@svix-onelson you were previously not convinced about this approach of using `#[ignore]`d tests plus extra `--ignored` test runs w.r.t. rabbitmq. How do you feel about it here given the diff?

### DIFF
--- a/server/run-tests.sh
+++ b/server/run-tests.sh
@@ -2,8 +2,6 @@
 
 # Run tests with various configurations:
 
-TEST_COMMAND="cargo test --all --all-features --all-targets $*"
-
 # Common variables:
 export DATABASE_URL="postgresql://postgres:postgres@localhost:5432/postgres"
 export SVIX_JWT_SECRET="test value"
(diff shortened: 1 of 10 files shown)


svix-jplatte:
Separately, any objections against disabling bash linting? Idk if it ever linted against something useful. Before the latest commit, it outputted a bunch of

```
      export SVIX_QUEUE_TYPE="redis"
             ^-------------^ SC2030 (info): Modification of SVIX_QUEUE_TYPE is local (to subshell caused by (..) group).
```

which is an "info" lint that I think should never cause CI to fail.

svix-jplatte:
While CI performance has quite high variability, there is a sign here that this improved CI runtime quite noticably: Here's page 1 of the most recent successful runs:

Screenshot 2024-03-11 at 20-02-37 Server CI · Workflow runs · svix_svix-webhooks

server/run-tests.sh
```diff
@@ -13,41 +11,52 @@ export SVIX_DB_POOL_MAX_SIZE="500"
...
-SVIX_QUEUE_TYPE="redis" \
-SVIX_CACHE_TYPE="redis" \
-SVIX_REDIS_DSN="redis://localhost:6379" \
-${TEST_COMMAND}
+(
+    export SVIX_QUEUE_TYPE="redis"
+    export SVIX_CACHE_TYPE="redis"
+    export SVIX_REDIS_DSN="redis://localhost:6379"
+    cargo test
+    cargo test -- --ignored redis
```
svix-onelson, line 19: Chiming in re: my gripe about possible different features being selected. Looks like this was fixed by not having a `TEST_COMMAND` as for the general case, which I think is fair.

This way, it should be clear if a given run needs to select features, it should happen for both `cargo test` invocations 👍🏻 

It's less DRY, but tests and related should not be DRY at the expense of information splitting.

The reliance on `#[ignore]` still feels a little fiddly to me, but I think this is better than it was when I last looked.

server/svix-server/src/core/cache/redis.rs
```diff
@@ -116,11 +116,16 @@ mod tests {
...
+            _ => panic!(
+                "This test should only be run when redis is configured as the cache provider"
+            ),
         }
     }
 
     #[tokio::test]
+    // run with `cargo test -- --ignored redis` only when redis is up and configured
+    #[ignore]
     async fn test_cache_crud_no_ttl() {
```
svix-james, line 129: Is this test captured by `cargo test -- --ignored redis` ?

server/svix-server/src/core/types/mod.rs
svix-onelson, line 1: Yeah, I wasn't sure why these tests started running suddenly in our other envs with the removal of `--all-targets --all`. Seems like doctests should have been included with those flags specified, but apparently not.

svix-onelson (approved):

server/svix-server/src/core/cache/redis.rs
svix-jplatte: Yes. That command will run all tests whose fully qualified name contains the string `redis`. In this case the fully qualified name is: `core::cache::redis::tests::test_cache_crud_no_ttl`.

server/svix-server/src/core/types/mod.rs
svix-jplatte: `--all-targets` specifically excludes doctests. I don't know what its purpose really is 🤷‍♂️

PR #1271, opened 2024-03-13
server: Use tracing_subscriber's TestWriter in tests

svix-jplatte:
… so that logs are only printed for failing tests.

### DIFF
--- a/server/svix-server/src/lib.rs
+++ b/server/svix-server/src/lib.rs
@@ -24,7 +24,7 @@ use tower_http::{
     cors::{AllowHeaders, Any, CorsLayer},
     normalize_path::NormalizePath,
 };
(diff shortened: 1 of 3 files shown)


svix-onelson (approved):

svix-onelson:
Does this help with my desire to log response bodies on status code mismatch? You mentioned some extra config was needed to get tracing to show up in tests (from the test code itself maybe?).

Either way, this seems good. Less output is better.

svix-jplatte:
This is just for the tracing subscriber that's installed for the server task. If the test code itself emits tracing events, those currently go nowhere because there is no subscriber. I can post a follow-up PR to change that (make the tests themselves also use `TestWriter`, but no extra otel stuff like we have for the server).

svix-aaron1011 (approved):

PR #1275, opened 2024-03-14
Some testing improvements

svix-jplatte:
Since #1244, we were no longer installing a global tracing subscriber that would print logs from (failing) tests. With this, we do again.

First two commits are drive-by improvements in response to local warnings from my nightly toolchain about redundant imports.


svix-james:
Looks fine, but can we add a description?

svix-jplatte:
Description added.

svix-james (approved):

PR #1283, opened 2024-03-25
Upgrade dependencies and adjust cargo-deny config

svix-jplatte:
We're currently getting quite a few warnings from cargo-deny. Most are about duplicated dependencies, but some are more important to fix.

Also do some upgrades that weren't related to cargo-deny warnings.


svix-gabriel (approved):

svix-jplatte:
Yeah, that's from `jwt-simple`. Currently assessing what to do about those, seems like the root issue is having both boringssl and openssl in the dependency tree: https://github.com/cloudflare/boring/issues/197

svix-jplatte:
`jwt-simple` upgrade backed out for now, and created an issue for it.

svix-gabriel (approved):

PR #1306, opened 2024-04-30
Bump version to 1.22.0 and update changelog

svix-lucho:
## Version 1.22.0
* Server: Don't require trailing slash at the end of request paths
* Server: Improve testing performance and reduce redundant test runs
* Server: Improve performance of `/api/v1/openapi.json` route
* Server: Upgrade dependencies and improve code formatting
* Server: Upgrade to Docker Compose v2
* Libs/Ruby: Add PATCH endpoints


svix-jplatte (approved):

svix-jplatte (approved):
Fix for bridge isn't _great_ but it's also obviously not making anything worse so let's get this merged and I'll look into why we're not using those fields separately.

PR #1308, opened 2024-05-02
Add oauth support to javascript

jaymell:
This updates the OpenApi spec/Javascript lib with new routes 
for oauth support. We are keeping the changes limited to Js
for now until the API is finalized.

### DIFF
--- a/javascript/src/index.ts
+++ b/javascript/src/index.ts
@@ -458,7 +458,7 @@ class Endpoint {
     );
   }
 
-  public updateOauthConfig(
+  public oauthConfigUpdate(
     appId: string,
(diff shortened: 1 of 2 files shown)


javascript/src/index.ts
```diff
@@ -456,6 +457,25 @@ class Endpoint {
       { appId, endpointId, eventExampleIn, ...options }
     );
   }
+
+  public updateOauthConfig(
```
tasn, line 461:
```suggestion
  public oauthConfigUpdate(
```
And same for the other ones.

svix-lucho (approved):

PR #1328, opened 2024-06-04
Redis updates

jaymell:
* Add `redisclusterunpooled` cache type

  Adds downstream changes to support unpooled redis cluster.
  Simply put, we don't gain anything by using connection pooling
  for clustered redis. The redis cluster client handles establishing
  connections, handling errors, etc., internally. As long as we don't
  make blocking redis requests, we are fine without connection pooling
  in the redis caching layer.

* Remove `rediscluster` cache backend

  Note: Also fixed some of the queue tests, which should match
  based on the _queue_ configuration, not the cache config.

* Rename `redisclusterunpooled` to `rediscluster`

  This is now our only clustered cache type, so simplify
  the name to match our prior clustered cache variant.

* Remove connection pool for `redis` cache variant

  Replace connection pooling with unpooled
  `redis::aio::ConnectionManager`. The ConnectionManager uses a single
  multiplexed connection to handle all requests and also handles
  reconnections directly, making connection pooling unnecessary.

  Note that these changes applying *only* to redis caching, not queuing.
  This is safe in the redis caching layer because we do not make
  blocking calls there. However, we *do* make blocking calls currently
  in our redis queue implementation, so we must continue to use
  connection pooling there to avoid blocking concurrent requests.

  Also note that names for our many methods of acquiring a `RedisPool`
  have been standardized for less ambiguity.

### DIFF
--- a/server/svix-server/src/redis/mod.rs
+++ b/server/svix-server/src/redis/mod.rs
@@ -1,15 +1,20 @@
 mod cluster;
 
+use std::time::Duration;
+
 use bb8::{Pool, RunError};
-use bb8_redis::RedisMultiplexedConnectionManager;
+use bb8_redis::RedisConnectionManager;
 use redis::{FromRedisValue, RedisError, RedisResult};
 
 pub use self::cluster::RedisClusterConnectionManager;
 use crate::cfg::Configuration;
 
+pub const REDIS_CONN_TIMEOUT: Duration = Duration::from_secs(2);
+
 #[derive(Clone, Debug)]
 pub enum RedisPool {
     Clustered(ClusteredRedisPool),
+    ClusteredUnpooled(ClusteredRedisUnpooled),
     NonClustered(NonClusteredRedisPool),
 }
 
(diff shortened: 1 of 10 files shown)


jaymell:
I went ahead and lumped this into one PR because the changeset is not really that large. The first commit, the addition of `redisclusterunpooled`, is just bringing in downstream changes here.  The second and third commits are cleanup to remove the no-longer-needed pooled cluster variant. The fourth commit is the most substantive new addition, but is fairly small. Nonetheless, I'm happy to break this into smaller PRs if so desired.

server/svix-server/src/redis/mod.rs
```diff
@@ -134,28 +213,92 @@ impl<'a> ClusteredPooledConnection<'a> {
...
+        let con = cli
+            .get_async_connection()
+            .await
+            .expect("Failed to get redis-cluster-unpooled connection");
+        RedisPool::ClusteredUnpooled(ClusteredRedisUnpooled { con })
+    } else {
+        let cli = redis::Client::open(redis_dsn).expect("Error initializing redis unpooled client");
+        let con = redis::aio::ConnectionManager::new_with_backoff_and_timeouts(
+            cli,
+            2,
```
jaymell, line 276: The values here match the library defaults, with exception of number of retries (1), which matches how we configure the clustered client.

svix-jplatte (approved):

PR #1333, opened 2024-06-11
bridge: Add Kafka as an input

svix-jplatte:
… that is, support converting Kafka messages into Svix API calls.

Part of https://github.com/svix/monorepo-private/issues/8508.

### DIFF
--- a/bridge/svix-bridge-plugin-kafka/src/input.rs
+++ b/bridge/svix-bridge-plugin-kafka/src/input.rs
@@ -1,0 +1,225 @@
+use std::{
+    str,
+    time::{Duration, Instant},
+};
+
+use rdkafka::{
+    consumer::{CommitMode, Consumer as _},
+    error::KafkaError,
+    Message as _,
+};
+use svix_bridge_types::{
+    async_trait, svix::api::Svix, CreateMessageRequest, JsObject, SenderInput, SenderOutputOpts,
+    TransformationConfig, TransformerInput, TransformerInputFormat, TransformerJob,
+    TransformerOutput, TransformerTx,
+};
+use tokio::task::spawn_blocking;
+
+use crate::{config::KafkaInputOpts, Error, Result};
+
+pub struct KafkaConsumer {
+    name: String,
+    opts: KafkaInputOpts,
+    transformation: Option<TransformationConfig>,
+    transformer_tx: Option<TransformerTx>,
+    svix_client: Svix,
+}
+
+impl KafkaConsumer {
+    pub fn new(
+        name: String,
+        opts: KafkaInputOpts,
+        transformation: Option<TransformationConfig>,
+        output: SenderOutputOpts,
+    ) -> Result<Self> {
+        Ok(Self {
+            name,
+            transformation,
+            transformer_tx: None,
+            opts,
+            svix_client: match output {
+                SenderOutputOpts::Svix(output) => {
+                    Svix::new(output.token, output.options.map(Into::into))
+                }
+            },
+        })
+    }
+
+    #[tracing::instrument(skip_all)]
+    async fn process(&self, msg: &rdkafka::message::BorrowedMessage<'_>) -> Result<()> {
+        let payload = msg.payload().ok_or_else(|| Error::MissingPayload)?;
+        let payload = if let Some(transformation) = &self.transformation {
+            let input = match transformation.format() {
+                TransformerInputFormat::Json => {
+                    let json_payload =
+                        serde_json::from_slice(payload).map_err(Error::Deserialization)?;
+                    TransformerInput::JSON(json_payload)
+                }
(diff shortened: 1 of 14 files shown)


svix-jplatte:
Tests are super flaky for me locally, but looks like CI is having no such problems after updating the port to 9094...

bridge/svix-bridge-plugin-kafka/src/input.rs
```diff
@@ -0,0 +1,215 @@
...
+            let consumer = opts.create_consumer()?;
+            tracing::debug!("Created StreamConsumer");
+
+            consumer.subscribe(&[&topic])?;
+            tracing::debug!(topic, "Subscribed");
+
+            Ok::<_, KafkaError>(consumer)
+        })
+        .await
+        // unwrap propagates any panics from create_consumer caught by tokio
```
svix-gabriel, line 127:
```suggestion
```

I think instead of a comment, it would be simpler to just embed the rationale in the panic message. e.g.

`.expect("create_consumer task panicked")`

Or something similar

bridge/svix-bridge-plugin-kafka/src/input.rs
```diff
@@ -0,0 +1,215 @@
...
+            // In that case, it's likely already doing batching reads (which don't seem to be
+            // a thing in the public API) internally.
+            // If not, we should likely do some sort of batching ourselves, e.g. have two separate
+            // tokio tasks, one which pulls messages from Kafka and one that processes them, with
+            // a bounded channel in between for backpressure.
+            let msg = consumer.recv().await?;
+            tracing::debug!("Received a message");
+
+            let mut process_error_count = 0;
+            while let Err(e) = self.process(msg.detach()).await {
```
svix-gabriel, line 142: Why did you decide to use `msg.detach()`? As opposed to having `process` take in a `BorrowedMessage` instead?

svix-gabriel (commented):
It would be nice to link the FIXMEs to actual tickets, or just drop the FIXME label if you don't think there's any concrete work to be done relating to the actual comments.

bridge/svix-bridge-plugin-kafka/src/input.rs
svix-jplatte: I failed to see that it implements a trait that lets you consume the contents without `unsafe`. After reading your comment I had another look and found that.

bridge/svix-bridge-plugin-kafka/src/input.rs
```diff
@@ -0,0 +1,214 @@
...
+            serde_json::from_slice(payload).map_err(Error::Deserialization)?
+        };
+
+        let CreateMessageRequest {
+            app_id,
+            message,
+            post_options,
+        } = payload;
+
+        // FIXME(jplatte): I think we should always set the idempotency key
```
svix-jplatte, line 78: So about this, do you agree @svix-gabriel? If so I'll update the PR. I'm unclear why the `CreateMessageRequest` in the queue would itself set the idempotency key, but it is part of `post_options` there. I was thinking we should always set it to something like `svix_bridge_kafka_{topic}_{stream_position}` such that a failure to commit and retry, or even a crash and restart after the webhook was sent doesn't re-send it.

svix-jplatte:
Created an issue for the two long FIXME's: https://github.com/svix/monorepo-private/issues/8556

bridge/svix-bridge-plugin-kafka/src/input.rs
svix-gabriel: To be honest, I'm not really sure. Your suggestion sounds reasonable to me.

I know @svix-onelson is much more familiar w/bridge and how it's used than I am, so he might have some insight.

bridge/svix-bridge-plugin-kafka/src/input.rs
```diff
@@ -0,0 +1,214 @@
...
+            // If not, we should likely do some sort of batching ourselves, e.g. have two separate
+            // tokio tasks, one which pulls messages from Kafka and one that processes them, with
+            // a bounded channel in between for backpressure.
+            let msg = consumer.recv().await?;
+            tracing::debug!("Received a message");
+
+            let mut process_error_count = 0;
+            while let Err(e) = self.process(&msg).await {
+                match e {
+                    // If the payload is invalid, log an error and continue.
```
svix-gabriel, line 143: I'm a bit confused by how the comments here line up with the code.
> log an error and continue.

and then later

>  break; // don't call process again

FWIW, I also think the code here is relatively easy to follow. The comment `It would fail the same way if retried` seems useful, but I don't think you need the other comments here.

svix-gabriel (commented):
I left one more question/comment, but everything else seems reasonable to me.

svix-jplatte:
Ready to merge from my POV, just needs an approval now!

PR #1337, opened 2024-06-13
server: Update Cargo.lock

svix-jplatte:
Fixes nightly builds that were running into inference errors in time.

Re. license: The [unicode.org FAQ](https://unicode.org/faq/unicode_license.html) describes it as a "permissive MIT type of license" and it is OSI approved. We need this because several new dependencies with this license are now being pulled in by the url crate (through idna).

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -27,6 +27,7 @@ allow = [
     "MIT",
     "MPL-2.0",
     "OpenSSL",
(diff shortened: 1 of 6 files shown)


svix-gabriel (approved):

svix-jplatte:
Finally, CI passed!

PR #1340, opened 2024-06-18
chore: Suppress 'unused' lint on beta

svix-aaron1011:


svix-jplatte (commented):
This error seems legit to me. Looks like we don't have a `Svix::background_task` method like we have methods to access all the other API route groups.

PR #1343, opened 2024-06-19
bridge: Reduce deno dependencies

svix-jplatte:
We were depending on deno_runtime, but only using items from its deno_core re-export.


svix-jplatte:
New kafka test made the test suit hang once, need to investigate but separately from this PR.

svix-onelson:
Fair. The `deno_runtime` dep was added in support of https://github.com/svix/svix-webhooks/pull/1004 which never landed. Can bring it back if/when it is needed.

PR #1350, opened 2024-06-25
Onelson/golang net bump

svix-onelson:
With the dependabot bump for the `net` package, a number of lints started to bark.

_After having caught a couple of wild geese_ I have learned that:
- we were invoking `golangci-lint` from a dedicated workflow _in addition to_ a "superlint" workflow.
- the configuration in the "superlint" case was way more aggressive than for the dedicated one.
- looks like we had been disabling superlint for any case where we had a dedicated workflow.

Refs: 
- https://golangci-lint.run/
- https://github.com/golangci/golangci-lint-action/

### DIFF
--- a/.github/workflows/go-lint.yml
+++ b/.github/workflows/go-lint.yml
@@ -10,7 +10,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: golangci-lint
-        uses: golangci/golangci-lint-action@v2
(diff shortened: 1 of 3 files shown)


svix-onelson:
Closed in favor of https://github.com/svix/svix-webhooks/pull/1351 which does the linter stuff separate from the bump. We should be able to rebase and merge the depbendabot PR as-is once the other PR has merged.

PR #1352, opened 2024-06-25
`Recover Failed Webhooks` use configurable `until`

jaymell:
Our API docs support an optional `until` parameter. This adds support for it in the backend.


svix-onelson (approved):
Looks familiar. :shipit:

PR #1355, opened 2024-07-01
Replace ctype_digit for PHP 8.1 deprecation of non-string arguments

jterry:
From [php.net's ctype_digit documentation](https://www.php.net/manual/en/function.ctype-digit.php):

> As of PHP 8.1.0, passing a non-string argument is deprecated. In the future, the argument will be interpreted as a string instead of an ASCII codepoint. Depending on the intended behavior, the argument should either be cast to [string](https://www.php.net/manual/en/language.types.string.php) or an explicit call to [chr()](https://www.php.net/manual/en/function.chr.php) should be made.

## Motivation

Deprecation is throwing a warning when passing a string into ctype_digit when checking the timestamp.

## Solution

Updated the integer check to move away from `ctype_digit`.

### DIFF
--- a/php/src/Webhook.php
+++ b/php/src/Webhook.php
@@ -70,7 +70,7 @@ class Webhook
 
     public function sign($msgId, $timestamp, $payload)
     {
-        $is_positive_integer = ctype_digit($timestamp);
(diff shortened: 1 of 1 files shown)


tasn:
Thanks for the PR! Will tag the team to take a look.

svix-onelson (approved):
Thanks for the patch! LGTM.

PR #1356, opened 2024-07-04
CI: add something to run php tests

svix-onelson:
## Motivation

We recently had a contribution to the PHP lib and verifying the change was sound turned out to be a bit of a wild goose chase.

## Solution

We already had some tests in place so it felt natural to get the suite running in CI.
After getting that going, there were a few warnings and failures that needed fixing up.

### DIFF
--- a/.github/workflows/php-ci.yml
+++ b/.github/workflows/php-ci.yml
@@ -1,0 +1,25 @@
+name: PHP CI
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'php/**'
+      - '.github/workflows/php-ci.yml'
+  pull_request:
+    paths:
+      - 'php/**'
+      - '.github/workflows/php-ci.yml'
+jobs:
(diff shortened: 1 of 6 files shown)


svix-james (approved):

.github/workflows/php-ci.yml
```diff
@@ -0,0 +1,25 @@
...
+      - 'php/**'
+      - '.github/workflows/php-ci.yml'
+jobs:
+  build-test:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: PHPUnit Tests
+        uses: php-actions/phpunit@v3
```
svix-jplatte, line 21: @svix-onelson Is there a reason you used older-than-latest versions of both these actions?

svix-onelson: @svix-jplatte no.

PR #1363, opened 2024-07-09
Bridge: add `/events` poller

svix-onelson:
Helps with https://github.com/svix/monorepo-private/issues/8654

Depends on client lib updates (not yet available).

I'd call this "quick and dirty" but it was actually a bit of a hefty lift to pull together. It is however fairly dirty.

The want is to be able to pull messages from `/events` and feed them into a receiver output.
The flow we'd see with `SenderInput` paired with `ReceiverOutput` is what we want, but the existing typing didn't allow for this.

Refactors were needed to:
- lift out the parts from the HTTP server that run payloads through transformations then forward to an output.
- update config to account for a new type of receiver: a "poller".
- new traits/types for the poller to differentiate it from the standard webhook receiver (to ensure we don't accidentally pass one through and try to bind it to a URL).
- Lots of "connective tissue" in the form of converters between config values and concrete ones that can actually do things.

Some of the "connective tissue" exists purely to mimic bits and pieces that existed for either the other receivers or senders (remember, this case is odd in that it's similar to both).

Refactorings aside, the poller itself boasts an exponential backoff for both error cases (either from `/events` or from the output) as well as for the case where the `/events` iterator is "done."

This diff comes with a promise that we will (soon) give these additions another look at clean up the stuff that doesn't make sense or feels redundant.

### DIFF
--- a/bridge/svix-bridge/src/config/mod.rs
+++ b/bridge/svix-bridge/src/config/mod.rs
@@ -15,9 +15,17 @@ use shellexpand::LookupError;
 use svix_bridge_plugin_kafka::{KafkaInputOpts, KafkaOutputOpts};
 use svix_bridge_plugin_queue::config::{QueueInputOpts, QueueOutputOpts};
 use svix_bridge_types::{
-    ReceiverInputOpts, ReceiverOutput, SenderInput, SenderOutputOpts, TransformationConfig,
+    svix::api::Svix, ReceiverInputOpts, ReceiverOutput, SenderInput, SenderOutputOpts, SvixOptions,
+    TransformationConfig,
 };
 use tracing::Level;
+
+#[derive(Deserialize)]
+#[serde(untagged)]
+pub enum EitherReceiver {
+    Webhook(WebhookReceiverConfig),
+    Poller(PollerReceiverConfig),
+}
 
 #[derive(Deserialize)]
 #[serde(deny_unknown_fields)]
@@ -27,7 +35,7 @@ pub struct Config {
     pub senders: Vec<WebhookSenderConfig>,
     /// Config for receiving webhooks and forwarding them to plugins.
     #[serde(default)]
-    pub receivers: Vec<WebhookReceiverConfig>,
+    pub receivers: Vec<EitherReceiver>,
     /// The log level to run the service with. Supported: info, debug, trace
     #[serde(default)]
     pub log_level: LogLevel,
@@ -81,18 +89,25 @@ impl Config {
             }
         }
 
-        for rc in &cfg.receivers {
-            if let Some(tc) = &rc.transformation {
-                crate::runtime::validate_script(tc.source().as_str()).map_err(|e| {
-                    Error::new(
-                        ErrorKind::Other,
-                        format!(
-                            "failed to parse transformation for receiver `{}`: {:?}",
-                            &rc.name, e,
-                        ),
-                    )
-                })?;
-            }
+        for (name, tc) in cfg.receivers.iter().filter_map(|either| match either {
+            EitherReceiver::Webhook(receiver) => receiver
+                .transformation
+                .as_ref()
+                .map(|tc| (&receiver.name, tc)),
+            EitherReceiver::Poller(receiver) => receiver
+                .transformation
+                .as_ref()
+                .map(|tc| (&receiver.name, tc)),
+        }) {
+            crate::runtime::validate_script(tc.source().as_str()).map_err(|e| {
+                Error::new(
+                    ErrorKind::Other,
+                    format!(
+                        "failed to parse transformation for receiver `{}`: {:?}",
+                        name, e,
+                    ),
+                )
+            })?;
         }
 
         Ok(cfg)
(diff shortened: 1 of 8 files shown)


bridge/svix-bridge/src/webhook_receiver/mod.rs
```diff
@@ -181,5 +197,206 @@ async fn transform(
...
+    } = &poller.input_opts;
+
+    let mut iterator = None;
+
+    loop {
+        tracing::trace!(app_id, subscription_id, "polling `/events`");
+        match poller
+            .svix_client
+            .message()
+            .events(V1MessageStreamParams {
```
svix-onelson, line 271: The new "subscription" endpoint is not yet available - this will need to switch as soon as it is.

bridge/svix-bridge/src/config/mod.rs
```diff
@@ -238,5 +253,61 @@ impl WebhookReceiverConfig {
     }
 }
 
+#[derive(Clone, Deserialize)]
+#[serde(tag = "type", rename_all = "kebab-case")]
+pub enum PollerInputOpts {
+    SvixEvents {
+        app_id: String,
+        subscription_id: String,
```
svix-onelson, line 261: These settings are subject to change pending https://github.com/svix/svix-webhooks-private/pull/2484

bridge/svix-bridge/src/config/mod.rs
```diff
@@ -238,5 +253,61 @@ impl WebhookReceiverConfig {
...
+            PollerInputOpts::SvixEvents {
+                svix_token,
+                svix_options,
+                ..
+            } => Some(Svix::new(
+                svix_token.clone(),
+                svix_options.clone().map(Into::into),
+            )),
+        }
+    }
```
svix-jplatte, line 280: Looks like this always returns `Some`, why wrap the `Svix` value in an `Option` at all?

bridge/svix-bridge/src/main.rs
```diff
@@ -326,9 +361,34 @@ async fn main() -> Result<()> {
...
+            EitherReceiver::Webhook(x) => {
+                acc.0.push(x);
+                acc
+            }
+            EitherReceiver::Poller(y) => {
+                acc.1.push(y);
+                acc
+            }
+        },
+    );
```
svix-jplatte, line 379: Can use `partition_map` from `itertools`.

bridge/svix-bridge/src/webhook_receiver/mod.rs
```diff
@@ -181,5 +197,161 @@ async fn transform(
...
+                    )
+                    .await
+                    {
+                        Err(status) => {
+                            tracing::error!(
+                                status = status.as_u16(),
+                                "error while parsing polled message"
+                            );
+                            // BACKOFF
+                            sleep_time = (sleep_time * 2).clamp(MIN_SLEEP, MAX_SLEEP);
```
svix-jplatte, line 306: Nice use of `clamp` :)

bridge/svix-bridge/src/webhook_receiver/mod.rs
```diff
@@ -181,5 +197,161 @@ async fn transform(
...
+                            status = status.as_u16(),
+                            "error while handling polled message"
+                        );
+                        // BACKOFF
+                        sleep_time = (sleep_time * 2).clamp(MIN_SLEEP, MAX_SLEEP);
+                    }
+                }
+                tracing::trace!(
+                    iterator =? &iterator,
+                    next_iterator =? &resp.iterator,
```
svix-jplatte, line 324: I find this spelling a bit obscure, though I guess I see its advantages. This should have exactly the same semantics and is how I usually see tracing fields written:

```suggestion
                    ?iterator,
                    next_iterator = ?resp.iterator,
```

bridge/svix-bridge/src/webhook_receiver/mod.rs
```diff
@@ -181,5 +197,161 @@ async fn transform(
...
+                    iterator =? &iterator,
+                    "request failed, retrying current iterator"
+                );
+                // BACKOFF
+                sleep_time = (sleep_time * 2).clamp(MIN_SLEEP, MAX_SLEEP);
+            }
+        }
+
+        if !sleep_time.is_zero() {
+            tracing::trace!(sleep_time=?sleep_time, "sleeping");
```
svix-jplatte, line 350:
```suggestion
            tracing::trace!(?sleep_time, "sleeping");
```

svix-jplatte (dismissed):
Nothing looks wrong, left some cleanup notes though (can be addressed in another PR).

svix-onelson (commented):
I'll take a look at making the itertools change while waiting on the rest of this stuff to coalesce. 🙏🏻

bridge/svix-bridge/src/webhook_receiver/mod.rs
```diff
@@ -181,5 +197,161 @@ async fn transform(
...
+                    NO_SLEEP
+                };
+            }
+
+            Err(err) => {
+                tracing::trace!(
+                    error =? err,
+                    iterator =? &iterator,
+                    "request failed, retrying current iterator"
+                );
```
svix-jplatte, line 343: Since you updated the other two, how about doing this one as well :smile: 

```suggestion
                tracing::trace!(
                    error = ?err,
                    ?iterator,
                    "request failed, retrying current iterator"
                );
```

svix-onelson:
## OSS Review: itertools

This popular extension crate adds new methods for iterators.
Useful in this case as it cleans up a previous call to `fold()` designed
to partition a collection into 2 collections that can't be expressed
cleanly with a plain `partition()`.

The selected version is not the newest, but it aligns with a selected
version already included in the project transitively.

cc @svix-james @jaymell

svix-onelson:
Some relatively new stuff since last review added, starting with 47e35a764ee8ef5286d703f44ae3e81e76de8481

- new config using the new "subscription token" concept
- revisions to the backoff/sleep handling in the loop

svix-james (approved):

PR #1366, opened 2024-07-10
Libs: bump spec

svix-onelson:


python/mypy.ini
```diff
@@ -1,2 +1,6 @@
 [mypy]
-strict = True
\ No newline at end of file
+strict = True
+exclude = (?x)(
```
tasn, line 3: I wouldn't fix it, let's just fix the issue...

svix-onelson:
Tried to build bridge from the rust sdk in this branch and got
```
error: could not compile `svix` (lib) due to 87 previous errors
```
so something's not right 😬

tasn (approved):

PR #1369, opened 2024-07-12
Bridge: use batch count=0 to signal backoff, not "done"

svix-onelson:
Looks like the `done` field doesn't signify we're at the end of the topic like I originally thought.

Using the count as a proxy for this works well enough for now, but may not in the future if we expose filters like channels/event types.

Follow-on from https://github.com/svix/svix-webhooks/pull/1363


svix-onelson:
I did check that the field is being set as I expect when we return each iterator and I think the problem I was seeing while testing the poller locally is just that create message rate limiting meant bridge was draining the stream faster than I was filling it.

I _do see better throughput_ by using count, which makes sense, but I'm also polling _much harder_ using this method. I think we should just leave it where it is for now and I should just raise my own rate limit 😅

PR #1374, opened 2024-07-25
bridge: Require type field for kafka input, output

svix-jplatte:
… same as for other inputs and outputs.
Also mention that kafka is supported in the `README.md`.


svix-onelson (approved):
Thanks for keeping the kitchen sink examples up to date! Easy to forget about them.

PR #1382, opened 2024-08-02
Bridge: rebuild rabbitmq producer on error

svix-onelson:
Previously there were two defects related to RabbitMQ connection handling as a receiver output:

1. When unable to connect to rabbitmq on startup, bridge would exit immediately.
2. When forwarding to rabbitmq resulted in an error, the connection could be left in a bad state and not recover leaving all subsequent deliveries to the node to fail until bridge was restarted.

To address both of these issues, `QueueForwarder` now initializes clients lazily, on demand, and will tear down the client when errors are encountered.

When the client is discarded, the output will re-initialize another client using the same connection parameters at the time of the next delivery to that output.

For webhook receiver and poller inputs, the HTTP client POSTing the input will see a 500 response in the case the publish to rabbit fails and can retry at their convenience. Whenever they do, the rabbitmq producer should either re-initialize (or fail trying). Rinse, repeat.

### DIFF
--- a/bridge/svix-bridge-plugin-queue/src/receiver_output/mod.rs
+++ b/bridge/svix-bridge-plugin-queue/src/receiver_output/mod.rs
@@ -1,38 +1,47 @@
 use std::sync::Arc;
 
-use omniqueue::DynProducer;
+use omniqueue::{DynProducer, QueueError};
 use svix_bridge_types::{async_trait, BoxError, ForwardRequest, ReceiverOutput};
+use tokio::sync::Mutex;
 
 use crate::{config::QueueOutputOpts, error::Result};
 
 #[derive(Clone)]
 pub struct QueueForwarder {
     name: String,
-    // FIXME: if we retain things like the queue name we can show this in the Debug impl
+    opts: QueueOutputOpts,
     // FIXME: raw payloads not yet supported for receivers, but probably should be.
-    sender: Arc<DynProducer>,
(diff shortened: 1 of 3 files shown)


bridge/svix-bridge-plugin-queue/src/receiver_output/mod.rs
```diff
@@ -44,7 +53,38 @@ impl ReceiverOutput for QueueForwarder {
...
+        // `QueueError::Generic` boxes the inner error so either we'd need to have a direct dep on
+        // `lapin` so we can downcast to inspect specifics, or otherwise we'd need omniqueue to
+        // re-export the error types.
+        //
+        // In practice multiple error types can show up that all mean the same thing so to start,
+        // just recycle for any rabbitmq error.
+        if let (Err(QueueError::Generic(_err)), QueueOutputOpts::RabbitMQ(_cfg)) =
+            (&res, &self.opts)
+        {
+            let _ = sender.take();
```
svix-aaron1011, line 83: Should we log the error here, so that we can tell if there's a bad configuration causing it to constantly reconnect?

svix-onelson: The error will get logged already on L86 (just below here). This is just checking to see if two things are true:

1. there was an error
2. we're talking to rabbit

... in which case we want to clear the sender so it can be rebuilt.

svix-aaron1011 (approved):

PR #1385, opened 2024-08-05
Improve clippy helpfulness

svix-jplatte:
- Make useful lints show up in more places (first and last commit)
- Disable a particular lint that has an annoying false positive case


svix-onelson (approved):
👍🏻

PR #1394, opened 2024-08-06
Libs(Python): fix datetime serialization for `before`/`after

svix-onelson:
The API server expects timestamps to be formatted a specific way (an RFC 3339, UTC timestamp "with a Z").

This diff tweaks how we produce strings for the `before` and `after` query params for message and attempt list requests, conforming to the expected format.

For `before` and `after` we have a convenient place to influence serialization by overriding the `to_dict` method, but a similar problem exists exists for `since` and `until` used in `EndpointStatsOptions`.

I didn't see a trivial way to influence the serialization in these cases without breaking the typing.

It seems like there should be a more uniform solution, but I don't yet know what it is. Ideally all datetimes we pass through to the openapi codegen API would be encoded the proper way.

### DIFF
--- a/python/svix/api.py
+++ b/python/svix/api.py
@@ -1,6 +1,6 @@
 import typing as t
 from dataclasses import asdict, dataclass, field
-from datetime import datetime
+from datetime import datetime, timezone
 
 from deprecated import deprecated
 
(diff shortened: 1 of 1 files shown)


svix-onelson:
Reading more of the openapi details in the generated code, I don't think this approach will work. There's a gap in the typing at the superficial callsites, but `datetime` values are expected deeper down. More research required for the right fit. Hopefully the correct thing will fit for `since`/`until` as well.

svix-onelson:
Reading more of the openapi details in the generated code, I don't think this approach will work. There's a gap in the typing at the superficial callsites, but `datetime` values are expected deeper down. More research required for the right fit. Hopefully the correct thing will fit for `since`/`until` as well.

svix-onelson:
https://github.com/svix/svix-webhooks/pull/1395 looks like it'll do what we need.

PR #1395, opened 2024-08-06
Libs(Python): ensure datetimes have a TZ

svix-onelson:
The server fails to parse timestamps without any timezone.

This diff treats any timestamp without a TZ set as if it's in UTC.

Folks who want to pass a local time need to be sure to set the relevant TZ themselves else we'll wind up treating it as UTC without shifting the various components.

### DIFF
--- a/python/svix/api.py
+++ b/python/svix/api.py
@@ -1,6 +1,6 @@
 import typing as t
 from dataclasses import asdict, dataclass, field
-from datetime import datetime
+from datetime import datetime, timezone
 
 from deprecated import deprecated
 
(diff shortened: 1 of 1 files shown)


python/svix/api.py
```diff
@@ -227,6 +236,14 @@ class MessageListOptions(ListOptions):
     channel: t.Optional[str] = None
     tag: t.Optional[str] = None
 
+    def to_dict(self) -> t.Dict[str, t.Any]:
+        d = super().to_dict()
+        if self.before is not None:
+            d["before"] = ensure_tz(self.before)
+        if self.after is not None:
+            d["after"] = ensure_tz(self.after)
```
svix-onelson, line 244: The super impl for `to_dict` prunes out None values so we have to guard these calls 😢

python/svix/api.py
```diff
@@ -549,8 +574,8 @@ async def get_stats(
             client=self._client,
             app_id=app_id,
             endpoint_id=endpoint_id,
-            since=options.since,
-            until=options.until,
+            since=ensure_tz(options.since),
+            until=ensure_tz(options.until),
```
svix-onelson, line 578: The None values would have been passed through here as-is previously so we can just "decorate" the options without guarding.

tasn (approved):

PR #1405, opened 2024-08-26

svix-jplatte:


svix-onelson:
https://github.com/svix/svix-webhooks/pull/1406 will hopefully fix the Go lint failure, but the python one I'm not sure about.

PR #1411, opened 2024-08-28
Bump version and update changelog

svix-lucho:
## Version 1.31.0
* Bridge: remove Beta tag
* Libs/Go: prune out unused import for "time" in codegen output
* Libs/C#: Fix appId, endpointId order in RotateSecretWithHttpInfoAsync
* Libs: Support passing `application` to `app-portal-access` endpoint


svix-onelson (approved):
Thanks for this 🙏🏻

Issue #1415, opened 2024-09-03
[Python] Create Event Type HTTP response fails to be parsed


anlambert:
Since `svix v1.29.0` and up to current release, the following error is reported when attempting to create an event type using Python:
```

swh_webhooks = <swh.webhooks.interface.Webhooks object at 0x7fb740857c90>
origin_create_event_type = EventType(name='origin.create', description='This event is triggered when a new software origin is added to the archiv...string', 'description': 'The URL of the newly created software origin', 'format': 'iri'}}, 'required': ['origin_url']})

    def test_create_valid_event_type(swh_webhooks, origin_create_event_type):
>       swh_webhooks.event_type_create(origin_create_event_type)

swh/webhooks/tests/test_webhooks.py:93: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
build/__editable__.swh.webhooks-0.3.1.dev3+g9b381c6.d20240903-py3-none-any/swh/webhooks/interface.py:251: in event_type_create
    self.svix_api.event_type.create(
../../../.virtualenvs/swh/lib/python3.11/site-packages/svix/api.py:885: in create
    return v1_event_type_create.request_sync(
../../../.virtualenvs/swh/lib/python3.11/site-packages/svix/internal/openapi_client/api/event_type/v1_event_type_create.py:154: in request_sync
    return request_sync_detailed(
../../../.virtualenvs/swh/lib/python3.11/site-packages/svix/internal/openapi_client/api/event_type/v1_event_type_create.py:125: in request_sync_detailed
    return _build_response(client=client, response=response)
../../../.virtualenvs/swh/lib/python3.11/site-packages/svix/internal/openapi_client/api/event_type/v1_event_type_create.py:72: in _build_response
    parsed=_parse_response(client=client, response=response),
[...]

PR #1416, opened 2024-09-04
Add deprecated field to event types

svix-jplatte:
## Motivation

Fixes the first bug reported in #1415.

## Solution

Backport event-type deprecated support.


server/svix-server/migrations/20240806212329_add_event_type_deprecated.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE eventtype ADD COLUMN deprecated BOOLEAN DEFAULT FALSE NOT NULL;
```
svix-jplatte, line 1: We haven't added a new migration to the OSS version in quite a long time, is it fine to add one here?

svix-aaron1011 (approved):

PR #1425, opened 2024-09-11
Repair broken tests

svix-jplatte:
## Motivation

Our test script was broken :facepalm: 
A lot of tests were failing because they were asserting wrong things about server responses.
We didn't see it in CI because the script was missing `set -e`.

## Solution

Make it actually fail if a command other than the last one fails. Repair broken tests.


svix-onelson (approved):
:shipit:

PR #1426, opened 2024-09-11
Add support for http proxies

svix-jplatte:
Adds support for HTTP(S) proxies, in addition to SOCKS5 proxies that we've been supporting for a little while.

Closes https://github.com/svix/monorepo-private/issues/8990.

### DIFF
--- a/server/svix-server/tests/it/e2e_proxy.rs
+++ b/server/svix-server/tests/it/e2e_proxy.rs
@@ -17,21 +17,38 @@ async fn test_message_delivery_via_socks5() {
     use crate::utils::start_svix_server_with_cfg;
 
     let mut cfg = get_default_test_config();
-    cfg.proxy_config = Some(proxy_config());
+    cfg.proxy_config = Some(socks_proxy_config());
     let (client, _) = start_svix_server_with_cfg(&cfg).await;
-    run_socks5_test(&client).await;
+    run_proxy_test(&client).await;
 }
 
-fn proxy_config() -> ProxyConfig {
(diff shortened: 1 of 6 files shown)


server/svix-server/tests/it/e2e_proxy.rs
```diff
@@ -17,21 +17,38 @@ async fn test_message_delivery_via_socks5() {
...
 
-fn proxy_config() -> ProxyConfig {
+fn socks_proxy_config() -> ProxyConfig {
     ProxyConfig {
         addr: ProxyAddr::new("socks5://localhost:1080").unwrap(),
     }
 }
 
-async fn run_socks5_test(client: &TestClient) {
+#[ignore] // requires an http proxy to be running
```
svix-jplatte, line 31: Unlike last time, I didn't even try to add `tinyproxy` to our docker compose stack. I could find way too many images for it on Docker Hub, but none looked sufficiently trustworthy. I tested this with a locally-running proxy instead. Suggestions for a Docker image to use would be welcome, but I think it's not required for merging this.

svix-jplatte:
Open source review: This PR adds a dependency on `hyper-proxy`.

It's not maintained and not by a well-known author, so I have audited it and found nothing suspicious. I also pinned it so even in the case of the owner's account being taken over, we are not impacted.

Once we upgrade to hyper 1.0, we will have to replace this crate, likely with `hyper-http-proxy` (which I couldn't use here because it _only_ supports hyper 1.0).

svix-james:
Open source: Approved

PR #1437, opened 2024-09-13
server: Make figment-jailed tests run sequentially

svix-jplatte:
## Motivation

figment's Jail does not actually isolate environments, it only stores the initial environment state and restores it after the test. This means that multiple tests using `jail.set_env` can break each other.

## Solution

Make these tests run sequentially.

### DIFF
--- a/server/Cargo.lock
+++ b/server/Cargo.lock
@@ -4026,6 +4026,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "scc"
+version = "2.1.17"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


svix-jplatte:
Open Source Review: `serial_test`

This crate is the fourth-most popular in the [Testing](https://crates.io/categories/development-tools::testing) category on crates.io. It has been downloaded millions of times.

svix-jplatte:
Actually I just found this comment in another of the config tests and now I want to just kill `figment::Jail` usage instead:

```rust
        // NOTE: Does not use `figment::Jail` like the above because set env vars will leak into
        // other tests overwriting real configurations
```

PR #1439, opened 2024-09-16
Fix issue with Python client version 1.34 - AttributeError: 'dict' object has no attribute 'to_dict'

qbzenker:
We run Svix in production and when we upgraded to the 1.34 version we started seeing the error:

```
AttributeError: 'dict' object has no attribute 'to_dict'
```

This PR aims to fix the issue - I think `payload` is already a dict when `to_dict` is called.

Thank you!

### DIFF
--- a/python/svix/api.py
+++ b/python/svix/api.py
@@ -1074,7 +1074,7 @@ class MessageAsync(ApiBase):
             with_content=False,
             **options.to_dict(),
         )
-        ret.payload = MessageOutPayload.from_dict(message_in.payload.to_dict())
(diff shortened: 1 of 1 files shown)


svix-jplatte:
Hey! Thanks for the PR and sorry for releasing a broken version! We're looking into it and will provide a fix soon.
There is more to it than the error this PR fixes, so we'll fix everything in one go on our own. Please subscribe to #1440 to get updates on this bug, or subscribe to releases on the repo to get notified when we publish a new release with the fix.

Issue #1440, opened 2024-09-16
[Python] Svix client v1.34 is broken (again)


anlambert:
I recently reported https://github.com/svix/svix-webhooks/issues/1415 and wanted to check if the issue was resolved `svix 1.34`.

After bumping svix server and Python client version to `1.34`, I got the following errors when executing `mypy` on the code using the svix Python client I am working on:
```
swh/webhooks/interface.py:255: error: Argument "schemas" to "EventTypeIn" has incompatible type "dict[str, dict[str, Any]]"; expected "Unset | EventTypeInSchemas | None"  [arg-type]
swh/webhooks/interface.py:265: error: Argument "schemas" to "EventTypeUpdate" has incompatible type "dict[str, dict[str, Any]]"; expected "Unset | EventTypeUpdateSchemas | None"  [arg-type]
swh/webhooks/interface.py:397: error: Argument "metadata" to "EndpointIn" has incompatible type "dict[str, Any]"; expected "Unset | EndpointInMetadata"  [arg-type]
swh/webhooks/interface.py:414: error: Argument "headers" to "EndpointHeadersIn" has incompatible type "dict[str, str]"; expected "EndpointHeadersInHeaders"  [arg-type]
swh/webhooks/interface.py:468: error: "EndpointOutMetadata" has no attribute "pop"  [attr-defined]
swh/webhooks/interface.py:571: error: Argument "payload" to "MessageIn" has incompatible type "dict[str, Any]"; expected "MessageInPayload"  [arg-type]
swh/webhooks/interface.py:718: error: "EndpointOutMetadata" has no attribute "get"  [attr-defined]
```
New Python types for svix API parameters seem to have been introduced so I updated my code accordingly, below is an example of using the new `MessageInPayload` type.
```diff
diff --git a/swh/webhooks/interface.py b/swh/webhooks/interface.py
index ea00983..4ab418c 100644
--- a/swh/webhooks/interface.py
+++ b/swh/webhooks/interface.py
@@ -31,6 +31,7 @@ from svix.api import (
     MessageAttemptListOptions,
     MessageAttemptOut,
     MessageIn,
+    MessageInPayload,
     MessageListOptions,
     Ordering,
     Svix,
@@ -568,7 +569,7 @@ class Webhooks:
                 app_uid,
                 MessageIn(
                     event_type=event_type_name,
-                    payload=dict(sorted(payload.items())),
+                    payload=MessageInPayload.from_dict(dict(sorted(payload.items()))),
[...]

PR #1442, opened 2024-09-17
Revert changes to Python code generation

svix-jplatte:
## Motivation

The Python library had a pretty significant breaking change. See #1440.

## Solution

Revert the Python codegen changes from #1423.

### DIFF
--- a/python/svix/api.py
+++ b/python/svix/api.py
@@ -1074,7 +1074,7 @@ class MessageAsync(ApiBase):
             with_content=False,
             **options.to_dict(),
         )
-        ret.payload = MessageOutPayload.from_dict(message_in.payload.to_dict())
+        ret.payload = message_in.payload
         return ret
 
     async def get(self, app_id: str, msg_id: str) -> MessageOut:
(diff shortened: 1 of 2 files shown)


python/svix/api.py
```diff
@@ -1074,7 +1074,7 @@ async def create(
             with_content=False,
             **options.to_dict(),
         )
-        ret.payload = MessageOutPayload.from_dict(message_in.payload.to_dict())
+        ret.payload = message_in.payload
```
tasn, line 1077:
```suggestion
        ret.payload = t.Cast(message_in.payload, t.Any)
```
Something like this?

python/templates/model.py.jinja
```diff
@@ -1,5 +1,5 @@
 {% set class_name = model.class_info.name %}
-{% if model.required_properties == [] and model.optional_properties == [] and False %}
+{% if model.required_properties == [] and model.optional_properties == [] %}
```
tasn, line 2: The revert looks correct, the `and False` looks silly and that it'll break stuff, so no wonder if broke stuff.

svix-jplatte:
Superseded by #1443.

PR #1448, opened 2024-09-24
Libs(Go): edit `EndpointIn` to allow null for channels, filter types

svix-onelson:
This is a stop-gap "quick fix" to allow callers to successfully create endpoints without specifying channels and filter types.

Today, they get a 422 response when passing an empty array (as required by our lib).

With this diff, the fields are set to pointers so they can be properly nulled out, as they were prior to #545.

Research required to see what we need to do so the codegen does the right thing given the spec (which produces good output for the other libs).

### DIFF
--- a/go/endpoint_test.go
+++ b/go/endpoint_test.go
@@ -1,0 +1,76 @@
+package svix_test
+
+import (
+	"encoding/json"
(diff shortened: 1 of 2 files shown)


svix-james (approved):

svix-james:
heckuva job!

PR #1449, opened 2024-09-24
Bump version and update changelog

svix-lucho:
## Version 1.36.0
* Libs(Go): edit `EndpointIn` to allow null for channels, filter types


svix-onelson (approved):
🙏🏻

PR #1450, opened 2024-09-24
Libs(Go): template overrides for nullable non-required containers

svix-onelson:
In #1448 the `EndpointIn` model was hand-edited such that `channels` and `filterTypes` were nullable, and thus correctly omitted from request bodies sent to the server.

This series of commits attempts to remove the need to keep reapplying this patch with each spec bump by updating the codegen template used to produce the model code.

This change is comprised of 3 commits:
1. the addition of a template override for `model_simple` ([src](https://github.com/OpenAPITools/openapi-generator/blob/v5.2.0/modules/openapi-generator/src/main/resources/go/model_simple.mustache)) as-is, taken from openapi at the v5.2.0 tag (matching the version of the generator we're using currently).
2. some modifications to said template, i.e. _our patch_.
3. updated go sources from running the generator.

When reviewing the template, it'll therefore be useful to go commit by commit so you can see the diff applied against the upstream.

This change has widespread impact as many models are changing. I'm still trying to figure out how we can prove everything is kosher and not causing accidental breakage. Open to ideas 🤔 

Existing tests are passing, and everything typechecks, so at the very least the codegen output is self-consistent.

### DIFF
--- a/go/internal/openapi/model_app_usage_stats_in.go
+++ b/go/internal/openapi/model_app_usage_stats_in.go
@@ -17,7 +17,7 @@ import (
 
 // AppUsageStatsIn struct for AppUsageStatsIn
 type AppUsageStatsIn struct {
-	AppIds []string `json:"appIds,omitempty"`
+	AppIds *[]string `json:"appIds,omitempty"`
 	Since time.Time `json:"since"`
 	Until time.Time `json:"until"`
 }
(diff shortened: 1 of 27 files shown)


svix-james:
It'd be great to have some basic tests before merging this / confirmation that they're passing. They don't need to be wired into CI now, but just having some that can be run manually would be a good starting point.

svix-onelson:
> It'd be great to have some basic tests before merging this / confirmation that they're passing. They don't need to be wired into CI now, but just having some that can be run manually would be a good starting point.

I raised https://github.com/svix/svix-webhooks/pull/1452 which will run any Go tests for this in CI.

Will need to add more tests to cover the 28 model sources impacted, and like you suggested, they may need to be skipped in CI until we get the infra side of this figured out.

It'll take me a minute to get that going. In the meantime, #1452 should de-risk an accidental revert of the hand-edit for `EndpointIn`.

svix-james:
So do you want this merged now, or hold off?

svix-onelson:
> So do you want this merged now, or hold off?

Hold off.

tasn (approved):

svix-james (approved):

PR #1470, opened 2024-10-02
Add 'message.attempt.recovered' operational webhook

### LINKED ISSUE #1453: the api docs show a "message.attempt.recovered" event - but the code doesn´t show this event
hi there,
i would like to now if the API docs are wrong or if I am missing something in the code,
because the API docs show a [message.attempt.recovered event](https://api.svix.com/docs#tag/Webhooks/operation/message.attempt.recovered), but the [code](https://github.com/svix/svix-webhooks/blob/main/server/svix-server/src/core/operational_webhooks.rs) does not.

image

svix-aaron1011:
This event is fired when a successful attempt is made after several attemtps have already failed.
It already exists in the hosted svix.com product

Fixes #1453


svix-onelson (approved):
LGTM

PR #1472, opened 2024-10-02
Add redis sentinel support

jaymell:
This adds support for using Redis Sentinel for queuing and caching.
Sentinel potentially requires a lot of additional configuration params,
so support for that that has been added here, although the `redis_dsn`
field has been reused and should point to a sentinel server if using the 
`redissentinel` queue/cache types.

### DIFF
--- a/server/svix-server/Cargo.toml
+++ b/server/svix-server/Cargo.toml
@@ -44,8 +44,8 @@ ed25519-compact = "2.1.1"
 chrono = { version="0.4.26", features = ["serde"] }
 reqwest = { version = "0.11.27", features = ["json", "rustls-tls", "hickory-resolver"], default-features = false }
 bb8 = "0.8"
-bb8-redis = "0.15.0"
-redis = { version = "0.25.4", features = ["tokio-comp", "tokio-native-tls-comp", "streams", "cluster-async", "tcp_nodelay", "connection-manager"] }
+bb8-redis = "0.16.0"
+redis = { version = "0.26", features = ["tokio-comp", "tokio-native-tls-comp", "streams", "cluster-async", "tcp_nodelay", "connection-manager", "sentinel"] }
 thiserror = "1.0.30"
 bytes = "1.1.0"
 blake2 = "0.10.4"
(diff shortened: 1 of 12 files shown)


server/svix-server/Cargo.toml
```diff
@@ -68,7 +68,7 @@ urlencoding = "2.1.2"
 form_urlencoded = "1.1.0"
 lapin = "2.1.1"
 sentry = { version = "0.32.2", features = ["tracing"] }
-omniqueue = { git = "https://github.com/svix/omniqueue-rs", rev = "5ae22000e2ea214ba707cac81657f098e5785a76", default-features = false, features = ["in_memory", "rabbitmq-with-message-ids", "redis_cluster"] }
+omniqueue = { git = "https://github.com/jaymell/omniqueue-rs", rev = "55f3bccb4298c5de75da1358675f6bfadb016222", default-features = false, features = ["in_memory", "rabbitmq-with-message-ids", "redis_cluster", "redis_sentinel"] }
```
jaymell, line 71: This will be updated once upstream PR is merged.

server/svix-server/src/redis/mod.rs
```diff
@@ -1,88 +1,177 @@
...
+}
+
 #[derive(Clone, Debug)]
 pub enum RedisManager {
     Clustered(ClusteredRedisPool),
     ClusteredUnpooled(ClusteredRedisUnpooled),
     NonClustered(NonClusteredRedisPool),
     NonClusteredUnpooled(NonClusteredRedisUnpooled),
+    Sentinel(SentinelPooled),
+    SentinelUnpooled(SentinelUnpooled),
```
svix-onelson, line 33: Not sure the implications, but I wonder how feasible it'd be to split pooled/unpooled out as a different facet entirely so we don't have to have 2 variants for each of these.

This would probably be an upheaval, but can "use a pooler" just be an option on each of these?

server/svix-server/src/redis/sentinel.rs
```diff
@@ -0,0 +1,53 @@
+use axum::async_trait;
+use redis::{
+    sentinel::{SentinelClient, SentinelNodeConnectionInfo, SentinelServerType},
+    ErrorKind, IntoConnectionInfo, RedisError,
+};
+use tokio::sync::Mutex;
+
+struct LockedSentinelClient(pub(crate) Mutex<SentinelClient>);
```
svix-onelson, line 8: I left a comment on the omniqueue side about favoring mutex over rwlock. Same question, different repo.

Any read-only access anticipated? If so, it might be a nice thing to switch to rwlock.

I guess even while the inner is pub, the outer is not so maybe we can quietly switch later without any impact.

jaymell: Commented on other PR, but I have a pending change in redis-rs that will hopefully eliminate the need for a Mutex here, so stay tuned.

server/svix-server/src/redis/mod.rs
jaymell: I have a PR forthcoming that will further refactor the this module. I'll consider this as well.

PR #1473, opened 2024-10-03
bump `path-to-regexp` via `openapi-generator-cli`

svix-onelson:
Addresses https://github.com/svix/svix-webhooks/security/dependabot/128 by bumping `@openapitools/openapi-generator-cli` to transitively bump `path-to-regexp`.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,38 +2,43 @@
 # yarn lockfile v1
 
 
+"@babel/runtime@^7.21.0":
+  version "7.25.7"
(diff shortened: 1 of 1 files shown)


svix-lucho (approved):
I'm assuming the openapi generator upgrade doesn't have any breaking changes since it's a minor version, right?

svix-onelson:
@svix-lucho The generator version and the cli version are two different things. The generator version should still be pinned to 5.2.x (via the `version` field in the `openapitools.json` file).

Ref: https://github.com/svix/svix-webhooks/blob/e794bdf4f4b7a921f74397ee768d58c928b558a5/openapitools.json#L5

PR #1477, opened 2024-10-04
More redis refactors

jaymell:
This removes a lot of the needless indirection/structs that were obscuring what should be a pretty simple thing -- i.e., the different variants for `ConnectionManager`s and `RedisConnection`s.

Note also that by moving some of the `query_async` calls to the `Cmd` rather than the connection, we can completely remove the `*Connection` types and associated macros.

There's definitely some more cleanup that can be done here, but I think this is good for one PR.


svix-onelson (approved):
👌🏻

PR #1478, opened 2024-10-04
Redis dlq support

jaymell:
Add support for moving failed messages to deadletter queue in
Redis. The queue can be re-driven via an undocumented endpoint,
`/api/v1/admin/redrive-dlq`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -389,6 +389,10 @@ To support graceful shutdown on the server, all running tasks are finished befor
 One of our main goals with open sourcing the Svix dispatcher is ease of use. The hosted Svix service, however, is quite complex due to our scale and the infrastructure it requires. This complexity is not useful for the vast majority of people and would make this project much harder to use and much more limited.
 This is why this code has been adjusted before being released, and some of the features, optimizations, and behaviors supported by the hosted dispatcher are not yet available in this repo. With that being said, other than some known incompatibilities, the internal Svix test suite passes. This means they are already mostly compatible, and we are working hard on bringing them to full feature parity.
 
+# Re-driving Redis DLQ
+We have an undocumented endpoint for re-driving failed messages that are DLQ'ed. You can do this by calling `POST /api/v1/admin/redrive-dlq/`.
+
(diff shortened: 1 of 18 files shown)


README.md
```diff
@@ -389,6 +389,8 @@ To support graceful shutdown on the server, all running tasks are finished befor
 One of our main goals with open sourcing the Svix dispatcher is ease of use. The hosted Svix service, however, is quite complex due to our scale and the infrastructure it requires. This complexity is not useful for the vast majority of people and would make this project much harder to use and much more limited.
 This is why this code has been adjusted before being released, and some of the features, optimizations, and behaviors supported by the hosted dispatcher are not yet available in this repo. With that being said, other than some known incompatibilities, the internal Svix test suite passes. This means they are already mostly compatible, and we are working hard on bringing them to full feature parity.
 
+# Re-driving Redis DLQ
+We have an undocumented endpoint for re-driving failed messages that are DLQ'ed. You can do this by calling `PUT /api/v1/admin/redrive-dlq/`.
```
svix-gabriel, line 393: This is maybe too nitpicky, but since this endpoint has side effects and technically isn't idempotent, shouldn't we use `POST`?

svix-gabriel (approved):
I left a minor nit, but otherwise LGTM.

Will we have admin endpoints for viewing the DLQ too?

README.md
jaymell: Nah, that seems reasonable.

jaymell:
> Will we have admin endpoints for viewing the DLQ too?

Yeah we probably need it. Elsewhere we monitor the queue depth, so at a minimum I should probably just add that for now. Otherwise it's completely opaque what's going on w/ the queue since we create/manage it internally.

jaymell:
For now, I've added metrics to monitor the DLQ (and other queues) via metrics (#1479). We can consider adding additional admin endpoints later, but for now the metrics will prevent DLQ contents from being _completely_ opaque to the user.

README.md
jaymell: Fixed.

svix-gabriel (approved):

PR #1486, opened 2024-10-16
soft-require typhoeus

### LINKED ISSUE #453: Remove Ruby gem hard-dependency on typhoeus
### Motivation

I'd be happy to use this gem to verify the webhook code, but I don't want to install typhoeus, which isn't needed for it. So, until it's removed, I'll probably just write the HMAC etc code myself.

### Proposal

MVP: make typhoeus a soft requirement for use in the api client, but not for installation. 

### Alternatives

Remove typhoeus altogether.

jjb:
resolves #453

### DIFF
--- a/ruby/Gemfile.lock
+++ b/ruby/Gemfile.lock
@@ -2,15 +2,11 @@ PATH
   remote: .
   specs:
     svix (1.38.0)
-      typhoeus (~> 1.0, >= 1.0.1)
 
 GEM
   remote: https://rubygems.org/
   specs:
     diff-lcs (1.4.4)
-    ethon (0.14.0)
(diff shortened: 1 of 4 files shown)


svix-gabriel (approved):
Thanks so much for tackling this!

jjb:
@svix-gabriel thanks for merging! I wasn't expecting it to get merged without discussion 😅. I didn't test it! 

- someone should test it
- maybe you want to adjust the copy
- maybe you want to inform your support / customer success / customer solutions / etc. teams, there may be additional support load from this

jjb:
@svix-gabriel when do you think the next ruby gem version can be published?

PR #1491, opened 2024-10-18
ci: Run lint jobs when the workflow definition changes

svix-jplatte:
## Motivation

When looking into https://github.com/svix/monorepo-private/issues/9262, I found that we were not testing workflow updates properly.

## Solution

Have the workflow run when the workflow definition changes.


svix-jplatte:
Looks like we were using a .NET Core version past EOL: https://versionsof.net/core/

svix-gabriel (approved):

PR #1492, opened 2024-10-18
Upgrade Node version in Javascript release action

svix-lucho:
The latest version of `openapi-generator-cli` requires a newer version of Node. We can just use LTS/latest, no need to pin to an older version. 

https://github.com/svix/svix-webhooks/actions/runs/11404529698/job/31733893512


jplatte (commented):
Let's do the same thing in the lint workflow, for consistency?

svix-jplatte (approved):

PR #1480, opened 2024-10-08
Libs(all): bump openapi generator from 5.2.0 to 7.9.0

svix-onelson:
With the existing generator versions, we've had issues during codegen for certain cases in the svix spec including:

- serialization of nullable fields
- enum (`oneOf`, `anyOf`) representations

Bumping the generators all the way up to the latest version (at time of writing, `7.9.0`) seems to resolve both of these, at least partially.

Some refactoring of extant template overrides has been required to "re-seat" our local patches on top of the newer templates.

There have also been cases of the generated symbol names using slightly different schemes than in earlier generator versions. The high-level API had to be updated to use the new names.

## Approach

Customization to the codegen comes in 2 flavors:
- edits to templates.
- manual edits where the codegen is checked-in to git and the paths are recorded in an `.openapi-generator-ignore` file.

Templates are better because they don't hide the customization as readily, but we've got a mix of these approaches throughout.

At a high-level, this change followed the following steps:
- replace templates entirely with the version tagged for 7.9.0.
- recreate our local patches _on top_ as a single rev.
- leave behind any customization made moot by upstream refactors.
- try to remove as many manual edits/ignores by migrating into templates where needed or just omitting the change entirely.
- add a light/gentle test to check specific behaviors things we had hacks in place for:
  - nullable containers (ie. Endpoint channels and filter types).
  - responses with empty bodies.

In addition to doing the "upstream merge" on template sources, I also had to sometimes look at the dependencies a lib depends on, upgrading or changing the package manifest, or build tooling.

When refering/comparing to the upstream templates:
- https://github.com/OpenAPITools/openapi-generator/blob/v7.9.0/modules/openapi-generator/src/main/resources/ (the new generator version tag)
- https://github.com/OpenAPITools/openapi-generator/blob/v5.2.0/modules/openapi-generator/src/main/resources/ (the older one)

## Status

Each lib came with its own challenges. Some I have higher or lower confidence in.
I'll call out the ones that are weaker so reviewers can give more scrutiny.

**Ruby**: ~~I wasn't able to get my local ruby setup in a healthy enough state to follow along with the readme instructions for building/testing.
This one needs a so-called "kitchen sink" test like I added to the other libs, and may need some dependencies bumped. Normally the dep updates required become known during testing.~~

[...]

PR #1496, opened 2024-10-29
Libs(Go): unestting nullable array fields via PATCH request

svix-onelson:
Work was done previously [1] to make nullable containers (arrays, maps) generate as refs. This avoided a problem where the JSON serialization code incrrectly sent an empty array for cases where a field was left unset. This would result in requests incorrectly failing with a 422 response when it should have been OK.

During the openapi generator update [2], the resulting codegen no longer had the 422 issue since it correctly omitted empty arrays from the body for PATCH requests. However, this now means we have no way to explicitly set a field to `null`. This specific issue was fixed in the some other langs with the work done to update the generator, but the Go generator seems to have fallen behind.

This rev adds tests to demonstrate the shortfall. We may need to rework the template patch made for the old 5.x generator to solve this in the meantime.

- 1: https://github.com/svix/svix-webhooks/pull/1450
- 2: https://github.com/svix/svix-webhooks/pull/1480

### DIFF
--- a/go/endpoint_test.go
+++ b/go/endpoint_test.go
@@ -7,7 +7,7 @@ import (
 	"testing"
 )
 
-func TestEndpoint_Serialization(t *testing.T) {
+func TestEndpointIn_Serialization(t *testing.T) {
 	testCases := []struct {
 		name            string
 		testEndpoint    *svix.EndpointIn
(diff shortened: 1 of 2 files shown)


svix-mman:
@svix-onelson I think our new Go lib supports this
See this test https://github.com/svix/svix-webhooks/blob/59f0490a027bc22f258bde9c30b1f0c22bcb82b2/go/svix_test.go#L564-L576

PR #1512, opened 2024-11-08

svix-lucho:


svix-james (approved):

go/internal/openapi/api_sink.go
svix-james, line 1: Why is this file deleted?

svix-lucho: It's because of the change in tags in the spec, it was basically renamed to `go/internal/openapi/api_stream.go`

PR #1518, opened 2024-11-11
Bridge: add a test to check process liveness

svix-onelson:
Mentioned in the script comments, we've seen segfaults in bridge coming from deno, and only after the interpreter has had some time to achieve steady-state.

This diff adds a script to run a check where we launch the process with an empty config (no senders or receivers registered), and send it to the background.
Bridge should be able to stay up indefinitely like this. If the process ends before we check on it (after sleeping a spell), that's probably a bug.

I reverted the last deno downgrade to try this script out in the failing state. It seemed to catch the issue well.

```
$ ./run-idle-test.sh
   Compiling svix-bridge v1.39.0 (/home/onelson/Projects/svix-webhooks/bridge/svix-bridge)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.29s
Monitoring PID=189293
fail: process terminated prematurely
```

### DIFF
--- a/.github/workflows/bridge-ci.yml
+++ b/.github/workflows/bridge-ci.yml
@@ -88,6 +88,11 @@ jobs:
     - name: Stop required services
       run: docker compose -f "bridge/testing-docker-compose.yml" down
 
+    - name: Idle Test
+      working-directory: bridge
+      run: ./run-idle-test.sh
+
+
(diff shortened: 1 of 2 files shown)


svix-james (approved):
Looks good!

PR #1524, opened 2024-11-18
js: Let tsc infer type of resp in isomorphic-fetch code

svix-jplatte:
Found while looking into https://github.com/svix/monorepo-private/issues/9283.


javascript/templates/http/isomorphic-fetch.mustache
```diff
@@ -49,7 +49,7 @@ export class IsomorphicFetchHttpLibrary implements HttpLibrary {
             credentials: isCredentialsSupported ? "same-origin" : undefined
             {{/browser}}
             {{/platforms}}
-        }).then((resp: any) => {
+        }).then(resp => {
```
svix-onelson, line 52: I don't see a good reason _not to do it_ but I wonder what motivated the change? Warning from the linter? Just you wondering about what the available fields were on the response object while hacking on the other ticket?

svix-onelson (approved):

PR #1530, opened 2024-11-20
Return 413 on large payloads

blaginin:
## Motivation

On large payloads, the server now returns 500

image

## Solution

Return 413 instead of 500. Also added a test that checks that.

### DIFF
--- a/server/svix-server/src/error.rs
+++ b/server/svix-server/src/error.rs
@@ -379,6 +379,14 @@ impl HttpError {
             detail.unwrap_or_else(|| "This API endpoint is not yet implemented.".to_owned()),
         )
     }
+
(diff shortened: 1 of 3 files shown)


tasn:
Thanks! @svix/engineering will take a look!

blaginin:
I think the [default](https://github.com/tokio-rs/axum/blob/main/axum-core/src/ext_traits/request.rs#L319) payload limit is quite small. Do you think it would make sense to increase it? Would it be better to make it configurable?

tasn:
TBH 2mb is HUGE for a webhook, we actually lowered it on the Svix cloud service. We kept it at the default here because whoever uses it controls the payload size they send anyway, so I don't think it matters too much to let people control it.

svix-james (approved):
Looks good to me. Thanks!

PR #1533, opened 2024-11-27
rust: Add MessageIn::{new_raw_payload, with_content_type}

svix-jplatte:
… and add some extra detail to the docs of the kotlin equivalent.

### DIFF
--- a/kotlin/lib/src/main/kotlin/Message.kt
+++ b/kotlin/lib/src/main/kotlin/Message.kt
@@ -76,10 +76,15 @@ class Message internal constructor(token: String, options: SvixOptions) {
 /**
  * Creates a [MessageIn] with the payload already being serialized.
  *
+ * The payload is not normalized on the server (usually whitespace outside
+ * of string literals, unnecessarily escaped characters in string and such
+ * are fixed up by the server), and is not even required to be JSON.
+ *
  * @param payload Serialized message payload
  * @param contentType Content type of the payload to send as a header. Defaults to `application/json`.
  *
  * See the class documentation for details about the other parameters.
+ *
  */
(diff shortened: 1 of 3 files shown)


kotlin/lib/src/main/kotlin/Message.kt
```diff
@@ -76,10 +76,20 @@ class Message internal constructor(token: String, options: SvixOptions) {
...
  * @param payload Serialized message payload
  * @param contentType Content type of the payload to send as a header. Defaults to `application/json`.
  *
  * See the class documentation for details about the other parameters.
+ *
+ * # OSS server compatibility
+ *
+ * This constructor uses a mechanism that is currently only available in
+ * Svix'es cloud and on-prem enterprise products. The open source server
+ * implementation does not support it as of v1.41.0.
```
tasn, line 92: No need to put it here. If we think it's important, let's fix OSS. Should be easy to add there.

rust/src/model_ext.rs
```diff
@@ -0,0 +1,48 @@
...
+    /// are fixed up by the server), and is not even required to be JSON.
+    ///
+    /// The default `content-type` of `application/json` will still be used,
+    /// unless overwritten with [`with_content_type`][Self::with_content_type].
+    ///
+    /// # OSS server compatibility
+    ///
+    /// This constructor uses a mechanism that is currently only available in
+    /// Svix'es cloud and on-prem enterprise products. The open source server
+    /// implementation does not support it as of v1.41.0.
```
tasn, line 21: No need to spam all of our docs with this notice. :)

svix-james:
LGTM

svix-james (approved):

PR #1538, opened 2024-11-28
go: Add convenient construction of rawPayload messages

svix-jplatte:

### DIFF
--- a/go/message.go
+++ b/go/message.go
@@ -96,3 +96,29 @@ func (m *Message) ExpungeContent(ctx context.Context, appId string, msgId string
 	res, err := req.Execute()
 	return wrapError(err, res)
 }
+
+// Instantiates a new MessageIn object with a pre-serialized payload.
+//
+// The payload is not normalized on the server (usually whitespace outside
+// of string literals, unnecessarily escaped characters in string and such
+// are fixed up by the server), and is not even required to be JSON.
+//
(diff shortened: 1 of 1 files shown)


go/message.go
```diff
@@ -96,3 +96,29 @@ func (m *Message) ExpungeContent(ctx context.Context, appId string, msgId string
 	res, err := req.Execute()
 	return wrapError(err, res)
 }
+
+// Instantiates a new MessageIn object with a pre-serialized payload.
+//
+// The payload is not normalized on the server (usually whitespace outside
```
svix-gabriel, line 102: From 102-104, what about something like
```
The payload is not normalized on the server. Normally, payloads are required to be valid
JSON, and Svix will minify the payload before sending the webhook (for example, by removing
extraneous whitespace or unnecessarily escaped characters in strings). With
`NewMessageInRaw`, the payload will be sent "as is", without any minification or
other processing. As a consequence, the `payload` does not even need to be valid JSON when using `NewMessageInRaw`.
```

It's a bit wordier, but maybe a bit clearer as well?

svix-jplatte: I think the last sentence is unnecessary given the paragraph afterwards mentions different content types. Together with changing "valid JSON" to just "JSON", I think it's sufficiently clear that non-JSON payloads are supported by this? See new commit.

PR #1540, opened 2024-11-29
python: Add convenient construction of rawPayload messages

svix-jplatte:


svix-lucho (dismissed):

svix-jplatte:
Tested and force-pushed to set a default for `content_type`.

svix-lucho (approved):

PR #1541, opened 2024-11-29
ruby: Add convenient construction of rawPayload messages

svix-jplatte:

### DIFF
--- a/ruby/lib/svix/message_api.rb
+++ b/ruby/lib/svix/message_api.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 module Svix
+  module_function
+
   class MessageAPI
     def initialize(api_client)
       @api = MessageApi.new(api_client)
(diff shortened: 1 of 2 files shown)


ruby/lib/svix/message_api.rb
```diff
@@ -22,4 +22,33 @@ def expunge_content(app_id, msg_id)
...
+  # of string literals, unnecessarily escaped characters in string and such
+  # are fixed up by the server), and is not even required to be JSON.
+  #
+  # `attributes[:payload]` must be a string. An extra attribute `content_type`
+  # is accepted that sets the `content-type` header of the payload, overwriting
+  # the default of `application/json` if specified. Other than that, the
+  # attributes are forwarded [`MessageIn.new`], so all the same ones are
+  # accepted.
+  def message_in_raw(attributes = {})
+    if !attributes.key?(:transformations_params)
```
svix-gabriel, line 38: It's been a minute since I've used ruby, but I think you can do

```ruby
attributes[:transformations_params] ||= {}
```

to avoid the if statement if you want.

svix-gabriel (dismissed):

ruby/lib/svix/message_api.rb
svix-jplatte: Thanks! Looks like it works.

svix-onelson (approved):
> I have no idea how to actually test this..

Seems like you managed just fine 😎

PR #1543, opened 2024-12-02
Use patched OpenAPI spec for Go, too

svix-jplatte:
Closes https://github.com/svix/monorepo-private/issues/9501.

### DIFF
--- a/go/svix_test.go
+++ b/go/svix_test.go
@@ -3,11 +3,12 @@ package svix_test
 import (
 	"context"
 	"errors"
-	svix "github.com/svix/svix-webhooks/go"
 	"net/http"
 	"net/url"
 	"os"
 	"testing"
+
+	svix "github.com/svix/svix-webhooks/go"
 )
 
 // Builds an API client for testing against an arbitrary API server with an arbitrary token.
@@ -96,9 +97,8 @@ func TestKitchenSink(t *testing.T) {
 }
 
 func TestStaticNullableString(t *testing.T) {
-	app := &svix.ApplicationIn{
-		Name: "test",
-		Uid:  svix.StaticNullableString("my-uid"),
+	app := &svix.ApplicationPatch{
+		Uid: svix.StaticNullableString("my-uid"),
 	}
 
 	if !app.Uid.IsSet() {
--- a/go/svix_test.go
+++ b/go/svix_test.go
@@ -2,6 +2,7 @@ package svix_test
 
 import (
 	"context"
+	"encoding/json"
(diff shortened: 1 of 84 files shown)


go/svix_test.go
```diff
@@ -114,3 +115,241 @@ func TestStaticNullableString(t *testing.T) {
...
+	ep_in.Uid = nil
+	ep_in.Metadata = &metadata
+	assertMarshalEq(ep_in, `{"metadata":{},"url":"http://example.local"}`, t)
+
+	ep_patch := svix.EndpointPatch{}
+	assertMarshalEq(ep_patch, `{}`, t)
+
+	// FIXME: setting channels to `null` not currently possible
+	ep_patch.Channels = []string{}
+	assertMarshalEq(ep_patch, `{"channels":[]}`, t)
```
svix-onelson, line 156: This looks like a regression (or an old bug I didn't notice?) to me. This was explicitly the case that caused folks to get 422 responses for endpoint create when they didn't want to change channels or filter types (because we validate that the array is not empty).

I have https://github.com/svix/svix-webhooks/blob/main/go/endpoint_test.go to check `EndpointIn`.

The patch behavior is only verified in the kitchen sink test, and it looks like I only got smart about asserting that PATCH didn't blank out unspecified fields later on in my process for the big generator bump PR.

https://github.com/svix/svix-webhooks/blob/087925540a0250e1b0c8458f0834771eae6c5c48/go/svix_test.go#L76-L95

In other langs, later, I have a flow where the create call sets 2 channels, and the patch adds 2 filter types. Then, I verify both of each are set.

Need to update the kitchen sink test to see if this is a new issue or an old one, but either way this looks wrong to me. We never want to actually send an empty array for PATCH. It's not valid on the backend.

svix-jplatte: I can guarantee you that there is no regression for patch because patch endpoints are not changed at all in this PR (see file list). But I agree that there's a bug here, hence the FIXME.

svix-onelson: Ah, I understand. You're comment here about not being able to set null is to say:

- you can send this body with a `nil` value.
- but it'll get stripped out (read: no regression).
- but you still can't send a `null` to the server to unset the value in the data store 😢 

It is possible to set the field to `nil` explicitly, it just doesn't let us send the `null`.

svix-onelson (approved):

go/svix_test.go
svix-jplatte: Exactly.

Issue #1483, opened 2024-10-11
Adding svix for webhook validation adds almost 1 MB to the JS bundle


mlafeldt:
Hey,

I've recently started using Svix to handle Clerk webhooks. For this, I tried adding the svix JS package to my Astro project to validate webhook signatures. Unfortunately, this ended up *doubling* my project's bundle size from 1 MB to 2 MB. 😱

As a workaround, I copied the following code and added `@stablelib/base64` + `fast-sha256` as dependencies:

https://github.com/svix/svix-webhooks/blob/d0443bda6fa6cccae17b5b3cc5cd420bb6d700e3/javascript/src/index.ts#L803-L939

Surprisingly, this had almost *no* impact on the bundle size.

I'm no expert on ESM or bundling, but there seems to be a problem with the code's structure that prevents proper tree shaking.

This is especially concerning in a constrained environment like Cloudflare Workers/Pages, where bundle size is (even more) important.

(Happy to provide more info if needed.)

mlafeldt:
Looks like I could also use https://github.com/standard-webhooks/standard-webhooks/tree/main/libraries/javascript as a lightweight alternative. However, it expects different non-branded headers...

svix-jplatte:
Some work has already happened on this by a contributor in https://github.com/svix/svix-webhooks/pull/1179.

svix-jplatte:
Unfortunately we've had to revert #1549 in v1.44.0, so re-opening this.

If you're using v1.43.0 without build issues though, you're welcome to stay on that version to enjoy the smaller bundle size while we figure this out.

mlafeldt:
I tried this again today because I was excited to see Clerk adding a `verifyWebhook` helper to their Astro package: https://github.com/clerk/javascript/blob/main/packages/astro/CHANGELOG.md#250

The bundle size situation has improved, but it's still 200 KB larger compared to using standardwebhooks.

tasn:
@mlafeldt, great to hear about the improvement, but annoying that it's so large. @svix-lucho, do you have any idea what's causing this?

tasn:
Bump

Danish-Belal:
is this still open to work ?

T-P-F:
Adding to @mlafeldt's solution, here's a function that I've confirmed is working for Svix (Clerk) webhooks in a deployed Cloudflare worker:

// https://www.npmjs.com/package/standardwebhooks
import { Webhook } from "standardwebhooks";

[...]

Issue #1505, opened 2024-11-07
code: 'ERR_REQUIRE_ESM'


ZONGMENG-kaito:
## Bug Report

I use `svix` in my npm project, but when I use the latest version, I got an error and the error code is `code: 'ERR_REQUIRE_ESM'`
### Version

Latest version

### Platform
Mac os

### Description

I believe is because this dependency's upgrade: https://github.com/svix/svix-webhooks/blob/main/javascript/package.json#L32, upgrade to `2.0.0`, which make the `@stablelib/base64` to be ESM only: https://github.com/StableLib/stablelib/blob/master/packages/base64/package.json#L6

![Image](https://github.com/user-attachments/assets/c86becd7-0c96-4899-ad31-bc2f320a1f36)

@svix-onelson I likely found the upgrade PR was made from you, would you mind downgrading or fixing it? Thanks a lot.

IbrahimMohammed47:
Until they fix it, I found that version `1.37.0` is the latest one using `@stablelib/base64` version 1
So you can use this in you package.json for now `"svix": "~1.37.0"` until we have a fix

ZONGMENG-kaito:
> Until they fix it, I found that version `1.37.0` is the latest one using `@stablelib/base64` version 1 So you can use this in you package.json for now `"svix": "~1.37.0"` until we have a fix or you may want to migrate your project to be `esm`

Thanks for your kinda reply, actually I am using this package: https://www.npmjs.com/package/@privy-io/server-auth which uses `svix` as a dep, so I finally got this error. I will try to modify my project to `esm`.

tasn:
I wonder if we should just downgrade the base64 lib. Thanks for the report!

ZONGMENG-kaito:
> I wonder if we should just downgrade the base64 lib. Thanks for the report!

Thanks, Tom, I strongly recommend that downgrade it to `^1.0.0`, think this way. should be the easiest way to fix it?

tasn:
Yeah, it's probably the right call (as long as there are no security issues or other incompatibilities). I wonder why upstream broke things this way. :|

ZONGMENG-kaito:
> Yeah, it's probably the right call (as long as there are no security issues or other incompatibilities). I wonder why upstream broke things this way. :|

Hope to see a quick rollback version, :D.

svix-onelson:
[...]

PR #1549, opened 2024-12-04
[Lib/JS] Compile svix package for ESM and CommonJS

svix-lucho:
Closes https://github.com/svix/svix-webhooks/issues/1483

Compile the svix JavaScript package in `ESM` and `CommonJS` simultaneously. This should solve the issues with tree-shaking (since it can only be done when the package is compiled with ESM). Also keeps the CommonJS version for backwards compatibility.

Tested this locally and all the usual imports (`import { EventTypeOut } from "svix"`) work as expected, with no changes whatsoever.

### DIFF
--- a/javascript/package.json
+++ b/javascript/package.json
@@ -1,10 +1,9 @@
 {
   "name": "svix",
-  "version": "1.42.0",
+  "version": "1.41.0",
   "description": "Svix webhooks API client and webhook verification library",
   "author": "svix",
   "repository": "https://github.com/svix/svix-libs",
-  "type": "commonjs",
   "keywords": [
     "svix",
     "diahook",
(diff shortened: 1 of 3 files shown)


svix-onelson (approved):
Will this help to surface problems like this?
https://github.com/svix/svix-webhooks/issues/1505

I'm still hoping we can track something in CI to catch this problem the next time it creeps in.

svix-lucho:
> Will this help to surface problems like this? #1505
> 
> I'm still hoping we can track something in CI to catch this problem the next time it creeps in.

I don't think this PR solves errors like that, because there's still a chance `svix` is used in `ESM` and we accidentally import a `commonJS` dependency. But we should be able to write a test and make the test runner interpret the code in both modes. I need to investigate how to do that.

PR #1553, opened 2024-12-06
publish wheels

dimbleby:
- direct invocation of setup.py is [deprecated](https://packaging.python.org/en/latest/discussions/setup-py-deprecated/)
- publish wheels

### DIFF
--- a/.github/workflows/python-release.yml
+++ b/.github/workflows/python-release.yml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   build:
-    name: Build source distribution
+    name: Build distributions
     runs-on: ubuntu-24.04
     environment: release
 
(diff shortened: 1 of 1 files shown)


svix-onelson (approved):

svix-onelson (commented):
Thanks for this!

connorads:
Hey @svix-onelson,

We've just started using `svix` and we're unable to install because we use [`--only-binary :all:`](https://discuss.python.org/t/why-build-wheels-for-pure-python-projects/33262/2).
Do you know when the next release will be with a wheel? Could you consider creating a new version? 🙏 

PS. Thanks for adding wheels @dimbleby

tasn:
@connorads, any idea what we need to do to fix this? Or is it just a matter of making a new release?

connorads:
> @connorads, any idea what we need to do to fix this? Or is it just a matter of making a new release?

I believe it's just a matter of making a new release because this PR means it builds the wheel as well as the sdist.

tasn:
Let me ask the team!

svix-lucho:
@connorads we've just published a new release, so this should be fixed now

dimbleby:
> we've just published a new release

you have not!

so far as I can see none of the github actions that are supposed to be triggered on tagged releases have run, so I think this is not my fault...

connorads:
Thanks @svix-lucho

But it still looks like the Python Release workflow still needs to run
https://github.com/svix/svix-webhooks/actions/workflows/python-release.yml

Still only at 1.44.0 on PyPI
https://pypi.org/project/svix/#history

tasn:
We changed how we do releases, it made the CI tasks not run for whatever reason. I just retriggered them, and we'll fix the root cause tomorrow.

connorads:
Thanks everyone!

PR #1558, opened 2024-12-10
rust: Move API resources into separate files

svix-jplatte:
… to prepare for these being replaced by generated files.


svix-onelson (approved):
:shipit:

PR #1559, opened 2024-12-11
Add optional clap feature

svix-jplatte:
… with some useful initial clap trait implementations.

New type definitions were generated with https://github.com/svix-jplatte/openapi-codegen/commit/fd0c102658511b3a5973b2101c89932c8bfb3e23.


svix-jplatte:
- `background_tasks` is excluded because it isn't well-integrated in our OpenAPI spec, it seems (so the tool doesn't pick it up)
- `message_attempts` is excluded because it's not clear how we'd best handle `MessageStatus` and `StatusCodeClass` in a clap context

svix-onelson (approved):
🤩

PR #1560, opened 2024-12-11
Add query parameter docs and reorder some things

svix-jplatte:
Adjust some field lists and impl block items to match the order in the spec.

This was done by importing the output of our new codegen tool, then reverting changes we don't (yet) want to import.

Version used: https://github.com/svix/openapi-codegen/commit/fd7dbc40af6badfe0dac8b6ccc1b889c2dd62d1f

Based on #1559.

### DIFF
--- a/rust/src/api/endpoint.rs
+++ b/rust/src/api/endpoint.rs
@@ -3,19 +3,31 @@ use crate::{apis::endpoint_api, error::Result, models::*, Configuration};
 
 #[derive(Default)]
 pub struct EndpointListOptions {
-    pub iterator: Option<String>,
+    /// Limit the number of returned items
     pub limit: Option<i32>,
-    pub order: Option<Ordering>,
-}
 
-pub struct Endpoint<'a> {
-    cfg: &'a Configuration,
+    /// The iterator returned from a prior invocation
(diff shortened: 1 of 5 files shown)


rust/src/api/endpoint.rs
```diff
@@ -3,21 +3,33 @@ use crate::{apis::endpoint_api, error::Result, models::*, Configuration};
...
 pub struct EndpointStatsOptions {
+    /// Filter the range to data starting from this date
+    ///
+    /// RFC3339 date string.
     pub since: Option<String>,
+
+    /// Filter the range to data ending by this date
+    ///
+    /// RFC3339 date string.
     pub until: Option<String>,
```
svix-onelson, line 26: There might be a way to tune it on the clap side, but fyi I think the help text only shows whatever is in the first paragraph (per markdown semantics) so having the blank line here means it'll be omitted.

svix-onelson (approved):

rust/src/api/endpoint.rs
svix-jplatte: Well, the CLI is gonna have its own types, so we can customize it however we want there :)

PR #1562, opened 2024-12-12
bridge: Update deno

jaymell:
This appears to fix mysterious transformation failures when run inside of a Docker container.


bridge/svix-bridge/src/runtime/mod.rs
```diff
@@ -119,20 +119,22 @@ fn run_script_inner(
...
     match res {
         Ok(global) => {
             let scope = &mut runtime.handle_scope();
             let local = v8::Local::new(scope, global);
             match serde_v8::from_v8::<JsObject>(scope, local) {
                 Ok(v) => Ok(TransformerOutput::Object(v)),
-                Err(serde_v8::Error::ExpectedObject(msg)) => {
-                    tracing::error!("{msg}");
+                Err(e @ serde_v8::Error::ExpectedObject(_)) => {
+                    tracing::error!("{e}");
```
jaymell, line 131: This error is a lot clearer when the text from the original error is printed.

bridge/svix-bridge/Cargo.toml
```diff
@@ -28,8 +28,8 @@ tracing-opentelemetry = "0.27.0"
 tracing-subscriber = { workspace = true, features = ["fmt", "json"] }
 # N.b. for newer deno versions (like this) the runtimes must be retained and reused since they will leak memory if you
 # create/drop them.
-deno_core = "0.204.0"
-deno_ast = "0.28.0"
+deno_core = "0.308.0"
```
svix-jplatte, line 31: Why this old-ish version? It seems close to the crashy one (though I guess it can't be exactly that wince we have a regression test).

bridge/svix-bridge/src/main.rs
```diff
@@ -315,7 +315,10 @@ async fn main() -> Result<()> {
             "Starting JS Transformation Workers: {}",
             cfg.transformation_worker_count
         );
-        let pooler = runtime::JsPooler::new(cfg.transformation_worker_count);
+
+        deno_core::JsRuntime::init_platform(None, false);
```
svix-onelson, line 319: I remember at one point we had this and then we didn't (like maybe it was invoked by something in deno) and there was something about it causing a segfault if not run _exactly once_.

Just wanted to mention in case there's a better spot in the program to do this. `main()` seems fine to me. Perhaps it could also go inside JsPooler::new() in a once cell or whatever?

svix-onelson (approved):

bridge/svix-bridge/Cargo.toml
jaymell: This matches the version we use downstream. I don't have a lot of confidence in this library, so for now I thought it would be safer to use a known-ok version from our perspective.

svix-jplatte: Oh, so this _is_ exactly the version that was causing us crashes in bridge before. I'm surprised it would work now.

jaymell: Adding the call to `init_platform` is what fixed it.

bridge/svix-bridge/src/main.rs
jaymell: I'm just going to leave it here for now.

PR #1567, opened 2024-12-13
Server: Increase default max pool sizes

jaymell:
Also improve documentation around tuning the parameters.

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,9 +245,9 @@ You can see more in [these instructions](./OpenTelemetry.md).
 
 ### Connection Pool Size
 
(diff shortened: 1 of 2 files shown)


svix-onelson (approved):

notnmeyer:
its just a documentation issue. i think the new default is fine when taken in context of your reference example in the compose file. some advice/suggestions around values suitable for the horizontal scaling situation would be a nice add as well.

or even, a really clear suggestion to use a connection pooler would have been great. in hindsight, i see in the spots in the readme that include pgbouncer and i get it now, but i think other new folks are likely to stumble in similar ways.

tasn:
Thanks for the feedback, we'll improve the docs!

PR #1569, opened 2024-12-16
Add FromStr impls for MessageStatus, StatusCodeClass

svix-jplatte:
Useful for the new Svix CLI written in Rust.


svix-onelson (approved):
🙏🏻

PR #1571, opened 2024-12-17
js: Add generated query parameter types

svix-jplatte:
And get rid of internal ListOptions type.

The ones without docs are not getting generated yet for various reasons, but I think this is a good enough start for replacing the high-level JS API with generated code.

### DIFF
--- a/javascript/src/util.ts
+++ b/javascript/src/util.ts
@@ -1,8 +1,3 @@
 export interface PostOptions {
     idempotencyKey?: string;
 }
-
-export interface ListOptions {
-    iterator?: string | null;
-    limit?: number;
-}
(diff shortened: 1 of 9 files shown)


svix-lucho (approved):

javascript/src/util.ts
```diff
@@ -1,8 +1,3 @@
 export interface PostOptions {
     idempotencyKey?: string;
 }
-
-export interface ListOptions {
-    iterator?: string | null;
-    limit?: number;
-}
```
svix-lucho, line 8: Are we exporting this from the package? If so, we need to document that this will be a breaking change

svix-jplatte: No, it was only exported for use in other modules.

PR #1577, opened 2024-12-19
CLI: rewrite in rust

svix-onelson:
To date, the [Go CLI](https://github.com/svix) has lived in its own repo and has received infrequent updates, failing to track our APIs as they change.

In an effort to improve on this, this diff adds new Rust version of the CLI here in the webhooks repo. The idea is to release the CLI along with the libs as a part of the same process (https://github.com/svix/monorepo-private/issues/9620).

Modules in the source tree under `cli_types/` and `cmds/api/` (excluding the `mod.rs`) are largely produced via our new codegen tooling (h/t @svix-jplatte).

Other modules under `cmds/` are written by hand, the most involved being `listen` which depends on greenfield code under `relay/`.

The relay code was largely inspired by the code seen in the Go CLI, but the overall shape of it has changed to account for GC vs borrow checker. Future work sound be planned to do a design pass now that the pieces are working together since breaking things apart to satisfy rustc left things in an odd shape.

## Differences

Comparing the Go and Rust CLIs, there are some differences. The list of features added in the Rust version that are lacking in Go's is long thanks to the new codegen.

It'll be more useful to focus on what _isn't here_.

### `--data-*` flags

In the Go CLI we exposed flags named `--data-foo` where `foo` would be a field in the expected JSON body sent with the request. These flags have not been carried forward. Instead users will be required to write out the full JSON body as a command-line argument.

Serde will give some feedback if the body fails to parse given the expected schema even before the request is made to the server.

Future work is planned to include an example body for each place we accept JSON in the long help.

### Import/Export

`svix import` and `svix export` have been left behind. For import, users are encouraged to leverage `svix event-type import-openapi` which should fill the same need.

For export, I think the best alternative is using `svix event-type list` and handling the responses as needed.

### Server URL no longer mandatory

The GO CLI required users to set the URL for the API server they wanted to interact with.
In the new Rust CLI, we will default to the issuer of the auth token. Typically this is going to be the safest default since manually specifying a Server URL should only be necessary when doing local development or talking to a non-production-SaaS server. The impacted group for this change would be enterprise customers and Svix employees and the fix is to do what they'd have needed to do all along: set the URL explicitly.

## Verification

Reviewers will note that there's a dramatic lack of unit tests 😇 
To this, I assure you there's technically more coverage on _this CLI_ than there is on the Go one 🤣.

The majority of the verification for this has been manual testing. If you have specific concerns, pitch a test and I can run through the scenario for you (or you can try it yourself!)

## OSS Reviews

There are several new-to-us crates included in this diff. I'll call each out in the comments below, looking for 👍🏻 from the security team.


svix-cli/src/cli_types/message_attempt.rs
```diff
@@ -0,0 +1,250 @@
...
+}
+
+impl From<MessageAttemptListAttemptedDestinationsOptions>
+    for api::MessageAttemptListAttemptedDestinationsOptions
+{
+    fn from(
+        MessageAttemptListAttemptedDestinationsOptions {
+                    limit,iterator,
+                }: MessageAttemptListAttemptedDestinationsOptions,
+    ) -> Self {
```
svix-jplatte, line 247: Does rustfmt fail to format this?

svix-cli/Cargo.lock
```diff
@@ -0,0 +1,2591 @@
...
+checksum = "686f825264d630750a544639377bae737628043f20d38bbc029e8f29ea968a7e"
+dependencies = [
+ "idna_adapter",
+ "smallvec",
+ "utf8_iter",
+]
+
+[[package]]
+name = "idna_adapter"
+version = "1.2.0"
```
svix-jplatte, line 960: Maybe we should pin this the same way we do in svix-server, to reduce dependencies / build times a bit.

svix-cli/src/cli_types/mod.rs
```diff
@@ -0,0 +1,41 @@
+//! Wrapper types around those in the SDK so they can act as args in the `clap` command tree.
+// FIXME: in the new codegen, we can look to add a `cli` feature that conditionally adds these
+//   attributes or trait impls to the wrapped types.
+//   At that point we can use the SDK types directly and remove this module.
```
svix-jplatte, line 4: You want to keep this idea alive?

svix-cli/src/cli_types/mod.rs
```diff
@@ -0,0 +1,41 @@
...
+}
+
+impl From<Ordering> for api::Ordering {
+    fn from(value: Ordering) -> Self {
+        match value {
+            Ordering::Ascending => api::Ordering::Ascending,
+            Ordering::Descending => api::Ordering::Descending,
+        }
+    }
+}
```
svix-jplatte, line 29: Is this still needed with `svix::api::Ordering` implementing `FromStr` now?

svix-cli/src/cmds/open.rs
```diff
@@ -0,0 +1,38 @@
...
+    pub command: OpenCommands,
+}
+
+const API_DOCS_URL: &str = "https://api.svix.com/docs";
+const DOCS_URL: &str = "https://docs.svix.com/";
+
+/// Quickly open Svix pages in your browser
+#[derive(Subcommand)]
+pub enum OpenCommands {
+    /// Open the Svix API documentation
```
svix-jplatte, line 18: Maybe this would make it easier to tell the difference between "api" and "docs"?

```suggestion
    /// Open the Svix API reference
```

svix-cli/src/cmds/signature.rs
```diff
@@ -0,0 +1,67 @@
...
+                secret,
+                signature,
+                msg_id,
+                timestamp,
+                payload,
+            } => {
+                let webhook = svix::webhooks::Webhook::new(&secret)?;
+                let mut headers = http::HeaderMap::with_capacity(3);
+                headers.insert("svix-id", msg_id.parse()?);
+                headers.insert("svix-timestamp", format!("{}", timestamp).parse()?);
```
svix-jplatte, line 60:
```suggestion
                headers.insert("svix-timestamp", timestamp.to_string().parse()?);
```

svix-cli/src/config.rs
```diff
@@ -0,0 +1,86 @@
...
+    server_url: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    debug_url: Option<String>,
+
+    // Relay stuff relates to the `listen` command.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub relay_token: Option<String>,
+    // FIXME: "url" isn't right. We expect a hostname, default is: `api.play.svix.com`
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub relay_debug_url: Option<String>,
```
svix-jplatte, line 25: Rename the field and use `#[serde(alias)]` or `#[serde(rename)]` to keep backwards compat?

svix-cli/src/main.rs
```diff
@@ -0,0 +1,148 @@
...
+struct Cli {
+    #[command(flatten)]
+    color: Color,
+    #[command(subcommand)]
+    command: RootCommands,
+}
+
+impl Cli {
+    /// Converts the selected `ColorChoice` from the CLI to a `ColorMode` as used by the JSON printer.
+    /// When the color choice is "auto", this considers whether stdout is a tty or not so that
```
svix-jplatte, line 39:
```suggestion
    /// Converts the selected `ColorChoice` from the CLI to a `ColorMode` as used by the JSON printer.
    ///
    /// When the color choice is "auto", this considers whether stdout is a tty or not so that
```

svix-cli/src/relay/mod.rs
```diff
@@ -0,0 +1,467 @@
...
+                r#"
+Webhook Relay is now listening at:
+{}
+
+All requests on this endpoint will be forwarded to your local URL:
+{}
+"#,
+                receive_url(&start_response.token),
+                self.local_url,
+            );
```
svix-jplatte, line 163: Could use [`printdoc!`](https://docs.rs/indoc/latest/indoc/macro.printdoc.html) here and below.

svix-cli/src/relay/mod.rs
```diff
@@ -0,0 +1,467 @@
...
+struct WsConnection {
+    stream: S,
+}
+
+impl WsConnection {
+    async fn new(websocket_url: &url::Url) -> Result<Self> {
+        let request = websocket_url.to_string().into_client_request()?;
+        let (stream, _resp) = connect_async(request)
+            .await
+            .with_context(|| "failed to connect to websocket server")?;
```
svix-jplatte, line 305:
```suggestion
            .context("failed to connect to websocket server")?;
```

svix-cli/src/relay/mod.rs
```diff
@@ -0,0 +1,467 @@
...
+    mut tx: SplitSink<S, Message>,
+) -> Result<()> {
+    while let Some(msg) = rx.recv().await {
+        tokio::time::timeout(
+            WRITE_WAIT,
+            tx.send(Message::Binary(
+                serde_json::to_vec(&msg)
+                    .expect("trivial serialization")
+                    .into(),
+            )),
```
svix-jplatte, line 368: Why not `Message::Text` / `serde_json::to_string`?

svix-cli/src/relay/mod.rs
```diff
@@ -0,0 +1,467 @@
...
+        .send()
+        .await?)
+}
+
+fn format_resp_headers(headers: &HeaderMap) -> Result<HashMap<String, String>> {
+    let mut out = HashMap::new();
+    const SKIP_USER_AGENT: HeaderValue = HeaderValue::from_static("Go-http-client/1.1");
+    for (k, v) in headers {
+        // FIXME: the Go cli used to unset the useragent if it matched the Go http client. Do we need something similar here? idk why this is needed.
+        if k == http::header::USER_AGENT && v == SKIP_USER_AGENT {
```
svix-jplatte, line 409: Haha, I hope we don't actually need this?

svix-cli/src/relay/mod.rs
```diff
@@ -0,0 +1,467 @@
...
+            continue;
+        }
+        out.insert(k.to_string(), v.to_str()?.to_string());
+    }
+    Ok(out)
+}
+
+//
```
svix-jplatte, line 417: Empty comment?

svix-cli/src/relay/mod.rs
svix-onelson: It's not clear to me what problem we were aiming to solve by removing this particular value in the first place. I'll remove it (since it clearly doesn't make sense today). If/when whatever problem shows up again, we can re-solve it 🤣

svix-cli/Cargo.toml
```diff
@@ -0,0 +1,32 @@
+[package]
+name = "svix-cli"
+description = "A CLI to interact with the Svix API."
+version = "0.1.0"
```
tasn, line 4: Sync the version with the rost of the libs, and please also update the version bump script to update Cargo.toml and cargo.lock like it does for the server. 🙏🏻

svix-cli/src/cli_types/mod.rs
svix-onelson: I think what we're doing today is fine, but I do worry a bit about keeping the two sets of templates aligned.

Maybe we can branch in the template rather than doing a cargo feature (which was raised as objectionable)? I'll remove the FIXME regardless. I think we are aware of the situation by now.

svix-cli/src/cli_types/mod.rs
svix-onelson: Probably not. I'll cut it.

svix-cli/Cargo.lock
svix-jplatte: This is what I'm talking about: https://github.com/svix/svix-webhooks/blob/7c188155e1501aa78cb28383e9fc24308282cfb1/server/svix-server/Cargo.toml#L22-L24

svix-cli/src/cli_types/mod.rs
svix-jplatte: Right. Yeah we can use macros or includes in the templates to share code between CLI and lib codegen.

svix-cli/src/config.rs
svix-onelson: `relay_hostname`? Sure. Why not fix it now, eh?

svix-cli/src/relay/mod.rs
tasn: What's the problem with binary?

svix-cli/Cargo.toml
svix-onelson: Sure. I was going to work that as a part of the releasing/cargo dist issue but I can work to get this in here instead.

svix-cli/src/relay/mod.rs
svix-jplatte: This code can only send UTF-8 because `serde_json` only produces UTF8, that's just clearer when using those. Pretty sure that everything is going to be the same on the wire.

svix-cli/Cargo.toml
tasn: It's fine then, do it as part of the relesae.

svix-onelson: Added an entry to the script for this. Looks like it should set the version the next time. I manually set it to `1.44.0` in the meantime.

svix-cli/src/relay/mod.rs
svix-onelson: Yeah, I think it's the same either way in this specific case.

It's a one-time per connection thing so there's no sense in trying to optimize (though that's not really what I was going for).

svix-onelson:
@svix-jplatte you mentioned in a separate conversation about wanting to tidy up imports (perhaps WRT codegen).

I wonder if you'd like to swoop in with a follow-up PR to get the rustfmt config setup how you like it. This still needs lints/CI etc. I can take a stab at it before merging this if not.

.github/workflows/cli-lint.yml
```diff
@@ -0,0 +1,77 @@
+name: CLI Lint
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'svix-cli/**'
```
svix-jplatte, line 8: What about Rust lib changes? Since we're using a path dependency now, I think it would be good to test CLI on Rust lib changes.

svix-onelson:
# OSS Review

## `tokio-tungstenite`

https://crates.io/crates/tokio-tungstenite/0.26.1

This is the websocket client used for the `listen` command.
Seems well maintained.

They have remarks in their readme about performance, specifically that some other libs might be better for cases where perf is mission critical. For this use case I think we're okay. Nobody should be running production systems tunneled through our relay server.


## `dialoguer`

https://crates.io/crates/dialoguer/0.11.0

This is used for the `login` command which prompts for user input to store in the user's config file.

Hasn't been updated in a while but there is a long list of contributors/maintainers on the github page. One piece of a larger "console" project.

## `colored_json`

https://crates.io/crates/colored_json/5.0.0

Used to optionally/conditionally color the JSON output from various API commands.

**This one comes with a drawback.** The license is `EPL-2.0` which was flagged by our security lints, requiring an exception. This is a slightly more restrictive license than we've seen in the past. Seems like it should be fine for the CLI since we're entirely in the open for it, but would need to make sure we don't introduce it as a dep to areas where we have proprietary builds happening.

I looked at 2 alternatives:
- https://github.com/saghm/json-color
- https://github.com/andrewhickman/termcolor-json

While both of these seemed able to give similar output, configurable colors, etc, each of these had issues with color mode detection and overriding.

We should revisit this at a later date. For now I added an exception to the `deny.toml` for this lib.


## `open`

https://crates.io/crates/open/5.3.1

Cross-platform "open this url in your default browser."

At first I thought to write this myself but since the Go CLI used a lib for this task, I elected to delegate.

The Windows one looked a bit fiddly when I peeked at the source, so I'm glad I did.

Project looks healthy. Regular updates and several maintainers.


## `home`

https://crates.io/crates/home/0.5.11

`home_dir()` in the stdlib currently has a deprecation warning since it can give the wrong answer for Windows systems.

An alternative version of the stdlib's `home_dir()` as used by `cargo` to better support Windows.  This is maintained by the cargo team.

In Rust 1.86 this will no longer be needed.

tasn:
Looks good to me. Thanks!

.github/workflows/cli-lint.yml
svix-onelson: Sold

PR #1584, opened 2024-12-20
rust: Regenerate high-level API

svix-jplatte:
Reverting parts of the changes to avoid unnecessary breakage.

### DIFF
--- a/rust/src/api/event_type.rs
+++ b/rust/src/api/event_type.rs
@@ -9,12 +9,15 @@ pub struct EventTypeListOptions {
     /// The iterator returned from a prior invocation
     pub iterator: Option<String>,
 
+    /// The sorting order of the returned items
+    pub order: Option<Ordering>,
+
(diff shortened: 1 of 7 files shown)


svix-onelson (approved):

rust/src/api/event_type.rs
```diff
@@ -101,28 +126,12 @@ impl<'a> EventType<'a> {
...
-        .await
-    }
-
     /// Archive an event type.
     ///
     /// Endpoints already configured to filter on an event type will continue to
     /// do so after archival. However, new messages can not be sent with it
     /// and endpoints can not filter on it. An event type can be unarchived
-    /// with the create operation.
+    /// with the [create operation][Self::create].
```
svix-onelson, line 134: This one is a problem for the CLI. Need to surface the params so we can use `--expunge`. This is supported in the Go CLI currently.

PR #1585, opened 2024-12-20
js: Regenerate high-level API

svix-jplatte:
Rust-style doc comments don't seem to work here :smile: 

Adjusted the codegen to output Java-style doc comments, which get picked up by VSCode for on-hover info, so that's a good sign.


svix-jplatte:
Ah, I missed the stats change. Will get back to this on Mon.

tasn (approved):

PR #1587, opened 2024-12-23
kotlin: Replace most query param classes with generated ones

svix-jplatte:
… and clean up the remaining ones to prepare for future replacement.

Also leave the default for the `limit` on pagination APIs up to the server. I don't think we need to mention this in the changelog since so far the default has been the same as on the server side; however if we were to change the limit for some API on the server side, it makes much more sense to me that the new default would apply immediately and for everyone who doesn't specify it, rather than requiring a libs update for Kotlin users.

Part of https://github.com/svix/monorepo-private/issues/9576.


tasn:
Yeah, no need to mention that. It's an implementation detail that was never public, and anyhow we won't change the server value anyway.

svix-james (approved):

PR #1589, opened 2024-12-24
Rust: derive Clone and Debug on Webhook struct

Samyak2:
## Motivation

- Since the Webhook struct is quite simple, there is no reason not to have Clone implemented on it
- My use case for Clone was to use Webhook as a `Data` in poem: https://docs.rs/poem/latest/poem/struct.Request.html#method.set_data

## Solution

- derive Clone on Webhook
- Also derive Debug because why not?

### DIFF
--- a/rust/src/webhooks.rs
+++ b/rust/src/webhooks.rs
@@ -30,6 +30,7 @@ pub enum WebhookError {
     InvalidPayload,
 }
 
+#[derive(Debug, Clone)]
 pub struct Webhook {
     key: Vec<u8>,
 }


Samyak2:
Actually, I see one reason for not implementing Debug. The struct stores the webhook secret key in a Vec. So the debug representation would expose this key when printed.

tasn:
Yeah, debug will expose that, and also clone means we'll be copying this value in memory a lot.
We can probably:
1. Debug: Implement a custom debug implementation that doesn't print the key.
2. Clone: implement Arc or something internally so that the clone doesn't close key material.

Happy to review such a PR if you think that would be helpful for you.

Samyak2:
Alright, that makes sense. I will open a separate PR if Clone is really needed. For now, I simply wrapped Webhook in an Arc to make it `Clone`. `Arc<Webhook>` works with Poem's `Data`.

Thank you for the response!

tasn:
With pleasure, and yeah, that feels like a better solution!

PR #1590, opened 2024-12-26
Libs(Rust, CLI): switch from `native-tls` to `rustls`

svix-onelson:
Using rustls simplifies some of the build/distribution concerns when packaging the CLI. Namely it allows us to produce binaries without having to ensure `openssl-sys` can build, and all that that involves.

Oddly, while `native-tls` is listed as an optional-but-default dep in the Rust SDK, specifying the dep on the SDK from the CLI with `default-features = false` didn't seem to remove `openssl-sys` from the dependency tree.

As a work around, I've updated the default features in the Rust SDK to target `rustls`, which does allow the sys crate to drop away.

This seems fine to me. Folks who still want to use `native-tls` can elect to do so, just not by default.

Some CLI features have been tuned up to support this effort, namely targeting specific individual features for `tokio`, and so on.

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -12,7 +12,10 @@ keywords = ["svix", "webhooks", "diahook"]
 categories = ["development-tools", "asynchronous", "network-programming", "web-programming"]
 
 [features]
-default = ["http1", "native-tls"]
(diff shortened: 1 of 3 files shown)


tasn:
TBH, I think rusttls is a better default for the rust sdk anyway...

tasn (approved):

PR #1593, opened 2024-12-26
CLI: build with `dist`

svix-onelson:
Adds workflows and supporting configs to use [dist](https://github.com/axodotdev/cargo-dist) for building and publishing binaries for the CLI.

To lean towards dist's happy path, a top-level `Cargo.toml` defines a workspace including any cargo projects we want to build and distribute (currently just `svix-cli`).

The `dist-workspace.toml` includes various parameters set by `dist init` to configure the generation of `.github/workflows/release.yml`.

`dist-workspace.toml` can be edited manually, but the `release.yml` should only be changed via `dist generate`.

## Installers

Currently, the `shell`, `powershell` and `msi` installers are configured.

`dist` also provides:

- `npm`
- `homebrew` (we'll definitely enable this one once we know everything looks good)

The current selection are essentially "curl to sh" (and the Windows equivalent) like is commonly done for `rustup` and seemed like the closest thing we'd see to a "hello world." My plan is to enable more in subsequent releases.

The `msi` installer includes a full installer wizard type deal.

## Releasing

These notes are for the folks doing releases once this lands. The gist is:

- run the version bump script as normal
- push the tag for the release
- create a **draft** Release for the tag
- manually run the `Release` workflow (which is managed by `dist`) specifying the tag.

image

The `Release` workflow uses the tag to find the draft release so it can publish its build artifacts there.

The last thing this workflow does is mark the Release as published, so be sure to update the release notes and so on before running this workflow.


dist-workspace.toml
```diff
@@ -0,0 +1,22 @@
...
+# Target platforms to build apps for (Rust target-triple syntax)
+targets = ["aarch64-apple-darwin", "aarch64-unknown-linux-gnu", "x86_64-apple-darwin", "x86_64-unknown-linux-musl", "x86_64-pc-windows-msvc"]
+# Whether dist should create a Github Release or use an existing draft
+create-release = false
+# Path that installers should place binaries in
+install-path = "~/.svix/bin"
+# Whether to install an updater program
+install-updater = true
+# Whether CI should trigger releases with dispatches instead of tag pushes
+dispatch-releases = true
```
svix-onelson, line 21: If we later decide we want to automatically build the cli on tag push, this is what we'd tweak. I elected to use a manual trigger to start so we have a bit more control over it.

svix-cli/Cargo.toml
```diff
@@ -1,12 +1,24 @@
...
 description = "A CLI to interact with the Svix API."
+homepage = "https://www.svix.com"
 authors = ["Svix Inc. <oss@svix.com>"]
 version = "1.44.0"
 edition = "2021"
 license = "MIT"
 keywords = ["svix", "webhooks", "diahook"]
 categories = ["development-tools", "asynchronous", "network-programming", "web-programming"]
+repository = "https://github.com/svix/svix-webhooks"
+publish = false
```
tasn, line 12: I think we do want to publish to crates.io?

.github/workflows/release.yml
```diff
@@ -0,0 +1,293 @@
+# This file was autogenerated by dist: https://opensource.axo.dev/cargo-dist/
```
svix-onelson, line 1: When the `dist-workspace.toml` changes, either via manual edits or by `dist init`, running `dist generate` will rewrite this workflow.

There are CI checks in place to catch uncommitted diffs, like we do for our openapi spec, etc.

svix-cli/Cargo.toml
svix-onelson: Yeah, I think we can (and should) but I wanted to do the publishes that involve external services as a 2nd step. Let's turn this one on when enabling homebrew 🙏🏻

Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[workspace]
```
tasn, line 1: Do we have to litter the toplevel with this? I'd much rather have everything be contained under svix-cli.

svix-onelson: I wasn't able to figure out how to make it work without this. I'm sure it's possible, somehow.

tasn: No worries then.

svix-onelson: I can ask around for clues (this wasn't a case described in the docs re: advanced layout/config) but we might end up being able to add bridge and so on into this workspace.

Could end up being an advantage to follow the happy path.

tasn: yeah, let's not touch it. You're right.

.github/workflows/release.yml
```diff
@@ -0,0 +1,293 @@
...
+#
+# If you push multiple tags at once, separate instances of this workflow will
+# spin up, creating an independent announcement for each one. However, GitHub
+# will hard limit this to 3 tags per commit, as it will assume more tags is a
+# mistake.
+#
+# If there's a prerelease-style suffix to the version, then the release(s)
+# will be marked as a prerelease.
+on:
+  pull_request:
```
tasn, line 42: Do we want it to run on PR?

.github/workflows/release.yml
```diff
@@ -0,0 +1,293 @@
...
+  workflow_dispatch:
```
tasn, line 43: How do we run this / how does it interact with releases?

svix-onelson: See the "Releasing" section of the PR description.

.github/workflows/release.yml
svix-onelson: Yes. The default behavior for this is to run `dist plan` which verifies the `dist-workspace.toml` configuration. No binaries are produced.

tasn (approved):

svix-cli/wix/main.wxs
```diff
@@ -0,0 +1,228 @@
+<?xml version='1.0' encoding='windows-1252'?>
```
tasn, line 1: What's this file btw?

svix-onelson: Generated to support the msi installer.

PR #1596, opened 2024-12-27
Docker: fix exposed port to be 8071.

tasn:
I don't think exposing 8080 ever did anything as the server is listening on 8071. It looks like a customer has been facing issues with their docker setup not letting them map the port because it doesn't match the exposed from the dockerfile. I'm not sure how it works for everyone else but not here though anyhow it should be fixed.

Ref: https://x.com/0xblacklight/status/1872430388158533871


svix-onelson (approved):
👍🏻

K-Mistele:
Thanks for the quick fix, guys!

PR #1600, opened 2024-12-31
CLI: remove `--no-logging` flag

svix-onelson:
The `listen` command at one point supported a mode where it would skip logging requests in Play's UI.

Today, this doesn't seem to be supported (Play rejects the request because it says the token is invalid).

Rather than work to repair this in Play, we're just going to remove the flag.


svix-onelson:
This one is ready for review, but sadly this is the 3rd PR in a chain. Will likely need another ✅ after the others have landed.

tasn (dismissed):

tasn (approved):

PR #1606, opened 2025-01-03
CLI: update generated code

svix-onelson:
~~During a previous codegen, functionality related to `logout` was erased. This command is meant to accept an arbitrary dashboard token and expire it, not the token configured for the client itself for regular API calls.~~ 

> The logout issue is a #wontfix. We're not carrying the Go CLI behavior forward in this case.

This diff also applies some pending changes made in the code generator templates, designed with the aim of reducing friction during codegen and as well as the number of hand-edits required to get the project building post-codegen.

- adding optional imports (which clippy can trivially remove as needed).
- combining the "types" and "commands" into a single file.
- special casing `before` and `after` so we don't have to go in and convert them to strings after the codegen runs.

Some hand-edits are still required to remove things that are not exposed in the SDK, but this should reduce some friction at least.


tasn (changes requested):
That's incorrect. Logout should expire the current token, it doesn't accept a special token parameter. We could potentially decide to allow passing a custom token, but this hasn't been the behavior so far and we don't need to change this.

svix-onelson:
That's fine but that's not how the Go CLI was written. If we want to just expire the configured token, that's easier for us.

svix-cli/src/cmds/api/message_attempt.rs
```diff
@@ -1,10 +1,227 @@
+use chrono::{DateTime, Utc};
 use clap::{Args, Subcommand};
-use colored_json::ColorMode;
+use svix::api::{self};
```
svix-jplatte, line 3:
```suggestion
use svix::api;
```

svix-onelson: I'm doing another round of template changes that should fix this, but I was going to apply that in a follow-up. Can it wait?

svix-jplatte: Sure.

PR #1609, opened 2025-01-03
Bump version to 1.45.1 and update changelog

svix-jplatte:
## What's Changed

* CLI: Rename the binary back from `svix-cli` to `svix`
* Libs/Java: Stop sharing one `ApiClient` between all `Svix` instances. Fixes a bug where multiple
  independently instantiated `Svix` instances would always use the same base path and auth token.


ChangeLog.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 
-## Unreleased
+## Version 1.45.1
+* CLI: Rename the binary from `svix-cli` to `svix`
```
tasn, line 4: Rename *back*, was always supposed to be svix

tasn (approved):

PR #1610, opened 2025-01-03
rust: Clean up internal request module

svix-jplatte:
Found some obvious improvements while getting refreshed on this to replace more of the current generated code.

### DIFF
--- a/rust/src/request.rs
+++ b/rust/src/request.rs
@@ -19,28 +19,23 @@ pub(crate) enum Auth {
 /// detected based on the configuration. This functionality is useful when the
 /// OpenAPI definition does not include an authorization scheme.
 pub(crate) struct Request {
-    auth: Option<Auth>,
     method: http1::Method,
     path: String,
     query_params: HashMap<String, String>,
     no_return_type: bool,
     path_params: HashMap<String, String>,
-    form_params: HashMap<String, String>,
     header_params: HashMap<String, String>,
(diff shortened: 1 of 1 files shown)


rust/src/request.rs
```diff
@@ -57,35 +53,35 @@ impl Request {
...
+    pub async fn execute<T: DeserializeOwned>(self, conf: &Configuration) -> Result<T, Error> {
+        match self.execute_inner(conf).await? {
+            // This is a hack; if there's no_ret_type, T is (), but serde_json gives an
+            // error when deserializing "" into (), so deserialize 'null' into it
+            // instead.
+            // An alternate option would be to require T: Default, and then return
+            // T::default() here instead since () implements that, but then we'd
+            // need to impl default for all models.
+            None => Ok(serde_json::from_str("null").expect("serde null value")),
+            Some(bytes) => Ok(serde_json::from_slice(&bytes).map_err(Error::generic)?),
```
svix-onelson, line 80: It's a shame this is in place, WRT CLI.

If we surfaced an `Option<T>` we could then `if let` to decide if the json printer should be used. Could be nice quality of life improvement if so.

svix-onelson (approved):

rust/src/request.rs
svix-jplatte: Yeah, this is just internal code. I'm sure the CLI stuff isn't actually going to be hard to handle, without any libs changes.

PR #1614, opened 2025-01-03
Libs: change release workflows to be triggered by release published

svix-onelson:
We had an issue where pushing tags named for version numbers caused unwanted side-effects.

Attempting to avoid this:
- no longer manually push tags.
- let the GitHub release automatically tag when it publishes.
- drive the other release workflows based on release: published events.

By moving the trigger off of tag push, this additionally opens the door for situations where we need to push a tag then create a release for it since the lib workflows won't be triggered until the release is finalized.

### DIFF
--- a/.github/workflows/csharp-release.yml
+++ b/.github/workflows/csharp-release.yml
@@ -1,9 +1,8 @@
 name: C# Release
 
 on:
-  push:
-    tags:
-      - 'v*'
+  release:
+    types: [published]
 
 jobs:
   dotnet:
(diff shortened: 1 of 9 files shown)


.github/workflows/csharp-release.yml
```diff
@@ -1,9 +1,8 @@
 name: C# Release
```
tasn, line 1: Missing bridge and server. (and I think that's it?)

svix-onelson: They are already configured to be exactly this 😎

tasn: Oh, so we even know it works! :P

tasn (approved):

PR #1615, opened 2025-01-04
CI: release libs, server, bridge with manual workflows

svix-onelson:
Changes all workflows related to Releases so they are manually triggered.

Continuing to iron out the kinks of the new `dist` based workflow used for building/releasing the CLI.

It looks as though the "release published" event we tried to trigger these on doesn't fire in some cases. Rumors suggest having a release published by an automation rather than a real person could be the difference.

For now as a workaround, let's trigger these manually. It's tedious, but we can likely add One More Workflow to kick them all off on our behalf as a 2nd step.

Additional: a new workflow called "Mega Releaser" aims to try and kick off all the rest. It might work. It might not. TBD.

### DIFF
--- a/.github/workflows/php-release.yml
+++ b/.github/workflows/php-release.yml
@@ -1,8 +1,12 @@
 name: PHP Release
 
 on:
-  release:
-    types: [published]
+  workflow_dispatch:
+  workflow_run:
+    workflows:
+      - Mega Releaser
+    types:
+      - completed
 
 jobs:
   packagist:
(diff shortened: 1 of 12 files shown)


tasn (approved):

.github/workflows/php-release.yml
```diff
@@ -1,8 +1,12 @@
 name: PHP Release
 
 on:
-  release:
-    types: [published]
+  workflow_dispatch:
```
tasn, line 4: Shouldn't we remove the dispatches?

PR #1616, opened 2025-01-05
csharp: improve performance and reduce memory footprint when creating and verifying webhook signatures

esskar:
Using latest csharp and dotnet features to increase performance and reduce memory footprint when creating and verifying webhook signatures

## Motivation

When using the webhook verification process we see lots of memory wasted and garbage collector kicking when our applications gets hit with a lot of webhook requests.

## Solution

* using ReadOnly<char> instead of string in Webhook.Verify and Webhook.Sign
* using stack alloctions instead of heap allocations when creating and verifing signatures as much as possible
* using new Base64 API to convert to base64 operating on spans

## Benchmarks

```csharp
[MemoryDiagnoser]
public class Tests
{
    private readonly string _msgId = "msg_p5jXN8AQM9LWM0D4loKWxJek";
    private readonly DateTimeOffset _timestamp = DateTimeOffset.FromUnixTimeSeconds(1614265330);
    private readonly string _payload = "{\"test\": 2432232314}";
    private readonly string _largePayload = Enumerable.Range(0, 1000).Select(i => i % 10).Aggregate(new StringBuilder(), (sb, i) => sb.Append(i), sb => sb.ToString());
    private readonly Webhook _webhook = new Webhook("whsec_MfKQ9r8GKYqrTwjUPD8ILPZIo2LaLaSw");
    private readonly WebHeaderCollection _headers;
    private readonly WebHeaderCollection _largeHeaders;
    
    public Tests()
    {
        var now = DateTimeOffset.UtcNow;
        _headers = new WebHeaderCollection
        {
            {"svix-id", _msgId},
            {"svix-timestamp", now.ToUnixTimeSeconds().ToString() },
            {"svix-signature", _webhook.Sign(_msgId, now, _payload)}
        };
        _largeHeaders = new WebHeaderCollection
        {
            {"svix-id", _msgId},
            {"svix-timestamp", now.ToUnixTimeSeconds().ToString() },
            {"svix-signature", _webhook.Sign(_msgId, now, _largePayload)}
        };
    }
    
    [Benchmark]
    public string? Sign()
    {
        return _webhook.Sign(_msgId, _timestamp, _payload);
    }
    
    [Benchmark]
    public string? SignLargePayload()
    {
        return _webhook.Sign(_msgId, _timestamp, _largePayload);
    }
    
    [Benchmark]
    public void Verify()
    {
        _webhook.Verify(_payload, _headers);
    }
    
    [Benchmark]
    public void VerifyLargePayload()
    {
        _webhook.Verify(_largePayload, _largeHeaders);
    }
}
```

Original Code

```
| Method             | Mean       | Error    | StdDev   | Gen0   | Allocated |
|------------------- |-----------:|---------:|---------:|-------:|----------:|
| Sign               |   945.0 ns |  3.75 ns |  3.13 ns | 0.1469 |     928 B |
| SignLargePayload   | 1,568.2 ns | 15.73 ns | 13.94 ns | 0.6142 |    3864 B |
| Verify             | 1,158.8 ns |  4.16 ns |  3.69 ns | 0.2117 |    1328 B |
| VerifyLargePayload | 1,772.6 ns |  5.76 ns |  4.49 ns | 0.6790 |    4264 B |
```

Code from this PR

```
| Method             | Mean       | Error   | StdDev  | Gen0   | Allocated |
|------------------- |-----------:|--------:|--------:|-------:|----------:|
| Sign               |   466.2 ns | 0.68 ns | 0.57 ns | 0.0267 |     168 B |
| SignLargePayload   |   885.4 ns | 3.20 ns | 2.68 ns | 0.0267 |     168 B |
| Verify             |   705.6 ns | 2.24 ns | 1.98 ns | 0.0095 |      64 B |
| VerifyLargePayload | 1,138.2 ns | 3.26 ns | 2.72 ns | 0.0095 |      64 B |
```


csharp/Svix/Utils.cs
```diff
@@ -29,7 +25,7 @@ public static bool SecureCompare(string a, string b)
             var result = 0;
             for (var i = 0; i < a.Length; i++)
             {
-                result |= a[i] ^ b[i];
+                result |= a[i].CompareTo(b[i]);
```
tasn, line 28: Is this constant time comparison? I guess it's a char so it probably is, but it's still risky as it could get optimized out. I don't think I'm comfortable with this change.

csharp/Svix/Webhook.cs
```diff
@@ -35,53 +41,65 @@ public Webhook(byte[] key)
             this.key = key;
         }
 
-        public void Verify(string payload, WebHeaderCollection headers)
+        public void Verify(ReadOnlySpan<char> payload, WebHeaderCollection headers)
+            => Verify(payload, headers.Get);
+        
+        public void Verify(ReadOnlySpan<char> payload, Func<string, string> headerProvider)
```
tasn, line 47: This breaks the public API, no? We probably need a string variant?

csharp/Svix/Webhook.cs
```diff
@@ -95,26 +113,67 @@ private static DateTimeOffset VerifyTimestamp(string timestampHeader)
...
+        private void CalculateSignature(ReadOnlySpan<byte> input, Span<byte> output)
         {
-            var toSign = $"{msgId}.{timestamp.ToUnixTimeSeconds().ToString()}.{payload}";
-            var toSignBytes = SafeUTF8Encoding.GetBytes(toSign);
             using (var hmac = new HMACSHA256(this.key))
             {
-                var hash = hmac.ComputeHash(toSignBytes);
-                var signature = Convert.ToBase64String(hash);
-                return $"v1,{signature}";
+                if (!hmac.TryComputeHash(input, output, out _))
```
tasn, line 175: Am I missing something, or is it not returning the same value? (missing the `v1,`)

csharp/Svix/Svix.csproj
```diff
@@ -7,6 +7,7 @@
     <Authors>Svix</Authors>
     <Company>Svix</Company>
     <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
+    <LangVersion>13</LangVersion>
```
tasn, line 10: This was just released, wasn't it? If so, this is way too new for us, we want to support versions that are at least 5 years old.

tasn (changes requested):
Hey, thanks for the contribution, and sorry to hear the current code was thrashing the GC.

I made some comments about specific changes (not all of them yet), though would you mind opening smaller PRs that are more focused on just one change so that things are easier to review?

At the moment there are 4+ completely separate changes here which make it very hard to: (1) understand the reasoning behind each change / need, and (2) understanding what's going on so that we can review it effectively.

Thanks a lot!

esskar:
> Hey, thanks for the contribution, and sorry to hear the current code was thrashing the GC.
> 
> I made some comments about specific changes (not all of them yet), though would you mind opening smaller PRs that are more focused on just one change so that things are easier to review?
> 
> At the moment there are 4+ completely separate changes here which make it very hard to: (1) understand the reasoning behind each change / need, and (2) understanding what's going on so that we can review it effectively.
> 
> Thanks a lot!

Will do. I will also make some comments to your questions here.

csharp/Svix/Svix.csproj
esskar: Your still supporting .net8.0 runtime; using C# 13 is just for your internal code; what your comsumers are using is not effected by this change

csharp/Svix/Webhook.cs
esskar: The signature value returned ist still having the `v1,` prefix, just for the internal usage the `v1,` is not prefixed; 
https://github.com/svix/svix-webhooks/pull/1616/files/f703640506ef1ec280a8a532a786cc99694208f8#diff-f5180021f8852f93bec73e950dfe608d4c4e35d73b12e0f6beafeca809ae6785R130

I noticed that during verification, you were generating the `v1,` prefix just to throw it away again.

csharp/Svix/Webhook.cs
esskar: There is an implicit conversion, but I provide it.

csharp/Svix/Utils.cs
esskar: reverted

csharp/Svix/Webhook.cs
esskar: I just checked: when the string overload functions are provided, dotnet will call them when string is passed, which would remove the benefit; when not providing them, dotnet will use implizit conversion, so consumers would benefit automatically without the need of changing there code

csharp/Svix/Webhook.cs
tasn: ah, thanks!

csharp/Svix/Svix.csproj
tasn: Gotcha, thanks!

csharp/Svix/Webhook.cs
tasn: OKie, thanks!

tasn:
Can you please rebase over `main`? I merged your other PR, and I'd love to see this PR without those changes for easier review.

csharp/Svix/Utils.cs
```diff
@@ -5,22 +5,18 @@ namespace Svix
 {
     internal static class Utils
     {
-
+        public static bool SecureCompare(string a, string b)
```
tasn, line 8: That's the one I was talking about where you also had the `string` overload, so are these not needed?

esskar: Removed the overload, it was needed before, as the generic version of this function made the overload needed for backwards compatibility; as this function is no longer generic, the overload is not needed.

csharp/Svix/Webhook.cs
```diff
@@ -35,63 +41,84 @@ public Webhook(byte[] key)
...
         {
-            ArgumentNullException.ThrowIfNull(headers);
+            if (payload == null)
+            {
+                throw new ArgumentNullException(nameof(payload));
+            }
+            if (headers == null)
+            {
+                throw new ArgumentNullException(nameof(headers));
+            }
```
tasn, line 53: This looks like you reverted your other changes by accident?

csharp/Svix/Webhook.cs
```diff
@@ -35,63 +41,84 @@ public Webhook(byte[] key)
...
             
             Verify(payload, headers.Get);
         }
 
-        public void Verify(string payload, Func<string, string> headersProvider)
+        public void Verify(ReadOnlySpan<char> payload, Func<string, string> headersProvider)
         {
-            ArgumentNullException.ThrowIfNull(payload);
-            ArgumentNullException.ThrowIfNull(headersProvider);
+            if (payload == null)
```
tasn, line 60: Same here? Accidentally reverted other changes?

csharp/Svix/Webhook.cs
esskar: ArgumentNullException.ThrowIfNull cannot be called on value stucts; thats why i use the "old" api.

csharp/Svix/Webhook.cs
esskar: See comments above

csharp/Svix/Webhook.cs
tasn: Ah, thanks.

csharp/Svix/Webhook.cs
```diff
@@ -105,26 +132,67 @@ private static DateTimeOffset VerifyTimestamp(string timestampHeader)
...
+            ReadOnlySpan<char> payload, 
+            Span<char> signature,
+            out int charsWritten)
+        {
+            // Estimate buffer size and use stackalloc for smaller allocations
+            int msgIdLength = SafeUTF8Encoding.GetByteCount(msgId);
+            int payloadLength = SafeUTF8Encoding.GetByteCount(payload);
+            int timestampLength = SafeUTF8Encoding.GetByteCount(timestamp);
+            
+            Span<byte> toSignBytes = stackalloc byte[msgIdLength + 1 + timestampLength + 1 + payloadLength];
```
tasn, line 167: This being a stack allocation is a problem. The C# stack is 1mb by default, and we see people with payloads that are larger than 1mb sometimes. This can very easily get to a stack overflow.

tasn (changes requested):
Made one last comment that I found in the latest review, but looks good otherwise.

csharp/Svix/Webhook.cs
esskar: Good pointed; added a upper limit for that.

tasn (approved):

tasn:
Thanks a lot!

PR #1618, opened 2025-01-06
kotlin: Use generated source file for most of the high-level API

svix-jplatte:
Like in other languages, `MessageAttempt` is a very large diff so I'd like to handle it separately. There's already enough breaking changes in this one PR :smiling_face_with_tear: 

Part of https://github.com/svix/monorepo-private/issues/9576.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## Unreleased
+* Libs/Kotlin **(Breaking)**: Mark `api` field of all API resource classes as `private` (previously
+  only some were private, accidentally)
+* Libs/Kotlin **(Breaking)**: Update `recover` to return `RecoverOut` (instead of nothing)
+* Libs/Kotlin **(Breaking)**: Update `replayMissing` to return `ReplayOut` (instead of nothing)
+* Libs/Kotlin **(Breaking)**: Update `sendExample` to return `MessageOut` (instead of nothing)
+* Libs/Kotlin: Fix the parameter names of `Endpoint.get` - `appId` and `endpointId` were swapped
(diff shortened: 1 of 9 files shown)


ChangeLog.md
```diff
@@ -1,5 +1,14 @@
 # Changelog
 
+## Unreleased
+* Libs/Kotlin **(Breaking)**: Mark `api` field of all API resource classes as `private` (previously
+  only some were private, accidentally)
+* Libs/Kotlin **(Breaking)**: Update `recover` to return `RecoverOut` (instead of nothing)
+* Libs/Kotlin **(Breaking)**: Update `replayMissing` to return `ReplayOut` (instead of nothing)
+* Libs/Kotlin **(Breaking)**: Update `sendExample` to return `MessageOut` (instead of nothing)
```
tasn, line 8: Are these breaking? I don't think so, not in Kotlin at least.

svix-jplatte: I'd expect they're "technically-breaking" where you can have an existing function `return` the result of a fn call which would break with mismatched return types, though there would be little reason for anybody to have written code like that.

svix-jplatte: Oh actually ISTR Kotlin also has an expression-oriented `fun x = expression` where `return` is implied, that could be a break for users too.

tasn: Maybe mark is "Potentially breaking" then, because I'd expect it to be rarely breaking for people.

svix-jplatte: Same applies to the very first breaking change. I added a note on top of the readme, wdyt?

tasn: TBH I think it's better to just call it "breaking" and be done with it. This breaks the format people expect from our changelog, and adds a lot of noise.

svix-james (approved):

PR #1621, opened 2025-01-06
CLI: update codegen for optional request bodies

svix-onelson:
For the API calls that require a body but all the fields are technically optional, the ideal case is to allow the CLI command to be invoked without a body argument at all.

This diff wraps the `JsonOf` fields in this situation with `Option` to this effect.

The generator template changes were "basic" in so much as the `Option`-wrapping happens uniformally for cases where passing nothing isn't especially useful (e.g. `PATCH` request).

With some additional special-casing, we can probably make those required arguments, but for now they are optional.

### DIFF
--- a/svix-cli/src/cmds/api/event_type.rs
+++ b/svix-cli/src/cmds/api/event_type.rs
@@ -73,7 +73,7 @@ pub enum EventTypeCommands {
     /// The importer will convert all webhooks found in the either the `webhooks` or `x-webhooks`
     /// top-level.
     ImportOpenapi {
-        event_type_import_open_api_in: JsonOf<EventTypeImportOpenApiIn>,
+        event_type_import_open_api_in: Option<JsonOf<EventTypeImportOpenApiIn>>,
         #[clap(flatten)]
         post_options: Option<PostOptions>,
     },
@@ -99,7 +99,7 @@ pub enum EventTypeCommands {
     /// Partially update an event type.
     Patch {
         event_type_name: String,
-        event_type_patch: JsonOf<EventTypePatch>,
+        event_type_patch: Option<JsonOf<EventTypePatch>>,
     },
 }
 
@@ -131,7 +131,9 @@ impl EventTypeCommands {
                 let resp = client
                     .event_type()
                     .import_openapi(
-                        event_type_import_open_api_in.into_inner(),
+                        event_type_import_open_api_in
+                            .map(|x| x.into_inner())
+                            .unwrap_or_default(),
                         post_options.map(Into::into),
                     )
                     .await?;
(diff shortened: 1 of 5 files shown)


svix-cli/src/cmds/api/event_type.rs
```diff
@@ -131,7 +131,9 @@ impl EventTypeCommands {
                 let resp = client
                     .event_type()
                     .import_openapi(
-                        event_type_import_open_api_in.into_inner(),
+                        event_type_import_open_api_in
+                            .map(|x| x.into_inner())
+                            .unwrap_or_default(),
```
svix-jplatte, line 136: `.unwrap_or_default().into_inner()` seems simpler and probably just needs `JsonOf` to derive `Default`?

svix-onelson: I wasn't confident that all the types we're wrapping actually provided `Default`. I happened to notice these specific bodies did.

But... checking this by adding `Default` to the type bounds, things seem to compile today. Maybe not forever.

I'll go back and make this change in the templates and it'll land here in a follow-up 👍🏻

svix-onelson: I updated the JsonOf typing to support this. The callsites will update during the next codegen.

svix-cli/src/json.rs
```diff
@@ -4,18 +4,18 @@ use anyhow::{Error, Result};
 use colored_json::{Color, ColorMode, ToColoredJson};
 use serde::{de::DeserializeOwned, Serialize};
 
-#[derive(Clone, Debug, PartialEq)]
-pub struct JsonOf<T: DeserializeOwned>(T);
+#[derive(Clone, Debug, Default, PartialEq)]
+pub struct JsonOf<T: DeserializeOwned + Default>(T);
```
svix-jplatte, line 8: This is unnecessary. As is the other bound, actually. These bounds should only be on the impl's that need them.

svix-onelson: I guess not!

svix-cli/src/json.rs
```diff
@@ -4,8 +4,8 @@ use anyhow::{Error, Result};
 use colored_json::{Color, ColorMode, ToColoredJson};
 use serde::{de::DeserializeOwned, Serialize};
 
-#[derive(Clone, Debug, PartialEq)]
-pub struct JsonOf<T: DeserializeOwned>(T);
+#[derive(Clone, Debug, Default, PartialEq)]
+pub struct JsonOf<T>(T);
 
 impl<T: DeserializeOwned> FromStr for JsonOf<T> {
```
svix-onelson, line 10: Looks like this was the only place where we _needed_ to add a type bound. The rest have been removed.

svix-jplatte (approved):

PR #1626, opened 2025-01-08
Make CI happy

svix-jplatte:
New lints dropped in beta.

### DIFF
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -6,7 +6,6 @@
 //! The main entry points of this library are the API client [`api::Svix`], and
 //! [`webhooks::Webhook`].
 
-#![warn(clippy::all)]
 #![forbid(unsafe_code)]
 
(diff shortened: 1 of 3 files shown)


rust/src/lib.rs
```diff
@@ -7,6 +7,7 @@
 //! [`webhooks::Webhook`].
 
 #![warn(clippy::all)]
+#![allow(clippy::literal_string_with_formatting_args)]
```
svix-gabriel, line 10: LMAO is this the lint that triggers from route definitions?

svix-gabriel (approved):

rust/src/lib.rs
svix-jplatte: Yes

svix-gabriel: Clippy really needs to run this stuff by us first smh

svix-gabriel (approved):

PR #1632, opened 2025-01-10
go: Replace most of the lib with generated code

svix-jplatte:
Closes https://github.com/svix/monorepo-private/issues/9576.


svix-onelson (approved):
LGTM

PR #1634, opened 2025-01-14
python: Split generated code into seperate files

svix-mman:
Splitting the generated python code into separate files in preparation for new codegen logic.


python/svix/api/application.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+class ApiBase:
+    _client: AuthenticatedClient
+
+    def __init__(self, client: AuthenticatedClient) -> None:
+        self._client = client
+
+
+class ApplicationAsync(ApiBase):
```
svix-jplatte, line 52: Since these API resource classes now live in different files, does that mean their import paths change? If so, is there a way to re-export them at the previous location?

python/svix/api/application.py
```diff
@@ -0,0 +1,141 @@
...
+    def to_dict(self) -> t.Dict[str, t.Any]:
+        return {k: v for k, v in asdict(self).items() if v is not None}
+
+
+@dataclass
+class PostOptions:
+    idempotency_key: t.Optional[str] = None
+
+    def to_dict(self) -> t.Dict[str, t.Any]:
+        return {k: v for k, v in asdict(self).items() if v is not None}
```
svix-jplatte, line 37: Why duplicate these per resource? `ListOptions` should probably be removed entirely, it just exists to reduce boilerplate in query parameter type definitions, and I removed it in the other libs. `PostOptions` can probably live in `api/__init__.py`?

svix-mman: regarding `ListOptions` I can just add the `limit` and `iterator` fields to the `ApplicationListOptions`

regarding `PostOptions`, good call. since this also breaks the `from svix.api import PostOptions`

svix-mman: on a second thought for now i think it's best just to focus on splitting the code up without refactoring (that much)
I will place the `PostOptions` and the `ListOptions` in a `api/common.py` file (can't place them in `api/__init__.py` since this will cause recursive imports)

python/svix/api/integration.py
```diff
@@ -11,47 +10,23 @@
...
 from ..internal.openapi_client.models.integration_in import IntegrationIn
 from ..internal.openapi_client.models.integration_key_out import IntegrationKeyOut
 from ..internal.openapi_client.models.integration_out import IntegrationOut
 from ..internal.openapi_client.models.integration_update import IntegrationUpdate
 from ..internal.openapi_client.models.list_response_integration_out import (
     ListResponseIntegrationOut,
 )
+from .common import ListOptions, PostOptions, ApiBase
 
 DEFAULT_SERVER_URL = "https://api.svix.com"
```
svix-jplatte, line 22: What about this one? Is it even used? Can it also be moved to the `common` module? (since it seems to be duplicated as well)

svix-mman: good call, the `DEFAULT_SERVER_URL` is only used in `api/__init__.py` in the `ClientBase`

svix-jplatte (approved):

PR #1635, opened 2025-01-14
Add previously available types to the __all__.

svix-mman:
these types were available to import before splitting the code into separate modules, so now they have to be re-exported


svix-james:
Can we correct the spelling in the commit title?

svix-mman:
> Can we correct the spelling in the commit title?

whops, sorry

svix-mman:
This is duplicated work, https://github.com/svix/svix-webhooks/pull/1640 accomplishes this

svix-jplatte:
For the record, I think we should still add stuff to `__all__`, but we can do that as part of generating the entire `__init__.py` file properly.

PR #1638, opened 2025-01-15
Remove duplicate openapi.json file

svix-jplatte:
We can point `include_str!` to a path outside the crate. This would not be a nice thing to do if we were publishing to crates.io, but we are not.


tasn:
I swear this never used to work. I wonder if it's because of the new workspace stuff.

svix-james:
My only question is, have we verified this works in the Docker image?

svix-jplatte:
Since this is `include_str!`, it either fails to build the container, or works. I did check `.dockerignore` which resulted in another PR, but didn't actually run a `docker build` :sweat_smile: will do so now.

svix-jplatte:
Docker image built successfully :+1:

svix-james (approved):

svix-james:
Aprproved, but small favor -- can we add a Python test that verifies the openapi and/or docs endpoint?

PR #1642, opened 2025-01-16
python: add generated client code for application.py

svix-mman:
Most of the code is generated by the codegen, excluding the `get_or_create` code and the `__all__` export

### DIFF
--- a/python/svix/api/application.py
+++ b/python/svix/api/application.py
@@ -1,6 +1,8 @@
 import typing as t
 from dataclasses import dataclass
 
+from .common import ApiBase, BaseOptions
+from ..internal.openapi_client import models
 
 from ..internal.openapi_client.api.application import (
     v1_application_create,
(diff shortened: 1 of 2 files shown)


svix-jplatte (dismissed):
Looks good to me! Any reason this is in draft state?

svix-jplatte:
Ah, I guess CI is complaining at the moment. And this is missing a changelog for the replacement of `PostOptions` with operation-specific option types.

svix-mman:
There is a bug in the codegen
the v1.application.delete does not have any return type, but the codegen shows applicationOut
let me hack on this

python/svix/api/application.py
```diff
@@ -1,53 +1,76 @@
...
 @dataclass
-class ApplicationListOptions(ListOptions):
-    order: t.Optional[Ordering] = None
+class ApplicationCreateOptions(BaseOptions):
+    idempotency_key: t.Optional[str] = None
+
+
+@dataclass
+class ApplicationGetOrCreateOptions(BaseOptions):
+    idempotency_key: t.Optional[str] = None
```
svix-jplatte, line 42: One more thought, these new options types should be exported from `svix.api` for consistency with other options types. Actually this might be something my import fix from yesterday _didn't_ fix, since those types moved into sub-modules.

svix-mman: I could add the new Options types to the `svix/api/__init__.py` by hand.
Or do you think it makes sense to find a way  to numerate all of the types that should be exported (without going through each file by hand)

svix-mman: I think for now I will just add these types to the `svix.api/__init__.py` by hand 
And in a separate pr i will modify the codegen to automatically generate the `__init__.py`

svix-jplatte (approved):

PR #1643, opened 2025-01-16
rust: Format models/*

svix-jplatte:
Reduce the diff with the upcoming new codegen.


rust/src/models/aggregate_event_types_out.rs
```diff
@@ -4,32 +4,31 @@
...
-                AggregateEventTypesOut {
-                    id,
-                    status,
-                    task,
-                    }
-                    }
-                    }
+#[derive(Clone, Default, Debug, PartialEq, Serialize, Deserialize)]
+pub struct AggregateEventTypesOut {
+    #[serde(rename = "id")]
```
svix-gabriel, line 18: I guess it's fine, but why are there all these NOOP serde(rename)s?

svix-gabriel (approved):

rust/src/models/aggregate_event_types_out.rs
svix-jplatte: 'cause old codegen is stupid

PR #1644, opened 2025-01-16
rust: Replace OpenAPI generator with openapi-codegen for model generation

svix-jplatte:
Part of https://github.com/svix/monorepo-private/issues/9599.


rust/src/models/environment_in.rs
```diff
@@ -1,34 +1,24 @@
...
-    #[serde(
-        rename = "transformationTemplates",
-        skip_serializing_if = "Option::is_none"
-    )]
-    pub transformation_templates: Option<Vec<models::TemplateIn>>,
+    #[serde(rename = "eventTypes")]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub event_types: Option<Vec<EventTypeIn>>,
+
+    pub settings: serde_json::Value,
```
tasn, line 12: No longer optional? Is that correct? (not taking a stance, just asking)

svix-jplatte: I made a point to go through all of the changes to non-deleted files, but glanced over this. It is shown as required here: https://api.svix.com/docs#tag/Environment/operation/v1.environment.import

I'll check why.

svix-jplatte: schemars' `schema_with` attribute disables some default behavior where `Option` fields are marked as not required apparently. I'll update this here by hand once, but also made a PR to fix the spec: https://github.com/svix/svix-webhooks-private/pull/3352

svix-james:
If you want me to hit the button, I'll do it, but I have no idea how to meaningfully review this :laughing:

svix-jplatte:
Reviewing the first commit and spot-checking a few files in the second one should be sufficient.

svix-james (approved):

PR #1645, opened 2025-01-16
python: Sort methods on python resources consistently

svix-mman:
These should be the only files that need sorting, the rest of resources have their methods sorted correctly.


svix-jplatte (approved):

svix-jplatte:
I think "consistently" is not very clear. For the record, this reorders the methods to match what our new codegen tool outputs (which I think it alphabetically ordered per endpoint path, then `create` > `read` > `update` > `delete` > `patch` > ...).

PR #1646, opened 2025-01-16
python: Use openapi-codegen for high-level python API client 

svix-mman:
The `openapi.json` file I am working with is incomplete. 
1. Some of the deprecated methods are not included in the `openapi.json`.
2. Some of the API resources are missing, `background_task` and `operational_webhook`. (I know that  `operational_webhook` is not a single resource)
3. Some API resources that are in the `openapi.json` are not currently in the high-level API client (`health` and `environment`).

Jonas is working on a fix for the `openapi.json` generator.

For now. 
1. I will manually add these deprecated calls by hand. And add a `@deprecated` decorator to those methods.
2. If the API resource is not included in the `openapi.json`, I will leave it as is.
3. For API resources that are not currently in the high-level API client, but are included in the `openapi.json`, I will ignore them for now.

Part of https://github.com/svix/monorepo-private/issues/9675

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,6 +1,11 @@
 # Changelog
 
 ## Unreleased
+* Libs/Python **(Breaking)**: `PostOptions` and `ListOptions` are no longer used in methods for `Authentication`,`Endpoint`,`EventType`,`Integration`,`MessageAttempt`,`Message` and `Statistics` resources. Instead each API call now has it's own `{Resource}{Operation}Options`. (Both sync and async)
+* Libs/Python: In `Application` the `dashboard_access` method is deprecated in favor of `app_portal_access`. (Both sync and async)
+* Libs/Python **(Breaking)**: `EndpointStatsOptions` is renamed to `EndpointGetStatsOptions`
+* Libs/Python **(Breaking)**: `MessageAttemptListOptions` is removed in favor of call specific `{Resource}{Operation}Options`
+* Libs/Python **(Breaking)**: For `Statistics` in the `aggregate_event_types` method the `task_id` parameter is removed, Please note that previously this parameter was ignored and had no affect (Both sync and async)
 * Libs/Kotlin **(Breaking)**: Mark `api` field of all API resource classes as `private` (previously
   only some were private, accidentally)
 * Libs/Kotlin **(Breaking)**: Update `recover` to return `RecoverOut` (instead of nothing)
--- a/python/svix/api/authentication.py
+++ b/python/svix/api/authentication.py
@@ -1,13 +1,43 @@
-from .common import PostOptions, ApiBase
+import typing as t
+from dataclasses import dataclass
+from deprecated import deprecated
+
+from .common import ApiBase, BaseOptions
+
 
 from ..internal.openapi_client.api.authentication import (
     v1_authentication_app_portal_access,
     v1_authentication_dashboard_access,
+    v1_authentication_expire_all,
     v1_authentication_logout,
 )
+
 from ..internal.openapi_client.models.app_portal_access_in import AppPortalAccessIn
 from ..internal.openapi_client.models.app_portal_access_out import AppPortalAccessOut
 from ..internal.openapi_client.models.dashboard_access_out import DashboardAccessOut
+from ..internal.openapi_client.models.application_token_expire_in import (
+    ApplicationTokenExpireIn,
+)
+
+
+@dataclass
+class AuthenticationAppPortalAccessOptions(BaseOptions):
+    idempotency_key: t.Optional[str] = None
+
+
+@dataclass
+class AuthenticationDashboardAccessOptions(BaseOptions):
+    idempotency_key: t.Optional[str] = None
+
+
+@dataclass
+class AuthenticationExpireAllOptions(BaseOptions):
+    idempotency_key: t.Optional[str] = None
+
+
+@dataclass
+class AuthenticationLogoutOptions(BaseOptions):
+    idempotency_key: t.Optional[str] = None
 
 
 class AuthenticationAsync(ApiBase):
@@ -15,8 +45,9 @@ class AuthenticationAsync(ApiBase):
         self,
         app_id: str,
         app_portal_access_in: AppPortalAccessIn,
-        options: PostOptions = PostOptions(),
+        options: AuthenticationAppPortalAccessOptions = AuthenticationAppPortalAccessOptions(),
     ) -> AppPortalAccessOut:
+        """Use this function to get magic links (and authentication codes) for connecting your users to the Consumer Application Portal."""
         return await v1_authentication_app_portal_access.request_asyncio(
             client=self._client,
             app_id=app_id,
@@ -24,14 +55,36 @@ class AuthenticationAsync(ApiBase):
             **options.to_dict(),
         )
 
+    @deprecated
     async def dashboard_access(
-        self, app_id: str, options: PostOptions = PostOptions()
+        self,
+        app_id: str,
+        options: AuthenticationDashboardAccessOptions = AuthenticationDashboardAccessOptions(),
     ) -> DashboardAccessOut:
         return await v1_authentication_dashboard_access.request_asyncio(
             client=self._client, app_id=app_id, **options.to_dict()
         )
 
-    async def logout(self, options: PostOptions = PostOptions()) -> None:
+    async def expire_all(
+        self,
+        app_id: str,
+        application_token_expire_in: ApplicationTokenExpireIn,
+        options: AuthenticationExpireAllOptions = AuthenticationExpireAllOptions(),
+    ) -> None:
+        """Expire all of the tokens associated with a specific application."""
+        return await v1_authentication_expire_all.request_asyncio(
+            client=self._client,
+            app_id=app_id,
+            json_body=application_token_expire_in,
+            **options.to_dict(),
+        )
+
+    async def logout(
+        self, options: AuthenticationLogoutOptions = AuthenticationLogoutOptions()
+    ) -> None:
+        """Logout an app token.
+
+        Trying to log out other tokens will fail."""
         return await v1_authentication_logout.request_asyncio(
             client=self._client, **options.to_dict()
         )
@@ -42,8 +95,9 @@ class Authentication(ApiBase):
         self,
         app_id: str,
         app_portal_access_in: AppPortalAccessIn,
-        options: PostOptions = PostOptions(),
+        options: AuthenticationAppPortalAccessOptions = AuthenticationAppPortalAccessOptions(),
     ) -> AppPortalAccessOut:
+        """Use this function to get magic links (and authentication codes) for connecting your users to the Consumer Application Portal."""
         return v1_authentication_app_portal_access.request_sync(
             client=self._client,
             app_id=app_id,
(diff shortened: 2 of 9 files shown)


svix-mman:
(I force-pushed a typo fix in the commit message)

python/svix/api/authentication.py
```diff
@@ -1,37 +1,90 @@
...
+        self,
+        app_id: str,
+        options: AuthenticationDashboardAccessOptions = AuthenticationDashboardAccessOptions(),
     ) -> DashboardAccessOut:
         return await v1_authentication_dashboard_access.request_asyncio(
             client=self._client, app_id=app_id, **options.to_dict()
         )
 
-    async def logout(self, options: PostOptions = PostOptions()) -> None:
+    async def expire_all(
```
tasn, line 68: There are multiple `expire_all`?

python/svix/api/authentication.py
```diff
@@ -42,23 +95,46 @@ def app_portal_access(
...
     ) -> AppPortalAccessOut:
+        """Use this function to get magic links (and authentication codes) for connecting your users to the Consumer Application Portal."""
         return v1_authentication_app_portal_access.request_sync(
             client=self._client,
             app_id=app_id,
             json_body=app_portal_access_in,
             **options.to_dict(),
         )
 
+    def expire_all(
```
tasn, line 108: There are multiple `expire_all`?

python/svix/api/endpoint.py
```diff
@@ -1,85 +1,123 @@
...
+    v1_endpoint_list,
     v1_endpoint_create,
-    v1_endpoint_delete,
     v1_endpoint_get,
-    v1_endpoint_get_headers,
-    v1_endpoint_get_secret,
-    v1_endpoint_get_stats,
-    v1_endpoint_list,
+    v1_endpoint_update,
+    v1_endpoint_delete,
```
svix-jplatte, line 14: Sure would be nice if `ruff format` sorted these imports :sweat_smile:
(I guess importing symbols can have side effects in Python so it can't really do that :melting_face:)

python/svix/api/endpoint.py
```diff
@@ -217,19 +274,20 @@ async def get_stats(
...
             client=self._client,
             app_id=app_id,
             endpoint_id=endpoint_id,
-            since=ensure_tz(options.since),
-            until=ensure_tz(options.until),
+            **options.to_dict(),
         )
 
-    async def transformations_get(
+    async def transformation_get(
```
svix-jplatte, line 287: Can you add this rename to the `ChangeLog.md` as a breaking change please? (same for the non-async one but doesn't need to be two separate entries in changelog)

python/svix/api/event_type.py
```diff
@@ -1,91 +1,139 @@
...
-            client=self._client,
-            **options.to_dict(),
+            client=self._client, **options.to_dict()
         )
 
     async def create(
-        self, event_type_in: EventTypeIn, options: PostOptions = PostOptions()
+        self,
+        event_type_in: EventTypeIn,
+        options: EventTypeCreateOptions = EventTypeCreateOptions(),
```
svix-jplatte, line 75: The replacement of `PostOptions` with operation-specific option classes should also be noted as a breaking change in the changelog.

python/svix/api/event_type.py
```diff
@@ -1,91 +1,139 @@
...
             client=self._client,
             event_type_name=event_type_name,
             json_body=event_type_update,
         )
 
-    async def delete(self, event_type_name: str) -> None:
+    async def delete(
+        self,
+        event_type_name: str,
+        options: EventTypeDeleteOptions = EventTypeDeleteOptions(),
```
svix-jplatte, line 121: I guess this one is _not_ a breaking change in Python, nice :)

Just a quick ping for @tasn here again - are we fine exposing the expunge parameter like this, on the basis that it's the easiest for codegen? I think we can always deprecate it later in favor of separate `archive` and `expunge` methods.

python/svix/api/message_attempt.py
```diff
@@ -1,64 +1,150 @@
...
+    status_code_class: t.Optional[models.StatusCodeClass] = None
 
 
 class MessageAttemptAsync(ApiBase):
     async def list_by_endpoint(
         self,
         app_id: str,
         endpoint_id: str,
-        options: MessageAttemptListOptions = MessageAttemptListOptions(),
+        options: MessageAttemptListByEndpointOptions = MessageAttemptListByEndpointOptions(),
```
svix-jplatte, line 139: Splitting `MessageAttemptListOptions` into several separate types is also sth. we should note in changelog (see existing entries for other libraries).

python/svix/api/statistics.py
```diff
@@ -1,43 +1,67 @@
...
+        retrieve the results of the operation."""
         return await v1_statistics_aggregate_app_stats.request_asyncio(
-            client=self._client,
-            json_body=app_usage_stats_in,
-            **options.to_dict(),
+            client=self._client, json_body=app_usage_stats_in, **options.to_dict()
         )
 
-    async def aggregate_event_types(self, task_id: str) -> AggregateEventTypesOut:
+    async def aggregate_event_types(self) -> AggregateEventTypesOut:
```
svix-jplatte, line 38: Found another breaking change, we had an unused parameter here before :sweat_smile:

svix-jplatte (commented):
Looks pretty good! I think we just need to document the breaking changes, then we can merge.

python/svix/api/event_type.py
tasn: Yeah, it's fine. We also don't need to expose an expunge function if there isn't one already.

python/svix/api/message.py
```diff
@@ -71,73 +89,117 @@ class MessageAsync(ApiBase):
...
-            **options.to_dict(),
+            client=self._client, app_id=app_id, **options.to_dict()
         )
 
     async def create(
-        self, app_id: str, message_in: MessageIn, options: PostOptions = PostOptions()
+        self,
+        app_id: str,
+        message_in: MessageIn,
+        options: MessageCreateOptions = MessageCreateOptions(),
```
tasn, line 110: @svix-jplatte, I wonder if we want to break this? I guess it's fine and we can just swallow this? Though if we do, just need to make sure it's the last Python break.

python/svix/api/endpoint.py
svix-mman: You are right, importing can cause side effects since importing will execute the `__init__.py`
But we could use this https://docs.astral.sh/ruff/formatter/#sorting-import

svix-mman:
> Looks pretty good! I think we just need to document the breaking changes, then we can merge.

I will go through the diffs one at a time and generate a change log for all breaking changes

ChangeLog.md
```diff
@@ -1,6 +1,16 @@
 # Changelog
 
 ## Unreleased
+* Libs/Python **(Breaking)**: `PostOptions` and `ListOptions` are no longer used in methods for `Authentication`,`Endpoint`,`EventType`,`Integration`,`MessageAttempt`,`Message` and `Statistics` resources. Instead each API call now has it's own ResourceMethodsOptions. (Both sync and async)
+* Libs/Python: `Application.dashboard_access` is not marked as deprecated. (Both sync and async)
```
svix-jplatte, line 5: Can you add "please use `app_portal_access instead`" or "in favor of `app_portal_access`"? I guess you wouldn't know this, but that's the replacement (or maybe the doc string says as much), and _if_ anybody reads this changelog after trying to upgrade and not figuring things out by themselves immediately, this would be how we can help them ^^

svix-jplatte:
```suggestion
* Libs/Python: `Application.dashboard_access` is now marked as deprecated. (Both sync and async)
```

also you have a typo there ;)

ChangeLog.md
```diff
@@ -1,6 +1,16 @@
...
+* Libs/Python **(Breaking)**: `EndpointStatsOptions` is renamed to `EndpointGetStatsOptions`
+* Libs/Python **(Breaking)**: For `Endpoint.patch_headers` the `endpoint_headers_in` argument is rename to `endpoint_patch_headers_in` (Both sync and async)
```
svix-jplatte, line 7: typo: "is rename"

svix-jplatte: This is a breaking change because every Python argument can be passed as `name=value` by default? I wonder if it's worth noting down..

ChangeLog.md
```diff
@@ -1,6 +1,16 @@
 # Changelog
 
 ## Unreleased
+* Libs/Python **(Breaking)**: `PostOptions` and `ListOptions` are no longer used in methods for `Authentication`,`Endpoint`,`EventType`,`Integration`,`MessageAttempt`,`Message` and `Statistics` resources. Instead each API call now has it's own ResourceMethodsOptions. (Both sync and async)
```
svix-jplatte, line 4: I'd call it `ResourceOperationOptions` (and maybe put as inline code too)

svix-jplatte: or even `{Resource}{Operation}Options`? I think py people would be familiar with the interpolation notation, so could be clearer than without.

ChangeLog.md
svix-jplatte: I'm seeing a bunch more like this below. I think they're decreasing the signal-noise ratio a bit. If you need sbd. else to make a decision ask Tom but I'd be inclined to skip these, seems extremely unlikely that people are using named arguments for these methods with very few parameters, and should be very easy to adjust to w/o reading the changelog as well.

ChangeLog.md
```diff
@@ -1,6 +1,16 @@
...
+* Libs/Python: `Application.dashboard_access` is not marked as deprecated. (Both sync and async)
+* Libs/Python **(Breaking)**: `EndpointStatsOptions` is renamed to `EndpointGetStatsOptions`
+* Libs/Python **(Breaking)**: For `Endpoint.patch_headers` the `endpoint_headers_in` argument is rename to `endpoint_patch_headers_in` (Both sync and async)
+* Libs/Python **(Breaking)**: `Endpoint.transformations_get` is renamed to `Endpoint.transformation_get` (Both sync and async)
+* Libs/Python **(Breaking)**: For`EventType.import_openapi` the `event_type_import_openapi_in` argument is renamed to `event_type_import_open_api_in` (Both sync and async)
+* Libs/Python **(Breaking)**: For `Integration.create` the `integ_in` argument is rename to `integration_in` (Both sync and async)
+* Libs/Python **(Breaking)**: For `Integration.update` the `integ_update` argument is rename to `integration_update` (Both sync and async)
+* Libs/Python **(Breaking)**: `MessageAttemptListOptions` is removed in favor of call specific Options
```
svix-jplatte, line 12: specific => operation-specific, should be a bit clearer

ChangeLog.md
```diff
@@ -1,6 +1,16 @@
...
+* Libs/Python **(Breaking)**: For `Statistics.aggregate_event_types` the `task_id` argument is removed (Both sync and async)
```
svix-jplatte, line 13: Could you reword to note that this parameter (not argument, arguments are at the call sites) was unused / not doing anything?

svix-jplatte (commented):
Lots of thoughts on these changelog entries :sweat_smile:

svix-jplatte (approved):

ChangeLog.md
tasn: TBH I wouldn't be surprised if people *are*, but I'm happy with making it more a blanket statement.

We just need to make sure to clarify that this is a one-time break, and that everything should be easy to adjust manually.

python/svix/api/endpoint.py
svix-mman: I will sort the imports in a separate PR

python/svix/api/message.py
svix-jplatte: If we're going to be consistent about these things, we need to. And it only affect users who were passing something for `options`. W.r.t. not breaking things again, yeah I don't see anything more we would want to change here.

PR #1647, opened 2025-01-20
python: Add basic openapi.json fetching test

svix-jplatte:
See https://github.com/svix/svix-webhooks/pull/1638#issuecomment-2599133399.


svix-onelson (approved):
Thanks for this 🙏🏻

PR #1648, opened 2025-01-20
Update redoc

svix-jplatte:
The next tag only contains pre-releases, and there haven't been any in a
while + we should be using a stable release anyways.

### DIFF
--- a/server/svix-server/src/static/docs.html
+++ b/server/svix-server/src/static/docs.html
@@ -21,6 +21,6 @@ ReDoc doesn't change outer page styles
 </head>
 <body>
 <redoc spec-url="/api/v1/openapi.json"></redoc>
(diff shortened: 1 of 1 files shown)


tasn (approved):

server/svix-server/src/static/docs.html
```diff
@@ -21,6 +21,6 @@
 </head>
 <body>
 <redoc spec-url="/api/v1/openapi.json"></redoc>
-<script src="https://cdn.jsdelivr.net/npm/redoc@next/bundles/redoc.standalone.js"> </script>
+<script src="https://cdn.jsdelivr.net/npm/redoc@2.3.0/bundles/redoc.standalone.js"> </script>
```
tasn, line 24:
```suggestion
<script src="https://cdn.jsdelivr.net/npm/redoc@latest/bundles/redoc.standalone.js"> </script>
```
Maybe better?

tasn (approved):

PR #1649, opened 2025-01-20
ci: Switch python tests to use public release image

svix-mman:
When running the python tests, the svix server docker image is executed.
Instead of building the server from scratch each time, use the publicly released image.

### DIFF
--- a/.github/workflows/python-tests.yml
+++ b/.github/workflows/python-tests.yml
@@ -20,8 +20,8 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Build svix server image
-        run: docker compose build
+      - name: Pull svix server image (and dependencies)
+        run: docker compose pull
         working-directory: ./server
 
       - uses: actions/setup-python@v5
--- a/python/tests/conftest.py
+++ b/python/tests/conftest.py
@@ -50,12 +50,19 @@ if not TOKEN and not SERVER_URL:
             # fallback on v1 otherwise
             return "docker-compose"
 
+    # `pytest-docker` reads this to override the location of the docker-compose.yml file
     @pytest.fixture(scope="session")
     def docker_compose_file():
         return [
             os.path.join(os.path.dirname(__file__), "../../server/docker-compose.yml"),
             os.path.join(os.path.dirname(__file__), "docker-compose.override.yml"),
         ]
+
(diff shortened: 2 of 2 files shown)


.github/workflows/python-tests.yml
```diff
@@ -20,8 +20,8 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Build svix server image
-        run: docker compose build
+      - name: Pull svix server image (and dependencies)
+        run: docker compose pull
```
tasn, line 24: Probably not the end of the word, but this will break if the libs rely on stuff that was just added to the server and not yet released.

svix-jplatte: Yeah, we talked about this in the retro. We do lots of libs-only releases now so it shouldn't hurt us if we have to do a server-only release between adding a new server feature and exposing it in the libs.

svix-jplatte:
Not sure what's happening with the test failures, but we're definitely still building the docker image here, but as part of the Python test. I think `python/tests/docker-compose.override.yml` needs to be adjusted to override the dependency on `server/Dockerfile` with the upstream image.

svix-mman:
Another problem here is the flaky tests.
The test runs 

> Not sure what's happening with the test failures, but we're definitely still building the docker image here, but as part of the Python test. I think `python/tests/docker-compose.override.yml` needs to be adjusted to override the dependency on `server/Dockerfile` with the upstream image.

I suspect the same, the startup time for this test is till very slow!
I will look into this.

python/tests/conftest.py
```diff
@@ -50,13 +50,21 @@ def docker_compose_command():
...
         return [
             os.path.join(os.path.dirname(__file__), "../../server/docker-compose.yml"),
             os.path.join(os.path.dirname(__file__), "docker-compose.override.yml"),
         ]
 
+    # `pytest-docker` will use this fixture to override the setup command
+    @pytest.fixture(scope="session")
+    def docker_setup():
+        # Don't include the default --build option in the setup
+        return  ["up -d"]
```
svix-jplatte, line 66: Double space? Is this normal?

svix-mman: Nope.

svix-jplatte (approved):

PR #1654, opened 2025-01-20
python: Replace OpenAPI generator http code with out own

svix-mman:
Switching the HTTP call logic with a single (a)sync function.

WIP:
- Don't use the old `internal.openapi_client` error types
- Migrate the resources to use this new client

ref: https://github.com/svix/monorepo-private/issues/9685


python/svix/api/application.py
```diff
@@ -31,124 +21,237 @@ class ApplicationListOptions(BaseOptions):
     # The sorting order of the returned items
     order: t.Optional[models.Ordering] = None
 
+    def query_params(self) -> t.Optional[t.Dict[str, t.Optional[t.Any]]]:
```
tasn, line 24:
```suggestion
    def _query_params(self) -> t.Optional[t.Dict[str, t.Optional[t.Any]]]:
```
Generate these with an underscore (both this and header params) so that it indicates that they are private.

python/svix/api/application.py
```diff
@@ -31,124 +21,237 @@ class ApplicationListOptions(BaseOptions):
...
+        return {
+            "limit": self.limit,
+            "iterator": self.iterator,
+            "order": self.order,
+        }
```
tasn, line 29: You should filter out nulls (I think we have a helper function already?) so that we don't make requests with `limit=null`.

svix-mman: I was checking for this in the HTTP client, but on a second thought I decided to move the logic in to the `_query_params`/`_header_params` methods

python/svix/api/application.py
```diff
@@ -21,41 +21,46 @@ class ApplicationListOptions(BaseOptions):
...
-    def query_params(self) -> t.Optional[t.Dict[str, t.Optional[t.Any]]]:
-        return {
+    def _query_params(self) -> t.Optional[t.Dict[str, str]]:
+        d = {
             "limit": self.limit,
             "iterator": self.iterator,
             "order": self.order,
         }
+        return {
+            k: str(sanitize_field(v)).lower() for k, v in d.items() if v is not None
```
tasn, line 31: * Why `lower()`? That feels wrong. Is it just because of ordering? If so you need to do that on the type itself instead of hacking it here (the enum value can be the casing you want it).
* Calling sanitize field on everything is silly, you just want to call it on the datetime value.
* Also, I don't think the `to str` serialization of datetime is what we want?
* Also also, you can just change the `to_dict()` function to accept optional list of keys and then everything will just be magically clean, e.g. this `_query_params` implementation will look like this:

```python
def _query_params(self) -> t.Optional[t.Dict[str, str]]:
        return self.to_dict(keys=["limit", "iterator", "order])
```
And the headers one will be similarly simple, and just change `to_dict` accordingly in the base class.

python/svix/api/common.py
```diff
@@ -44,3 +51,125 @@ class ApiBase:
...
+        raise HttpError.init_exception(response.json(), response.status_code)
+    if response.status_code == HTTPStatus.NOT_FOUND:
+        raise HttpError.init_exception(response.json(), response.status_code)
+    if response.status_code == HTTPStatus.CONFLICT:
+        raise HttpError.init_exception(response.json(), response.status_code)
+    if response.status_code == HTTPStatus.UNPROCESSABLE_ENTITY:
+        raise HTTPValidationError.init_exception(response.json(), response.status_code)
+    if response.status_code == HTTPStatus.TOO_MANY_REQUESTS:
+        raise HttpError.init_exception(response.json(), response.status_code)
+    raise UnexpectedStatus(response.status_code, response.content)
```
tasn, line 175: I know it doesn't matter because it raises, but `elif` and the last one being `else` is much more descriptive of what's going on here (or just `or` statements).

Though really, this is a leftover from the OpenAPI generator, tbh we just want to change all of this to this (without any ifs):
```
        raise HttpError.init_exception(response.json(), response.status_code)
```
No more of this annoying "UnexpectedStatus".

svix-mman: I think this is the best way to go about this, Just a single error for HTTP exceptions.

tasn: We may want to add new errors at some point, but for now HttpError vs. UnexpectedStatus is silly for us, and the rest of the `if` branches are all identical for you so doing `if` there doesn't make sense.

python/svix/api/application.py
svix-mman: > Why `lower()`. 

Funny enough this is for boolean options, calling `__str__` (using `str()`) does not convert them to lower case, and the server only accepts `true`/`false`

> Calling sanitize field on everything is silly, you just want to call it on the datetime value.

`sanitize_field` does the `if isinstance(v,datetime)`.

> Also, I don't think the to str serialization of datetime is what we want?

Does it make sense to have a single `serialize_str` function that converts each type to it's serialized string representation?
In my head, there are 3 cases that I want to control the string serialization. 
1. `bool`: Should be converted to lowercase
2. `datetime`: Should be a RFC3339 date string
3. `Oredring` (and other custom types): These implement `__str__`, so it's as simple as calling `str(val)` (without calling `lower()` on them

> Also also, I guess this means that to_dict is no longer used? So can probably also drop that while we are at it?

Yup!

tasn: So just fix sanitize_field to handle more than just `isinstance datetime` add to all the types, yeah. Even if generated we don't want too much redundant code so maybe have a function that sanitizes fields and removes nulls and just call that from these places?

svix-mman: Yup, a single `serialize_query_param` function. (the headers are always strings)

tasn: params*, with an s. Just patch the whole dictionary. :)

svix-mman: But I don't think I can have this `serialize_query_param` check for nulls, since a function will always return `None` if there is no explicit return value

tasn: That's why it should accept the whole dictionary, so that it can filter out nulls.

python/svix/api/common.py
```diff
@@ -44,3 +49,111 @@ class ApiBase:
...
+        path_params: t.Optional[t.Dict[str, str]],
+        query_params: t.Optional[t.Dict[str, str]],
+        header_params: t.Optional[t.Dict[str, str]],
+        json_body: t.Optional[t.Dict[str, t.Any]],
+    ) -> t.Dict[str, t.Any]:
+        if path_params is None:
+            url = f"{self._client.base_url}{path}"
+        else:
+            formatted_path = path.format(**path_params)
+            url = url = f"{self._client.base_url}{formatted_path}"
```
tasn, line 66:
```suggestion
            url = f"{self._client.base_url}{formatted_path}"
```
Though also, I'll make another suggestion.

tasn, line 66:
```suggestion
        if path_params is not None:
            path = path.format(**path_params)
        url = f"{self._client.base_url}{path}"
```
Essentially what you want is: if the path params are not none you need to pre process the path, otherwise you don't. Then you take the path and create a URL from it, so this is really what you want.

python/svix/api/common.py
```diff
@@ -27,6 +27,21 @@ def sanitize_field(v: t.Any) -> t.Any:
     return v
 
 
+def _serialize_single_param(val: t.Any) -> str:
+    if isinstance(val, datetime):
+        if val.tzinfo is None:
+            val.replace(tzinfo=timezone.utc)
+        return val.isoformat()
+    elif isinstance(val, bool):
+        return str(val).lower()
```
tasn, line 36:
```suggestion
        return "true" if val else "false"
```
Be explicit with what you want. You don't want a lower case string representation of boolean, you want "true" if val, and "false" otherwise.

python/svix/api/common.py
```diff
@@ -22,11 +27,32 @@ def sanitize_field(v: t.Any) -> t.Any:
...
 @dataclass
 class BaseOptions:
     def to_dict(self) -> t.Dict[str, t.Any]:
         return {k: sanitize_field(v) for k, v in asdict(self).items() if v is not None}
 
+    def _query_params(self) -> t.Dict[str, str]:
+        return {}
+
+    def _header_params(self) -> t.Dict[str, str]:
+        return {}
```
tasn, line 54:
```suggestion
    def _query_params(self) -> t.Optional[t.Dict[str, str]:
        return None

    def _header_params(self) -> t.Optional[t.Dict[str, str]:
        return None
```
It looks like None is accepted by httpx helpers? If so, we can just do that for the empty case.

svix-mman: I think it makes sense to keep the signature consistent with the generated `_query_params`/`_header_params` method, its signature is `def _query_params(self) -> t.Dict[str, str]:`.
And I don't think it makes sense to have the generated function hint a `t.Optional[t.Dict[..]]` if it will always return a `t.Dict[..]`

tasn: It doesn't always return a dict, if it's OOP and the base function returns None so it's very consistent. That's how people do function overloading in OOP when the default is None. I don't want to nitpick too much, so do whatever, just calling it out.

svix-mman:
@tasn The last 2 commits are breaking change, but they add new methods that are deprecated, and the methods have "`deprecated`" in the method name
This feels wrong to me, we should not be adding new methods that are already deprecated

python/svix/api/application.py
```diff
@@ -31,124 +21,234 @@ class ApplicationListOptions(BaseOptions):
...
 class ApplicationCreateOptions(BaseOptions):
     idempotency_key: t.Optional[str] = None
 
+    def _header_params(self) -> t.Dict[str, str]:
+        return serialize_params(
+            {
+                "idempotency-key": self.idempotency_key,
+            }
+        )
+
```
svix-jplatte, line 44: Why is the call to `serialize_params` repeated within every `_header_params` function, rather than being part of `_request_sync` / `_request_asyncio`?

python/svix/api/authentication.py
```diff
@@ -72,21 +89,52 @@ async def expire_all(
...
+    @deprecated
+    async def dashboard_access(
+        self,
+        app_id: str,
+        options: AuthenticationDashboardAccessOptions = AuthenticationDashboardAccessOptions(),
+    ) -> DashboardAccessOut:
+        """DEPRECATED: Please use `app-portal-access` instead.
+
+        Use this function to get magic links (and authentication codes) for connecting your users to the Consumer Application Portal."""
+        # ruff: noqa: F841
```
svix-jplatte, line 113: What does this do?

python/svix/api/authentication.py
```diff
@@ -112,11 +166,16 @@ def expire_all(
...
-            json_body=application_token_expire_in,
-            **options.to_dict(),
+        # ruff: noqa: F841
+        response = self._request_sync(
+            method="post",
+            path="/api/v1/auth/app/{app_id}/expire-all",
+            path_params={
+                "app_id": app_id,
+            },
+            query_params=options._query_params(),
```
svix-jplatte, line 176: Why not inline this (generated) function? Same for `_header_params`.

python/svix/api/background_task.py
```diff
@@ -1,45 +1,91 @@
...
-        return list_background_tasks.request_sync(
-            client=self._client, **options.to_dict()
+        """List background tasks executed in the past 90 days."""
+        # ruff: noqa: F841
+        response = self._request_sync(
+            method="get",
+            path="/api/v1/background-task",
+            path_params={},
+            query_params=options._query_params(),
+            header_params=options._header_params(),
```
svix-jplatte, line 77: Side note, I like the explicitness of using `name=value` for all parameters to the `_request` methods :+1:

python/svix/api/authentication.py
svix-mman: I feel this is cleaner, I would not like the idea of duplicating the `_query_params`/`_header_params` in both the sync/async version of each function

svix-jplatte: Right, that makes sense.

python/svix/api/application.py
tasn: @svix-jplatte, fwiw, I think it makes sense here, as here is where the structure returns normalized values. We may have fancier logic here at some point (e.g. some nulls should maybe be kept after all), so doing it here makes sense.

tasn:
> @tasn The last 2 commits are breaking change, but they add new methods that are deprecated, and the methods have "`deprecated`" in the method name This feels wrong to me, we should not be adding new methods that are already deprecated
> 
> The reason we only get this change now (instead of when I was working on the high-level API), is because I am working with a new `openapi.json` file

Whatever @svix-jplatte thinks is fine by me.

python/svix/api/background_task.py
svix-mman: :smile:

python/svix/api/authentication.py
svix-mman: Fixed!

python/svix/api/application.py
svix-jplatte: Right, you're saying it's not a function of the types alone? I guess that makes sense.

svix-jplatte (approved):

svix-jplatte (approved):

PR #1656, opened 2025-01-21
python: Move `OperationalWebhook` to new codegen

svix-mman:
And rename to `OperationalWebhookEndpoint`
This change was not done in https://github.com/svix/svix-webhooks/pull/1646 because the `openapi.json` file was not ready.

Splitting this off from the main [new HTTP client PR](https://github.com/svix/svix-webhooks/pull/1654) since I am also breaking some of the function arguments


svix-mman:
@svix-jplatte Do you think we should add a breaking change notice for the function params changing?

python/svix/api/operational_webhook_endpoint.py
```diff
@@ -82,61 +72,108 @@ async def list(
...
     ) -> None:
-        return await rotate_operational_webhook_endpoint_secret.request_asyncio(
-            client=self._client,
-            endpoint_id=endpoint_id,
-            json_body=operational_webhook_endpoint_secret_in,
-            **options.to_dict(),
+        """Rotates an operational webhook endpoint's signing secret.
+
+        The previous secret will remain valid for the next 24 hours."""
+        # ruff: noqa: F841
```
svix-jplatte, line 167: I thought the latest version of the template no longer outputs these `noqa` directives?

svix-jplatte (commented):
Which breaking change exactly? The parameter names / replacement of `PostOptions` / something else? The param names and `PostOptions` already have changelog entries, right? If so, let's just update those.

python/svix/api/operational_webhook_endpoint.py
svix-mman: Whops, this was not generated using the latest template.

ChangeLog.md
```diff
@@ -1,7 +1,8 @@
 # Changelog
 
 ## Unreleased
-* Libs/Python **(Breaking)**: `PostOptions` and `ListOptions` are no longer used in methods for `Authentication`,`Endpoint`,`EventType`,`Integration`,`MessageAttempt`,`Message` and `Statistics` resources. Instead each API call now has it's own `{Resource}{Operation}Options`. (Both sync and async)
+* Libs/Python **(Breaking)**: The `OperationalWebhook` resource is renamed to `OperationalWebhookEndpoint`
+* Libs/Python **(Breaking)**: `PostOptions` and `ListOptions` are no longer used in methods for `Authentication`,`Endpoint`,`EventType`,`Integration`,`MessageAttempt`,`Message`, `Statistics` and `OperationalWebhookEndpoint` resources. Instead each API call now has it's own `{Resource}{Operation}Options`. (Both sync and async)
```
svix-jplatte, line 5:
```suggestion
* Libs/Python **(Breaking)**: `PostOptions` and `ListOptions` are no longer used in methods for `Authentication`,`Endpoint`,`EventType`,`Integration`,`MessageAttempt`,`Message`, `Statistics` and `OperationalWebhookEndpoint` resources. Instead each API call now has its own `{Resource}{Operation}Options`. (Both sync and async)
```

ChangeLog.md
```diff
@@ -1,7 +1,8 @@
 # Changelog
 
 ## Unreleased
-* Libs/Python **(Breaking)**: `PostOptions` and `ListOptions` are no longer used in methods for `Authentication`,`Endpoint`,`EventType`,`Integration`,`MessageAttempt`,`Message` and `Statistics` resources. Instead each API call now has it's own `{Resource}{Operation}Options`. (Both sync and async)
+* Libs/Python **(Breaking)**: The `OperationalWebhook` resource is renamed to `OperationalWebhookEndpoint`
```
svix-jplatte, line 4: This doesn't seem right. The resource was always named that, just the module name changed (which was never part of a release, so we don't need to mention it here).

svix-mman: Yeah, I think you are right.

svix-jplatte (approved):

PR #1659, opened 2025-01-23
rust / cli: Merge query and header parameter structs

svix-jplatte:
Same thing we did for Python. Lets us expose query parameters on POST operations in a nice way.


svix-cli/src/cli_types/mod.rs
svix-onelson, line 1: 👍🏻

svix-onelson (approved):
👍🏻

PR #1661, opened 2025-01-23
python: Switch to explicit import for custom types

svix-mman:
Before I get working on generating the models, I want this ready first

### DIFF
--- a/python/svix/api/application.py
+++ b/python/svix/api/application.py
@@ -2,13 +2,13 @@
 import typing as t
 from dataclasses import dataclass
 
-from ..internal.openapi_client import models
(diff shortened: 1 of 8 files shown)


svix-jplatte (commented):
Why are we doing this? I think I left a comment before on the template side already a few days ago, because we are relying on ruff fixing up otherwise invalid code there and I don't understand why we wouldn't just keep doing `models.Type`.

svix-mman:
@svix-jplatte I guess this PR is not really needed, Let's merge it into #1654

svix-jplatte:
Okay.

PR #1666, opened 2025-01-24
python: Switch api model generation to our new codegen

svix-mman:
`codegen.json`
```json
{
  "file-generated-at": "2025-01-27T14:39:43.964452136+00:00",
  "git-rev": "bbde42595953457be710de1653c22789ec056219",
  "openapi.json-sha256": "9fa16f8eb28f617b8495cafc9bd529626145df7af01e5c09acdc2b98e39a4725"
}
```

ref: https://github.com/svix/monorepo-private/issues/9686


python/setup.py
```diff
@@ -31,6 +31,7 @@ def read_file(filepath):
 META_PATH = os.path.join(PKG_DIR, PKG_NAME, "__init__.py")
 META_CONTENTS = read_file(META_PATH)
 PKG_REQUIRES = [
+    "pydantic >=2.10",
```
tasn, line 34: Can probably lax this requirement to `>=2`? No need to force our customers to be on the bleeding edge.

svix-mman: Good call!

python/svix/api/.codegen.json
```diff
@@ -0,0 +1,6 @@
+{
+  "file-generated-at": "2025-01-27T14:39:43.964452136+00:00",
+  "git-rev": "bbde42595953457be710de1653c22789ec056219",
+  "openapi-codegen-version": "0.1.0",
+  "openapi.json-sha256": "9fa16f8eb28f617b8495cafc9bd529626145df7af01e5c09acdc2b98e39a4725"
+}
```
svix-jplatte, line 6: Maybe just drop this in the PR description like I did in #1667? I don't think we should have one of these per directory with generated files, we should have a central one that's updated together with all the generated code through a small script down the line.

python/svix/api/client.py
```diff
@@ -0,0 +1,67 @@
...
+        return attr.evolve(self, timeout=timeout)
+
+
+@attr.s(auto_attribs=True)
+class AuthenticatedClient(Client):
+    """A Client which has been authenticated for use on secured endpoints"""
+
+    token: str
+    prefix: str = "Bearer"
+    auth_header_name: str = "Authorization"
```
svix-jplatte, line 62: Why have these as class attributes?

python/svix/api/http_error.py
```diff
@@ -0,0 +1,73 @@
...
+T = TypeVar("T", bound="HttpError")
+
+
+@attr.s(auto_attribs=True)
+class HttpError(Exception):
+    """
+    Attributes:
+        code (str):
+        detail (str):
+    """
```
svix-jplatte, line 15: This seems like a rather useless doc string?

python/svix/models/background_task_data.py
```diff
@@ -0,0 +1,6 @@
+# TODO - remove this special case when we fix the generated code for empty openapi structs
+from typing import Any, Dict, TypeVar
+
+T = TypeVar("T", bound="BackgroundTaskData")
```
svix-jplatte, line 4: What is this for?

python/svix/models/common.py
```diff
@@ -0,0 +1,15 @@
+from pydantic import BaseModel, ConfigDict
+from re import sub
+
+
+def to_camelcase(s: str) -> str:
+    s = sub(r"(_|-)+", " ", s).title().replace(" ", "").replace("*", "")
+    return "".join([s[0].lower(), s[1:]])
```
svix-jplatte, line 7: o_O

The name is descriptive... But where is this ugly thing coming from? Surely you didn't write it yourself?

python/svix/models/endpoint_in.py
```diff
@@ -0,0 +1,23 @@
...
+
+from .common import SvixBaseModel
+
+
+class EndpointIn(SvixBaseModel):
+    channels: t.Optional[t.Set[str]] = None
+    """List of message channels this endpoint listens to (omit for all)."""
+    description: t.Optional[str] = None
+    disabled: t.Optional[bool] = None
+    filter_types: t.Optional[t.Set[str]] = None
```
svix-jplatte, line 12: I'm wondering whether we should put empty lines between attributes, because I find the docstrings below fields to really impair readability, and maybe that would increase it back again a little.

python/tests/test_client.py
```diff
@@ -159,9 +159,9 @@ def test_svix_endpoint_create(
     )
     assert endpoint.url == endpoint_url
     assert endpoint.uid == endpoint_uid
-    assert endpoint.filter_types == [event_type.name]
+    assert endpoint.filter_types == {event_type.name}
```
svix-jplatte, line 162: What is this? We changed the type of this field?

python/scripts/generate_openapi.sh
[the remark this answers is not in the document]
svix-mman: Since CI is running workflows from main, it will attempt to run it.
Once we merge this branch, we should be able to delete this script

python/tests/test_client.py
svix-mman: Yup, all `openapi.json` sets are now python sets

python/svix/models/background_task_data.py
svix-mman: This is used in `BackgroundTaskOut`, it has a field that the `openapi.json` just calls `"type": "object"` with nothing else.
It's the only warning the codegen spits out

python/svix/api/client.py
svix-mman: This is 1/2 files that were copied verbatim from the old codegen

python/svix/api/http_error.py
svix-mman: This is 2/2 files that were copied verbatim from the old codegen, And I have not looked at them in depth

python/svix/models/endpoint_in.py
svix-mman: I can try. I will update these and let me know how you feel

svix-mman: How do you feel about something like this?
```python
class EndpointIn(SvixBaseModel):
    channels: t.Optional[t.Set[str]] = None
    """List of message channels this endpoint listens to (omit for all)."""

    description: t.Optional[str] = None

    disabled: t.Optional[bool] = None

    filter_types: t.Optional[t.Set[str]] = None
```

python/svix/api/.codegen.json
svix-mman: Yeah, I think this makes sense.

python/svix/models/endpoint_in.py
svix-jplatte: Yeah, that's what I meant. I think it looks better, what do you think?

python/svix/models/common.py
svix-jplatte: Right.. I'm really not a fan on relying on this sort of thing to be honest. IIRC we even have some fields that are `snake_case` accidentally, that we didn't change for backwards compatibility.

How would you feel about switching to [explicit aliases](https://docs.pydantic.dev/2.10/concepts/fields/#field-aliases) whereever the field name is not the same as the serialized name, like we do in the Rust lib?

python/tests/test_client.py
svix-jplatte: Was this discussed anywhere? I mean it makes sense if we ignored backwards compat, I'm just wondering whether it makes sense to have this extra breaking change that seems like it would be very easy to avoid.

python/svix/models/message_status.py
```diff
@@ -0,0 +1,12 @@
+# this file is @generated
+from enum import IntEnum
+
+
+class MessageStatus(IntEnum):
+    VALUE_0 = 0
+    VALUE_1 = 1
+    VALUE_2 = 2
+    VALUE_3 = 3
```
svix-jplatte, line 9: These names aren't great, especially because there are no docs. I checked the spec: We actually have `x-enum-varnames` in there, so we can generate useful names instead of `VALUE_N`s here.

python/svix/models/endpoint_in.py
tasn: As a side note, why don't we have doc strings for description, disabled, filter types?

svix-jplatte: We don't have extensive docs for JSON models in general. Some things have docs, some things don't. For example, the very first endpoint in our docs (list applications): pagination query parameters are documented, response fields (`ApplicationOut`) aren't:

Screenshot 2025-01-28 at 15-09-04 Svix API Reference

python/tests/test_client.py
svix-jplatte: I agree it probably doesn't help to have it as a set. Though I don't think the "they want ordering" bit makes sense - the ordering is erased when deserializing the parameters into a `BTreeSet` (maybe `HashSet` in some cases still) server-side.

python/svix/models/common.py
svix-mman: This was my thought exactly, I did this originally (I already wrote the template for this)
This is what `EndpointIn` will look like
```python
# this file is @generated
import typing as t

from pydantic import Field

from .common import SvixBaseModel


class EndpointIn(SvixBaseModel):
    channels: t.Optional[t.Set[str]] = None
    """List of message channels this endpoint listens to (omit for all)."""

    description: t.Optional[str] = None

    disabled: t.Optional[bool] = None

    filter_types: t.Optional[t.Set[str]] = Field(default=None, alias="filterTypes")

    metadata: t.Optional[t.Dict[str, str]] = None

    rate_limit: t.Optional[int] = Field(default=None, alias="rateLimit")

    secret: t.Optional[str] = None
    """The endpoint's verification secret.

    Format: `base64` encoded random bytes optionally prefixed with `whsec_`.
    It is recommended to not set this and let the server generate the secret."""

    uid: t.Optional[str] = None
    """Optional unique identifier for the endpoint."""

    url: str

    version: t.Optional[int] = None

```
What do you think?

python/tests/test_client.py
svix-mman: Okie, I will change this to a list

python/svix/models/common.py
svix-mman: @tasn I know you had thoughts about this ^^

python/scripts/generate_openapi.sh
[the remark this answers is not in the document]
svix-jplatte: > Since CI is running workflows from main

Are you sure about this? I'm actually pretty sure CI runs on a merge commit as-if the PR had been merged, including any changes to CI workflows.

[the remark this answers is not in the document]
svix-mman: Odd, I could swear I tried this, but it did not work.
I just checked the CI running on this branch and indeed; the `generate-openapi.sh` is not running.
I will delete the file.

python/svix/models/common.py
svix-mman: For now, I will use the pydantic alias instead of assuming that every field is camelCase

python/svix/models/status_code_class.py
```diff
@@ -0,0 +1,22 @@
...
+    """The different classes of HTTP status codes:
+    - CodeNone = 0
+    - Code1xx = 100
+    - Code2xx = 200
+    - Code3xx = 300
+    - Code4xx = 400
+    - Code5xx = 500"""
+
+    CODE_NONE = 0
+    CODE1XX = 100
```
svix-jplatte, line 15: Would be nice if our `to_upper_snake_case` helper turned this into `CODE_1XX`, but probably not worth the effort to change that.

svix-jplatte (approved):
Any (breaking) changes in this that we should include in the changelog?

PR #1670, opened 2025-01-28
python: Add new operational webhook API

svix-jplatte:
New API routes just dropped :)
This change is already out in a pre-release of the Python library, we'll add it to the other libraries before the next release.


svix-mman:
Currently, we are importing all models to `svix/api/__init__.py`. This is because we want to maintain the ability to import `from svix.api import OperationalWebhookEndpointHeadersIn`
This is sadly still done by hand :(

svix-mman (dismissed):
lgtm
Add the 2 `OperationalWebhookEndpointHeadersIn`, `OperationalWebhookEndpointHeadersOut` models to `svix/api/__init__.py` and merge!

svix-mman (approved):

PR #1671, opened 2025-01-28
js: Replace OpenAPI generator http code with our own

svix-jplatte:
Part of https://github.com/svix/monorepo-private/issues/9598.

### DIFF
--- a/javascript/src/api/application.ts
+++ b/javascript/src/api/application.ts
@@ -29,7 +29,7 @@ export class Application {
     request.setQueryParam("iterator", options?.iterator);
     request.setQueryParam("order", options?.order);
 
-    return request.send(this.requestCtx);
+    return request.send(this.requestCtx, "ListResponseApplicationOut");
   }
 
   /** Create a new application. */
@@ -40,9 +40,9 @@ export class Application {
     const request = new SvixRequest(HttpMethod.POST, "/api/v1/app");
 
     request.setHeaderParam("idempotency-key", options?.idempotencyKey);
-    request.body = applicationIn;
+    request.setBody(applicationIn, "ApplicationIn");
 
-    return request.send(this.requestCtx);
+    return request.send(this.requestCtx, "ApplicationOut");
   }
 
   /** Get the application with the UID from `applicationIn`, or create it if it doesn't exist yet. */
@@ -52,11 +52,11 @@ export class Application {
   ): Promise<ApplicationOut> {
     const request = new SvixRequest(HttpMethod.POST, "/api/v1/app");
 
-    request.body = applicationIn;
     request.setQueryParam("get_if_exists", true);
     request.setHeaderParam("idempotency-key", options?.idempotencyKey);
+    request.setBody(applicationIn, "ApplicationIn");
 
-    return request.send(this.requestCtx);
+    return request.send(this.requestCtx, "ApplicationOut");
   }
 
   /** Get an application. */
@@ -65,7 +65,7 @@ export class Application {
 
     request.setPathParam("app_id", appId);
 
-    return request.send(this.requestCtx);
+    return request.send(this.requestCtx, "ApplicationOut");
   }
 
   /** Update an application. */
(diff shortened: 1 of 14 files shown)


javascript/src/api/application.ts
```diff
@@ -20,73 +19,93 @@ export interface ApplicationListOptions {
...
+    const request = new SvixRequest(HttpMethod.POST, "/api/v1/app");
 
-  /** Get the application with the UID from `applicationIn`, or create it if it doesn't exist yet. */
-  public getOrCreate(
+    request.body = applicationIn;
+    request.setHeaderParam("idempotency-key", options?.idempotencyKey);
+
+    const responseBody: any = await request.send(this.requestCtx);
+    return responseBody as ApplicationOut;
+  } /** Get the application with the UID from `applicationIn`, or create it if it doesn't exist yet. */
```
tasn, line 50: Looks wrong?

javascript/src/request.ts
```diff
@@ -0,0 +1,176 @@
...
+    // This work around that. Source:
+    // https://github.com/cloudflare/workers-sdk/issues/2514#issuecomment-2178070014
+    const isCredentialsSupported = "credentials" in Request.prototype;
+
+    const response = await sendWithRetry(url, {
+      method: this.method.toString(),
+      body,
+      headers: {
+        accept: "application/json, */*;q=0.8",
+        "user-agent": `svix-libs/${LIB_VERSION}/javascript`,
```
tasn, line 116: Probably want to make this a static instead of generating it every time?

javascript/src/request.ts
```diff
@@ -0,0 +1,177 @@
...
+   *
+   * If the server returns a 422 error, an `ApiException<HTTPValidationError>` is thrown.
+   * If the server returns another 4xx error, an `ApiException<HttpErrorOut>` is thrown.
+   *
+   * If the server returns a 5xx error, the request is retried up to two times with exponential backoff.
+   * If retries are exhausted, an `ApiException<HttpErrorOut>` is thrown.
+   */
+  public async send<R>(ctx: SvixRequestContext): Promise<R> {
+    const responseBody = await this.sendInner(ctx);
+    return JSON.parse(responseBody);
```
svix-lucho, line 86: does `JSON.parse` deserialize `Date` correctly?

svix-lucho: I don't think it does, the current middleware has this somewhere

```
    public static deserialize(data: any, type: string, format: string): any {
        // polymorphism may change the actual type.
        type = ObjectSerializer.findCorrectType(data, type);
        if (data == undefined) {
            return data;
        } else if (primitives.indexOf(type.toLowerCase()) !== -1) {
            return data;
        } else if (type.endsWith(nullableSuffix)) {
            let subType: string = type.slice(0, -nullableSuffix.length); // Type | null => Type
            return ObjectSerializer.deserialize(data, subType, format);
        } else if (type.endsWith(optionalSuffix)) {
            let subType: string = type.slice(0, -optionalSuffix.length); // Type | undefined => Type
            return ObjectSerializer.deserialize(data, subType, format);
        } else if (type.startsWith(arrayPrefix)) {
            let subType: string = type.slice(arrayPrefix.length, -arraySuffix.length); // Array<Type> => Type
            let transformedData: any[] = [];
            for (let date of data) {
                transformedData.push(ObjectSerializer.deserialize(date, subType, format));
            }
            return transformedData;
        } else if (type.startsWith(mapPrefix)) {
            let subType: string = type.slice(mapPrefix.length, -mapSuffix.length); // { [key: string]: Type; } => Type
            let transformedData: { [key: string]: any } = {};
            for (let key in data) {
                transformedData[key] = ObjectSerializer.deserialize(
                    data[key],
                    subType,
                    format,
                );
            }
            return transformedData;
        } else if (type === "Date") {
            return new Date(data);
        } else {
            if (enumsMap.has(type)) {// is Enum
                return data;
            }

            if (!typeMap[type]) { // dont know the type
                return data;
            }
            let instance = new typeMap[type]();
            let attributeTypes = typeMap[type].getAttributeTypeMap();
            for (let attributeType of attributeTypes) {
                let value = ObjectSerializer.deserialize(data[attributeType.baseName], attributeType.type, attributeType.format);
                if (value !== undefined) {
                    instance[attributeType.name] = value;
                }
            }
            return instance;
        }
    }
```

svix-jplatte: Dang, yes. I forgot all about this. Thanks for catching!

svix-jplatte: Good thing we can use the existing `ObjectSerializer` for now. Need to think about how we are going to replace this down the line...

svix-lucho: Ahh good idea. Yeah, also enums are going to need a solution.

svix-lucho (approved):

PR #1672, opened 2025-01-29
rust: Regenerate lib

svix-jplatte:
Some generator and spec improvements have happened. We've also got two new endpoints.

This required almost zero manual changes, and the last one I did have to do is about to be fixed by a spec update! :tada:
(however, `models/mod.rs` and `api/mod.rs` are not yet being updated by codegen)


rust/src/models/message_attempt_trigger_type.rs
```diff
@@ -1,44 +1,37 @@
...
-impl Default for MessageAttemptTriggerType {
-    fn default() -> MessageAttemptTriggerType {
-        Self::Scheduled
+impl fmt::Display for MessageAttemptTriggerType {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let value = match self {
+            Self::Scheduled => "0",
+            Self::Manual => "1",
+        };
+        f.write_str(value)
```
svix-mman, line 35: Does it make sense to `f.write_str(&serde_json::to_string(self).unwrap())`, this way we don't have to generate code that `serde_repr` already knows about?

svix-mman (approved):

rust/src/models/message_attempt_trigger_type.rs
svix-jplatte: I'd like to avoid an `unwrap` like that. I have another idea though..

PR #1679, opened 2025-01-30
python: Add special case where the server does not send a camelCase field

svix-mman:
@svix-jplatte [here](https://github.com/svix/svix-webhooks-private/blob/dec537950635d73531734c2980811163191c4d32/server/svix-server/src/v1/endpoints/event_type/mod.rs#L720) is where this field is defined

### DIFF
--- a/python/svix/models/event_type_import_open_api_out_data.py
+++ b/python/svix/models/event_type_import_open_api_out_data.py
@@ -1,5 +1,6 @@
-# this file is @generated
 import typing as t
+
+from pydantic import Field
 
 from .common import BaseModel
 from .event_type_from_open_api import EventTypeFromOpenApi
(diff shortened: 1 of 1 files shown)


tasn (dismissed):
If the server is bad, fix the server.

svix-jplatte:
> If the server is bad, fix the server.

That would break all existing code parsing the reasons of that endpoint, though?

tasn:
Ah, it's "out", I thought it was in. Sorry.

I guess this is something we should handle in the generator automatically, so something like "if property name is not camelcase then generate an alias here explicitly".

Though can we please *also* fix it server side by calling it something completely separate and returning the field twice (and deprecating the old field or better yet, hiding it altogether)? E.g. call it "potential" or "applicable" or something.

svix-mman:
> I guess this is something we should handle in the generator automatically, so something like "if property name is not camelcase then generate an alias here explicitly".

I added this PR that generates this code. https://github.com/svix/openapi-codegen/pull/51

svix-mman:
@tasn As far as I am aware, this is the only field that does not follow the camleCase naming convention.
Can we merge this?

svix-jplatte (approved):

PR #1685, opened 2025-02-03
Svix CLI: release the CLI as a docker image.

tasn:
It could be useful in some environments.

### DIFF
--- a/svix-cli/Dockerfile
+++ b/svix-cli/Dockerfile
@@ -1,0 +1,62 @@
+# Base build
+FROM rust:1.81-slim-bookworm AS build
+
+RUN apt-get update && apt-get install -y \
+    build-essential=12.* \
+    checkinstall=1.* \
+    curl=7.* \
+    libssl-dev=* \
+    pkg-config=1.8.* \
+    zlib1g-dev=1:* \
+    cmake=3.* \
+    --no-install-recommends
+
+RUN set -ex ; \
+        mkdir -p /app ;\
+        useradd appuser ;\
+        chown -R appuser: /app ;\
+        mkdir -p /home/appuser ;\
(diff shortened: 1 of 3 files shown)


svix-cli/Dockerfile
```diff
@@ -0,0 +1,62 @@
...
+    update-ca-certificates; \
+    rm -rf /var/lib/apt/lists/*
+
+USER appuser
+
+COPY --from=build /app/svix-cli/target/release/svix /usr/local/bin/svix
+COPY bridge/scripts/check-deps.sh /usr/local/bin/check-deps.sh
+
+# Verify all the dynamic libs we depend on are present in the runtime stage
+RUN /usr/local/bin/check-deps.sh /usr/local/bin/svix
```
svix-onelson, line 60: nice 😎

svix-cli/Dockerfile
```diff
@@ -0,0 +1,62 @@
...
+
+RUN apt-get update && apt-get install -y \
+    build-essential=12.* \
+    checkinstall=1.* \
+    curl=7.* \
+    libssl-dev=* \
+    pkg-config=1.8.* \
+    zlib1g-dev=1:* \
+    cmake=3.* \
+    --no-install-recommends
```
svix-onelson, line 12: Some of these I bet we could ditch, but maybe that'd mean more cache evictions on the GHA side. Pretty sure `cmake` was needed for `rdkafka` and nothing else.

svix-onelson (approved):
Let's try it!

svix-cli/Dockerfile
tasn: Probably right, though not worth even thinking about, this is just for the build image, not the deploy image.

PR #1686, opened 2025-02-03
Remove additionalProperties: true in regen_openapi.sh

svix-jplatte:
… to allow committing a spec file without this hack, which is useful for our new code generator.


svix-onelson (approved):
Wondering about the downside of removing this (even temporarily)?

Edit: oh, I see. This was previously being removed in the generator or something. The larger change is to move the responsibility to this script so the generator doesn't need to hack around to do it.

Issue #1652, opened 2025-01-20
Allow to configure request timeout on javascript client


zepedrotm:
### Motivation

We have recently found that the current request timeout using the Svix javascript client is > 30s. It would great to have the option of configure the timeout e.g. implement strategies, such as failing fast and rely on retries.

### Proposal

Add a config property to set the desired request timeout upon initializing the Svix client.

### Alternatives

The alternative is to drop the usage of the javascript client and implement a custom one using one of the JS libs for HTTP integrations.

tasn:
Just to add: we probably also want a sane default.

PR #1687, opened 2025-02-03
js: Support setting a request timeout

svix-jplatte:
Closes https://github.com/svix/svix-webhooks/issues/1652.


svix-onelson (approved):
Q: what's the behavior when timeout is `<= 0`?

svix-jplatte:
Nasal demons.





(I'll check tomorrow)

svix-jplatte:
Similar errors on Firefox and Chrome:

> Uncaught TypeError: Failed to execute 'timeout' on 'AbortSignal': Value is outside the 'unsigned long long' value range.

MDN just says it must be between 0 and `Number.MAX_SAFE_INTEGER`, not what happens if it isn't.

PR #1688, opened 2025-02-03
python: Add `None` fallback if the server does not send an error

svix-mman:


svix-mman:
> I really wonder what sort of error the server returned there... If it was an empty payload, the `response.json()` going into this would have failed, right?

I would think so, But I guess not?
This is the `response.json()`

```python
    def json(self, **kwargs: typing.Any) -> typing.Any:
        if self.charset_encoding is None and self.content and len(self.content) > 3:
            encoding = guess_json_utf(self.content)
            if encoding is not None:
                return jsonlib.loads(self.content.decode(encoding), **kwargs)
        return jsonlib.loads(self.text, **kwargs)
```

PR #1690, opened 2025-02-03
Rust: Add retries

jaymell:
Default of 2 retries if setting is not explicitly specified.


rust/tests/it/kitchen_sink.rs
```diff
@@ -0,0 +1,233 @@
...
+    assert_eq!(want_channels, got_channels);
+    assert_eq!(want_filter_types, got_filter_types);
+
+    // Should complete without error if the deserialization handles empty bodies
+    // correctly.
+    client.endpoint().delete(app.id, ep.id).await.unwrap();
+}
+
+#[tokio::test]
+async fn test_default_retries() {
```
svix-onelson, line 130: Fine to put it here, but the kitchen sink stuff was intended to run against a live svix-server.

I wonder if it should be in a different module.

rust/src/api/mod.rs
```diff
@@ -75,6 +75,13 @@ pub struct SvixOptions {
     ///
     /// Default: 15 seconds.
     pub timeout: Option<std::time::Duration>,
+    /// Number of retries
+    ///
+    /// The number of times the client will retry if a server-side error
+    /// or timeout is received.
+    ///
+    /// Default: 2
+    pub num_retries: Option<u32>,
```
svix-onelson, line 84: 2 seems like a small number. Is that consistent with the rest of the libs?

rust/src/api/mod.rs
```diff
@@ -105,6 +113,7 @@ impl Svix {
             // These fields will be set by `with_token` below
             base_path: String::new(),
             bearer_access_token: None,
+            num_retries: options.num_retries.unwrap_or(2),
```
svix-onelson, line 116: Maybe hoist this out as a const? idk

rust/tests/it/kitchen_sink.rs
jaymell: Maybe a less-ambiguous name than kitchen sink. I really didn't know what that meant.

svix-onelson: Patches welcome.

rust/src/api/mod.rs
jaymell: It is consistent with JS, anyway. I have not done a comprehensive check.

svix-onelson (approved):

rust/tests/it/kitchen_sink.rs
jaymell: Or I could just leave it, which is what I'm going to do.

PR #1696, opened 2025-02-04
Some CLI cleanup

svix-jplatte:


svix-cli/src/cmds/api/endpoint.rs
```diff
@@ -126,7 +126,6 @@ pub enum EndpointCommands {
     /// List the application's endpoints.
     List {
         app_id: String,
-
```
svix-james, line 129: But this was my favorite newline...

svix-james (approved):

PR #1700, opened 2025-02-05
Refactor JS lib to use our new codegen

svix-jplatte:
TODO: 
- [x] Export all models from `src/index.ts`
- [x] Decide what we want to do about undefined fields in the response body
	 Currently if the server omits a field from the response body, the json deserialization wont fail. And anyone using the js/ts lib will just get `undefined` if they try to access that field

## Breaking changes
- `endpoint.oauthUpdate` and `endpoint.oauthDelete` are no longer generated
- Deprecated endpoint  `MessageAttempt.list` is removed
- Generated exceptions and models are no longer exported from `webhook.ts`

## New resources
I figured since they are publicly documented in api.svix.com/docs we should add them here.

- Environment
- Health

ref: 


javascript/src/models/app_portal_access_out.ts
```diff
@@ -0,0 +1,22 @@
+// this file is @generated
+
+export interface AppPortalAccessOut {
+  token: string;
+  url: string;
+}
+
+export namespace AppPortalAccessOutUtil {
```
svix-lucho, line 8: maybe call this (and the others) `Serializer` or something like that instead of `Util`?

javascript/src/models/environment_out.ts
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  export function _toJsonObject(self: EnvironmentOut): any {
+    return {
+      createdAt: self.createdAt,
+      eventTypes: self.eventTypes.map((item) => EventTypeOutUtil._toJsonObject(item)),
+      settings: self.settings,
+      transformationTemplates: self.transformationTemplates.map((item) =>
+        TemplateOutUtil._toJsonObject(item)
+      ),
```
svix-lucho, line 35: can be just 

```suggestion
      transformationTemplates: self.transformationTemplates.map(TemplateOutUtil._toJsonObject),
```

and you get rid of the implicit `any` I think

javascript/src/request.ts
```diff
@@ -77,8 +77,8 @@ export class SvixRequest {
     this.headerParams[name] = value;
   }
 
-  public setBody(value: any, type: string) {
-    this.body = JSON.stringify(ObjectSerializer.serialize(value, type, ""));
+  public setBody(value: any) {
+    this.body = JSON.stringify(value);
```
svix-lucho, line 81: Maybe not a big deal, but I feel like this should work like `send`, where you pass the object and the serializer instead of just the object. To avoid accidentally sending the object without first serializing it.

svix-jplatte: Sure, can change it. In case it's not clear, this file can be edited directly in the repo, no codegen involved.

javascript/src/models/environment_out.ts
svix-mman: Maybe later  :)

svix-mman:
## Open source review
Dev dependency

Replace 
- `"@typescript-eslint/eslint-plugin": "^4.15.2"`
- `"@typescript-eslint/parser": "^4.15.2"`
- `"@typescript-eslint/typescript-estree": "^4.15.2"`

With `"typescript-eslint": "^8.24.0"`

svix-mman:
## Open Source review
[eslint-plugin-unused-imports](https://github.com/sweepline/eslint-plugin-unused-imports)

-  MIT license 
- Dev dependency

Automatically removes unused imports

javascript/src/models/app_portal_access_out.ts
svix-mman: fixed

svix-mman:
## Open source review
Dev dependency

https://github.com/httptoolkit/mockttp

-  Apache-2.0 license 

Used to test that out going requests are de/serialized correctly

javascript/src/models/messageEndpointOut.ts
```diff
@@ -0,0 +1,61 @@
...
+  /** List of message channels this endpoint listens to (omit for all). */
+  channels?: string[] | null;
+  createdAt: Date | null;
+  /** An example endpoint name. */
+  description: string;
+  disabled?: boolean;
+  filterTypes?: string[] | null;
+  /** The Endpoint's ID. */
+  id: string;
+  nextAttempt?: Date | null | null;
```
svix-lucho, line 15: This is weird. I see it in some other models. Any chance it becomes a problem?

javascript/src/mockttp.test.ts
```diff
@@ -0,0 +1,292 @@
...
+      .forPost(/\/api\/v1\/app.*/)
+      .thenReply(200, ApplicationOut);
+    const svx = new Svix("token", { serverUrl: mockServer.url });
+
+    await svx.application.create({ name: "test" }, { idempotencyKey: "random-key" });
+    const requests = await endpointMock.getSeenRequests();
+    expect(requests.length).toBe(1);
+    expect(requests[0].headers["idempotency-key"]).toBe("random-key");
+  });
+
```
svix-lucho, line 112: since we have all this nice test setup, can we also add a test to make sure the token and the user agent are sent?

svix-mman: done

javascript/src/models/messageEndpointOut.ts
svix-mman: Indeed this is odd, Fixed.

svix-lucho (approved):

PR #1701, opened 2025-02-05
js: Use UpperCamelCase for generated JS files

svix-jplatte:
Following https://github.com/svix/openapi-codegen/pull/58.


svix-mman:
@svix-jplatte lucho mentioned that js uses camelCase (not UpperCamelCase) for filenames
So the new lib uses camelCase for all file names
Can we close this?

PR #1706, opened 2025-02-09
Replace Go lib with lib generated using openapi-codegen

svix-mman:
Steps:
- [x] Generate high level API client
- [x] Switch to new low-level HTTP client (instead of old openapi-generator)
- [x] Generate models
	- [x] structs
	- [x] enums
	- [x] nullable fields for `*Patch`
- [x] Delete all old openapi-generator code

```json
{
  "file-generated-at": "2025-02-09T19:04:32.725115411+00:00",
  "git-rev": "c15cdd07311368502bd4684637fd20ef69fa8f57",
  "openapi.json-sha256": "27c305c4c890291e9dd2506cebe18422317faea9963f3d9abf89a405731e3206"
}
```

ref:


go/svix_http_client.go
```diff
@@ -0,0 +1,217 @@
...
+
+	v, localErr := serializeQueryParam(val, key)
+	if localErr != nil {
+		*err = localErr
+	} else {
+		d[key] = v
+	}
+}
+
+func serializeQueryParam(val interface{}, key string) (string, error) {
```
svix-onelson, line 171: I wonder if it'd be worth bringing in a lib for this like https://github.com/google/go-querystring

go/svix_test.go
svix-onelson, line 1: The new test cases added here look fair. I'd like to see the new nullable wrapper put through its paces against a live server.

`getTestClient()` can be used for this. That helper will build a client if the token and server url are set in env, or otherwise skip the test. See the "kitchen sink" tests for usage.

go/utils/nullable.go
```diff
@@ -0,0 +1,69 @@
...
+	return v.isSet
+}
+
+func (v *Nullable[T]) Unset() {
+	v.val = nil
+	v.isSet = false
+}
+
+// Set the nullable value to nil, the when json serializing a null will be emitted
+func (v *Nullable[T]) SetNil() {
```
svix-onelson, line 43: I think the API for `Nullable` seems fair. Is there a reason  we need `SetNil` instead of just passing a nil to `Set`?

svix-onelson:
I need to look over some of the http client impl a little more closely, but hopefully I've left some actionable notes in the meantime.

svix-onelson:
OSS Review (httpmock): approved ✅

go/application.go
```diff
@@ -1,124 +1,214 @@
...
+	}
+	headerMap := map[string]string{}
+	var jsonBody []byte
+
+	if o != nil {
+		var err error
+		SerializeParamToMap("idempotency-key", o.IdempotencyKey, headerMap, &err)
+		if err != nil {
+			return nil, err
+		}
```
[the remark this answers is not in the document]
svix-mman: I did not mean to make `SerializeParamToMap` public, it is now private.

I wrote it like this for cases where I need to add a lot of params, And since each param can fail, I am able to write something like this without checking the error on each call to `serializeParamToMap`

```go
// in Message.List
if o != nil {
	serializeParamToMap("limit", o.Limit, queryMap, &err)
	serializeParamToMap("iterator", o.Iterator, queryMap, &err)
	serializeParamToMap("channel", o.Channel, queryMap, &err)
	serializeParamToMap("before", o.Before, queryMap, &err)
	serializeParamToMap("after", o.After, queryMap, &err)
	serializeParamToMap("with_content", o.WithContent, queryMap, &err)
	serializeParamToMap("tag", o.Tag, queryMap, &err)
	serializeParamToMap("event_types", o.EventTypes, queryMap, &err)
	if err != nil {
		return nil, err
	}
}
```
And `serializeParamToMap` checks if the error was set by the last invocation of `serializeParamToMap` before mutating the `queryMap`/`headerMap`
```go
func serializeParamToMap(key string, val interface{}, d map[string]string, err *error) {
	// I pass the error in here so I don't have to "if err != nil" for every query param
	if *err != nil {
		return
	}
	...
}
```

go/application.go
```diff
@@ -1,124 +1,214 @@
...
+		SerializeParamToMap("order", o.Order, queryMap, &err)
+		if err != nil {
+			return nil, err
 		}
 	}
-
-	ret, res, err := req.Execute()
-	if err != nil {
-		return nil, wrapError(err, res)
+	ret, apiErr := executeRequest[models.ListResponseApplicationOut](
```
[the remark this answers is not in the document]
svix-mman: `DELETE` requests are that wrinkle, `Delete` functions return `error` and `executeRequest` returns `(*T, error)`.
I can't return `executeRequest` from `Delete`
But I updated the template to generate nicer `err` code :)

ChangeLog.md
```diff
@@ -1,6 +1,13 @@
 # Changelog
 
 ## Unreleased
+* Libs/Go **(Breaking)**: Excluding specific fields on the *Patch models (`ApplicationPatch` for example), all `Nullable{Type}` removed from the models
+* Libs/Go **(Breaking)**: All `Nullable{Type}` (for example `NullableString`) are replaced with a new generic `Nullable[T]` type, the new type can be imported from `github.com/svix/svix-webhooks/go/utils`
+* Libs/Go **(Breaking)**: All custom model types are now imported from `github.com/svix/svix-webhooks/go/models` instead of `github.com/svix/svix-webhooks/go`
+* Libs/Go **(Breaking)**: Some all methods that previously had a `{Method}WithOptions` counterpart, Now take in an extra argument. These new arguments can be set to `nil` if unneeded
+* Libs/Go **(Breaking)**: All `*WithOptions` functions are now removed in replaced with their regular counterparts. For example in `Endpoint.RecoverWithOptions` is now `Endpoint.Recover`
+* Libs/Go **(Breaking)**: All uses of `PostOptions` are removed, and renamed to `{Resource}{Operation}Options`. For example in `Endpoint.RecoverWithOptions` you would now use `Endpoint.Recover` and pass in the `EndpointListOptions`
+
```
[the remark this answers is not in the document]
svix-mman: I think you mean I merge the C#/Go note about `{Resource}{Operation}Options`?
I did that. let me know if I misunderstood

go/application.go
```diff
@@ -1,124 +1,214 @@
...
+	if o != nil {
+		var err error
+		SerializeParamToMap("idempotency-key", o.IdempotencyKey, headerMap, &err)
+		if err != nil {
+			return nil, err
 		}
 	}
-
-	ret, res, err := req.Execute()
+	jsonBody, err := json.Marshal(applicationIn)
```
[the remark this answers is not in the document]
svix-mman: Indeed, this does not need to be part of each request
Fixed.

go/application.go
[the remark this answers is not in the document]
svix-mman: After some template elbow grease, the generated code will simply return `executeRequest` if the response body is defined

go/models/app_usage_stats_in.go
```diff
@@ -0,0 +1,13 @@
...
+
+import "time"
+
+type AppUsageStatsIn struct {
+	// Specific app IDs or UIDs to aggregate stats for.
+	//
+	// Note that if none of the given IDs or UIDs are resolved, a 422 response will be given.
+	AppIds []string  `json:"appIds,omitempty"`
+	Since  time.Time `json:"since"`
+	Until  time.Time `json:"until"`
```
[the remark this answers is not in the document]
svix-mman: Good call, I added some tests for de/serialization of `time.Time`
Go's documentation states the to `UnmarshalJSON` `time.Time` "The time must be a quoted string in the RFC 3339 format."

go/models/background_task_type.go
```diff
@@ -0,0 +1,43 @@
...
+const (
+	BACKGROUNDTASKTYPE_ENDPOINT_REPLAY      BackgroundTaskType = "endpoint.replay"
+	BACKGROUNDTASKTYPE_ENDPOINT_RECOVER     BackgroundTaskType = "endpoint.recover"
+	BACKGROUNDTASKTYPE_APPLICATION_STATS    BackgroundTaskType = "application.stats"
+	BACKGROUNDTASKTYPE_MESSAGE_BROADCAST    BackgroundTaskType = "message.broadcast"
+	BACKGROUNDTASKTYPE_SDK_GENERATE         BackgroundTaskType = "sdk.generate"
+	BACKGROUNDTASKTYPE_EVENT_TYPE_AGGREGATE BackgroundTaskType = "event-type.aggregate"
+)
+
+var AllowedBackgroundTaskType = []BackgroundTaskType{
```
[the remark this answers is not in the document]
svix-mman: Nope, this does not need to be public.
Fixed

go/models/application_out.go
```diff
@@ -0,0 +1,16 @@
...
+	CreatedAt time.Time `json:"createdAt"`
+	// The app's ID
+	Id        string            `json:"id"`
+	Metadata  map[string]string `json:"metadata"`
+	Name      string            `json:"name"`
+	RateLimit *uint16           `json:"rateLimit,omitempty"`
+	// The app's UID
+	Uid       *string   `json:"uid,omitempty"`
+	UpdatedAt time.Time `json:"updatedAt"`
+}
```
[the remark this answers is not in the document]
svix-mman: For comments that don't have a newline, I now place them after the field.
(It looks good on my IDE, github destroyed the alignment)
```go
type ApplicationOut struct {
	CreatedAt time.Time         `json:"createdAt"`
	Id        string            `json:"id"` // The app's ID
	Metadata  map[string]string `json:"metadata"`
	Name      string            `json:"name"`
	RateLimit *uint16           `json:"rateLimit,omitempty"`
	Uid       *string           `json:"uid,omitempty"` // The app's UID
	UpdatedAt time.Time         `json:"updatedAt"`
}
```

image

go/svix_http_client.go
svix-mman: The problem is that the `*Options` struct can contains header params as well
So we can't just pass the whole struct into `query.Values(*Options).Encode()`

svix-mman: And this should technically be called `serializeQueryOrHeaderParam`
(Renamed it)

go/utils/nullable.go
svix-mman: Nope, this was me coping the methods from the4 old none generic `Nullable{Type}` types, Fixed.

I also removed `NewExplicitlySetNilNullable` now you can just call `NewNullableFromPtr` to create a `Nullable[string]` that is set to nil

go/application.go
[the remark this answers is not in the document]
svix-onelson: Another way might be to do a loop over the key/value pairs and bail on the earliest err.

Anyway, like I said, if this is private we'll have room to change this as we see fit. I think it's fine for now, but it'd be good to do another pass on this (especially in the case where we might leverage a lib for the query string).

go/models/app_usage_stats_in.go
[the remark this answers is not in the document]
svix-onelson: The nice thing about having a test is two years from now when there's some new fancy datetime lib that someone brings in, we'll know the behavior is consistent. 👍🏻

go/svix_http_client.go
```diff
@@ -0,0 +1,221 @@
...
+	// If val is null don't add it to the query params map
+	if val == nil || (reflect.ValueOf(val).Kind() == reflect.Ptr && reflect.ValueOf(val).IsNil()) {
+		return
+	}
+
+	v, localErr := serializeQueryOrHeaderParam(val, key)
+	if localErr != nil {
+		*err = localErr
+	} else {
+		d[key] = v
```
svix-onelson, line 171: Should this be percent-encoded? I'm not sure if there's a difference in how headers and query params are handled, but I thought query params _should be_. 

Perhaps this is handled by the client or somewhere else? 

> Edit: I spotted the `query.Add` call above after leaving _this comment_.

go/svix_http_client.go
```diff
@@ -0,0 +1,221 @@
...
+	query := parsedURL.Query()
+	for key, value := range params {
+		query.Add(key, value)
+	}
+
+	parsedURL.RawQuery = query.Encode()
+	return parsedURL.String(), nil
+}
+
+func serializeParamToMap(key string, val interface{}, d map[string]string, err *error) {
```
svix-onelson, line 157: Wondering about the choice of giving `d` a `map[string]string` type. Since headers and query params can be repeated, isn't this a lossy operation?

Does the client lib expect a `map[string][string]`, forcing the choice?

Probably it doesn't matter much since the user input is coming to us via fields on a struct, therefore the keys are known and distinct. If later we changed the representation in the public API, the `map` storage could end up quietly dropping/replacing keys 🤔

go/svix_http_client.go
```diff
@@ -0,0 +1,221 @@
...
+
+func addQueryParams(baseURL string, params map[string]string) (string, error) {
+	parsedURL, err := url.Parse(baseURL)
+	if err != nil {
+		return "", err
+	}
+
+	query := parsedURL.Query()
+	for key, value := range params {
+		query.Add(key, value)
```
svix-onelson, line 150: Maybe this one is the call that does percent encoding? Worth checking.

go/svix_http_client.go
```diff
@@ -0,0 +1,221 @@
...
+	urlWithPath := client.BaseURL + replacePathKeys(path, pathParams)
+	urlStr, err := addQueryParams(urlWithPath, queryParams)
+	if err != nil {
+		return nil, err
+	}
+	encodedBody, err := json.Marshal(jsonBody)
+	if err != nil {
+		return nil, err
+	}
+	req, err := http.NewRequestWithContext(ctx, method, urlStr, bytes.NewBuffer(encodedBody))
```
svix-onelson, line 59: Wanted to mention passing `Context` all the way down to this call here is what gives the caller the ability to set a deadline for the request.

Double check that it's actually possible to feed a context in from the public API and have it passed along all the way down here.

(I didn't see anything to indicate this wasn't already happening, but wanted to mention it in case you hadn't been thinking about this use case)

go/svix_test.go
svix-onelson: I think `EndpointPatch` is a better candidate for testing this stuff since it has nullable arrays (which application doesn't, right?)

go/svix_http_client.go
svix-mman: For query params, we don't allow repeating params
We comma separate any List[T] of params

For headers, I don't expect any request specific headers to repeat.

go/svix_test.go
svix-onelson: Actually, the kitchen sink test is pretty close to what we want if we ignore the fact it's sort of a gross pile of assertions bundled together out of convenience.

IIRC the gist was to create an app, some event types, some endpoints then set channels/event_types using `PUT` and `PATCH`.

Adding something to show unsetting either/or via `PATCH` and asserting the other one is left as-is would give is the last piece for nullable arrays.

Remember that nullable arrays are _awkward_ because sending an empty one results in an error on the server, so we need to prove that the `null` is what gets sent, not `[]`.

go/svix_http_client.go
svix-mman: Later down in the same function we percent encode
`parsedURL.RawQuery = query.Encode()`

ChangeLog.md
```diff
@@ -1,6 +1,13 @@
 # Changelog
 
 ## Unreleased
+* Libs/Go **(Breaking)**: Excluding specific fields on the *Patch models (`ApplicationPatch` for example), all `Nullable{Type}` removed from the models
+* Libs/Go **(Breaking)**: All `Nullable{Type}` (for example `NullableString`) are replaced with a new generic `Nullable[T]` type, the new type can be imported from `github.com/svix/svix-webhooks/go/utils`
+* Libs/Go **(Breaking)**: All custom model types are now imported from `github.com/svix/svix-webhooks/go/models` instead of `github.com/svix/svix-webhooks/go`
+* Libs/Go **(Breaking)**: Some all methods that previously had a `{Method}WithOptions` counterpart, Now take in an extra argument. These new arguments can be set to `nil` if unneeded
+* Libs/Go **(Breaking)**: All `*WithOptions` functions are now removed in replaced with their regular counterparts. For example in `Endpoint.RecoverWithOptions` is now `Endpoint.Recover`
```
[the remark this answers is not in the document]
svix-onelson:
```
* Libs/Go **(Breaking)**: All methods that previously had a `{Method}WithOptions` counterpart, Now take in an extra argument. These new arguments can be set to `nil` if unneeded
* Libs/Go **(Breaking)**: All `*WithOptions` functions are now removed in replaced with their regular counterparts. For example in `Endpoint.RecoverWithOptions` is now `Endpoint.Recover`
```

These two still look like they are describing the same breaking change. One says _functions_ and the other says _methods_, but they're talking about the same thing, right?

go/application.go
```diff
@@ -1,156 +1,208 @@
...
 
 type Application struct {
-	api *openapi.APIClient
+	client *SvixHttpClient
 }
 
 type ApplicationListOptions struct {
 	// Limit the number of returned items
-	Limit *int32
+	Limit *uint64
```
svix-onelson, line 16: Raising this to u64 seems like an odd choice, just because we expect these to be maybe triple digits, if that, per validation rules.

I guess you just care about removing the negative range? Zero is still an odd case. I probably would have just left this as int32. 🤷🏻

ChangeLog.md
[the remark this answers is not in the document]
svix-mman: Yup talking about the same thing

go/application.go
svix-mman: This is not raising to u64 per-say, It's just that this is the type on the `openapi.json`

go/svix_test.go
svix-mman: check out this new tests
[`1a95f3c` (#1706)](https://github.com/svix/svix-webhooks/pull/1706/commits/1a95f3cadceb285ef5dec86466003baa786cbd6b)
What do you think?

ChangeLog.md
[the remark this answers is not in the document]
svix-mman: Corrected

go/svix_test.go
```diff
@@ -488,3 +493,106 @@ func TestNullableAgainstServer(t *testing.T) {
...
+		},
+	)
+	patch := models.EndpointPatch{
+		Uid: utils.NewNullable("here so a body is sent"),
+		// this will not be sent to server
+		Channels: utils.NewUnsetNullable[[]string](),
+	}
+
+	_, err := svx.Endpoint.Patch(ctx, "app1", "endp1", patch)
+	assertExpectedError(t, err, "unexpected end of JSON input")
```
svix-onelson, line 597: I get what's going on here, but it felt a bit confusing when I first saw assertions about errors to verify the impl passed the test here 🙃

I wonder if there's anything we can do to make this feel less backwards. One option would be to have a hardcoded response that is a full `EndpointOut` for the OK. A bit annoying, maybe, but you can use the out model to keep things all aligned as the API evolves.

go/svix_test.go
svix-onelson: Left a note about the assertions in some spots but these test cases look good to me.

go/svix_test.go
svix-mman: fixed.

svix-onelson (approved):

PR #1708, opened 2025-02-11
Replace Kotlin lib with lib generated using openapi-codegen

svix-mman:
TODO:
- [x] Generate high level API client
- [x] Switch to new low-level HTTP client (instead of old openapi-generator)
- [x] Generate models
	- [x] structs
	- [x] enums
	- [x] nullable fields for `*Patch`
- [x] Delete all old openapi-generator code
- [x] Delete debug statements
- [x] Document breaking changes

Test coverage
- [x] Model de/serialization tests
    - [x] Unset/Null/Some
    - [x] DateTime RFC3339 and timezone
- [x] Retry logic
- [x] Headers: User-Agent and req-id / retry-count
- [x] `{Operation}{Method}Options` serialization tests
    - [x] Headers
    - [x] Query params
        - [x] Enums
        - [x] kotlinx.datetime.Instant
        - [x] List
        - [x] Bool

## Change log
* Libs/Kotlin **(VERY IMPORTANT)**: The parameter order `appId` and `msgId` were swapped on `Message.get` and `Message.expungeContent`
* Libs/Kotlin **(Breaking)**: All uses of `ListOptions`/`PostOptions` are removed, and renamed to `{Resource}{Operation}Options`. For example in `Endpoint.List` you would now use `EndpointListOptions`
* Libs/Kotlin **(Breaking)**: In the 4 `*Patch` patch models, nullable fields are of type `MaybeUnset<T>` instead of `T`. call `MaybeUnset.Present(val)` to initialize this value
* Libs/Kotlin **(Breaking)**: `SvixOptions` no longer has `initialRetryDelayMillis` or `numRetries` instead use `retrySchedule`
* Libs/Kotlin **(Breaking)**: All `{Resource}{Operation}Options` and model classes (`ApplicationIn`/`MessageOut`) are now data classes
* Libs/Kotlin **(Breaking)**: Deprecated functions `MessageAttempt.list` and `MessageAttempt.listAttemptsForEndpoint` are removed
* Libs/Kotlin **(Breaking)**: All uses of `java.time.OffsetDateTime` replaced with `kotlinx.datetime.Instant`
* Libs/Kotlin **(Breaking)**: All uses of `java.net.URL` in request/response models are replaced with `String`

```json
{
  "file-generated-at": "2025-02-19T21:44:05.890303513+00:00",
  "git-rev": "98dbc5b090a5c8d72fe50962ee04b46fb9d7db20",
  "openapi.json-sha256": "33fb5ed57fd9afa4c737cebd0f66244436813844fb00e2a057b06525819af8d7"
}
```

ref:


svix-mman:
## Open source review
[wiremock](https://github.com/wiremock/wiremock)
- Apache-2.0 license

Allows me to easily test the API client

ChangeLog.md
```diff
@@ -1,6 +1,10 @@
 # Changelog
 
 ## Unreleased
+* Libs/Kotlin **(Breaking)**  Deprecated functions `MessageAttempt.list` and `MessageAttempt.listAttemptsForEndpoint` is removed
```
tasn, line 4: Are removed*

ChangeLog.md
```diff
@@ -1,6 +1,10 @@
...
+* Libs/Kotlin **(Breaking)**: All uses of `java.time.OffsetDateTime` replaced with `kotlinx.datetime.Instant`
+* Libs/Kotlin **(Breaking)**: All uses of `java.net.URL` in request/response models are replaced with `String`
```
tasn, line 6: Is this a good idea? I guess it's what we do elsewhere and it's just endpoint url, right?

svix-mman: Yup, We don't use this url for anything client side

svix-james (approved):

ChangeLog.md
```diff
@@ -1,12 +1,19 @@
 # Changelog
 
 ## Unreleased
+* Libs/C#, Libs/Go and Libs/Kotlin **(Breaking)**: All uses of `ListOptions`/`PostOptions` are removed, and renamed to `{Resource}{Operation}Options`. For example in `Endpoint.List` you would now use `EndpointListOptions`
+* Libs/Kotlin **(Breaking)**: `SvixOptions` no longer has `initialRetryDelayMillis` or `numRetries` instead use `retrySchedule`
+* Libs/Kotlin: Fix the parameter names of `Endpoint.get` - `appId` and `endpointId` were swapped
```
tasn, line 6: This one is potentially scary, it's not an innocent break as this will fail for people at *runtime* is there a way to keep the old one but mark it as deprecated and add a new one instead that's fixed?

ChangeLog.md
```diff
@@ -1,12 +1,19 @@
...
+* Libs/Kotlin **(Breaking)**: All `{Resource}{Operation}Options` and model classes (`ApplicationIn`/`MessageOut`) are now data classes
+* Libs/Kotlin **(Breaking)**: Deprecated functions `MessageAttempt.list` and `MessageAttempt.listAttemptsForEndpoint` are removed
+* Libs/Kotlin **(Breaking)**: All uses of `java.time.OffsetDateTime` replaced with `kotlinx.datetime.Instant`
+* Libs/Kotlin **(Breaking)**: All uses of `java.net.URL` in request/response models are replaced with `String`
+* Libs/Kotlin **(Breaking)**: All uses of `Map<String,Any>` in request/response models are replaced with `kotlinx.serialization.json.JsonObject`
```
tasn, line 11: I could have sworn I had a comment for this already. This feels wrong to me, we shouldn't be exposing the fact that it's a json object internally (I'm not even sure how common this is). Where is it used exactly?

ChangeLog.md
svix-mman: Now that I look at this there are 2 more swaps that I forgot to document (Added them)
I am not sure how I want to approach this, I understand this will cause many systems to panic at runtime

ChangeLog.md
tasn: This leaks an internal implementation detail though, and it's not really the types that people expect in an SDK.

ChangeLog.md
tasn: Not panic, but fail silently at runtime (well 404s), which is terrible. Can we avoid breaking it please?

.github/workflows/kotlin-lint.yml
```diff
@@ -21,12 +21,7 @@ jobs:
...
-      - name: Regen openapi libs
-        run: |
-          yarn
-          ./regen_openapi.sh
-
       - name: Build
         run: |
           cd kotlin
-          ./gradlew build -x test
+          ./gradlew build
```
tasn, line 27: This no longer runs the tests?

svix-mman: This no longer disables the tests
`-x, --exclude-task                 Specify a task to be excluded from execution.
`

kotlin/lib/src/main/kotlin/Application.kt
```diff
@@ -1,106 +1,103 @@
...
-    fun limit(limit: Int) = apply { this.limit = limit }
+data class ApplicationCreateOptions(val idempotencyKey: String? = null)
 
-    /** The iterator returned from a prior invocation */
-    fun iterator(iterator: String) = apply { this.iterator = iterator }
-
-    /** The sorting order of the returned items */
-    fun order(order: Ordering) = apply { this.order = order }
-}
-
```
tasn, line 26: We forgot about docs?

kotlin/lib/src/main/kotlin/Integration.kt
```diff
@@ -75,46 +63,46 @@ class Integration internal constructor(token: String, options: SvixOptions) {
...
+        client.executeRequest<Any, Boolean>("DELETE", url.build())
     }
 
     /**
      * Get an integration's key.
      *
      * @deprecated
      */
-    @Deprecated(message = "This endpoint is deprecated.")
+    @Deprecated("")
```
tasn, line 86: Is it intentional that we are not using a deprecation message?

kotlin/lib/src/main/kotlin/Application.kt
svix-mman: Yup :sweat_smile:  fixed.

.github/workflows/kotlin-lint.yml
tasn: ah lol.

PR #1710, opened 2025-02-12
Add User-Agent to C# lib

svix-mman:
The version number is defined in `csharp/Svix/Svix.csproj`. And we are able to access it at runtime

Also change the `svix-req-id` header from a u32 to a u64 (because why not)

### DIFF
--- a/csharp/Svix.Tests/WiremockTests.cs
+++ b/csharp/Svix.Tests/WiremockTests.cs
@@ -149,6 +149,22 @@ namespace Svix.Tests
         }
 
         [Fact]
+        public void UserAgentIsSent()
+        {
+            stub.Given(Request.Create().WithPath("/api/v1/app/*"))
+                .RespondWith(Response.Create().WithStatusCode(200).WithBody(applicationOutJsonStr));
+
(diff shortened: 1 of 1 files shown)


csharp/Svix/SvixHttpClient.cs
```diff
@@ -170,6 +173,8 @@ HttpRequestMessage BuildRequest(
 
             request.Headers.Add("Authorization", $"Bearer {_token}");
             request.Headers.Add("svix-req-id", req_id.ToString());
+            // For some reason "svix-libs/1.56.0/csharp" is not a valid user-agent :|
```
tasn, line 176: What's this note? Should it be removed?

csharp/Svix/SvixHttpClient.cs
```diff
@@ -170,6 +173,10 @@ HttpRequestMessage BuildRequest(
 
             request.Headers.Add("Authorization", $"Bearer {_token}");
             request.Headers.Add("svix-req-id", req_id.ToString());
+
+
```
tasn, line 177:
```suggestion


```
No need for all of this extra space.

tasn (approved):

PR #1716, opened 2025-02-13
Fix docker tags

jaymell:
We don't trigger this job on release so need a different way of getting the tags. This should work.


svix-onelson (approved):
👌🏻

Issue #1721, opened 2025-02-15
[rust]  ListResponseXXXOut.iterator should be optional


azhur:
### Version

1.57.0

### Platform

macosx

### Description

Rust `ListResponseXXXOut.iterator` is now String but should be Option<String> instead?

tasn:
Thanks @azhur for the report, also looping in @svix-mman.
Yes, it should 100% be `Option<String>` as it's nullable.

svix-mman:
@azhur Thanks again for reporting this issue.
This should be fixed in the new version [1.58.0](https://crates.io/crates/svix/1.58.0).

PR #1722, opened 2025-02-16
rust: Fix ListResponseXXXOut models

svix-mman:
ref: https://github.com/svix/svix-webhooks/issues/1721

- Pull in new updates from `lib-openapi.json`
    - Doc comments
    - New models

Breaking changes: 
	- For all `ListResponse*Out` structs, the `iterator` field is now `Option<String>,` instead of `String`
	- `EnvironmentOut.settings` is now a `Option<serde_json::Value>` instead of `serde_json::Value`


tasn (approved):

tasn:
Technically not a breaking change, but an unbreaking change.

PR #1723, opened 2025-02-16
Update go lib with some new fixes

svix-mman:
- Pull in new updates from `lib-openapi.json`
    - Doc comments
    - New models
- Correctly mark nullable types as `*T`

Breaking changes: 
	- For all `ListResponse*Out` models, the `Iterator` is now `*string` instead of `string`
	- `EnvironmentOut.Settings` is now a `*map[string]any` instead of `map[string]any

Some breaking changes on environment related models, But this is fine since we don't have the environment as part of our Go lib yet

```json
{
  "file-generated-at": "2025-02-19T20:45:38.376408871+00:00",
  "git-rev": "98dbc5b090a5c8d72fe50962ee04b46fb9d7db20",
  "openapi.json-sha256": "33fb5ed57fd9afa4c737cebd0f66244436813844fb00e2a057b06525819af8d7"
}
```


go/models/list_response_application_out.go
```diff
@@ -4,6 +4,6 @@ package models
 type ListResponseApplicationOut struct {
 	Data         []ApplicationOut `json:"data"`
 	Done         bool             `json:"done"`
-	Iterator     string           `json:"iterator"`
+	Iterator     *string          `json:"iterator,omitempty"`
```
tasn, line 7: Why is it sometimes `Nullable` and sometimes *?

svix-mman: For now we are only using `Nullable` on user created types (`EndpointPatch`)
But this seems like this could become a bug
Let me get back with a fix

svix-gabriel (approved):

tasn (approved):

PR #1729, opened 2025-02-18
ci: Update to ubuntu-22.04

svix-mman:
ref: https://github.com/actions/runner-images/issues/11101
> The Ubuntu 20.04 Actions runner image will begin deprecation on 2025-02-01 and will be fully unsupported by 2025-04-01

### DIFF
--- a/dist-workspace.toml
+++ b/dist-workspace.toml
@@ -24,3 +24,6 @@ tap = "svix/homebrew-svix"
 # Publish jobs to run in CI
 publish-jobs = ["homebrew"]
 
+# FIXME: for now override the runner in ci
+# remove when cargo-dist updates the runner to newer version
+allow-dirty = ["ci"]
--- a/dist-workspace.toml
+++ b/dist-workspace.toml
@@ -24,6 +24,8 @@ tap = "svix/homebrew-svix"
 # Publish jobs to run in CI
 publish-jobs = ["homebrew"]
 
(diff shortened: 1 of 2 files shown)


svix-james (dismissed):

dist-workspace.toml
```diff
@@ -7,6 +7,7 @@ members = ["cargo:svix-cli"]
 cargo-dist-version = "0.28.0"
 # CI backends to support
 ci = "github"
+allow-dirty = ["ci"]
```
svix-james, line 10: Sounds gross. What does this actually do

svix-mman: Allows me to override the runner manually, Not sure how else to do this

svix-james: Sorry I'm ignorant here and mostly prefer to remain so. What does "override the runner manually" mean?

svix-mman: nvm: I found A fix for this that does not include allowing dirty
> What does "override the runner manually" mean?
The release.yml file is generated by [cargo-dist](https://github.com/axodotdev/cargo-dist), and they (cargo-dist) use the old ubuntu-20.04.
When I tried to manually edited the file, dist was complaining (it was expected 20.04), so allow-dirty made dist not check the runners

But all of this does not matter now, since I found a way to fix this

svix-james (approved):

PR #1731, opened 2025-02-18
go: Fix null ptr derf

svix-mman:


go/svix.go
```diff
@@ -30,20 +30,22 @@ type (
...
+			}
+		} else {
+
+			svixHttpClient.BaseURL = options.ServerUrl.String()
 		}
-	} else {
-		svixHttpClient.BaseURL = options.ServerUrl.String()
 	}
-
 	if options != nil {
```
tasn, line 49: You're checking for options != nil just below.

svix-mman: Let me clean this up a bit

tasn (approved):

PR #1733, opened 2025-02-19
Add rust lib regeneration to regen_openapi.sh

svix-jplatte:
I'm not back from vacation yet, but I had a free minute and wanted to get this branch that I had ready locally into a PR.

The first commit includes a typo that's carried over from the OpenAPI spec: `ExpungAllContentsOut` should be `ExpungeAllContentsOut`. Should be fixed separately.

### DIFF
--- a/.github/workflows/cli-lint.yml
+++ b/.github/workflows/cli-lint.yml
@@ -49,6 +49,11 @@ jobs:
     steps:
     - uses: actions/checkout@v4
 
+    # the codegen uses `rustfmt +nightly`
+    - uses: dtolnay/rust-toolchain@nightly
+      with:
+        components: rustfmt
+
     - name: Regen openapi libs
       run: |
         yarn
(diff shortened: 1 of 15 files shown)


svix-jplatte:
I don't get what's wrong in CI. The script generates a usable Rust lib locally..

svix-mman:
The problem is that the codegen is trying to [run](https://github.com/svix/openapi-codegen/blob/98dbc5b090a5c8d72fe50962ee04b46fb9d7db20/src/postprocessing.rs#L53-L56) `rustfmt +nightly` and is [failing](https://github.com/svix/svix-webhooks/actions/runs/13409570820/job/37456384707?pr=1733#step:3:225).

svix-mman (dismissed):

svix-jplatte:
Needs another approval from somebody else because the approval is from the same person who also pushed most recently, so does not count.

svix-lucho (approved):

svix-mman (approved):

PR #1736, opened 2025-02-20
Bump to version 1.59.0

svix-mman:
## Version 1.59.0

* Libs/Kotlin **(VERY IMPORTANT)**: The parameter order `appId` and `msgId` were swapped on `Message.get` and `Message.expungeContent`
* Libs/Kotlin **(Breaking)**: All uses of `ListOptions`/`PostOptions` are removed, and renamed to `{Resource}{Operation}Options`. For example in `Endpoint.List` you would now use `EndpointListOptions`
* Libs/Kotlin **(Breaking)**: In the 4 `*Patch` patch models, nullable fields are of type `MaybeUnset<T>` instead of `T`. call `MaybeUnset.Present(val)` to initialize this value
* Libs/Kotlin **(Breaking)**: `SvixOptions` no longer has `initialRetryDelayMillis` or `numRetries` instead use `retrySchedule`
* Libs/Kotlin **(Breaking)**: All `{Resource}{Operation}Options` and model classes (`ApplicationIn`/`MessageOut`) are now data classes
* Libs/Kotlin **(Breaking)**: Deprecated functions `MessageAttempt.list` and `MessageAttempt.listAttemptsForEndpoint` are removed
* Libs/Kotlin **(Breaking)**: All uses of `java.time.OffsetDateTime` replaced with `kotlinx.datetime.Instant`
* Libs/Kotlin **(Breaking)**: All uses of `java.net.URL` in request/response models are replaced with `String`
* Libs/JavaScript **(Breaking)**: Removed non-public `endpoint.oauthUpdate` and `endpoint.oauthDelete` operations.
* Libs/JavaScript **(Breaking)**: Removed deprecated operation `MessageAttempt.list`
* Libs/JavaScript **(Breaking)**: Exceptions and models are no longer exported from `webhook.ts`, import them from `index.ts` instead


ChangeLog.md
```diff
@@ -9,6 +9,9 @@
 * Libs/Kotlin **(Breaking)**: Deprecated functions `MessageAttempt.list` and `MessageAttempt.listAttemptsForEndpoint` are removed
 * Libs/Kotlin **(Breaking)**: All uses of `java.time.OffsetDateTime` replaced with `kotlinx.datetime.Instant`
 * Libs/Kotlin **(Breaking)**: All uses of `java.net.URL` in request/response models are replaced with `String`
+* Libs/JavaScript **(Breaking)**: Removed none-public `endpoint.oauthUpdate` and `endpoint.oauthDelete` operations.
```
svix-lucho, line 12:
```suggestion
* Libs/JavaScript **(Breaking)**: Removed non-public `endpoint.oauthUpdate` and `endpoint.oauthDelete` operations.
```

svix-mman: whops, fixed.

PR #1748, opened 2025-02-24
ci: Add workflow to check for uncommitted codegen changes

svix-jplatte:
Tested successfully: https://github.com/svix/svix-webhooks/actions/runs/13500955055/job/37718984950

### DIFF
--- a/.github/workflows/codegen.yml
+++ b/.github/workflows/codegen.yml
@@ -1,0 +1,57 @@
+name: Codegen
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+# When pushing to a PR, cancel any jobs still running for the previous head commit of the PR
+concurrency:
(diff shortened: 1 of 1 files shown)


svix-mman:
@svix-jplatte 
Is the plan here to eventually check for uncommitted changes on all of the languages. Or is this workflow only for rust?

svix-mman:
Is the long term plan is not to track the generated libs in git, Or we will regen in CI so we can verify that committed libs are up to date?

svix-jplatte:
The long term plan is to continue tracking the generated libs in git, and have CI verify that any changes to the spec or templates are reflected in the libs / no manual changes were made to libs that the script would overwrite again. This PR does adds that CI checks, though it only affects Rust for now because that's the only lang the regen script affects that has its source committed. Makes sense?

svix-mman (approved):

PR #1749, opened 2025-02-25
Fix bug in bump_version.js script

svix-mman:
The script reads the `.version` file to get the `<old_version>` It then uses the `<new_version>` from `argv[2]` to replace `<old_version>` with `<new_version>` for all of the files in `filesPaths`

There is a bug if for whatever reason the `<old_version>` in the target file does not match the `<old_version>` in `.version`

I added a quick check to see if the file was modified, if it was not exit with status 1

### DIFF
--- a/tools/bump_version.js
+++ b/tools/bump_version.js
@@ -36,8 +36,6 @@ const filesPaths = [
     // Ruby
     "ruby/Gemfile.lock",
     "ruby/lib/svix/version.rb",
-    // OpenAPI spec
-    "server/openapi.json",
     // Cloud OpenAPI spec - not necessary but any other time of updating seems weirder
     "lib-openapi.json",
 ];
@@ -63,6 +61,10 @@ filesPaths.forEach((relativePath) => {
     const filePath = join(rootDir, relativePath);
     const content = readFileSync(filePath, 'utf8');
     const updated_content = content.replace(replaceRegExp, newVersion);
+    if (content === updated_content) {
+        console.error("New version was not written to " + filePath)
+        process.exit(1);
+    }
     writeFileSync(filePath, updated_content);
 })
 


server/openapi.json
```diff
@@ -2158,7 +2158,7 @@
     "info": {
         "description": "",
         "title": "Svix API",
-        "version": "1.1.1",
+        "version": "1.59.2",
```
tasn, line 2161: This is on purpose.

tasn (changes requested):
It's not a bug, it's on purpose. Unless you have a reason why you want to change that?

svix-jplatte (approved):
Nice. I don't remember why we use `1.1.1` for `server/openapi.json` but I guess we can always add it back to the list in `bump_version.js` if we decide we want to put the regular version in that spec file.

tools/bump_version.js
```diff
@@ -36,8 +36,6 @@ const filesPaths = [
     // Ruby
     "ruby/Gemfile.lock",
     "ruby/lib/svix/version.rb",
-    // OpenAPI spec
-    "server/openapi.json",
```
tasn, line 40: This should be kept there, I was just talking about the script. Sorry.

svix-mman: If we keep it here, then it means we are going to update the version field in the `server/openapi.json`
But the `./generate-openapi.sh` will simply override the new version with `1.1.1`, as well as fail in CI (`./generate-openapi.sh --check`)

tasn: Yeah, the 1.1.1 is just for the SDK generation anyway.

tasn: Updating the spec is required for CI I believe so we want it there.

PR #1755, opened 2025-02-27
Remove Ruby codegen from regen_openapi.sh

svix-jplatte:
It is soon going to be replaced (#1753), and fails on the latest version of our OpenAPI specification.


svix-jplatte:
Can't do this because we don't commit the generated source, duh :facepalm:

PR #1756, opened 2025-02-27
python: Add new codegen templates

svix-jplatte:
Imported from openapi-codegen repo.

### DIFF
--- a/python/svix/api/message.py
+++ b/python/svix/api/message.py
@@ -3,7 +3,7 @@ import typing as t
 from dataclasses import dataclass
 from datetime import datetime
 
-from ..models import ListResponseMessageOut, MessageIn, MessageOut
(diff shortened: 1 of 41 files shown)


python/svix/api/message.py
```diff
@@ -242,9 +269,26 @@ def create(
             header_params=options._header_params(),
             json_body=message_in.model_dump_json(exclude_unset=True, by_alias=True),
         )
-        ret = response.json()
-        ret["payload"] = message_in.payload
-        return MessageOut.model_validate(ret)
```
svix-jplatte, line 247: Pretty sure that this was the only manual change we had, and as we discussed a while ago, there is really no good reason for this to be here.

svix-mman (approved):

PR #1760, opened 2025-02-27
js: Add new codegen templates

svix-jplatte:
One tiny change from the templates currently in https://github.com/svix/openapi-codegen: `DateTime` query parameters  are hardcoded to accept `null` (they were broken since a recent PR correctly changed `Date | null` => `Date` in JS type translation).

### DIFF
--- a/.github/workflows/codegen.yml
+++ b/.github/workflows/codegen.yml
@@ -25,6 +25,12 @@ jobs:
     steps:
     - uses: actions/checkout@v4
 
+    # JS codegen uses biome
+    - name: Setup Biome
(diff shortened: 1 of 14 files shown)


svix-jplatte:
I don't know for sure. But it would be a breaking change to not allow `null` anymore.

svix-mman:
Why is this change not showing up in the `Check for uncommitted changes` workflow?

svix-jplatte:
Because I already adjusted the template to emit exactly what we had committed before.

svix-mman (approved):

PR #1768, opened 2025-02-28
cli: Add codegen template and regenerate

svix-jplatte:
The codegen template added here has some slight modifications to the one we were using before, as can be seen from the diff of src/cmds/api.

### DIFF
--- a/regen_openapi.sh
+++ b/regen_openapi.sh
@@ -55,6 +55,25 @@ fi
     rm rust/src/api/{environment,health}.rs
 )
 
+# CLI
+(
+    # Print commands we run
+    set -x
+
+    openapi-codegen generate \
+        --template svix-cli/templates/api_resource.rs.jinja \
+        --input-file lib-openapi.json \
+        --output-dir svix-cli/src/cmds/api
+
+    # Our CLI templates currently output some unused imports. Get rid of them.
+    cargo fix --manifest-path svix-cli/Cargo.toml --allow-dirty
+    # `cargo fix` can leave the source in an inconsistently-formatted state.
+    cargo fmt --manifest-path svix-cli/Cargo.toml
+
+    # Remove APIs we may not (yet) want to expose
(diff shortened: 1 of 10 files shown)


regen_openapi.sh
```diff
@@ -55,6 +55,25 @@ fi
...
+        --input-file lib-openapi.json \
+        --output-dir svix-cli/src/cmds/api
+
+    # Our CLI templates currently output some unused imports. Get rid of them.
+    cargo fix --manifest-path svix-cli/Cargo.toml --allow-dirty
+    # `cargo fix` can leave the source in an inconsistently-formatted state.
+    cargo fmt --manifest-path svix-cli/Cargo.toml
+
+    # Remove APIs we may not (yet) want to expose
+    rm svix-cli/src/cmds/api/{background_task,environment,health,operational_webhook_endpoint,statistics}.rs
```
svix-gabriel, line 74: Are we going to have to remember to update this anytime we add an API that we aren't ready to be exposed?

svix-gabriel: Yeah... it just seems like an easy thing to forget.

It'd be nice if the public APIs were whitelisted, rather than blacklisting the secret APIs. If that's impractical though I suppose it's fine.

svix-gabriel (approved):

regen_openapi.sh
svix-mman: > Yeah... it just seems like an easy thing to forget.

I think I may have already added some of these APIs by mistake :(

svix-jplatte: To be clear, none of these APIs are secret. The real secret ones are already not emitted. If you look at the list, all of these are part of the public docs. We just hadn't exposed them in the CLI before (for the libs, the list is shorter) and I didn't want to make the diff larger for this PR.

I'm thinking we probably do want `environment` added to all libs sooner rather than later. `health` just doesn't seem that useful, but maybe it is?

PR #1769, opened 2025-02-28

svix-jplatte:
This does the bare minimum to make CI green with the latest OpenAPI spec. In particular:

- This does not update any languages we don't have in `regen_openapi.sh` yet - we're going to add the remaining languages over the coming week or two
- This does not expose the ingest endpoint API in JS and Python - we want to generate an `ingest` resource API that has an accessor for an `endpoint` sub-resource, but the templates don't support it yet (this is also coming soon); for Rust I wrote this by hand to get rid of some "unused pub" warnings


svix-mman:
@svix-jplatte I want to merge and release this today so the `headers` field on `EndpointIn` is released (For a JS customer)
Since we don't expose `ingest` yet, there should not be a problem committing the new struct enum logic without testing it first right?

svix-mman (approved):

svix-jplatte:
This version of the spec doesn't contain any struct enums yet anyways.

PR #1770, opened 2025-02-28
Bump to version 1.61.0

svix-mman:
## Version 1.61.0
* Libs/Ruby **(Breaking)**: Ruby version changed from `2.7` to `3.4.2`
* Libs/Ruby **(Breaking)**: Deprecated methods `MessageAttempt.list` and `MessageAttempt.list_attempts_for_endpoint` are removed


svix-jplatte (approved):
Maybe add a changelog entry for the `headers` thing?

svix-mman:
> Maybe add a changelog entry for the headers thing?

I thought of this, but we would only add it for python/js/rust
I think it's better to add the changelog about headers once we have it for all of the libs

PR #1777, opened 2025-03-01
ruby: Pull in change from latest codegen

svix-mman:
Pull in Ruby templates from the `openapi-codegen` repo
Also pull in latest changes from `lib-openapi.json`

For now `regen_openapi.sh` for Ruby is disabled in CI


ruby/templates/api_resource.rb.jinja
```diff
@@ -0,0 +1,72 @@
...
+      {%- if op | has_query_or_header_params -%}
+      options = {}
+      {%- endif -%}
+    {% endset -%}
+
+    def {{ op.name | to_snake_case }}({{ func_args | strip_trailing_comma }})
+      {% if op | has_query_or_header_params -%}
+      options = options.transform_keys(&:to_s)
+      {% endif -%}
+      path = "{{ op.path | generate_ruby_path_str(op.path_params) }}"
```
svix-jplatte, line 37: Why have a local variable for this? Looks like it's only used in a single place.

ruby/templates/api_extra/application_create.rb
```diff
@@ -0,0 +1,16 @@
...
+    "POST",
+    path,
+    query_params:{
+        "get_if_exists"=> true
+    },
+    headers: {
+        "idempotency-key" => options["idempotency-key"]
+    },
+    body: application_in
+    )
```
svix-jplatte, line 14: Parameters not indented?

svix-jplatte: Or rather, the first few and last few lines of this function should probably be un-indented by two spaces?

ruby/templates/component_type.rb.jinja
```diff
@@ -0,0 +1,7 @@
+{% if type.kind == "struct" -%}
+    {% include "types/struct.rb.jinja" -%}
+{%- elif type.kind == "string_enum"-%}
+    {% include "types/string_enum.rb.jinja" -%}
+{%- elif type.kind == "integer_enum" %}
+{%- include "types/integer_enum.rb.jinja" -%}
```
svix-jplatte, line 6:
```suggestion
{%- elif type.kind == "integer_enum" -%}
    {% include "types/integer_enum.rb.jinja" -%}
```

ruby/templates/summary.rb.jinja
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
```
svix-jplatte, line 1: Why do we need this?

ruby/templates/types/struct.rb.jinja
```diff
@@ -0,0 +1,87 @@
+# frozen_string_literal: true
+# This file is @generated
```
svix-jplatte, line 2: If we need this for all types, might as well move it up to `component_type.rb.jinja`?

ruby/templates/api_resource.rb.jinja
svix-mman: Indeed we don't need a variable here

ruby/templates/summary.rb.jinja
svix-mman: This was in the old lib, It basically ensures that strings can't be modified at runtime. [source](https://stackoverflow.com/a/37799399)

ruby/templates/types/struct.rb.jinja
svix-mman: I don't like putting stuff in the parent `component_type.xx.jinja`, I feel like it's confusing.

svix-jplatte: Hm, maybe we should split up the `component_type` template entirely then? It would result in a tiny bit of duplication, but I've also thought about it before. Then we don't need the repeated type `kind` comparions either.

svix-jplatte: Okay, let's leave things as-is for now and I'll write it down for later consideration.

svix-jplatte (approved):

Issue #1779, opened 2025-03-03
Java SDK: `Message.create` escapes JSON string since v1.60.0


mroloux:
### Version

v1.60.0

### Description

Since v1.60.0, `Message.create()` escapes quotes in the payload. So if you pass in a JSON string, quotes get escaped - but they shouldn't. This worked fine until v1.59.0.

It's not an option for us to pass in a POJO that Svix transforms to a JSON string. We use GSON as JSON library, and Svix doesn't seem to be able to serialize GSON `JsonObject` objects.

Would be nice to get the old behaviour back, or at least a way to specify we're passing in a JSON string.

svix-mman:
@mroloux Thanks for the report. you can use `Message.messageInRaw` to pass in a raw json encoded object.
In version `1.60.0` we replaced `GSON` with `jackson`, so I don't think we would be able to serialize GSON objects

tasn:
We are discussing this internally, will get back to you soon.

tasn:
@mroloux, sorry about this break, it was not intended. We are going to release a new version that completely removes opinionated choices about the serialization library as that has been a constant pain for everyone. Instead we will change this function to accept a string (pre-serialized by you, and untouched by us) that we will send verbatim.

This will be a compile time error that you'll need to address, but I believe will be a more robust solution for existing customers and new customers going forward.

mroloux:
Sounds good, thanks 👍

svix-mman:
@mroloux we just released `v1.61.1`,  you can now pass in a json encoded string into `MessageIn.payload`.

mroloux:
👍

mroloux:
Maybe I'm doing something wrong, but now I get an error 422 when sending a message:

```json
{
  "detail" : [ {
    "loc" : [ "body", "payload" ],
    "msg" : "Payload must be an object.",
    "type" : "value_error"
  } ]
}
```

The code looks like this:

```java
MessageIn message = new MessageIn()
        .eventId(/* event ID */)
        .eventType((/* event type */)
        .payload(/* a JSON string */);
svix.getMessage().create(Long.toString(applicationId), message);
```

svix-mman:
@mroloux I apologize about this, this is a bug.
I am working on a fix right now

svix-mman:
@mroloux `v1.61.2` was released with a fix for this bug

PR #1783, opened 2025-03-03
java/kotlin: Switch the type of `MessageIn.payload` to a string

svix-mman:
Switch to type of `MessageIn.payload` to a string, this allows customers to send pre-serialized json string

fixes: https://github.com/svix/svix-webhooks/issues/1779

### DIFF
--- a/java/lib/src/main/java/com/svix/api/Message.java
+++ b/java/lib/src/main/java/com/svix/api/Message.java
@@ -135,6 +135,20 @@ public class Message {
         if (options.idempotencyKey != null) {
             headers.put("idempotency-key", options.idempotencyKey);
         }
+        if (messageIn.getTransformationsParams() != null) {
+            if (messageIn.getTransformationsParams().get("rawPayload") == null) {
+                // transformationsParams may be immutable
+                HashMap<String, Object> trParams =
+                        new HashMap<>(messageIn.getTransformationsParams());
+                trParams.put("rawPayload", messageIn.getPayload());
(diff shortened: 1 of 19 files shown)


svix-mman:
Needs more work

svix-james (approved):

java/lib/src/test/com/svix/test/WiremockTests.java
```diff
@@ -428,7 +438,7 @@ public void msgInRaw() throws Exception {
         svx.getMessage().create("app1", msg);
 
         String expectedBody =
-                "{\"eventType\":\"event.ended\",\"payload\":{},\"transformationsParams\":{\"rawPayload\":\"<xml>{no"
```
svix-jplatte, line 431: This is wrong. We must still send an (empty) object for the `payload` property, potentially by first converting the updated `MessageIn` type from the public API to an internal type with a differently-typed `payload`.

java/lib/src/main/java/com/svix/api/Message.java
```diff
@@ -233,7 +247,7 @@ public static MessageIn messageInRaw(final String payload, final String contentT
         trParam.put("rawPayload", payload);
         trParam.put("headers", Collections.singletonMap("content-type", contentType));
         MessageIn msg = new MessageIn();
-        msg.setPayload(new HashMap<>());
+        msg.setPayload("");
```
svix-jplatte, line 250: This seems like a no-op given `setTransformationsParams` sets the actual body.

PR #1786, opened 2025-03-04
java/kotlin: Send empty struct instead of empty string

svix-mman:
Fixes regression introduced in https://github.com/svix/svix-webhooks/pull/1783

### DIFF
--- a/java/lib/src/main/java/com/svix/api/Message.java
+++ b/java/lib/src/main/java/com/svix/api/Message.java
@@ -1,13 +1,18 @@
 // this file is @generated
 package com.svix.api;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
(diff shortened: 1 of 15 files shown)


java/lib/src/main/java/com/svix/api/Message.java
```diff
@@ -251,4 +258,36 @@ public static MessageIn messageInRaw(final String payload, final String contentT
         msg.setTransformationsParams(trParam);
         return msg;
     }
+
+    @ToString
+    @EqualsAndHashCode
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    @NoArgsConstructor
+    @Getter
+    @Setter
```
svix-jplatte, line 267: For the internal class, should we maybe make the fields public and access them directly instead?

svix-jplatte (approved):

PR #1793, opened 2025-03-05
kotlin: Update codegen templates to support struct enums

svix-mman:
The API looks like this
```kotlin
val sourceIn = IngestSourceIn(
    name = "name",
    uid = "uuiidd",
    config = IngestSourceInConfig.Cron(
        CronConfig(
            contentType = "ads123",
            payload = "payload",
            schedule = "* * * * *",
        )
    ),
)
// read the payload field
if (sourceIn.config is Cron) {
    // smart cast
    println(sourceIn.config.cron.payload)
}

// without any enum fields
val sourceIn = IngestSourceIn(
    name = "name",
    uid = "uuiidd",
    config = IngestSourceInConfig.Genericwebhook,
)
```

### DIFF
--- a/kotlin/lib/src/test/com/svix/kotlin/SerializationTests.kt
+++ b/kotlin/lib/src/test/com/svix/kotlin/SerializationTests.kt
@@ -1,0 +1,59 @@
+package com.svix.kotlin
+
+import com.svix.kotlin.models.CronConfig
+import com.svix.kotlin.models.IngestSourceIn
+import com.svix.kotlin.models.IngestSourceInConfig
+import com.svix.kotlin.models.IngestSourceInConfig.Cron
+import kotlin.test.assertEquals
+import kotlinx.serialization.json.Json
+import org.junit.jupiter.api.Test
+import kotlin.test.assertIs
+
+class SerializationTests {
+    @Test
+    fun structEnumWithFields() {
+        val jsonSource =
+            """{"name":"name","uid":"uuiidd","type":"cron","config":{"contentType":"web-app/json","payload":"totally json fr fr","schedule":"* * * * *"}}"""
+        val sourceIn =
+            IngestSourceIn(
+                name = "name",
+                uid = "uuiidd",
+                config = Cron(
+                    CronConfig(
+                        contentType = "web-app/json",
+                        payload = "totally json fr fr",
+                        schedule = "* * * * *",
+                    )
+                ),
+            )
+        // de/serialization works both ways
+        assertEquals(jsonSource, Json.encodeToString(sourceIn))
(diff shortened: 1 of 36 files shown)


kotlin/lib/src/test/com/svix/kotlin/SerializationTests.kt
```diff
@@ -0,0 +1,52 @@
...
+//    @Test
+//    fun structEnumWithFields() {
+//        val jsonSource =
+//            """{"name":"name","uid":"uuiidd","type":"cron","config":{"contentType":"web-app/json","payload":"totally json fr fr","schedule":"* * * * *"}}"""
+//        val sourceIn =
+//            IngestSourceIn(
+//                name = "name",
+//                uid = "uuiidd",
+//                type =
+//                    IngestSourceInType.Cron(
```
svix-jplatte, line 25: Isn't it that in all of Py, JS and Rust I chose to name the sub-type (and field where applicable) based on the content field rather than the discriminator field? Why use the discriminator field for the name here?

svix-mman: In my head `config = IngestSourceInConfig.Cron(CronConfig(...))` looks odd. (double nesting of Config)
`type = IngestSourceInType.Cron(CronConfig(..))` looks better

svix-jplatte: That's exactly what we have in the server code as well though. Also `type` is a keyword in many languages, so I think the existing solution is easier all around. If you're still unconvinced, maybe we should have a small vote on Slack to get this bikeshed painted quickly?

svix-mman: No need, I will use content-field :+1:

kotlin/lib/src/test/com/svix/kotlin/SerializationTests.kt
```diff
@@ -0,0 +1,59 @@
...
+    fun readStructEnumField() {
+        val jsonSource =
+            """{"name":"name","uid":"uuiidd","type":"cron","config":{"contentType":"web-app/json","payload":"totally json fr fr","schedule":"* * * * *"}}"""
+        val sourceIn = Json.decodeFromString(IngestSourceIn.serializer(), jsonSource)
+
+        assertIs<IngestSourceInConfig.Cron>(sourceIn.config)
+        // the assertIs smart casted for us
+        assertEquals("* * * * *", sourceIn.config.cron.schedule)
+    }
+}
```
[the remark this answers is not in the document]
svix-jplatte: VSCode: `"files.insertFinalNewline": true,`

PR #1801, opened 2025-03-06
python: Support field-less object types

svix-jplatte:
We've got some field-less object types in the next version of our API spec.


svix-mman (approved):
I really don't like merging template changes without seeing how they affect the generated code :(

svix-mman:
that works :+1:

PR #1808, opened 2025-03-07
Dependency + cargo deny updates

svix-jplatte:
Should turn CI green again.


svix-jplatte:
Okay something broke with Kafka, not upgrading dependencies other than ring for now.

PR #1814, opened 2025-03-11
server: Run recover in batches

future-highway:
The intent of the recovery loop seems to be to recover up to the `RECOVERY_LIMIT` in increments of `BATCH_SIZE`, but the limit arg is currently set to `RECOVERY_LIMIT`.

There also seems to be a typo in how the loop breaks, which could cause an excessive number of messages to be recovered: `num_done > RECOVERY_LIMIT` instead of `num_done >= RECOVERY_LIMIT` (equal should also break the loop).

## Motivation

This change should result in a more reasonable load per request on external resources (i.e., db & queue) as seemingly originally intended. However, it will likely make the recovery process slower as the number of messages to recover approaches the limit; what was 1 loop will now be up to 100, at the limit.  

## Solution

Set limit to `BATCH_SIZE` and stop when `num_done >= `RECOVERY_LIMIT`.


tasn:
Thank you for the PR, I asked the team to review!

svix-gabriel (approved):
Thanks for fixing this!

Issue #1796, opened 2025-03-06
regen_openapi.py does not consider dependencies


svix-jplatte:
Our new `regen_openapi.py` script can currently produce some spurious errors when generating the CLI because it depends on the Rust lib via `path` dependency, but since generation happens in parallel, the lib might not yet be done generating.

PR #1817, opened 2025-03-12
Hardcode a workaround for cli codegen dependency on rust codegen

svix-mman:
The cli task depends on the rust task to complete, there is no way to express this dependency in our codegen The long term plan is to have a customer rust frontend cli to the codegen, but for now this will do

Fixes https://github.com/svix/svix-webhooks/issues/1796

I want to merge this before enabling codegen for cli

### DIFF
--- a/regen_openapi.py
+++ b/regen_openapi.py
@@ -228,6 +228,17 @@ def parse_config():
                     "template_dir": language_config["template_dir"],
                 }
             )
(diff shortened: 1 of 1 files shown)


svix-jplatte (approved):
If the steps within one lang block are run serially, maybe we should just update the toml file to the structure the py code generates internally here? It doesn't seem so bad.

PR #1820, opened 2025-03-13
server: Replace usage of once_cell with equivalent from std

future-highway:
Drop the direct dependency on `once_cell` as equivalent functionality was released in `std` for Rust 1.80 (July 25, 2024). 

## Motivation

Eventually this will hopefully lead to 1 less dependency, however for now `once_cell` still remains as an indirect dependency since it is used widely in the ecosystem.

## Solution

Replace usage of `once_cell::sync::Lazy` with `std::sync::LazyLock`.

### DIFF
--- a/server/Cargo.lock
+++ b/server/Cargo.lock
@@ -4827,7 +4827,6 @@ dependencies = [
  "lapin",
  "num_enum",
  "omniqueue",
- "once_cell",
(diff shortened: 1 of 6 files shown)


svix-jplatte (approved):
Sure, why not :+1:

PR #1823, opened 2025-03-17
js: Add internal lib

svix-mman:
Add an internal lib for frontend.

The non-public models are not exported from 'src/models/index.ts'

### DIFF
--- a/codegen.toml
+++ b/codegen.toml
@@ -18,16 +18,33 @@ output_dir = "rust/src/models"
 
 [javascript]
 template_dir = "javascript/templates"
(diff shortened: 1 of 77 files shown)


svix-jplatte:
If this is for our internal usage, why would it be in this repo?

svix-lucho (approved):
We discussed it, but just making sure since it's not in the description: `api_internal` is not exported from the root index, right?

svix-mman:
> We discussed it, but just making sure since it's not in the description: `api_internal` is not exported from the root index, right?

Correct

PR #1829, opened 2025-03-20
go: Add support for struct enums

svix-mman:
This is what the API will look like
```go
// struct enum with fields
sourceIn := models.IngestSourceIn{
	Name: "Mendy",
	Type: models.IngestSourceInTypeCron,
	Config: models.CronConfig{
		Payload:  "Hello from space",
		Schedule: "0 0 0 * *",
	},
}

// struct enum without fields
uid := "very unique"
sourceIn := models.IngestSourceIn{
	Name: "Mendy",
	Uid: &uid
	Type: models.IngestSourceInTypeGenericWebhook,
}

// read a field
if cronConfig, ok := sourceIn.Config.(models.CronConfig); ok {
	println(cronConfig.Payload)
}
```

edit(jplatte): Closes https://github.com/svix/monorepo-private/issues/9981, closes https://github.com/svix/monorepo-private/issues/10059

### DIFF
--- a/go/svix_test.go
+++ b/go/svix_test.go
@@ -664,3 +664,86 @@ func TestListResponseOutModels(t *testing.T) {
 	}
 
 }
+
+func TestStructEnumWithFields(t *testing.T) {
+	expectedJson := `{"name":"Mendy","type":"cron","config":{"payload":"Hello from space","schedule":"0 0 0 * *"}}`
+	sourceIn := models.IngestSourceIn{
+		Name: "Mendy",
+		Type: models.IngestSourceInTypeCron,
+		Config: models.CronConfig{
+			Payload:  "Hello from space",
+			Schedule: "0 0 0 * *",
+		},
+	}
+	sourceInJson, err := json.Marshal(sourceIn)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if string(sourceInJson) != expectedJson {
+		t.Errorf("Expected dumped json to match")
+	}
+}
+
+func TestStructEnumWithNoFields(t *testing.T) {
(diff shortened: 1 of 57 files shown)


go/svix_test.go
```diff
@@ -664,3 +664,86 @@ func TestListResponseOutModels(t *testing.T) {
 	}
 
 }
+
+func TestStructEnumWithFields(t *testing.T) {
+	expectedJson := `{"name":"Mendy","type":"cron","config":{"payload":"Hello from space","schedule":"0 0 0 * *"}}`
+	sourceIn := models.IngestSourceIn{
+		Name: "Mendy",
+		Type: models.IngestSourceInTypeCron,
+		Config: models.CronConfig{
```
svix-jplatte, line 673: Why do we have the `Type` field? Should we not infer the `type` JSON field from the type of `Config` when serializing?

openapi-templates/go/component_type.go.jinja
```diff
@@ -1,7 +1,9 @@
 {% if type.kind == "struct" -%}
-	{% include "types/struct.go.jinja" ignore missing -%}
+	{% include "types/struct.go.jinja" -%}
```
svix-jplatte, line 2: :+1:

go/svix_test.go
svix-mman: Because of `SvixConfig`, just by looking at `Config` I am not able to know what the `Type` should be

svix-jplatte: Ooh, right. The `Config` field is just for the inner data, which does not have a 1:1 correspondence with types. I was thinking this was similar to what we have in Python, but it's not :+1:

svix-jplatte (approved):

PR #1830, opened 2025-03-20
Bridge: update to use new rust lib, refactor poller input

svix-onelson:
This diff updates bridge's `PollerInput` to be based on the newer stream API exposed by Polling Endpoints.

### DIFF
--- a/bridge/svix-bridge-types/Cargo.toml
+++ b/bridge/svix-bridge-types/Cargo.toml
@@ -9,7 +9,7 @@ async-trait = "0.1"
 tokio.workspace = true
 serde.workspace = true
 serde_json.workspace = true
-svix = { version = "1.25.0", features = ["svix_beta"] }
+svix = { path = "../../rust", features = ["svix_beta"] }
 
 [lints]
 workspace = true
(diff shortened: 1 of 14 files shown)


bridge/svix-bridge-types/Cargo.toml
```diff
@@ -9,7 +9,7 @@ async-trait = "0.1"
 tokio.workspace = true
 serde.workspace = true
 serde_json.workspace = true
-svix = { version = "1.25.0", features = ["svix_beta"] }
+svix = { path = "../../rust", features = ["svix_beta"] }
```
svix-jplatte, line 12: If we do this, do we need to update a CI workflow file to run bridge CI on Rust lib changes?

bridge/svix-bridge-types/src/lib.rs
```diff
@@ -234,28 +237,13 @@ pub struct SvixSenderOutputOpts {
...
-
 /// Senders convert messages into Create Message API calls so the JSON pulled out of message queues
 /// or produced by transformations need to conform to this shape.
 #[derive(Clone, Deserialize, Serialize)]
 #[serde(rename_all = "camelCase")]
 pub struct CreateMessageRequest {
     pub app_id: String,
     pub message: MessageIn,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub post_options: Option<PostOptions>,
```
svix-jplatte, line 258: This seems like an unrelated change, no?

svix-onelson: These relate to updating the rust sdk from the very old (openapi-generator) sdk to the new one.

svix-jplatte (approved):

PR #1833, opened 2025-03-20
CLI: add support for Polling Endpoint Consumers

svix-onelson:
Adds support for using `v1.events-public.consumer` and `v1.events-public.consumer-seek`, but gated behind a `svix_beta` cargo feature.

### DIFF
--- a/svix-cli/src/cmds/api/events_public.rs
+++ b/svix-cli/src/cmds/api/events_public.rs
@@ -1,4 +1,3 @@
-use chrono::{DateTime, Utc};
 use clap::{Args, Subcommand};
 use svix::api::*;
 
--- a/codegen.toml
+++ b/codegen.toml
@@ -50,7 +50,7 @@ extra_codegen_args = ["--include-mode=public-and-hidden"]
 template_dir = "svix-cli/templates"
 extra_mounts = { "svix-cli/.rustfmt.toml" = "/app/.rustfmt.toml" }
 extra_shell_commands = [
-   "cargo fix --manifest-path svix-cli/Cargo.toml --allow-dirty --allow-staged",
+   "cargo fix --all-targets --all-features --manifest-path svix-cli/Cargo.toml --allow-dirty --allow-staged",
    "cargo fmt --manifest-path svix-cli/Cargo.toml",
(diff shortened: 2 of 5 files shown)


svix-cli/src/cmds/api/events_public.rs
```diff
@@ -1,4 +1,3 @@
-use chrono::{DateTime, Utc};
```
svix-onelson, line 1: CI is complaining that these unused imports are unused, even while I've removed them here. Not sure why. Ideas to fix this? /cc @svix-mman

svix-mman: The CI runs `regen_openapi.py` to ensure that whatever code is committed is identical to the generated code. so it regenerates that import, simply removing it by hand won't work.
The codegen runs `cargo fix` to remove the unused imports, I pushed a fix to this PR adding `--all-targets --all-features` to `cargo fix`

svix-mman (dismissed):

svix-gabriel:
We should probably rename the endpoint to something like `stream_events`, if that's not already taken? IIRC `public_events` was just the chosen name because we didn't know what to name it at the time.

svix-onelson:
> We should probably rename the endpoint to something like `stream_events`, if that's not already taken? IIRC `public_events` was just the chosen name because we didn't know what to name it at the time.

Yeah, this is specifically why I _only exposed this in the rust client_ and _only behind the `svix_beta` flag_. The name right now is not great, even with the historical context.

svix-cli/src/cmds/api/events_public.rs
svix-onelson: Ah, so it's the feature preventing it from cleaning this up during the lint. Makes sense.

svix-mman (approved):

svix-onelson:
Looks like @svix-mman is not yet a code owner here 🤔

svix-mman:
> Looks like @svix-mman is not yet a code owner here 🤔

I think it's because I pushed the fix for `cargo fix`, so it ignores my review :shrug:

svix-onelson:
Oh! I didn't realize you pushed to _my branch_. I thought you merged that to `main`.

svix-mman:
> Oh! I didn't realize you pushed to _my branch_. I thought you merged that to `main`.

My apologizes, I guess it's was not the correct etiquette. :sweat:

svix-jplatte (approved):

PR #1836, opened 2025-03-21
java: Update template to support sub-resources

svix-jplatte:
Part of https://github.com/svix/monorepo-private/issues/10043.

### DIFF
--- a/java/templates/api_resource.java.jinja
+++ b/java/templates/api_resource.java.jinja
@@ -17,7 +17,10 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
 
-{% if resource.name == "message"%}
+{% if resource.subresources | length > 0 -%}
+import lombok.Getter;
+{% endif -%}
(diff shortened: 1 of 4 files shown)


java/templates/api_resource.java.jinja
```diff
@@ -17,7 +17,10 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
 
-{% if resource.name == "message"%}
+{% if resource.subresources | length > 0 -%}
+import lombok.Getter;
+{% endif -%}
```
svix-mman, line 22: FWIW, google-java-format removes explicit unused imports (i think...)

svix-mman (approved):

java/templates/api_resource.java.jinja
svix-jplatte: Yeah, I think I've seen it do that. But I like this better, still.

PR #1843, opened 2025-03-21
go: Pull in changes form latest codegen

svix-mman:
Will fix CI on main


svix-onelson (approved):
👍🏻

go/events_public.go
svix-jplatte, line 1: We really need to rename this thing before the next release!

PR #1846, opened 2025-03-28
svix-cli: Add interactive login

svix-lucho:
Adds a selector when using `svix login` to log in using the dashboard. 

**Demo:**

https://github.com/user-attachments/assets/ec5a6480-723b-4111-a9bb-4c14f090071a

### DIFF
--- a/svix-cli/src/cmds/login.rs
+++ b/svix-cli/src/cmds/login.rs
@@ -1,25 +1,40 @@
-use anyhow::Result;
+use std::time::{Duration, Instant};
+
+use anyhow::{Context, Result};
 use dialoguer::Input;
+use reqwest::Client;
+use serde::Deserialize;
 
 use crate::{config, config::Config};
 
-pub fn prompt() -> Result<()> {
+pub async fn prompt(_cfg: &Config) -> Result<()> {
     print!("Welcome to the Svix CLI!\n\n");
 
-    let auth_token = Input::new()
-        .with_prompt("Auth Token")
-        .validate_with({
-            move |input: &String| -> Result<()> {
-                if !input.trim().is_empty() {
-                    Ok(())
-                } else {
-                    Err(anyhow::anyhow!("auth token cannot be empty"))
+    let selections = &["Login in dashboard.svix.com", "Input token manually"];
+    let selection = dialoguer::Select::new()
+        .with_prompt("How would you like to authenticate?")
+        .items(selections)
+        .default(0)
+        .interact()?;
+
+    let auth_token = if selection == 0 {
+        dashboard_login().await?
+    } else {
+        Input::new()
+            .with_prompt("Auth Token")
+            .validate_with({
+                move |input: &String| -> Result<()> {
+                    if !input.trim().is_empty() {
+                        Ok(())
+                    } else {
+                        Err(anyhow::anyhow!("auth token cannot be empty"))
+                    }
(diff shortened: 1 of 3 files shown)


svix-james (approved):

svix-cli/src/cmds/login.rs
```diff
@@ -45,3 +59,100 @@ pub fn prompt() -> Result<()> {
...
+            "{LOGIN_SERVER_URL}/dashboard/cli/login/session/start"
+        ))
+        .send()
+        .await
+    {
+        Ok(res) => res.json::<CliStartLoginSessionOut>().await?,
+        Err(e) => {
+            eprintln!("Failed to get session ID: {}", e);
+            return Err(anyhow::anyhow!("Failed to get session ID"));
+        }
```
svix-jplatte, line 94: Why print to stderr _and_ return an error? Would the error not be printed otherwise?

svix-cli/src/cmds/login.rs
```diff
@@ -45,3 +59,100 @@ pub fn prompt() -> Result<()> {
...
+            .await
+        {
+            Ok(resp) => resp,
+            Err(e) => {
+                return Err(anyhow::anyhow!(
+                    "Failed to connect to authentication server: {}",
+                    e
+                ));
+            }
+        };
```
svix-jplatte, line 129:
```suggestion
        let response = client
            .post(&poll_url)
            .json(&serde_json::json!({ "sessionId": session_id }))
            .send()
            .awai
            .context("Failed to connect to authentication server")?;
```

(need to `use anyhow::Context`)

svix-cli/src/cmds/login.rs
```diff
@@ -45,3 +59,100 @@ pub fn prompt() -> Result<()> {
...
+
+        if response.status().is_success() {
+            match response.json::<AuthTokenOut>().await {
+                Ok(data) => {
+                    println!("Authentication successful!\n");
+                    return Ok(data.token);
+                }
+                Err(_) => return Err(anyhow::anyhow!("Failed to parse auth token")),
+            }
```
svix-jplatte, line 138: Can also use `.context()` here (unless you think we should be hiding the deserialization error?)

svix-cli/src/cmds/login.rs
```diff
@@ -45,3 +59,100 @@ pub fn prompt() -> Result<()> {
...
+            let error_message = match response.json::<serde_json::Value>().await {
+                Ok(json) => json
+                    .get("detail")
+                    .and_then(|d| d.as_str())
+                    .unwrap_or("Unknown error")
+                    .to_string(),
+                Err(_) => "Unknown error".to_string(),
+            };
+
+            return Err(anyhow::anyhow!("Authentication failed: {}", error_message));
```
svix-jplatte, line 150:
```suggestion
            anyhow::bail!("Authentication failed: {error_message}");
```

svix-cli/src/cmds/login.rs
svix-lucho: yep, this was unnecessary

svix-cli/src/cmds/login.rs
svix-lucho: ohh cool, I didn't know that. Updated the code to use `Context`

svix-james (approved):

PR #1847, opened 2025-03-31
Add Ingest API to all libs

svix-jplatte:
CLI support will be a separate PR because it requires slightly more work.
All the new files are generated (as indicated by comments), but changes to existing files were made by hand.


svix-mman (changes requested):
@svix-jplatte can you un-comment the tests in `javascript/src/mockttp.test.ts`(379-410) and delete line 98 in `javascript/src/index`

svix-mman (approved):

PR #1854, opened 2025-03-31
Bump to v1.63.0

svix-mman:
## Version 1.63.0
* Libs/Cli: New Ingest source and Ingest endpoint APIs are now available.
* Libs: New environment APIs (import, export) are now available.
* Libs/Ruby: Fix bug causing `application.get_or_create` to raise a `TypeError`


ChangeLog.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## Version 1.63.0
+* Libs/Cli: New Ingest source and Ingest endpoint APIs are now available.
```
svix-jplatte, line 4: I think we shouldn't use the same format as `Libs/{Lang}` here, this kinda sounds like it's only for the CLI.

```suggestion
* Libs, CLI: New Ingest source and Ingest endpoint APIs are now available.
```

svix-jplatte (approved):

PR #1863, opened 2025-04-03
python: Remove match statement in favor of if-elif-else

svix-mman:
`match` was introduced in python 3.10, and we support python >= 3.6


svix-jplatte (approved):

tasn:
Python 3.8 is EOL, and 3.9 is almost EOL. Just FYI.

PR #1864, opened 2025-04-03
Bump to version 1.63.1

svix-mman:
## Version 1.63.1
* Libs/Python: Removed internal use of `match` statement, allowing python versions older than 3.10 to be used


ChangeLog.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## Version 1.63.1
+* Libs/Python: Removed internal use of `match` statement, allowing python versions older then 3.10 to be used
```
svix-jplatte, line 4:
```suggestion
* Libs/Python: Removed internal use of `match` statement, allowing python versions older than 3.10 to be used
```

svix-jplatte (approved):

PR #1866, opened 2025-04-04
Replace tokio mutex with std mutex

### LINKED ISSUE #1062: Replace tokio mutex with std mutex
IMHO, I suggest that replacing `tokio::sync::Mutex` with `std::sync::Mutex` can potentially lead to better performance outcomes. This assertion is based on the design differences between the two mutex implementations, particularly in asynchronous contexts.

For further insights, check out the following resources:

[Discussion](https://hachyderm.io/@rain/110737366912995499) on this topic. *Never use Tokio mutexes.*
[tokio::sync::Mutex](https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use
) documentation.

CodeMan62:
## Motivation
1. Performance: `std::sync::Mutex` is generally more efficient when the critical sections are short and contention is low, as it avoids the overhead of async machinery.
2. Simplicity: Standard mutex provides a simpler implementation that's more straightforward to reason about.
3. Best Practices: As noted in the [Tokio documentation](https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use), it's recommended to use `std::sync::Mutex` when:
   - The critical sections are short
   - The mutex isn't held across .await points
   - Performance is a priority

## Solution
The solution involves:

1. Replacing instances of `tokio::sync::Mutex` with `std::sync::Mutex` throughout the codebase
2. Updating the corresponding lock/unlock patterns to use the standard library implementation
3. Ensuring that no mutex is held across .await points to maintain correct async behavior
4. Verifying that all critical sections are short-lived to maximize performance benefits

Thanks for reviewing it let me know if any changes required

### DIFF
--- a/bridge/svix-bridge-plugin-queue/src/receiver_output/mod.rs
+++ b/bridge/svix-bridge-plugin-queue/src/receiver_output/mod.rs
@@ -58,7 +58,7 @@ impl ReceiverOutput for QueueForwarder {
         // it gets in a bad state.
         // When None, rebuild the sender.
         if sender.is_none() {
-            *sender = Some(Self::build_sender(&self.opts).unwrap()?);
+            *sender = Some(Self::build_sender(&self.opts).await.unwrap());
         }
 
         let res = sender
--- a/bridge/svix-bridge-plugin-queue/src/receiver_output/mod.rs
+++ b/bridge/svix-bridge-plugin-queue/src/receiver_output/mod.rs
@@ -2,7 +2,7 @@ use std::sync::Arc;
 
 use omniqueue::{DynProducer, QueueError};
 use svix_bridge_types::{async_trait, BoxError, ForwardRequest, ReceiverOutput};
-use tokio::sync::Mutex;
+use std::sync::Mutex;
 
 use crate::{config::QueueOutputOpts, error::Result};
 
@@ -53,12 +53,12 @@ impl ReceiverOutput for QueueForwarder {
     }
 
     async fn handle(&self, request: ForwardRequest) -> Result<(), BoxError> {
-        let mut sender = self.sender.lock().await;
+        let mut sender = self.sender.lock().unwrap();
         // `QueueForwarder` initializes its sender lazily, but also the sender can be discarded if
         // it gets in a bad state.
         // When None, rebuild the sender.
         if sender.is_none() {
-            *sender = Some(Self::build_sender(&self.opts).await?);
+            *sender = Some(Self::build_sender(&self.opts).unwrap()?);
         }
 
         let res = sender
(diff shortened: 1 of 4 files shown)


bridge/svix-bridge-plugin-queue/src/receiver_output/mod.rs
```diff
@@ -53,12 +53,12 @@ impl ReceiverOutput for QueueForwarder {
     }
 
     async fn handle(&self, request: ForwardRequest) -> Result<(), BoxError> {
-        let mut sender = self.sender.lock().await;
+        let mut sender = self.sender.lock().unwrap();
```
tasn, line 56: This seems very wrong? The std mutex seems to be sync, and tokio is the async version. It looks like we should be using the tokio one and that this code is correct.

Any evidence to the contrary?

tasn (changes requested):
https://github.com/svix/svix-webhooks/pull/1866/files#r2029365945

bridge/svix-bridge-plugin-queue/src/receiver_output/mod.rs
CodeMan62: Hey @tasn thanks for review but it is almost 2 am in night I will do the changes in morning thanks

svix-jplatte: Yeah, the sender is by necessity held over await points so we must **not** use a std Mutex for it.

server/svix-server/src/redis/sentinel.rs
```diff
@@ -34,7 +34,7 @@ impl bb8::ManageConnection for RedisSentinelConnectionManager {
     type Error = RedisError;
 
     async fn connect(&self) -> Result<Self::Connection, Self::Error> {
-        self.client.0.lock().await.get_async_connection().await
+        self.client.0.lock().unwrap().get_async_connection().unwrap()
```
svix-jplatte, line 37: I don't see why the second await is suddenly an unwrap. I kicked off CI, I suspect that this doesn't actually compile.

svix-jplatte (changes requested):
All of the changes outside of the test code still look wrong to me. I'm pretty sure we need the async mutex in all of those cases. And for the test code, it doesn't seem all that useful to switch either.

I will trigger CI to run again to make sure, but unless you have a specific reason to believe there is a worthwhile change here, I'd encourage you to close this PR and I'll close the associated issue since it doesn't look like there was anything of substance there really.

CodeMan62:
> associated

Hii @svix-jplatte thanks for review i am closing this PR. please close the associated issue too.

PR #1868, opened 2025-04-07
Lib/JS: Fix deserialization for optional arrays

svix-lucho:
When deserializing an array, use `?.map` instead of just `.map` to account for cases when element is optional.


svix-jplatte (approved):
This adds the `?` even if the property is not marked optional, right? Even then though, let's merge (we can remove unnecessary `?`s later).

svix-lucho:
> This adds the `?` even if the property is not marked optional, right? Even then though, let's merge (we can remove unnecessary `?`s later).

@svix-jplatte yes. It was the simplest fix I knew how to make quickly.

PR #1869, opened 2025-04-07
python: Fix minimum version for pydantic

svix-jplatte:
We are using ModelWrapValidatorHandler, which was introduced in 2.10

Also add maximum versions because it seems wrong that we don't have them.

### DIFF
--- a/python/requirements.txt
+++ b/python/requirements.txt
@@ -1,7 +1,7 @@
-python_dateutil >= 2.5.3
-setuptools >= 21.0.0
-urllib3 >= 1.25.3
-Deprecated >= 1.2.13
-types-deprecated >= 1.2.5
-types-python-dateutil >= 2.8.9
-pydantic >= 2.10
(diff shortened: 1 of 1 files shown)


mrbcmorris:
@svix-jplatte Just ran into this in my application today so I've pinned svix to 16.2.0.  Do we also need to declare pydantic > 2.10 in the setup.cfg file so pip is aware of this requirement?

I'm stuck on pydantic 2.9.2 until I update my application for compatibility with 2.10 so when I upgrade svix to 16.4.0 my application started failing.

svix-jplatte:
You shouldn't have to declare `pydantic > 2.10` yourself, no. I was not aware that 2.10 contains breaking changes. That's unfortunate. I hope it's not much of an issue for you to pin the svix dependency.

svix-jplatte:
Oh, that's a good point. We probably should, I'll send a PR!

PR #1883, opened 2025-04-14
go: Remove `EventsPublic`. it's not being used anywhere

svix-mman:
This struct (`EventsPublic`) is not being used anywhere

And it's not possible to create this struct manually, because the `client` field is private. So I am sure this code is not being used


svix-jplatte (approved):
Yeah, this was left over from a previous spec version. It was never public. We likely have it for other languages too. For Rust, check with Owen whether the new public poller routes are usable in CLI before removing it there / in the Rust SDK.

PR #1884, opened 2025-04-14
Move `SvixHttpClient` to internal package

svix-mman:
This is done in preparation for the new internal API client

In Go, the `internal` package has special semantics. https://pkg.go.dev/cmd/go#hdr-Internal_Directories
So we can define public  method for our SDK to consume, but external packages wont be able to import the internal package
For example this is the error I get when trying to import the internal package in the TF provider
```
package github.com/svix/terraform-provider-svix
        main.go:10:3: use of internal package github.com/svix/svix-webhooks/go/internal not allowed
```

This is technically a breaking change, because the `svix.SvixHttpClient` struct will no longer be available 
However, this is not an issue. Since there is no use in defining the `svix.SvixHttpClient` by hand, because the only use case it has is to be part of the resource structs
```go
type Application struct {
	client *SvixHttpClient
}
```
But a user of our SDK would not be define the `Application` struct, because `client` is not an exported field.

### DIFF
--- a/go/application.go
+++ b/go/application.go
@@ -4,14 +4,15 @@ package svix
 import (
 	"context"
 
+	"github.com/svix/svix-webhooks/go/internal"
 	"github.com/svix/svix-webhooks/go/models"
 )
 
 type Application struct {
-	client *SvixHttpClient
+	client *internal.SvixHttpClient
 }
 
-func newApplication(client *SvixHttpClient) *Application {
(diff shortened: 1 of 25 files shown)


svix-mman:
> Why were these public in the first place? 

Not sure :shrug: 

> Should they not have been `lowerCamelCase` private symbols?

Yup

PR #1885, opened 2025-04-14
Create new internal Go lib

svix-mman:
Create new Go internal API client

This client only has the limited set of operations needed for the terraform plugin

note: This is branched from https://github.com/svix/svix-webhooks/pull/1884, I will rebase on main once https://github.com/svix/svix-webhooks/pull/1884 is  merged

### DIFF
--- a/codegen.toml
+++ b/codegen.toml
@@ -116,6 +116,7 @@ output_dir = "java/lib/src/main/java/com/svix/models"
 template_dir = "openapi-templates/go"
 extra_shell_commands = [
    "rm go/health.go",
+   "sed -i 's/package svix/package internalapi/g' go/internalapi/management*"
 ]
 [[go.task]]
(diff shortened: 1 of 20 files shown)


svix-jplatte:
Why add this to the SDK, rather than the terraform package?

svix-mman:
> Why add this to the SDK, rather than the terraform package?

I kinda want to dog food the SDK, it feels bad to generate an extra lib.
Is there any downside to create this internal api?

svix-mman:
> I was talking about only generating code there for the internal stuff. I don't see a big downside to this approach, one thing is that it doesn't allow us to later remove the hidden stuff from this repo, but that doesn't seem so important.

I don't like the idea of exposing the internal workings (`executeRequest`, etc...) of the SDK publicly

svix-jplatte:
Don't need to do that, could just duplicate.

Issue #1138, opened 2023-12-06
Handling of operational webhooks server name


boothj5:
After having an issue running local operational webhooks (see the following pull request [https://github.com/svix/svix-webhooks/pull/1137](https://github.com/svix/svix-webhooks/pull/1137)) it looks as though handling the server url could do with a little more validation.

Comments from @tasn in Slack chat:

"we should probably have validation (or rather URL sanitization?) to fix this class of issues going forward."

"we should have also checked if it's a 404 because the org is not found or if it's just a bad path, so we should have improved that too."

The following discussion also has some more information: [https://github.com/svix/svix-webhooks/discussions/1125](https://github.com/svix/svix-webhooks/discussions/1125)

svix-jplatte:
We now do basic validation of the URL, and trailing slashes are automatically trimmed too, as of [#1887](https://github.com/svix/svix-webhooks/pull/1887).

PR #1887, opened 2025-04-15
Add URL validation to operational server webhooks

CodeMan62:
## Motivation

This PR addresses issue #1138  "Handling of operational webhooks server name" which was identified after users experienced problems configuring and troubleshooting operational webhooks in local development environments.

The current implementation of operational webhooks has two key issues:

1. The server URL provided in `SVIX_OPERATIONAL_WEBHOOK_ADDRESS` is not validated or sanitized before use, leading to problems when users provide URLs with trailing slashes or incorrect formats.

2. When operational webhooks fail to deliver with a 404 error, the system logs a generic "Operational webhooks are enabled but no listener set for {org_id}" message, without distinguishing between different failure scenarios (e.g., incorrect URL format, missing application, or missing endpoint).

These issues make it difficult to diagnose and resolve operational webhook configuration problems, as referenced in discussion #1125 where a user couldn't get operational webhooks working locally.

## Solution

This PR implements three key improvements to the operational webhooks handling:

1. **URL Validation**: Added validation for the `operational_webhook_address` configuration value in `cfg.rs` that verifies:
   - The URL has a valid format
   - The URL uses http or https scheme
   - The URL includes a valid host

2. **URL Sanitization**: Enhanced the `OperationalWebhookSenderInner::new` method in `operational_webhooks.rs` to properly sanitize URLs by:
   - Trimming whitespace
   - Removing trailing slashes
   - Preserving the essential URL structure

3. **Improved Error Handling**: Expanded the error handling in `send_operational_webhook` to provide more detailed diagnostics:
   - Distinguishes between "no application exists" and "no endpoint configured" scenarios for 404 errors
   - Adds specific handling for BAD_REQUEST errors that often indicate URL format issues
   - Includes the actual URL in error messages to aid debugging

### DIFF
--- a/server/svix-server/src/cfg.rs
+++ b/server/svix-server/src/cfg.rs
@@ -76,34 +76,33 @@ fn default_redis_pending_duration_secs() -> u64 {
     45
 }
 
-fn validate_operational_webhook_url(url: &Option<String>) -> Result<(), ValidationError> {
-    if let Some(url_str) = url {
-        match Url::parse(url_str) {
-            Ok(url) => {
-                // Verify scheme is http or https
-                if url.scheme() != "http" && url.scheme() != "https" {
-                    return Err(validation_error(
-                        Some("operational_webhook_address"),
-                        Some("URL scheme must be http or https"),
-                    ));
-                }
-
-                // Verify there's a host
-                if url.host().is_none() {
-                    return Err(validation_error(
-                        Some("operational_webhook_address"),
-                        Some("URL must include a valid host"),
-                    ));
-                }
-            }
-            Err(_) => {
+fn validate_operational_webhook_url(url: &str) -> Result<(), ValidationError> {
+    match Url::parse(url) {
+        Ok(url) => {
+            // Verify scheme is http or https
+            if url.scheme() != "http" && url.scheme() != "https" {
                 return Err(validation_error(
                     Some("operational_webhook_address"),
-                    Some("Invalid URL format"),
+                    Some("URL scheme must be http or https"),
+                ));
+            }
+
+            // Verify there's a host
+            if url.host().is_none() {
+                return Err(validation_error(
+                    Some("operational_webhook_address"),
+                    Some("URL must include a valid host"),
                 ));
             }
         }
+        Err(_) => {
+            return Err(validation_error(
+                Some("operational_webhook_address"),
+                Some("Invalid URL format"),
+            ));
+        }
     }
+
     Ok(())
 }
 
--- a/server/svix-server/src/cfg.rs
+++ b/server/svix-server/src/cfg.rs
@@ -12,7 +12,7 @@ use ipnet::IpNet;
 use serde::{Deserialize, Deserializer};
 use tracing::Level;
 use url::Url;
-use validator::{Validate, ValidationError, ValidationErrors};
+use validator::{Validate, ValidationError};
 
 use crate::{
     core::{cryptography::Encryption, security::JwtSigningConfig},
--- a/server/svix-server/src/cfg.rs
+++ b/server/svix-server/src/cfg.rs
@@ -11,11 +11,13 @@ use figment::{
 use ipnet::IpNet;
 use serde::{Deserialize, Deserializer};
 use tracing::Level;
-use validator::{Validate, ValidationError};
+use url::Url;
+use validator::{Validate, ValidationError, ValidationErrors};
 
 use crate::{
     core::{cryptography::Encryption, security::JwtSigningConfig},
     error::Result,
+    v1::utils::validation_error,
 };
 
 fn deserialize_main_secret<'de, D>(deserializer: D) -> Result<Encryption, D::Error>
@@ -74,6 +76,37 @@ fn default_redis_pending_duration_secs() -> u64 {
     45
 }
 
+fn validate_operational_webhook_url(url: &Option<String>) -> Result<(), ValidationError> {
+    if let Some(url_str) = url {
+        match Url::parse(url_str) {
+            Ok(url) => {
+                // Verify scheme is http or https
+                if url.scheme() != "http" && url.scheme() != "https" {
+                    return Err(validation_error(
+                        Some("operational_webhook_address"),
+                        Some("URL scheme must be http or https"),
+                    ));
+                }
+
+                // Verify there's a host
+                if url.host().is_none() {
+                    return Err(validation_error(
+                        Some("operational_webhook_address"),
+                        Some("URL must include a valid host"),
+                    ));
+                }
+            }
+            Err(_) => {
+                return Err(validation_error(
+                    Some("operational_webhook_address"),
+                    Some("Invalid URL format"),
+                ));
+            }
+        }
+    }
+    Ok(())
+}
+
 #[derive(Clone, Debug, Deserialize, Validate)]
 #[validate(
     schema(function = "validate_config_complete"),
@@ -85,6 +118,7 @@ pub struct ConfigurationInner {
 
     /// The address to send operational webhooks to. When None, operational webhooks will not be
     /// sent. When Some, the API server with the given URL will be used to send operational webhooks.
+    #[validate(custom = "validate_operational_webhook_url")]
     pub operational_webhook_address: Option<String>,
 
     /// The main secret used by Svix. Used for client-side encryption of sensitive data, etc.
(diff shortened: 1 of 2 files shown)


server/svix-server/src/core/operational_webhooks.rs
```diff
@@ -116,9 +116,25 @@ pub struct OperationalWebhookSenderInner {
 
 impl OperationalWebhookSenderInner {
     pub fn new(keys: Arc<JwtSigningConfig>, url: Option<String>) -> Arc<Self> {
+        // Sanitize the URL if present
+        let sanitized_url = url.as_ref().map(|url_str| {
+            // Remove trailing slashes
+            let mut cleaned_url = url_str.trim().to_string();
```
svix-jplatte, line 122: Why trim? If there were any spaces, wouldn't the URL verification have failed already?

server/svix-server/src/core/operational_webhooks.rs
```diff
@@ -116,9 +116,25 @@ pub struct OperationalWebhookSenderInner {
...
+            while cleaned_url.ends_with('/') {
+                cleaned_url.pop();
+            }
+            
+            // Return the cleaned URL or original if empty
+            if !cleaned_url.is_empty() {
+                cleaned_url
+            } else {
+                url_str.clone()
+            }
```
svix-jplatte, line 132: Why?

svix-jplatte (changes requested):
Thanks for the PR!

The URL validation part seems fine to me.

Regarding URL sanitization, see inline comments.

Regarding the error handling,

> When operational webhooks fail to deliver with a 404 error, the system logs a generic "Operational webhooks are enabled but no listener set for {org_id}" message, without distinguishing between different failure scenarios (e.g., incorrect URL format, missing application, or missing endpoint).

This does not make sense to me. Incorrect URL format is something we should be catching earlier, so no need to try to distinguish this when receiving a 404. No endpoints for the app does not result in an error at all.

server/svix-server/src/core/operational_webhooks.rs
CodeMan62: thanks for adressing let me remove trim

server/svix-server/src/core/operational_webhooks.rs
CodeMan62: sorry it was mistake and thanks for adressing

server/svix-server/src/core/operational_webhooks.rs
```diff
@@ -171,20 +182,36 @@ impl OperationalWebhookSenderInner {
...
+                        .application()
+                        .get(recipient_org_id.clone())
+                        .await
+                        .ok()
+                    {
+                        // App exists but endpoint not found
+                        tracing::warn!(
+                            "Operational webhooks are enabled, but no endpoint is configured for organization {} (app exists)",
+                            recipient_org_id,
+                        );
```
svix-jplatte, line 201: I still don't think this is right. If there are no endpoints, that wouldn't lead to a 404.

CodeMan62: makes sense

CodeMan62: Done!

svix-jplatte: I'm sorry if I wasn't clear: I don't think there's any value in this extra API call in the 404 case at all.
Maybe I'm wrong though? Have you been able to get the server to produce both warnings in different circumstances locally?

CodeMan62: I would suggest responding to svix-jplatte with something like this:

---

You're right, and I appreciate the clarification. After thinking about it, I agree there's probably limited value in making that extra API call just to determine whether it's the app missing or the endpoint missing.

I haven't been able to reliably test both scenarios locally in different circumstances. The real goal was to make the error message more helpful for diagnosing configuration issues, but your point makes sense - a simple 404 handler without the extra API call would be cleaner and more efficient.

I've simplified the code to use a single warning message that covers both cases: "Operational webhooks are enabled, but no listener found for organization {}". This removes the unnecessary API call while still providing a useful message when webhooks can't be delivered.

Let me know if this approach works better!

server/svix-server/src/core/operational_webhooks.rs
```diff
@@ -151,6 +161,7 @@ impl OperationalWebhookSenderInner {
             .to_string();
 
         let recipient_org_id = recipient_org_id.to_string();
+        let url_clone = url.clone(); // Clone for use in the async block
```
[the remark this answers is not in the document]
CodeMan62: SO you mean i have to remove changes i have made after this line??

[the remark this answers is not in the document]
svix-jplatte: Yeah, this clone is no longer necessary, and I don't think the comment and tracing changes are all that useful.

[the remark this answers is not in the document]
CodeMan62: Done sorry for late

svix-jplatte (approved):

server/svix-server/src/cfg.rs
```diff
@@ -74,6 +76,37 @@ fn default_redis_pending_duration_secs() -> u64 {
...
+            Err(_) => {
+                return Err(validation_error(
+                    Some("operational_webhook_address"),
+                    Some("Invalid URL format"),
+                ));
+            }
+        }
+    }
+    Ok(())
+}
```
svix-jplatte, line 108: As per CI, it looks like this function should actually accept `&String` / `&str` for the `Validate` derive macro to do its thing.

```suggestion
fn validate_operational_webhook_url(url: &str) -> Result<(), ValidationError> {
    match Url::parse(url) {
        Ok(url) => {
            // Verify scheme is http or https
            if url.scheme() != "http" && url.scheme() != "https" {
                return Err(validation_error(
                    Some("operational_webhook_address"),
                    Some("URL scheme must be http or https"),
                ));
            }

            // Verify there's a host
            if url.host().is_none() {
                return Err(validation_error(
                    Some("operational_webhook_address"),
                    Some("URL must include a valid host"),
                ));
            }
        }
        Err(_) => {
            return Err(validation_error(
                Some("operational_webhook_address"),
                Some("Invalid URL format"),
            ));
        }
    }

    Ok(())
}
```

server/svix-server/src/core/operational_webhooks.rs
```diff
@@ -116,9 +116,19 @@ pub struct OperationalWebhookSenderInner {
 
 impl OperationalWebhookSenderInner {
     pub fn new(keys: Arc<JwtSigningConfig>, url: Option<String>) -> Arc<Self> {
```
svix-jplatte, line 118:
```suggestion
    pub fn new(keys: Arc<JwtSigningConfig>, mut url: Option<String>) -> Arc<Self> {
```

server/svix-server/src/core/operational_webhooks.rs
```diff
@@ -116,9 +116,19 @@ pub struct OperationalWebhookSenderInner {
...
+        // Sanitize the URL if present
+        let sanitized_url = url.as_ref().map(|url_str| {
+            // Remove trailing slashes
+            let mut cleaned_url = url_str.to_string();
+            while cleaned_url.ends_with('/') {
+                cleaned_url.pop();
+            }
+            cleaned_url
+        });
```
svix-jplatte, line 127:
```suggestion
        if let Some(url) = &mut url {
            // Remove trailing slashes
            while cleaned_url.ends_with('/') {
                cleaned_url.pop();
            }
        }
```

server/svix-server/src/cfg.rs
```diff
@@ -11,11 +11,13 @@ use figment::{
 use ipnet::IpNet;
 use serde::{Deserialize, Deserializer};
 use tracing::Level;
-use validator::{Validate, ValidationError};
+use url::Url;
+use validator::{Validate, ValidationError, ValidationErrors};
```
svix-jplatte, line 15:
```suggestion
use validator::{Validate, ValidationError};
```

server/svix-server/src/core/operational_webhooks.rs
```diff
@@ -115,7 +115,15 @@ pub struct OperationalWebhookSenderInner {
 }
 
 impl OperationalWebhookSenderInner {
-    pub fn new(keys: Arc<JwtSigningConfig>, url: Option<String>) -> Arc<Self> {
+    pub fn new(keys: Arc<JwtSigningConfig>, mut url: Option<String>) -> Arc<Self> {
+        // Sanitize the URL if present
+        if let Some(url) = &mut url {
+            // Remove trailing slashes
+            while curl.ends_with('/') {
```
svix-jplatte, line 122:
```suggestion
            while url.ends_with('/') {
```

CodeMan62:
Thank you @svix-jplatte  for all you helps

PR #1892, opened 2025-04-16
go: Set `Content-type` on put requests

svix-mman:
Also fix bug where GET request would send `null` as the body

### DIFF
--- a/go/internal/svix_http_client.go
+++ b/go/internal/svix_http_client.go
@@ -52,13 +52,23 @@ func ExecuteRequest[ReqBody any, ResBody any](
 	if err != nil {
 		return nil, err
 	}
-	encodedBody, err := json.Marshal(jsonBody)
-	if err != nil {
-		return nil, err
-	}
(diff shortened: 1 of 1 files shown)


go/internal/svix_http_client.go
```diff
@@ -52,13 +52,23 @@ func ExecuteRequest[ReqBody any, ResBody any](
...
+	if jsonBody != nil {
+		encodedBody, err := json.Marshal(jsonBody)
+		if err != nil {
+			return nil, err
+		}
+		req, err = http.NewRequestWithContext(ctx, method, urlStr, bytes.NewBuffer(encodedBody))
+		if err != nil {
+			return nil, err
+		}
+		req.Header.Set("Content-type", "application/json")
```
svix-james, line 65: Nit, but can we either put the proper title-cased name of the header here or just use lower case?

svix-mman: Good call :sweat_smile:

svix-james (approved):

PR #1895, opened 2025-04-17
go: Emit `null` instead of `{}` for when struct enum config field is `nil`

svix-mman:
Let's wait for https://github.com/svix/svix-webhooks/pull/1894


svix-jplatte:
What's the motivation for this?

svix-mman:
Not necessary, this was a tmp fix for me to continue working on the terraform side

PR #1897, opened 2025-04-18
libs: Add `source_id` to ingest resources

svix-mman:
Due to an oversight, this was not generated in the original `openapi.json`, fixed in https://github.com/svix/svix-webhooks-private/pull/3874

I am not worried that this broke anyone, because ingest resource were unusable, so no one used them ;)


kotlin/lib/src/main/kotlin/IngestEndpoint.kt
```diff
@@ -81,32 +84,33 @@ class IngestEndpoint(private val client: SvixHttpClient) {
...
         client.executeRequest<Any, Boolean>("DELETE", url.build())
     }
 
     /** Get the additional headers to be sent with the ingest. */
-    suspend fun getHeaders(endpointId: String): IngestEndpointHeadersOut {
+    suspend fun getHeaders(sourceId: String, endpointId: String): IngestEndpointHeadersOut {
         val url =
             client
                 .newUrlBuilder()
-                .encodedPath("/ingest/api/v1/source/{source_id}/endpoint/$endpointId/headers")
```
svix-james, line 97: Just curious -- how was this even compiling before?

svix-james (approved):

kotlin/lib/src/main/kotlin/IngestEndpoint.kt
svix-mman: In this case `.../{source_id}/...` is a valid string
The codegen did not know about `source_id` and it simply ignored it

PR #1901, opened 2025-04-22
Switch C# TargetFramework to netstandard2.0

svix-jplatte:
After re-reading https://learn.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-1-0, I'm thinking this might be the best target, if it works.


svix-jplatte:
Okay, so it seems we can't use the new `required` feature from C# 11 if we target .NET Standard (see also https://stackoverflow.com/questions/74447497/is-it-possible-to-use-the-c11-required-modifier-with-net-framework-4-8-and).

Putting this PR on hold.

PR #1905, opened 2025-04-28
go: Add `SetUserAgentSuffix`

svix-mman:
This allows users of the go lib to suffix the `user-agent` with whatever custom string they chose

This will be used in the terraform provider, to indicate that requests are coming from the provider

The providers user agent will looks like this

`svix-libs/1.64.1/go/tf-provider-v0.1.1`

### DIFF
--- a/go/internalapi/internalapi.go
+++ b/go/internalapi/internalapi.go
@@ -16,12 +16,12 @@ type (
 	}
 )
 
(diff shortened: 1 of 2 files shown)


svix-onelson (commented):
I'm wondering if there's any way to achieve this without making this part of the public API?

If we add it anyway, should we also add it to the rest of the libs for parity? I.E. should it be something we handle in codegen?

svix-onelson (approved):

PR #1906, opened 2025-04-29
Libs: update spec

svix-mman:
Pull in changes from openapi

### DIFF
--- a/java/lib/src/main/java/com/svix/models/ConnectorOut.java
+++ b/java/lib/src/main/java/com/svix/models/ConnectorOut.java
@@ -128,7 +128,7 @@ public class ConnectorOut {
     }
 
     /**
-     * The TransformationTemplate's ID.
+     * The Connector's ID.
      *
      * @return id
      */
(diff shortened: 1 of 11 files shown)


java/lib/src/main/java/com/svix/models/ConnectorOut.java
```diff
@@ -128,7 +128,7 @@ public ConnectorOut id(String id) {
     }
 
     /**
-     * The TransformationTemplate's ID.
+     * The Connector's ID.
```
svix-jplatte, line 131: Why does the diff not have the same thing for kotlin, ruby? :raised_eyebrow:

svix-mman: No doc comments in these languages :(
Ruby I can understand, but I am not sure why Kotlin does not have any doc comments.
I will post a PR for that

svix-jplatte: Would it be hard to add doc comments to Ruby? Does Ruby not have the concept of doc comments? Should we add regular comments at least then?

svix-jplatte (approved):

PR #1909, opened 2025-04-29
js: Remove internal js lib

svix-mman:
This is not used in the frontend anymore
There is no risk of breaking the API since non of these models were exported from `javascript/src/models/index.ts`


codegen.toml
```diff
@@ -21,30 +21,16 @@ output_dir = "rust/src/models"
...
-   "-e=v1.sink.create",
-]
-[[javascript.task]]
 template = "codegen-templates/javascript/component_type_summary.ts.jinja"
 output_dir = "javascript/src/models"
 [[javascript.task]]
 template = "codegen-templates/javascript/component_type.ts.jinja"
 output_dir = "javascript/src/models"
-extra_codegen_args = ["--include-mode=public-and-hidden"]
+extra_codegen_args = ["--include-mode=only-public"]
```
svix-jplatte, line 33: Is this not the default? Do we need `extra_codegen_args` still?

svix-jplatte (dismissed):
> There is no risk of breaking the API since non of these models were exported from `javascript/src/models/index.ts`

People can import from sub-modules, no?
Obligatory link: https://www.hyrumslaw.com/ :wink:

codegen.toml
svix-mman: Correct, we don't need this.
I will remove it

svix-jplatte (approved):

PR #1913, opened 2025-05-05
Add CI for libs to push on main

svix-mman:
Some of these CI workflows were not triggered on a push to main

### DIFF
--- a/.github/workflows/go-ci.yml
+++ b/.github/workflows/go-ci.yml
@@ -8,6 +8,15 @@ on:
       - "go.mod"
       - "lib-openapi.json"
       - ".github/workflows/go-ci.yml"
+  push:
+    branches:
+      - main
+    paths:
+      - "go/**"
+      - "go.sum"
+      - "go.mod"
+      - "lib-openapi.json"
+      - ".github/workflows/go-ci.yml"
 
 jobs:
   build:
--- a/.github/workflows/go-lint.yml
+++ b/.github/workflows/go-lint.yml
@@ -3,7 +3,18 @@ on:
   pull_request:
     paths:
       - "go/**"
+      - "go.sum"
+      - "go.mod"
       - "lib-openapi.json"
+  push:
+    branches:
+      - main
+    paths:
+      - "go/**"
+      - "go.sum"
+      - "go.mod"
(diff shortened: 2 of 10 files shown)


.github/workflows/go-ci.yml
```diff
@@ -4,11 +4,15 @@ on:
   pull_request:
     paths:
       - "go/**"
-      - "go.sum"
-      - "go.mod"
```
svix-jplatte, line 8: I think these should still be included? For the push trigger too.

.github/workflows/go-ci.yml
```diff
@@ -4,11 +4,15 @@ on:
...
       - ".github/workflows/go-ci.yml"
-
+  push:
+    branches:
+      - main
+    paths:
+      - "go/**"
+      - "lib-openapi.json"
+      - ".github/workflows/go-ci.yml"
 jobs:
```
svix-jplatte, line 16: Please let's have a newline here like before ;)

.github/workflows/go-lint.yml
```diff
@@ -4,6 +4,13 @@ on:
     paths:
       - "go/**"
       - "lib-openapi.json"
```
svix-jplatte, line 6: Looks like the repo-root go files are missing here.

.github/workflows/rust-lint.yml
```diff
@@ -7,6 +7,7 @@ on:
     paths:
       - 'rust/**'
       - '.github/workflows/rust-lint.yml'
+      - "lib-openapi.json"
```
svix-jplatte, line 10: Kind of infuriating how we're mixing single quotes and double quotes. But can fix in a follow-up PR.

.github/workflows/go-ci.yml
svix-mman: I thought we already had a `- "go/**"` push trigger, But as I am writing this I remember these files are in the root of the repo
I will revert this :sweat_smile:

.github/workflows/rust-lint.yml
svix-mman: I will post a PR after this one

svix-jplatte (approved):

PR #1914, opened 2025-05-05
Use double quotes consistently

svix-mman:
Don't use a mix of single and double quotes

### DIFF
--- a/.github/workflows/bridge-ci.yml
+++ b/.github/workflows/bridge-ci.yml
@@ -5,15 +5,15 @@ on:
     branches:
       - main
     paths:
-      - 'bridge/**'
-      - 'rust/**'
-      - '.github/workflows/bridge-ci.yml'
+      - "bridge/**"
+      - "rust/**"
+      - ".github/workflows/bridge-ci.yml"
   pull_request:
     paths:
-      - 'bridge/**'
-      - '.github/workflows/bridge-ci.yml'
-      - 'rust/**'
-      - 'lib-openapi.json'
+      - "bridge/**"
+      - ".github/workflows/bridge-ci.yml"
+      - "rust/**"
+      - "lib-openapi.json"
 
 # When pushing to a PR, cancel any jobs still running for the previous head commit of the PR
 concurrency:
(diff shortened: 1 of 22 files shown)


.github/workflows/php-release.yml
```diff
@@ -18,7 +18,7 @@ jobs:
       - name: Update Packagist
         run: |
           curl -X POST "https://packagist.org/api/update-package?username=svix&apiToken=$PACKAGIST_API_KEY" \
-            -H 'content-type: application/json' \
-            -d '{"repository":{"url":"https://packagist.org/packages/svix/svix"}}'
+            -H "content-type: application/json" \
+            -d "{"repository":{"url":"https://packagist.org/packages/svix/svix"}}"
```
svix-jplatte, line 22: This looks like it may break things.

svix-mman: whops :sweat_smile: fixed.

svix-jplatte (approved):

.github/workflows/bridge-ci.yml
svix-onelson, line 1: Sort of mirroring Jonas' comment, I would have encouraged us to go single-quote just because it feels safer in CI where shell invocations sometimes happen (and _unexpected variable substitution_ could be _unexpected_). Single quotes are "safey on" with an explicit escalation to double quotes where needed.

Maybe that's just me?

svix-mman:
```bash
cd .github/workflows/
find -type f -name "*.yml" | xargs sed -i  "s/\"/'/g"
```
Okie, I may as well get it right :smiling_face_with_tear:

svix-mman: nvm, This is going to take more then 5 minutes

svix-mman: @svix-onelson does your comment block this PR, or can I merge?

svix-onelson: My preference is to just close the PR. I did a quick search and there are 108-ish occurrences of `"` and the majority of them are in places where variable substitution are in play.

PR #1917, opened 2025-05-06
csharp: Mark BaseUrl as deprecated in favor of ServerUrl on `SvixOptions`

svix-mman:


svix-lucho (approved):

svix-jplatte (approved):

csharp/Svix/SvixOptions.cs
```diff
@@ -2,17 +2,24 @@ namespace Svix
...
+            string serverUrl,
             int timeoutMilliseconds = 15000,
-            List<int>? retryScheduleMilliseconds = null
+            List<int>? retryScheduleMilliseconds = null,
+            string? baseUrl = null
         )
         {
-            BaseUrl = baseUrl;
+            ServerUrl = baseUrl ?? serverUrl;
+            BaseUrl = baseUrl ?? serverUrl;
```
tasn, line 22: Why are we setting both? The structure should only have one variable and this gets both.
Additionally, `serverUrl` should be the priority, not `baseUrl`...

csharp/Svix/SvixOptions.cs
```diff
@@ -2,17 +2,24 @@ namespace Svix
...
+        public string ServerUrl { get; }
         public List<int> RetryScheduleMilliseconds { get; } = [50, 100, 200];
         public int TimeoutMilliseconds { get; } = 15000;
 
+        /// <param name="serverUrl">The server URL to connect to.</param>
+        /// <param name="baseUrl">[Deprecated] Please use serverUrl parameter instead.</param>
+#pragma warning disable CS0618
         public SvixOptions(
-            string baseUrl,
+            string serverUrl,
```
tasn, line 15: This should be optional as well because:
1. Otherwise it'll break for people that use baseUrl as named (this won't be set).
2. It's optional anyway, we have a default...

svix-jplatte (approved):

PR #1919, opened 2025-05-06
Make sed command work both on macos/linux

svix-mman:
We use `sed` to do a quick text replace as part of the libs codegen

Macos ships a BSD `sed` and (most) linux distros ship a GNU `sed`, they are not exactly 1-1 compatible 

This PR makes the `sed` command work on both GNU/BSD sed

### DIFF
--- a/codegen.toml
+++ b/codegen.toml
@@ -104,7 +104,8 @@ output_dir = "java/lib/src/main/java/com/svix/models"
 template_dir = "codegen-templates/go"
 extra_shell_commands = [
    "rm go/health.go",
-   "sed -i 's/package svix/package internalapi/g' go/internalapi/management*"
+   "sed -i.bak 's/package svix/package internalapi/g' go/internalapi/management* && rm go/internalapi/management*.bak"
+
 ]
 [[go.task]]
 template = "codegen-templates/go/api_resource.go.jinja"


codegen.toml
```diff
@@ -104,7 +104,10 @@ output_dir = "java/lib/src/main/java/com/svix/models"
 template_dir = "codegen-templates/go"
 extra_shell_commands = [
    "rm go/health.go",
-   "sed -i 's/package svix/package internalapi/g' go/internalapi/management*"
+   # same thing as `sed -i 's/package svix/package internalapi/g' go/internalapi/management*`, but this will work both on macos(bsd-sed) and linux(gnu-sed)
+   "python -c \"import glob; [(lambda f: open(f, 'w').write(content))(f) for f in glob.glob('go/internalapi/management*') if (content := open(f, 'r').read().replace('package svix', 'package internalapi'))]\""
+
+
```
tasn, line 110:
```suggestion
sed -i.bak 's/package svix/package internalapi/g' go/internalapi/management* && rm go/internalapi/management*.bak
```
If it's cursed, maybe switch to something not cursed?

svix-mman: > If it's cursed, maybe switch to something not cursed?

Good call. I will use this instead of the python monstrosity

> Also, why do we even need this sed?

Because I want to do something that the codegen won't let me, and this is easier then going into the codegen and re-architecting it

svix-jplatte (approved):
I took a look at bsd-sed docs and don't really get why the previous thing wouldn't work there, but alright :shrug:

PR #1920, opened 2025-05-06
Bump codegen image

svix-mman:
Latest change https://github.com/svix/openapi-codegen/pull/109
Rubyfmt is not compiled from source each time


svix-jplatte (approved):
Linking to relevant changes would be nice ;)

PR #1924, opened 2025-05-13
Remove cmake hack again

svix-jplatte:
The upstream change was reverted.


svix-mman (approved):
in before they upgrade to cmake 4 in a couple of months

Issue #1926, opened 2025-05-18
Allow configuring proxies in python sdk


guywillett:
### Motivation

I need to be able to add `proxies` option when initialising Svix(), and not by environment variable

### Proposal

Add `proxies` to SvixOptions, which is passed to underlying httpx client.

EG
proxies = {
    "http://": "https://your_proxy:port",
    "https://": "https://your_proxy:port",
}
options = SvixOptions(proxies=proxies)
svix = Svix('SECRET-KEY', options)

### Alternatives

I might be missing something, and this is already possible somehow

tasn:
Sounds very reasonable. CC @svix-mman

PR #1928, opened 2025-05-19
python: Enable configuring proxies in `SvixOptions`

svix-mman:
ref: https://github.com/svix/svix-webhooks/issues/1926

Adds a new `proxy` field to `SvixOptions`, to be used as both `http` and `httpx` proxy


svix-jplatte:
> Adds a new `proxy` field to `SvixOptions`, to be used as both `http` and `httpx` proxy

You mean both `http` and `https`? Also, why? AFAIU #1926 asks for being able to configure both as their own thing, and I could there being use cases for only setting one of the two.

svix-mman:
> You mean both `http` and `https`? 

Yes :sweat_smile:

> Also, why? AFAIU #1926 asks for being able to configure both as their own thing, and I could there being use cases for only setting one of the two.

If they are using our client, they are talking to **https://** api.svix.com, so we don't need to support both

svix-jplatte:
Then we should be naming this `https_proxy` and leaving out HTTP proxy support for now, IMO :)

svix-mman:
@svix-jplatte we are keeping this single `proxy` field.

PR #1931, opened 2025-05-22
Publish version 1.66.0

svix-jplatte:
## Version 1.66.0
* Libs/JavaScript: Fix response processing code for endpoints with an optional datetime field in the response body
* Libs/PHP: Update minimum version of PHP to 8.0
* Server: Support sending raw (pre-formatted) JSON payloads


svix-jplatte:
Wrong branch.

PR #1934, opened 2025-05-25
make the retry schedule parameters configurable 

### LINKED ISSUE #782: Libs: make the retry schedule parameters configurable
For example in Kotlin we have:
```
        var sleepTime = 50L
        val numRetries = 3
```
Would be great to make it possible for people to configure them (and other things?) when creating the Svix client.

CodeMan62:
Closes #782 
cc @svix-jplatte  need some further help

### DIFF
--- a/rust/src/api/mod.rs
+++ b/rust/src/api/mod.rs
@@ -25,6 +25,7 @@ mod message_poller;
 mod operational_webhook;
 mod operational_webhook_endpoint;
 mod statistics;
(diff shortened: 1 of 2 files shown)


svix-jplatte:
I'm sorry, but this isn't really a useful change. See also my reply on the associated issue: https://github.com/svix/svix-webhooks/issues/782#issuecomment-2934607410

PR #1936, opened 2025-06-02
go: Disable HTTP/2.0

svix-jplatte:
## Motivation

Some customers are seeing HTTP 400 responses from our load balancer, only with the Go SDK and only with HTTP/2.0.

## Solution

Disable HTTP/2.0 for the Go SDK.

### DIFF
--- a/go/internal/svix_http_client.go
+++ b/go/internal/svix_http_client.go
@@ -3,6 +3,7 @@ package internal
 import (
 	"bytes"
 	"context"
+	"crypto/tls"
 	"encoding/json"
 	"fmt"
 	"io"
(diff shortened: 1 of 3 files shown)


go/internal/svix_http_client.go
```diff
@@ -26,12 +27,21 @@ type SvixHttpClient struct {
 }
 
 func DefaultSvixHttpClient(defaultBaseUrl string) SvixHttpClient {
+	// Disable HTTP/2.0
+	tr := http.DefaultTransport.(*http.Transport).Clone()
+	tr.ForceAttemptHTTP2 = false
```
svix-gabriel, line 32: Based on https://pkg.go.dev/net/http#example-Protocols-Http1or2, it seems like you can just do

`tr.SetHttp2(false)`

Why do we also have to mess with the tls config?

svix-jplatte: I googled and after a lot of back and forth I found that this was the recommended way.

I'm guessing maybe they later added that `Protocols` field to make this easier? It says added in Go 1.24.0 though, and we support Go 1.20 as per `go.mod`.

PR #1941, opened 2025-06-18
js: Generate the index.ts

svix-mman:
Pulls in some changes from monorepo, and modifies the summary template.
The `src/index.ts` is now fully generated.

I will open a PR in monorepo to sync (again) the changes I make here to the template.

### DIFF
--- a/codegen-templates/javascript/summary.ts.jinja
+++ b/codegen-templates/javascript/summary.ts.jinja
@@ -1,19 +1,9 @@
 // this file is @generated
-import { Application } from "./api/application";
-import { Authentication } from "./api/authentication";
-import { BackgroundTask } from "./api/backgroundTask";
-import { Endpoint } from "./api/endpoint";
-import { Environment } from "./api/environment";
-import { EventType } from "./api/eventType";
-import { Health } from "./api/health";
-import { Ingest } from "./api/ingest";
-import { Integration } from "./api/integration";
-import { Management } from "./api/management";
-import { Message } from "./api/message";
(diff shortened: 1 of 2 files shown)


codegen-templates/javascript/summary.ts.jinja
```diff
@@ -0,0 +1,61 @@
...
+
+
+  {% for resource in api.resources -%}
+  public get {{ resource.name | to_lower_camel_case }}(){
+    return new {{ resource.name | to_upper_camel_case }}(this.requestCtx);
+  }
+
+  {% endfor -%}
+  {# hardcoded for backwards compat #}
+  public get operationalWebhookEndpoint() {
```
svix-jplatte, line 58: Let's deprecate this, then you can also remove the comment above.

svix-mman: I don't see a reason to break this (only here), if we want to deprecate `operationalWebhookEndpoint` we should do it in all of the libs.

svix-jplatte: I think it's already deprecated in some of them :stuck_out_tongue:

svix-mman: I just checked
- C# not deprecated
- Go not deprecated
- Java not deprecated
- JS not deprecated
- Kotlin not deprecated
- Python (a/sync) not deprecated
- Ruby not deprecated
- Rust Is the only one deprecated

svix-jplatte (approved):
Alright, can do the deprecation separately.

PR #1942, opened 2025-06-19
[PHP] Handle Badly Formatted Signatures

rodnaph:
Currently a notice will be thrown because of the invalid array access after splitting the signature. Would an exception be better here? signalling unexpected data from svix?

## Motivation

I ran into this while testing Svix integration in our app, and throwing random test data at it. I understand this data should come from Svix, but it seems like it should be handled gracefully anyway.

## Solution

If the signature in the header doesn't at least appear well formatted, skip it.

### DIFF
--- a/php/tests/WebhookTest.php
+++ b/php/tests/WebhookTest.php
@@ -58,7 +58,7 @@ final class WebhookTest extends \PHPUnit\Framework\TestCase
         $this->expectExceptionMessage("No matching signature found");
 
         $testPayload = new TestPayload(time());
-        $testPayload->header['svix-signature'] = 'BAD_SIG_NA_TURE';
+        $testPayload->header['svix-signature'] = 'BAD_SIG_NATURE';
 
         $wh = new \Svix\Webhook($testPayload->secret);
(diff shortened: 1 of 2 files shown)


tasn (approved):
Thanks for the submission!

php/tests/WebhookTest.php
```diff
@@ -52,6 +52,18 @@ public function testInvalidSignatureThrowsException()
         $wh->verify($testPayload->payload, $testPayload->header);
     }
 
+    public function testBadlyFormattedSignatureThrowsException()
+    {
+        $this->expectException(\Svix\Exception\WebhookVerificationException::class);
+        $this->expectExceptionMessage("No matching signature found");
+
+        $testPayload = new TestPayload(time());
+        $testPayload->header['svix-signature'] = 'BAD_SIG_NA_TURE';
```
tasn:
```suggestion
        $testPayload->header['svix-signature'] = 'BAD_SIG_NATURE';
```
Just doing this to make the typo checker happy.

svix-james (approved):

PR #1944, opened 2025-06-23
Automatically send an idempotency key on all requests

svix-mman:
... If the user of the SDK did not already add an idempotency key

Closes: https://github.com/svix/monorepo-private/issues/10619

### DIFF
--- a/ruby/lib/svix/svix_http_client.rb
+++ b/ruby/lib/svix/svix_http_client.rb
@@ -50,6 +50,11 @@ module Svix
       # Add headers
       headers.each { |key, value| request[key] = value }
 
+      # Check if idempotency-key header already exists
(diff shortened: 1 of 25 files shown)


ruby/lib/svix/svix_http_client.rb
```diff
@@ -50,6 +50,11 @@ def execute_request(method, path, **kwargs)
       # Add headers
       headers.each { |key, value| request[key] = value }
 
+      # Check if idempotency-key header already exists
+      if !request.key?("idempotency-key")
+        request["idempotency-key"] = "auto_" + rand(0...(2 ** 64)).to_s
```
tasn: UUID everywhere.

svix-james:
OSS Review: Approved

PR #1947, opened 2025-06-24
Only send `idempotency-key` on post requests

svix-mman:
No need to send this on requests where the server won't use the `idempotency-key` at all

### DIFF
--- a/csharp/Svix/SvixHttpClient.cs
+++ b/csharp/Svix/SvixHttpClient.cs
@@ -59,7 +59,7 @@ namespace Svix
             {
                 headerParams = new Dictionary<string, string>();
             }
-            if (!headerParams.ContainsKey("idempotency-key"))
+            if (!headerParams.ContainsKey("idempotency-key") && method == HttpMethod.Post)
             {
                 headerParams["idempotency-key"] = "auto_" + Guid.NewGuid().ToString();
             }
--- a/go/internal/svix_http_client.go
+++ b/go/internal/svix_http_client.go
@@ -84,7 +84,7 @@ func ExecuteRequest[ReqBody any, ResBody any](
 	}
 
 	key, ok := headerParams["idempotency-key"]
-	if !ok || key == "" {
+	if (!ok || key == "") && strings.ToUpper(method) == "POST" {
 		req.Header.Set("idempotency-key", "auto_"+uuid.New().String())
 	}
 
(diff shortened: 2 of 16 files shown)


csharp/Svix/SvixHttpClient.cs
```diff
@@ -61,7 +61,10 @@ public async Task<ApiResponse<T>> SendRequestAsync<T>(
             }
             if (!headerParams.ContainsKey("idempotency-key"))
             {
-                headerParams["idempotency-key"] = "auto_" + Guid.NewGuid().ToString();
+                if (method == HttpMethod.Post)
```
svix-jplatte: Why no `&&`? Also, probably cheaper to check the method first (not that it matters).

go/internal/svix_http_client.go
```diff
@@ -85,7 +85,9 @@ func ExecuteRequest[ReqBody any, ResBody any](
 
 	key, ok := headerParams["idempotency-key"]
 	if !ok || key == "" {
-		req.Header.Set("idempotency-key", "auto_"+uuid.New().String())
+		if method == "POST" {
+			req.Header.Set("idempotency-key", "auto_"+uuid.New().String())
+		}
 	}
```
svix-jplatte: Same here, nesting doesn't seem necessary. (or nest all the existing code in the method check)

java/lib/src/main/java/com/svix/SvixHttpClient.java
```diff
@@ -56,7 +56,9 @@ public <Req, Res> Res executeRequest(
 
         String idempotencyKey = headers == null ? null : headers.get("idempotency-key");
         if (idempotencyKey == null || idempotencyKey.isEmpty()) {
-            reqBuilder.addHeader("idempotency-key", "auto_" + UUID.randomUUID().toString());
+            if (method.toUpperCase() == "POST") {
```
svix-jplatte: Can you guess my review comment on this :stuck_out_tongue:

rust/src/request.rs
```diff
@@ -113,7 +113,8 @@ impl Request {
         self.header_params
             .insert("svix-req-id", rand::rng().random::<u32>().to_string());
 
-        if !self.header_params.contains_key("idempotency-key") {
+        if self.method == http1::Method::POST && !self.header_params.contains_key("idempotency-key")
+        {
```
svix-jplatte, line 117: I bet clippy complained when you did the same nesting here ^^

svix-mman: > I bet clippy complained when you did the same nesting here ^^

Yup :smile: 

meh

I will clean this up :muscle:

svix-jplatte (approved):

PR #1949, opened 2025-06-26
feat(svix-server): server healthchecks

y-nk:
## Motivation

As discussed in the community slack, the server lacks of healthcheck which is easily available for instrumentalisation (docker-compose, kubernetes, etc)

## Solution

I'm personally inclined not to have a command in the svix-server binary, mostly because it would not self-introspect a running server but rather "just fetch the healthcheck endpoint", therefore adding curl to the docker image seems lighter and less invasive imho.

I've implemented both solutions:

1. a healthcheck command which makes a `HEAD /v1/health` request and returns `Ok()` on 200 (error otherwise)
2. added curl in the server's Dockerfile
3. implemented both tests in the docker-compose

## Note

1. As requested i've also added the healthcheck commands and condition status of the dependencies in the docker-compose file
2. Please be gentle in the review as I'm not fluent in rust

### DIFF
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -14,6 +14,7 @@ RUN cargo chef prepare --recipe-path recipe.json
 FROM chef AS build
 
 RUN apt-get update && apt-get install -y \
+        curl=7.* \
         build-essential=12.* \
         checkinstall=1.* \
         zlib1g-dev=1:* \
--- a/server/docker-compose.yml
+++ b/server/docker-compose.yml
@@ -13,11 +13,18 @@ services:
     ports:
       - "8071:8071"
     depends_on:
-    - pgbouncer
-    - redis
+      pgbouncer:
+        condition: service_healthy
+      redis:
+        condition: service_healthy
 
   postgres:
     image: "docker.io/postgres:13.4"
+    healthcheck:
+      test: ["CMD-SHELL", "pg_isready -U postgres"]
+      interval: 1s
+      timeout: 1s
+      retries: 600
     volumes:
       - "postgres-data:/var/lib/postgresql/data/"
     environment:
@@ -36,10 +43,16 @@ services:
       DB_PASSWORD: "postgres"
       MAX_CLIENT_CONN: 500
     depends_on:
-    - postgres
+      postgres:
+        condition: service_healthy
 
   redis:
     image: "docker.io/redis:7-alpine"
+    healthcheck:
+      test: ["CMD", "redis-cli", "ping"]
+      interval: 1s
+      timeout: 1s
+      retries: 600
     # Enable persistence
     command: "--save 60 500 --appendonly yes --appendfsync everysec"
     volumes:
--- a/server/docker-compose.yml
+++ b/server/docker-compose.yml
@@ -5,6 +5,12 @@ services:
       context: .
       dockerfile: Dockerfile
     image: svix/svix-server
+    healthcheck:
+      test: ["CMD-SHELL", "curl -f http://localhost:8071/api/v1/health"]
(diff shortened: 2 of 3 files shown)


server/Dockerfile
```diff
@@ -14,6 +14,7 @@ RUN cargo chef prepare --recipe-path recipe.json
 FROM chef AS build
 
 RUN apt-get update && apt-get install -y \
+        curl=7.* \
```
tasn: Won't work as this is the build step, not the runtime step, though I see below that you did a different healthcheck command. Let me ask the team to take a look. Thanks!

server/docker-compose.yml
```diff
@@ -5,6 +5,12 @@ services:
       context: .
       dockerfile: Dockerfile
     image: svix/svix-server
+    healthcheck:
+      test: ["CMD-SHELL", "curl -f http://localhost:8071/api/v1/health"]
+      # test: ["CMD-SHELL", "svix-server healthcheck"]
```
tasn:
```suggestion
      test: ["CMD-SHELL", "svix-server healthcheck http://localhost:8071"]
```
Let's maybe have the healthcheck get a server URL as a parameter? This will make it very simple for the cases where we change the listen port.

server/svix-server/src/main.rs
```diff
@@ -113,6 +117,22 @@ async fn main() -> anyhow::Result<()> {
     dotenv().ok();
 
     let args = Args::parse();
+
+    // Handle commands that don't need configuration first
+    if let Some(Commands::Healthcheck) = args.command {
+        let client = reqwest::Client::new();
+        let response = client
+            .head("http://localhost:8071/api/v1/health")
```
tasn: See comment about, should potentially be a parameter. Also, we can probably use `get` rather than `head` and potentially print the information.

tasn: @svix-mman, can you please take a look?

y-nk: @tasn i would understand the parameter if this was the svix-cli, but as svix-server i was assuming self-introspection only.

do you still prefer having this in the bin rather than handling this externally? i personally find the curl approach would be better/less invasive but since we discussed it in slack i was not sure which you'd like.

i need to trim the PR after review.

tasn: @svix-jplatte, but curl is not in the docker image, and pulling it in sounds like unnecessary bloat. It's very little rust.

tasn: @y-nk, the alternative would be to read the configuration to detect the hostname and port, or maybe we can just default to localhost:8071 and have the additional param as optional for people that change the port?
I guess we already have code that does exactly this, as we do it to know the listen address.

server/docker-compose.yml
svix-jplatte: And if it's somehow possible for this to reuse the port from the `ports` part of this service definition, that would be even better. :)

server/svix-server/src/main.rs
svix-jplatte: Already deleted my comment about curl :sweat_smile:

Yeah, let's do it this way but have the server URL (without the `/api/v1/health`) as a command-line argument.

@y-nk That is, just change the new `Healthcheck` variant to

```rust
Healthcheck {
    server_url: String,
}
```

and you can use it here with `Some(Commands::Healthcheck { server_url })` + `format!("{server_url}/api/v1/health")`.

server/docker-compose.yml
tasn: @svix-jplatte, that I don't think we can do unfortunately, though I think/hope people won't change the *internal* port.

svix-mman (approved):

Issue #258, opened 2022-02-24
format!() should not be used inside loops


rimutaka:
`format!()` macro should be avoided in places where it is called repeatedly, e.g. inside loops like this https://github.com/svix/svix-webhooks/blob/main/server/svix-server/src/core/types.rs#L429. Even if that line is not inside a loop it will be called repeatedly on every property it serializes.

An alternative can be picked from this list https://github.com/hoodie/concatenation_benchmarks-rs and also check this discussion on Reddit https://www.reddit.com/r/rust/comments/t06hk7/string_concatenations_benchmarks_updated/.

Additional info: https://github.com/hoodie/concatenation_benchmarks-rs/pull/11 (not merged yet at the time of reporting, but contains useful info on crates implementing fast concatenation).

Some of the macros benchmarked in that study have very good performance, but I would check their source and what they expand to before picking the winner. They all serve a slightly different purpose.

tasn:
Oh nice, thanks for the report, and you're most definitely right!

PR #1951, opened 2025-07-01
server: Reduce allocations

fluiderson:
## Motivation

Closes https://github.com/svix/svix-webhooks/issues/258.

## Solution

I've gone over the server code and replaced the most obvious parts with unnecessary allocations with no or less allocating counterparts. I was really bothered with how signatures string are generated, so I payed more attention to their function, I think there's still a room for improvement, but it'll need more changes all over the server codebase. For now, let's start from this.

I've also updated `base64` to the latest version. The current version doesn't have needed for this PR function to calculate an encoded string length.

Review commit by commit.

### DIFF
--- a/server/svix-server/src/worker.rs
+++ b/server/svix-server/src/worker.rs
@@ -3,7 +3,6 @@
 
 use std::{
     collections::HashMap,
-    fmt::Write,
     sync::{
         atomic::{AtomicU64, AtomicUsize, Ordering},
         Arc, LazyLock,
@@ -16,6 +15,7 @@ use base64::{engine::general_purpose::STANDARD, Engine};
 use chrono::Utc;
 use futures::future;
 use http::{HeaderValue, StatusCode, Version};
+use itertools::Itertools;
 use rand::Rng;
 use sea_orm::{
     prelude::DateTimeUtc, ActiveModelBehavior, ActiveModelTrait, ColumnTrait, DatabaseConnection,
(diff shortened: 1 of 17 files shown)


server/svix-server/src/worker.rs
```diff
@@ -167,21 +168,51 @@ fn sign_msg(
...
+                    const ED25519_LEN: usize = "v1a".len() + base64::encoded_len(64, true).unwrap();
+
+                    ",".len()
+                        + match x.type_() {
+                            EndpointSecretType::Hmac256 => HMAC_LEN,
+                            EndpointSecretType::Ed25519 => ED25519_LEN,
+                        }
+                })
+                .sum::<usize>()
+                + endpoint_signing_keys.len() * " ".len(),
```
svix-jplatte: I am not a fan of this complex length pre-calculation to save a few reallocs. I think you actually got it wrong because this is not counting the first signature (since the first element of `endpoint_signing_keys` has been split off at this point in the code).

How about only getting rid of the intermediate string allocations and skipping the `String::with_capacity`? That should get this function back to close to its initial readability.

svix-jplatte (commented):
Thanks for the PR!

server/svix-server/src/worker.rs
```diff
@@ -169,18 +171,25 @@ fn sign_msg(
...
-            };
-            format!("{version},{}", base64::encode(sig))
+
+    let mut iter = endpoint_signing_keys.iter().map(|x| {
+        let sig = x.sign(main_secret, to_sign.as_bytes());
+        let version = match x.type_() {
+            EndpointSecretType::Hmac256 => "v1",
+            EndpointSecretType::Ed25519 => "v1a",
+        };
+        format!("{version},{}", STANDARD.encode(sig))
```
svix-jplatte: This is the intermediate allocation I was saying could (still) be avoided. Maybe check out [`Itertools::format_with`](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.format_with)? We depend on itertools indirectly anyways.

server/svix-server/src/worker.rs
fluiderson: I've removed this calculation. Yeah, I'd be better if I added a test for it, I overlooked the first sig. It's a shame that the current version doesn't pre-allocate a string considering that sigs are of a fixed length, but you're right that it greatly affects readability.

server/svix-server/src/worker.rs
fluiderson: Will do! I've tried to not introduce more deps. With `itertools`, I think I can remove this folding in my last commit, too.

server/svix-server/src/worker.rs
```diff
@@ -177,10 +180,13 @@ fn sign_msg(
...
                 EndpointSecretType::Ed25519 => "v1a",
             };
-            format!("{version},{}", base64::encode(sig))
+
+            (version, STANDARD.encode(sig))
         })
-        .collect::<Vec<String>>()
-        .join(" ")
+        .format_with(" ", |(version, sign), f| {
+            f(&format_args!("{version},{sign}"))
```
svix-jplatte: One last change request: Can you roll up the `map` into `format_with`?

PR #1952, opened 2025-07-06
fix(go): fix execute with retry empty body bug

mixnblend:
## Motivation

- I am occasionally seeing the Sentry error 

```
Post "<svix-server-url>": http: ContentLength=2461 with Body length 0
```

in our Sentry as a result of an error with the Golang Svix Client.

When reading through the code it seems like this behaviour is caused when calling executeRequestWithRetries when a retry occurs and the body, which is a ReadCloser is read once on the initial call, there's nothing left in the stream, and then a retry results in the above error.

## Solution

- preserve the request body at the start of executeRequestWithRetries
- set it on initial try and on subsequent retries

You can verify this behaviour by running the tests against the current code and observing the failure as below, and then applying the changes and checking the tests pass.

Screenshot 2025-07-06 at 19 57 15

### DIFF
--- a/go/internal/svix_http_client.go
+++ b/go/internal/svix_http_client.go
@@ -128,7 +128,6 @@ func ExecuteRequest[ReqBody any, ResBody any](
 }
 
 func executeRequestWithRetries(client *SvixHttpClient, request *http.Request) (*http.Response, error) {
-
 	var bodyBytes []byte
 	if request.Body != nil {
 		var err error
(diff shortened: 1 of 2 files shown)


svix-jplatte:
Thanks for the PR! Looks like CI is failing, but it's only about the test code. You should fix it of course, but we'll review this soon either way :)

mixnblend:
Thanks so much for getting back to me so quickly. Will check why that failed and fix. Was passing on local.

PR #1965, opened 2025-07-22
cli: add seed command to generate test application, endpoints, & events

### LINKED ISSUE #1950: seed data at startup
### Motivation

When starting svix-server in local (for local dev, mostly) there's often a need to create a project and sometimes additional data (events) to start working. This can be automated with using `svix-cli` either as a separate step, or even integrated into a `docker-compose` job after the boot of the svix-server.

1. an additional step with `svix-cli` would mean changing the startup command from `docker compose up --detach --wait` to be `docker compose up --detach --wait && ./seed-svix.sh`

2. an additional step with docker-compose means more or less the same thing but as a docker service. note that if using `--wait`, the job would end and therefore the exit code would be != 0 (which can be deceptive)

Solution works but somehow doesn't _feel_ clean, or at least _as clean as others_.

### Proposal

Multiple possibilities:

1. do as stack-auth and [propose a limited set of seed data built from env vars](https://github.com/stack-auth/stack-auth/blob/9f794854106b32bf43f87b7dc042e4ef2dbc3113/docker/server/.env#L9-L17). that would mean extra coding and env var support in the seed script.

2. do as postgres/mongo/most-db: add a client in the server container and let users add an `init.d/` script. that would mean adding `svix-cli` in the `svix-server` image and good api support from the svix-cli.

The 2nd solution would also mean that healthcheck command could be added to the `svix-cli` which looks like a better fit for this (rather than within `svix-server`)

KranzAklilu:
This pr aims to close #1950

## Motivation
It would be cool if svix-cli had a way to seed data friction

## Solution
svix seed now:
- creates a test application
- generates predefined user.* event types (signup, signin, signout, deleted)
- with `--reset` flag to delete all existing applications and event types (after confirmation)
- creates a new test application and N endpoints (via --endpoint-count)
- outputs created resources in JSON format
-

### DIFF
--- a/svix-cli/src/cmds/mod.rs
+++ b/svix-cli/src/cmds/mod.rs
@@ -3,4 +3,5 @@ pub mod completion;
 pub mod listen;
 pub mod login;
 pub mod open;
+pub mod seed;
 pub mod signature;
--- a/svix-cli/src/cmds/seed.rs
+++ b/svix-cli/src/cmds/seed.rs
@@ -1,0 +1,180 @@
+use anyhow::Context;
+use clap::Args;
+use serde::{Deserialize, Serialize};
+use svix::api::*;
+
+#[derive(Args)]
+struct SeedOptions {
+    /// Will clear out all the applications and event types
+    #[arg(long, default_value = "false")]
+    pub reset: bool,
+
+    /// The number of endpoints to create (0-10)
+    #[arg(long, value_parser = clap::value_parser!(u8).range(..=10) , default_value = "2")]
+    pub endpoint_count: u8,
+}
+
+#[derive(Args)]
+pub struct SeedArgs {
+    #[clap(flatten)]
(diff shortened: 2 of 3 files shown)


svix-lucho (approved):
This is great! thanks! I think it would be nice if the seed also sent a bunch of messages to the endpoints as an example. But it can be another PR, or feel free to add it to this one.

KranzAklilu:
@svix-lucho I've added the ability to send messages. Let me know if there's anything else. I'm happy to work on this :)

svix-lucho:
Also, since you asked 😄 , the last thing I think would be good to have here are schemas in the event types (see https://api.svix.com/docs#tag/Event-Type/operation/v1.event-type.create) and then the messages can use some example payload that follows that schema instead of `{}`.

svix-lucho (approved):

Issue #1154, opened 2024-01-04
Make sure all of the libraries correctly handle trailing slashes


tasn:
Both of these should work:
1. https://api.svix.com/
2. https://api.svix.com

At the moment only the latter works in JS. Not sure about the rest.

PR #1969, opened 2025-07-27
Fix: Handle trailing slashes in server URLs across all libraries

jaiganeshs21:
PR Summary

  ## Summary
  • Implements consistent trailing slash handling across JavaScript, Python, Ruby, and Rust libraries
  • Adds robust URL normalization to prevent double slashes in API requests
  • Maintains backward compatibility while fixing edge cases with multiple trailing slashes
  • Ensures uniform behavior across all SDK languages for better developer experience

Implementation details 
  - Add consistent trailing slash normalization in JavaScript, Python, Ruby, and Rust
  - JavaScript: Implement regex-based URL and path normalization
  - Python: Use rstrip('/') for clean URL handling
  - Ruby: Apply regex substitution for multiple slash removal
  - Rust: Loop-based trailing slash elimination
  - Add comprehensive test suites for all libraries
  - Ensure backward compatibility and consistent behavior

Closes
Issue - https://github.com/svix/svix-webhooks/issues/1154

### DIFF
--- a/javascript/src/index.ts
+++ b/javascript/src/index.ts
@@ -49,7 +49,11 @@ export class Svix {
 
   public constructor(token: string, options: SvixOptions = {}) {
     const regionalUrl = REGIONS.find((x) => x.region === token.split(".")[1])?.url;
-    const baseUrl: string = options.serverUrl ?? regionalUrl ?? "https://api.svix.com";
+    const baseUrl: string = (
+      options.serverUrl ??
+      regionalUrl ??
+      "https://api.svix.com"
+    ).replace(/\/+$/, "");
 
     this.requestCtx = { baseUrl, token, timeout: options.requestTimeout };
   }
--- a/javascript/src/request.ts
+++ b/javascript/src/request.ts
@@ -109,7 +109,10 @@ export class SvixRequest {
   }
 
   private async sendInner(ctx: SvixRequestContext): Promise<Response> {
-    const url = new URL(ctx.baseUrl + this.path);
+    // Ensure path always starts with a single slash
+    const normalizedPath = this.path.replace(/^\/+/, "/");
+    // Combine normalized base URL and path
+    const url = new URL(ctx.baseUrl + normalizedPath);
     for (const [name, value] of Object.entries(this.queryParams)) {
       url.searchParams.set(name, value);
     }
--- a/python/tests/test_trailing_slash.py
+++ b/python/tests/test_trailing_slash.py
@@ -1,0 +1,37 @@
+import pytest
+from svix.api import Svix, SvixOptions
+
+
+def test_trailing_slash_handling():
+    """Test that trailing slashes are properly removed from server URLs."""
+    
+    # Test with trailing slash
+    svix_with_slash = Svix("key", SvixOptions(server_url="http://localhost:8000/"))
+    assert svix_with_slash._client.base_url == "http://localhost:8000"
+    
+    # Test without trailing slash
+    svix_without_slash = Svix("key", SvixOptions(server_url="http://localhost:8000"))
+    assert svix_without_slash._client.base_url == "http://localhost:8000"
+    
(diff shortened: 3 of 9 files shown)


jaiganeshs21:
@tasn Can you please review this PR?

javascript/src/request.ts
```diff
@@ -109,7 +109,10 @@ export class SvixRequest {
   }
 
   private async sendInner(ctx: SvixRequestContext): Promise<Response> {
-    const url = new URL(ctx.baseUrl + this.path);
+    // Ensure path always starts with a single slash
+    const normalizedPath = this.path.replace(/^\/+/, "/");
```
svix-jplatte, line 113: Why do this, and only for JavaScript? This is an internal API, and I'm pretty sure it's only ever called with paths that already begin with (one) slash, I think?

javascript/src/index.ts
```diff
@@ -49,7 +49,11 @@ export class Svix {
 
   public constructor(token: string, options: SvixOptions = {}) {
     const regionalUrl = REGIONS.find((x) => x.region === token.split(".")[1])?.url;
-    const baseUrl: string = options.serverUrl ?? regionalUrl ?? "https://api.svix.com";
+    const baseUrl: string = (
+      options.serverUrl ??
+      regionalUrl ??
+      "https://api.svix.com"
+    ).replace(/\/+$/, "");
```
svix-jplatte, line 56: I think it's cleaner to only run the `replace` if the user specified a server URL. The fallback values all don't end in `/` and I wouldn't want us to become sloppy with things like that internally.

Same for all the other SDKs.

python/tests/test_trailing_slash.py
svix-jplatte, line 1: Can you add a final newline to the end of this file?

ruby/spec/trailing_slash_spec.rb
svix-jplatte, line 1: Same here.

ruby/spec/trailing_slash_spec.rb
```diff
@@ -0,0 +1,80 @@
+RSpec.describe "Trailing Slash Handling" do
```
svix-jplatte, line 1: I'm surprised you copy-pasted bits from the client initialization here rather than actually calling the client initialization. This way the test could easily get out of sync with the actual code. This being Ruby, should it not be simple to check the internal state of the client after construction, like you did with Python? (I haven't seriously used Ruby, but I know it allows all sorts of monkey patching)

rust/tests/trailing_slash_test.rs
svix-jplatte, line 1: If you make this a unit test rather than an integration test, you can make assertions about the internal state of the `Svix` client. The way it works is you create an inline module within `src/api/client.rs` like

```rust
#[cfg(test)]
mod tests {
    // write tests here
}
```

rust/tests/trailing_slash_test.rs
```diff
@@ -0,0 +1,62 @@
...
+        ("http://localhost:8000///", "http://localhost:8000"),
+    ];
+
+    for (input, expected) in test_cases {
+        let mut result = input.to_string();
+        while result.ends_with('/') {
+            result.pop();
+        }
+        
+        assert_eq!(result, expected, "Failed for input: {}", input);
```
svix-jplatte, line 17:
```suggestion
        assert_eq!(result, expected, "Failed for input: {input}");
```

rust/tests/trailing_slash_test.rs
```diff
@@ -0,0 +1,62 @@
...
+        println!("✅ {} → {}", input, result);
```
svix-jplatte, line 18: stdout is only printed if a test fails, so this seems kind of pointless.

svix-jplatte (commented):
Thanks for the PR! Left some comments.

PR #1971, opened 2025-07-28
fix: typo

KranzAklilu:
## Motivation
fix a typo

### DIFF
--- a/server/svix-server/src/core/idempotency.rs
+++ b/server/svix-server/src/core/idempotency.rs
@@ -299,7 +299,7 @@ async fn lock_loop(
     }
 }
 
-/// Resolves the service and cache the result assuming the response is successful
+/// Resolve the service and cache the result assuming the response is successful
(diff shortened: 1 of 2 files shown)


server/svix-server/src/core/idempotency.rs
```diff
@@ -299,7 +299,7 @@ async fn lock_loop(
     }
 }
 
-/// Resolves the service and chaches the result assuming the response is successful
+/// Resolves the service and cache the result assuming the response is successful
```
svix-jplatte: This makes the sentence inconsistent. It needs to be either "Resolves [...] and caches" or "Resolve [...] and cache". I guess the latter might be a bit nicer.

```suggestion
/// Resolve the service and cache the result assuming the response is successful
```

svix-jplatte (approved):

svix-mman (approved):

PR #1974, opened 2025-07-31
js: Add support for retry schedule

KranzAklilu:
Add support for retry schedule in the JavaScript lib

## Motivation
Work in progress for #782

Would be cool if you could specify the number of retries for failed requests plus exact delay schedule for retries.
Currently, number of retries is hardcoded (2) retry delays are hard coded.

## Solution
Provide the ability for the user to specify delays as an option (retryScheduleInMs)

**Updated the number of retry logic**. if user specifies numOfRetries = 3, we actually make 4 requests as we 
shouldn't count the initial request.


javascript/src/index.ts
```diff
@@ -34,6 +34,15 @@ export interface SvixOptions {
...
   requestTimeout?: number;
+  /** List of delays (in milliseconds) to wait before each retry attempt.
+   *  Takes precedence over numRetries.
+   */
+  retryScheduleInMs?: number[];
+  /** The number of times the client will retry if a server-side error
+   *  or timeout is received.
+   *  Default: 2
+   */
+  numRetries?: number;
```
svix-lucho: I think this should be enforced in the type itself (allowing either `retryScheduleInMs` or `numRetries`, but not both)

KranzAklilu: `retryScheduleInMs` is meant for users that need a bit more customization. I don't think removing `numRetries` makes a lot of sense because most people are probably going to use that. I also think with `retryScheduleInMs` you get more customization that'll come in handy for users that need a bit more delay between retires.

svix-lucho: I'm not saying you remove `numRetries`. I mean that the Typescript type should enforce that you pass one **or** the other, but not both at the same time. 

`new Svix(token, { numRetries: 2, retryScheduleInMs: [1, 2, 3] })` should fail compilation

`new Svix(token, { numRetries: 2 })` and `new Svix(token, { retryScheduleInMs: [1, 2, 3] })` should both work

svix-jplatte:
Thanks!

PR #1975, opened 2025-07-31
rust: Add support for retry schedule

KranzAklilu:
Add support for retry schedule in the rust lib

## Motivation
Work in progress for #782

Would be cool if one could specify an exact delay schedule for retries.
Currently, retry delays are hard coded.

## Solution

Updated the number of retry logic. if user specifies num_retries = 3, we actually make 4 requests as we
shouldn't count the initial request.

### DIFF
--- a/rust/src/request.rs
+++ b/rust/src/request.rs
@@ -90,7 +90,7 @@ impl Request {
     }
 
     pub async fn execute<T: DeserializeOwned>(self, conf: &Configuration) -> Result<T, Error> {
-        match self.execute_with_backoff(conf, conf.num_retries).await? {
+        match self.execute_with_backoff(conf).await? {
             // This is a hack; if there's no_ret_type, T is (), but serde_json gives an
             // error when deserializing "" into (), so deserialize 'null' into it
             // instead.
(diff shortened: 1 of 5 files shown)


rust/src/request.rs
```diff
@@ -102,26 +102,34 @@ impl Request {
...
             self.header_params
                 .insert("idempotency-key", format!("auto_{}", uuid::Uuid::new_v4()));
         }
 
+        let mut retries: u32 = conf
+            .retry_schedule_in_ms
+            .as_ref()
+            .map(|v| v.len() as u32 + 1)
+            .unwrap_or(conf.num_retries);
+
```
svix-jbrown: I like what you're doing here, but this logic feels a little messy to me.

What do you think if we instead make a single iterator describing how many retries to do? Something like

```rust
    let mut retries = conf.retry_schedule_in_ms
      .map(|s| box_retry_iterator(s.clone()))
      .unwrap_or(box_retry_iterator(
        std::iter::successors(Some(Duration::from_millis(20)), |last_backoff| {
            Some(MAX_BACKOFF.min(*last_backoff * 2))
        })
        .take(conf.num_retries),
    ));
```

where `box_retry_iterator` is just 

```rust
fn box_retry_iterator<I: IntoIterator<Item = Duration> + 'static>(
    i: I,
) -> Box<dyn Iterator<Item = Duration>> {
    Box::new(i.into_iter())
}
```

Then inside the loop, instead of doing `retries == 0`, we could do `let next_backoff = retries.next()` and break if that's None.

What do you think?

KranzAklilu: Yup, that looks super clean, thanks. Let me know if there are any other suggestions.

KranzAklilu:
Hello @svix-jbrown. I've implemented the requested changes. Let me know if you've anything else in mind. I would love to get this merged. Thanks

svix-jbrown (approved):

PR #1980, opened 2025-08-05
Re-add endpoint.transformation-partial-update as deprecated

svix-jplatte:


codegen/codegen.toml
```diff
@@ -112,6 +111,7 @@ output_dir = "go/models"
 [[go.task]]
 template = "templates/go/api_resource.go.jinja"
 output_dir = "go/internalapi"
+input_files = ["lib-openapi.json"]
```
svix-jplatte, line 114: This is only required because the filtering CLI arguments below don't get applied to the `.ron` file. I should fix that, but this isn't _wrong_ either, so I think we can merge as-is.

PR #1984, opened 2025-08-06
codegen: Remove usage of `cargo fix` from the `extra_shell_commands` …

svix-mman:
This speeds up the `regen_openapi.py` by a bunch.
Now we don't need to wait for `cargo fix`


codegen/templates/svix-cli/api_resource.rs.jinja
svix-jplatte, line 1: We should also add a `@generated` comment at the top of this file.

svix-jplatte (dismissed):
I would kinda prefer to add conditional imports, but can do that after. This is definitely an improvement.

svix-jplatte (approved):

PR #1985, opened 2025-08-06
Log codegen duration

svix-mman:
I want to see how long it takes to pull the image / run the codegen in CI.

No downsides, only upside.


svix-jplatte (commented):
`time.time()` is wall clock, it can go backwards if the system time changes (daylight savings and other shenanigans). Use [`perf_counter`](https://docs.python.org/3/library/time.html#time.perf_counter).

PR #1993, opened 2025-08-11
Update to latest debian version

svix-mman:
Also updated the rust version of the dockerfile to 1.89.0 (there was no rust-1.86-trixie image)

### DIFF
--- a/svix-cli/Dockerfile
+++ b/svix-cli/Dockerfile
@@ -1,10 +1,10 @@
 # Base build
-FROM rust:1.86-slim-bookworm AS build
+FROM rust:1.89-slim-trixie AS build
 
 RUN apt-get update && apt-get install -y \
     build-essential=12.* \
     checkinstall=1.* \
-    curl=7.* \
+    curl=8.* \
     libssl-dev=* \
     pkg-config=1.8.* \
     zlib1g-dev=1:* \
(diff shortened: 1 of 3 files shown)


svix-cli/Dockerfile
```diff
@@ -37,7 +37,7 @@ RUN touch src/main.rs
 RUN cargo build --release --frozen
 
 # Production
-FROM debian:bookworm-slim AS prod
+FROM debian:trixie-slim AS prod
```
svix-jbrown, line 40: somewhat out of scope here, but we actually don't dynamically link anything except libc, so we could do a way more minimal image for `svix-cli`. do we need `ncurses` and `perl` and `bash` and `findutils` and even `apt`? probably not

svix-mman: Correct, I think we should do what `uv` does for their docker image https://github.com/astral-sh/uv/blob/main/Dockerfile

svix-mman: I will open a new PR for that.

svix-mman: @svix-jbrown On a second thought, I don't think it's worth the time
1. Our images are 90Mb, and about 40Mb compressed. 
2. No one uses these images. (https://hub.docker.com/r/svix/svix-cli)

PR #1997, opened 2025-08-12
Bump codegen image

svix-mman:
Also fixes issue where biome deleted the `this file is @generated` comment from `javascript/src/api/operationalWebhook.ts`, I was able to fix this by updating biome, and adding an extra newline into the template.

Also added the `this file is @generated` comment to some rust and python code.

### DIFF
--- a/javascript/src/index.ts
+++ b/javascript/src/index.ts
@@ -11,8 +11,8 @@ import { Integration } from "./api/integration";
 import { Message } from "./api/message";
 import { MessageAttempt } from "./api/messageAttempt";
 import { OperationalWebhook } from "./api/operationalWebhook";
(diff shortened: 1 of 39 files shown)


javascript/src/index.ts
svix-jplatte, line 1: Uhhh.. The previous ordering of the imports made way more sense??

svix-jplatte: Is this a bug in biome? :sweat_smile:

svix-mman: > Uhhh.. The previous ordering of the imports made way more sense??

New default in biome :shrug: 

> Is this a bug in biome? 😅

The deleting comments? yup.

PR #1998, opened 2025-08-12
Log generated codegen files

svix-mman:
The goal is
1. If the same file was generated in more than 1 codegen run, log an error
2. If a file was not generated with the `this file is @generated` comment in the first 2 lines, log an error
3. For every file on disk with `this file is @generated`, if that file was not generated in the latest `./regen_openapi.py`, log an error


regen_openapi.py
```diff
@@ -290,6 +314,51 @@ def pull_image():
...
+            if path in processed_files and path not in allowed_to_be_generated_twice:
+                raise Exception(
+                    f"Unexpected. same file `{path}` was generated twice, in 2 different codegen tasks"
+                )
+            processed_files.add(path)
+
+            # ensure each file has `this file is @generated` in the first (2) lines
+            with open(path, "r") as f:
+                first_line = "".join(f.readlines()[0:2])
+                assert "this file is @generated" in first_line.lower(), (
```
svix-jplatte: Actually, there's that `@` there because `@generated` is already special-cased by some tools. Let's only grep for that, then we can change the wording too, without changing this script.

```suggestion
                assert "@generated" in first_line.lower(), (
```

PR #1999, opened 2025-08-13
js: Use native fetch

### LINKED ISSUE #1447: Deprecation Warning: Punycode Module
### Version
Svix version 1.35.0

### Platform
Node.js v22.9.0 
Express.js version 4.21.0.

### Description
I'm currently using the Svix library in my project and have encountered a deprecation warning related to the punycode module. After investigating, I identified the following dependency chain:

Svix depends on svix-fetch
svix-fetch depends on node-fetch (version 2.7.0)
node-fetch relies on whatwg-url, which in turn depends on punycode
Could you provide guidance on how to address this warning? Is there an upcoming update or an alternative approach I should consider to avoid this deprecation?

 I tried this code:

"dev": "nodemon src/index.js"

I expected to see this happen: Run the server without warnings.

Instead, this happened: 
(node:2701) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)

Thank you in advance for your assistance!

svix-jplatte:
## Motivation

Fixes #1447.

## Solution

Use native `fetch` instead of `svix-fetch` which wraps `node-fetch`. This was used to polyfill this API for node.js users, but node.js has been shipping a native `fetch` implementation since v18, and v20 is currently the oldest supported release.

I have verified that node.js supports keep-alive and enables it by default (which is what `svix-fetch` added over plain `node-fetch`).

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## Unreleased
+* Libs/JavaScript: Use native `fetch` API
+
 ## Version 1.71.0
 * Bridge: Add `http` output to `receivers`
   * See [`bridge/svix-bridge.example.receivers.yaml`](./bridge/svix-bridge.example.receivers.yaml) for a usage example
--- a/javascript/package.json
+++ b/javascript/package.json
@@ -33,7 +33,6 @@
     "@types/node": "^22.7.5",
     "es6-promise": "^4.2.8",
     "fast-sha256": "^1.3.0",
-    "svix-fetch": "^3.0.0",
     "url-parse": "^1.5.10",
     "uuid": "^10.0.0"
   },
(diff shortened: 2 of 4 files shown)


svix-jplatte:
For posterity, I had a small internal conversation about this re. keep-alive being on by default. There are barely any docs out there, and Node's older [`http` API](https://nodejs.org/api/http.html) having keep-alive off by default doesn't help either. The only docs I could find regarding `fetch` is the docs for the underlying `undici` library, which says [here](https://github.com/nodejs/undici/blob/main/docs/docs/api/Dispatcher.md#parameter-dispatchoptions) that its `reset` parameter (opposite of keepalive) is off by default.

svix-lucho (approved):

PR #2002, opened 2025-08-14
libs: Ignore unknown keys

svix-mman:
If the server responds with extra unknown keys, ignore them.

Java and Kotlin, this involves adding a setting to the deserializer "object thingy" 

For the rest, we already ignore unknown keys. So I added some tests.

### DIFF
--- a/kotlin/lib/src/main/kotlin/SvixHttpClient.kt
+++ b/kotlin/lib/src/main/kotlin/SvixHttpClient.kt
@@ -14,6 +14,7 @@ internal constructor(
     private val baseUrl: HttpUrl,
     private val defaultHeaders: Map<String, String>,
     private val retrySchedule: List<Long>,
+    private val jsonDeserializer: Json = Json { ignoreUnknownKeys = true }
 ) {
     private val client: OkHttpClient = OkHttpClient()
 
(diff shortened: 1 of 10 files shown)


kotlin/lib/src/main/kotlin/SvixHttpClient.kt
```diff
@@ -44,8 +45,8 @@ internal constructor(
         }
 
         if (headers?.get("idempotency-key") == null && method == "POST") {
-                val uuid = UUID.randomUUID().toString()
-                reqBuilder.addHeader("idempotency-key", "auto_" + uuid)
+            val uuid = UUID.randomUUID().toString()
+            reqBuilder.addHeader("idempotency-key", "auto_" + uuid)
```
svix-jplatte, line 49: So we don't have formatting checks in CI for Kotlin, huh?

rust/tests/it/wiremock_tests.rs
```diff
@@ -131,3 +131,33 @@ async fn test_client_provided_idempotency_key_is_not_overridden() {
...
+    );
+
+    svx.application().list(None).await.unwrap();
+
+    let requests = mock_server
+        .received_requests()
+        .await
+        .expect("we should have sent a request");
+
+    assert_eq!(1, requests.len());
```
svix-jplatte: Pretty sure you can set the expected number of requests on the `Mock` object and only call `verify` at the end (or not, I think it's even verified on `Drop` of the `MockServer`).

svix-jplatte (approved):

kotlin/lib/src/main/kotlin/SvixHttpClient.kt
svix-mman: > So we don't have formatting checks in CI for Kotlin, huh?

Nope.

rust/tests/it/wiremock_tests.rs
svix-mman: Correct. TIL!

svix-jplatte (approved):

PR #2005, opened 2025-08-18
style: use pre-commit to enforce line endings/etc

svix-jbrown:
This sets up [`pre-commit`](https://pre-commit.com/) to enforce line endings, trailing whitespace, bad symlinks, and other surface-level details. It also fixes all existing violations.


codegen/templates/go/types/struct_enum.go.jinja
```diff
@@ -25,7 +25,7 @@ import (
 //   - {% for v in vars %}"{{ v.name }}"{% if not loop.last %},{% endif %}{% endfor %}: {{ use_type }}
 {%- endfor -%}
 type {{ ty_name }} struct {
-    {% include "types/struct_fields.go.jinja" -%}	
+    {% include "types/struct_fields.go.jinja" -%}
```
svix-jplatte: Mind also changing the four spaces at the beginning here to a tab, like in the rest of the file?

server/svix-server/migrations/20221104170041_add_applicationmetadata.down.sql
svix-jplatte, line 1: Please exclude the migrations directory, unless you know for a fact that SQLx ignores trailing whitespace when computing the migration's hash (I don't think it does).

svix-jbrown: I don't really want to exclude it forever; otherwise we'll end up with more bad content in the future. I can exclude these specific files, though.

PR #2030, opened 2025-08-25
Publish version 1.75.0

svix-jbrown:
## What's Changed

* Bridge, CLI, Server: Modify Dockerfiles to use cache mounts for improved build time; these now require Docker 1.2 or later to build
* Libs/JS: Add support for custom retry schedule (thanks @KranzAklilu)


svix-jplatte:
Is the dependency upgrade actually relevant for users? Not really a major dependency IMHO.

svix-jbrown:
I'm fine removing it.

We also need to redo this on top of #2031 to fix the lint problems.

svix-jplatte (dismissed):

svix-jplatte (approved):

PR #2036, opened 2025-08-28
No more bitnami/kafka

svix-mman:


svix-jbrown (approved):
something something left pad something something

PR #2047, opened 2025-09-02
bridge: Simplify webhook_receiver::route

svix-jplatte:
This function was overly complex. Found while looking into https://github.com/svix/monorepo-private/issues/11136.

### DIFF
--- a/bridge/svix-bridge/src/webhook_receiver/mod.rs
+++ b/bridge/svix-bridge/src/webhook_receiver/mod.rs
@@ -92,39 +92,27 @@ async fn route(
         transformer_tx,
     }): State<InternalState>,
     req: SerializableRequest<Unvalidated>,
-) -> http::StatusCode {
-    if let Some(IntegrationState {
+) -> Result<http::StatusCode, http::StatusCode> {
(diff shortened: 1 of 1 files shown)


bridge/svix-bridge/src/webhook_receiver/mod.rs
```diff
@@ -92,39 +92,27 @@ async fn route(
...
+    })?;
+
+    let payload = parse_payload(
+        req.payload(),
+        transformation.as_ref(),
+        transformer_tx.clone(),
+    )
+    .await?;
+
+    handle(payload, output.clone()).await
```
svix-jbrown, line 115: nit: I think I prefer `Arc::clone(output)` over `output.clone()` to make it clear that we aren't cloning the inner type.  I feel like there used the be a clippy for calling `.clone` on an Arc, but maybe I'm misremembering.

svix-jplatte: There is a lint, but it's opt-in (allow-by-default). Can try doing that as another PR.

Issue #93, opened 2021-07-10
Allow signature validation without timestamp tolerance enforcement


svix-frank:
Should keep it enabled by default, but we might want to allow custom tolerances or allow to disable the tolerance enforcement altogether.

Can be useful if you want to verify old payloads or if you have a bad clock.

masahiro-yamamoto-rc:
@svix-frank 
Would you let me know why this issue was closed, please?

We are in a situation where we need to use a custom tolerance.
We are going to create a subclass of `__verify_timestamp`, but wondering why such a simple method is prefixed with double underscore.
https://github.com/svix/svix-webhooks/blob/03ab8b993df6094329ac47f7e2f130fd53e8fc24/python/svix/webhooks.py#L67-L79

dimaqq:
Asking for the same tweak but for a different reason than OP: sometimes, things go bad in production and webhooks need to be retried for business reasons.

Ref: https://news.ycombinator.com/item?id=27823109

tasn:
I reopened it. No idea why it was closed, but at the very least it's useful for tests (if not a variety of other use-cases too).

> Asking for the same tweak but for a different reason than OP: sometimes, things go bad in production and webhooks need to be retried for business reasons.

You should generate a new signature when you retry (Svix already does it).

svix-jplatte:
There's a function `VerifyIgnoringTimestamp` for this nowadays. Please let us know if we closed this in error.

seanarnold:
@svix-jplatte it appears `VerifyIgnoringTimestamp` is only in the Go library, is that intended?

ogenodisho:
Hi @svix-jplatte , I raised a PR to address this issue: #2055

PR #2055, opened 2025-09-04
[Python] Make the webhook tolerance configurable.

ogenodisho:
## Motivation

This is to address #93

The motivation is described in detail in the issue.

## Solution

Introduce a `timedelta` parameter in the Webhook init function which specifies the tolerance.

Consumers may leave this blank and it will default to 5 minutes, or provide `timedelta.max` to essentially bypass tolerance checking.

This change is backwards compatible.


python/svix/webhooks.py
```diff
@@ -18,8 +18,9 @@ class WebhookVerificationError(Exception):
 class Webhook:
     _SECRET_PREFIX: str = "whsec_"
     _whsecret: bytes
+    _whtolerance: timedelta
 
-    def __init__(self, whsecret: t.Union[str, bytes]):
+    def __init__(self, whsecret: t.Union[str, bytes], whtolerance: timedelta = timedelta(minutes=5)):
```
svix-lucho, line 23: I think a better approach would be adding the `VerifyIgnoringTimestamp` method, like we have in Go, as it was discussed in https://github.com/svix/svix-webhooks/issues/93

svix-lucho: My concern with this is that it might make it 'easy' to use potentially insecure tolerances. With a separate method it's clear that the security implications are different.

ogenodisho: Hi @svix-lucho, thanks for the review.

I did see that part in the related issue about verifying the webhook while ignoring the timestamp. However, my reasoning for doing it this way is that I think it's actually safer (and more flexible) than allowing consumers to disable the timestamp verification entirely.

If a consumer disables it entirely, they're much more susceptible to replay attacks. However, if the tolerance is configurable, they have the flexibility of tuning the tolerance to suit their use cases, while still being protected against potential replay attacks to a certain degree.

In regards to your second message, I wouldn't be opposed to introducing a `verify_using_custom_tolerance` method instead.

Please let me know your thoughts.

PR #2059, opened 2025-09-09
Remove msg dest 1

jaymell:


server/svix-server/migrations/20250905195339_messageattempt_updates.up.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE messageattempt DROP CONSTRAINT fk_messageattempt_msg_dest_id_messagedestination;
```
svix-jplatte, line 1: Why is this constraint dropped, actually? Does it not make cleanup of the old entries harder? Does the expired message cleaner still clean up everything for the older messages?

server/svix-server/src/v1/endpoints/endpoint/recovery.rs
```diff
@@ -128,7 +141,9 @@ pub(super) async fn recover_failed_webhooks(
     let db = db.clone();
     let queue_tx = queue_tx.clone();
     tokio::spawn(async move {
-        bulk_recover_failed_messages(db, queue_tx, app, endp, since, until).await
+        if let Err(e) = bulk_recover_failed_messages(db, queue_tx, app, endp, since, until).await {
+            tracing::error!("Error recovering failed messages: {}", e);
```
svix-jplatte:
```suggestion
            tracing::error!("Error recovering failed messages: {e}");
```

server/svix-server/migrations/20250905195339_messageattempt_updates.up.sql
svix-james: Initially I was just trying to remove as many traces of messagedestination as possible, but I think you're probably right in that removing it is not necessary. I don't think the expired message cleaner touches anything but message payloads, though, so keeping it won't actually help there. At this point I'd like to go ahead and remove it for consistency with downstream.

PR #2061, opened 2025-09-11
Fix sending status

jaymell:


server/svix-server/Cargo.toml
```diff
@@ -100,6 +100,7 @@ tikv-jemallocator = { version = "0.6", optional = true }
 assert_matches = "1.5.0"
 axum-server = { version = "0.7.1", features = ["tls-openssl"] }
 ctor = "0.2.7"
+wiremock = { version = "=0.6.4" }
```
jaymell, line 103: The latest version (0.6.5) would not compile for me locally, so I pinned it.

PR #2066, opened 2025-09-17
Replace prettier with biome

svix-mman:
Replace `prettier` with biome for the whole JS sdk
The codegen already uses biome to format the generated code.

### DIFF
--- a/javascript/src/HttpErrors.ts
+++ b/javascript/src/HttpErrors.ts
@@ -1,6 +1,6 @@
 export class HttpErrorOut {
-  "code": string;
-  "detail": string;
+  code!: string;
+  detail!: string;
 
   static readonly discriminator: string | undefined = undefined;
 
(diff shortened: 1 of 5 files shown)


javascript/src/HttpErrors.ts
```diff
@@ -38,15 +38,15 @@ export class ValidationError {
   /**
    * The location as a [`Vec`] of [`String`]s -- often in the form `[\"body\", \"field_name\"]`, `[\"query\", \"field_name\"]`, etc. They may, however, be arbitrarily deep.
    */
-  "loc": Array<string>;
+  loc!: Array<string>;
   /**
    * The message accompanying the validation error item.
    */
-  "msg": string;
+  msg!: string;
```
svix-mman, line 45: these properties will be assigned before they're used.

svix-lucho (approved):

PR #2072, opened 2025-09-22
ts: Fix all outstanding biome lint rules

svix-mman:
Review this PR commit by commit. each commits fixes a different rule 

2. [Fix `lint/correctness/noUnusedFunctionParameters`](https://github.com/svix/svix-webhooks/pull/2072/commits/0fd10198caff546cda41486f31785eeb17adc9cc), renamed params from `<param>` to `_<param>`  
3. [Fix `lint/style/noNonNullAssertion`](https://github.com/svix/svix-webhooks/pull/2072/commits/b0b86dabb550fea74b50fe7dff94264351bf02d0), Remove some not null assertions from the tests
4. [Fix(ignore) `lint/suspicious/noImplicitAnyLet`](https://github.com/svix/svix-webhooks/pull/2072/commits/c71fb196ccd7cd9a69c80fee8b0cb1f3fc6d11a9): Ignore this warning (since this code is generated, I am comfortable with ignoring the type system) 
5. [Fix(ignore) `lint/suspicious/noEmptyInterface`](https://github.com/svix/svix-webhooks/pull/2072/commits/52bceca38d7cc7ee237fdd916ba2253dc218138c), we could technically export these as  types instead of interfaces, but that would be a breaking change

### DIFF
--- a/codegen/templates/javascript/types/struct_enum.ts.jinja
+++ b/codegen/templates/javascript/types/struct_enum.ts.jinja
@@ -72,6 +72,7 @@ export const {{ type.name | to_upper_camel_case }}Serializer = {
     },
 
     _toJsonObject(self: {{ type.name | to_upper_camel_case }}): any {
+        // biome-ignore lint/suspicious/noImplicitAnyLet: the return type needs to be any
         let {{ content_field_name }};
         switch (self.{{ disc_field_name }}) {
         {%- for variant in type.variants %}
--- a/codegen/templates/javascript/types/struct_enum.ts.jinja
+++ b/codegen/templates/javascript/types/struct_enum.ts.jinja
@@ -17,6 +17,7 @@ interface _{{ type_name }}Fields {
 {# export the types for enum variants without a config -#}
 {% for variant in type.variants %}
     {% if variant.schema_ref is not defined %}
+// biome-ignore lint/suspicious/noEmptyInterface: backwards compat
 interface {{ type_name }}{{ variant.name | to_upper_camel_case }}{{ type.content_field | to_upper_camel_case }} {}
     {% endif %}
 {% endfor %}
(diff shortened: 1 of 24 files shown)


svix-mman:
(the codegen task is not running because this PR is not targeting main)

codegen/templates/javascript/types/struct_enum.ts.jinja
```diff
@@ -65,7 +65,7 @@ export const {{ type.name | to_upper_camel_case }}Serializer = {
             {{ disc_field_name }},
             {{ content_field_name }}:get{{ content_field_name | to_upper_camel_case }}({{ disc_field_name }}),
             {% for field in type.fields -%}
-                {% set field_expr %}object['{{ field.name }}']{% endset -%}
```
svix-lucho, line 68: just calling it out that this won't work if field name is `hello-world`, for example. Not sure if we'll ever care about that (and it doesn't affect us now), but just worth noting.

svix-jplatte: Yeah I'd prefer if we kept the universally applicable syntax.

svix-mman: I think if that ever happens, I will just modify the templates back :shrug: 
but for now I just want to make biome happy :face_holding_back_tears:

svix-lucho: maybe it's fine to ignore that rule, since it actually makes sense to use the literal keys in codegen

svix-mman: whoops, I missed @svix-jplatte's comment when I replied.
I guess there is no harm in ignoring that 1 rule, especially when it makes our templates more robust.

svix-jplatte (approved):

PR #2080, opened 2025-10-02
go: Add a test for NewMessageInRaw return value serialization

svix-jplatte:
There's a report of this method not doing what it's supposed to do. Here's a test for the JSON form of its output. Looks fine to me, but I'm probably missing something.


svix-jplatte:
I see https://github.com/svix/svix-webhooks/pull/2053 already added a more comprehensive test.

PR #2081, opened 2025-10-03
Replace the `TWINE_PASSWORD` with trusted publishing

svix-mman:
Instead of using the `TWINE_PASSWORD`, Tom added support for oidc/trusted publishing.

The `TWINE_PASSWORD` was also removed. So this is now the only way to publish to our pypi package.

### DIFF
--- a/.github/workflows/python-release.yml
+++ b/.github/workflows/python-release.yml
@@ -10,6 +10,8 @@ on:
 
 jobs:
   build:
+    permissions:
+      id-token: write
     name: Build distributions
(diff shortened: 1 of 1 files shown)


.github/workflows/python-release.yml
```diff
@@ -10,6 +10,8 @@ on:
 
 jobs:
   build:
+    permissions:
+      id-token: write
```
svix-mman, line 14: This is needed for oidc publishing
https://github.com/pypa/gh-action-pypi-publish#usage

PR #2091, opened 2025-10-14
js: Add option to configure the fetch method

svix-mman:
Add new option in the js `SvixOptions`. `fetch`.
Allows user to override the fetch impl, while still keeping the default retry/idempotency headers behavior. 

This came from a community slack request. https://svixcommunity.slack.com/archives/C022BJ2EBHV/p1760452704322919

### DIFF
--- a/javascript/src/request.ts
+++ b/javascript/src/request.ts
@@ -24,6 +24,11 @@ export type SvixRequestContext = {
   token: string;
   /** Time in milliseconds to wait for requests to get a response. */
   timeout?: number;
+  /**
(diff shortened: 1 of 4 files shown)


javascript/src/request.ts
```diff
@@ -24,6 +24,7 @@ export type SvixRequestContext = {
   token: string;
   /** Time in milliseconds to wait for requests to get a response. */
   timeout?: number;
+  fetch?: typeof fetch;
```
svix-lucho, line 31: can you add a comment here like the rest of the options?

PR #2093, opened 2025-10-21
Publish version 1.78.0

svix-mman:
## Version 1.78.0
* Libs/Python: Fix bug preventing the generic webhook ingest source from being deserialized
* Libs/Python: Add API for the new [Svix Stream](https://www.svix.com/stream/)


ChangeLog.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## Version 1.78.0
+* Libs/Python: Fix bug preventing ingest sources from being deserialized
```
svix-jplatte: Only the generic webhook type was actually affected, right?

svix-mman: Now that I think about it, the poller sink is also a empty config struct enum

svix-mman: but the "streaming" section was not added to the public Svix client yet

PR #2095, opened 2025-10-22
server: Add support for creating apps in CMG

svix-mman:
Add support for creating app in CMG

part of: https://github.com/svix/monorepo-private/issues/11626

### DIFF
--- a/server/openapi.json
+++ b/server/openapi.json
@@ -1931,6 +1931,11 @@
             },
             "MessageIn": {
                 "properties": {
+                    "application": {
+                        "$ref": "#/components/schemas/ApplicationIn",
+                        "description": "Optionally creates a new application alongside the message.\n\nIf the application id or uid that is used in the path already exists, this argument is ignored.",
+                        "nullable": true
+                    },
                     "channels": {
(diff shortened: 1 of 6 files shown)


server/svix-server/src/v1/endpoints/application.rs
svix-jbrown, line 293: presumably this will behave poorly if people do concurrent requests, since we're just doing a regular `insert` and not any kind of upsert. It seems more likely that there might be concurrent requests with this kind of merged API. Do we want to try to handle them better here?

svix-mman: > Do we want to try to handle them better here?

I don't think so

svix-jbrown (approved):
don't say I didn't warn you if people complain about getting conflict/deadlock errors from concurrent requests to this API. :-)

server/svix-server/src/v1/endpoints/application.rs
svix-mman: maybe I will re-vist this later, but for now I don't see a reason.
This create_app logic will only be triggered if we don't find the app using the path app_id, so in 99% of cases we won't hit this logic

Issue #2056, opened 2025-09-05
Allow configuring redis queue prefixes


turip:
### Motivation

When cohosting multiple services into the same redis, one way to isolate those services' keys is to use different prefixes and [Redis ACLs](https://redis.io/docs/latest/operate/oss_and_stack/management/security/acl/).

Right now Svix has support to define custom prefixes for it's queues for integration tests (see [here](https://github.com/svix/svix-webhooks/blob/main/server/svix-server/src/queue/redis.rs#L45), comment [here](https://github.com/svix/svix-webhooks/blob/main/server/svix-server/src/lib.rs#L118-L122))

However in the redis db the queues would look like this when using the self-hosted version: 
`{queue}_svix_v3_main`

### Proposal

Introduce a new configuration/environment variable called SVIX_QUEUE_PREFIX. If this is set, then [here](https://github.com/svix/svix-webhooks/blob/main/server/svix-server/src/lib.rs#L102-L105):

```
pub async fn run(cfg: Configuration) {
    let _metrics = setup_metrics(&cfg);
    run_with_prefix(None, cfg, None).await
}
```

We would specify the actual prefix instead of None.

### Alternatives

This proposal ensures that this is a backward compatible change.

If a breaking change is fine, most probably `svix_v3_main` should be the default queue name, and if SVIX_QUEUE_PREFIX is set only then the key should be `PREFIX_svix_v3_main`

PR #2107, opened 2025-10-30
feat: allow specifying queue prefix

turip:
## Motivation

This allows to colocate more services inside the same redis cluster and isolate them using ACLs. The exact issue is documented in this issue: https://github.com/svix/svix-webhooks/issues/2056

## Solution

The solution relies on the already existing facility for queue prefixes, it just exposes that as a config option/environment variable.

### DIFF
--- a/server/svix-server/src/cfg.rs
+++ b/server/svix-server/src/cfg.rs
@@ -185,7 +185,10 @@ pub struct ConfigurationInner {
     /// The DSN for the Redis-backed queue. Overrides `redis_dsn`. (can be left empty if not using
     /// redis)
     pub queue_dsn: Option<String>,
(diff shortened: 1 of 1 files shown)


server/svix-server/src/cfg.rs
```diff
@@ -185,6 +185,8 @@ pub struct ConfigurationInner {
     /// The DSN for the Redis-backed queue. Overrides `redis_dsn`. (can be left empty if not using
     /// redis)
     pub queue_dsn: Option<String>,
+    /// The prefix to use for the queue. If not provided it will be empty.
```
svix-jplatte:
```suggestion
    /// Optional prefix to use for the queue.
    ///
    /// Allows multiple instances of svix-server to share a single queue backend
    /// without interfering with each other.
```

PR #2110, opened 2025-10-31
feat: allow otel resource setting from env var

vanhtuan0409:
## Motivation

Right now SVIX server does not support adding arbitrary otel resource except for service name. Such configuration can be read from env var as standard otel support

## Solution

Enable otel resource detector and reading from env var


vanhtuan0409:
I put it via slack but repeated here for anyone came by. We can make a new configurable just for `environment.name` but i feel it is very limited. Otel semver contains a lot of attribute and it keep changing, I think it is better to leave the operator decide which attribute they want to add

Ref: https://opentelemetry.io/docs/concepts/semantic-conventions/

vanhtuan0409:
@svix-james sure things. I'm open a new PR at https://github.com/svix/svix-webhooks/pull/2115

PR #2114, opened 2025-11-05
svix-cli: add tracing support

svix-jbrown:
This is helpful for debugging. Rather than use the `RUST_LOG` env var, I've used the standard Unix `-v` / `--verbose` flags.

### DIFF
--- a/svix-cli/src/main.rs
+++ b/svix-cli/src/main.rs
@@ -37,6 +37,13 @@ const BIN_NAME: &str = env!("CARGO_BIN_NAME");
 struct Cli {
     #[command(flatten)]
     color: Color,
+    #[arg(
+        short,
+        long,
+        action = clap::ArgAction::Count,
+        help = "Log more. This option may be repeated up to 3 times"
+    )]
+    verbose: u8,
     #[command(subcommand)]
     command: RootCommands,
 }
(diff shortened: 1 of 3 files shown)


svix-cli/src/main.rs
```diff
@@ -37,6 +37,8 @@ const BIN_NAME: &str = env!("CARGO_BIN_NAME");
 struct Cli {
     #[command(flatten)]
     color: Color,
+    #[arg(short, long, action = clap::ArgAction::Count, help="Log more. This option may be repeated up to 3 times")]
```
svix-jplatte:
```suggestion
    #[arg(
        short,
        long,
        action = clap::ArgAction::Count,
        help = "Log more. This option may be repeated up to 3 times"
    )]
```

svix-jbrown: some day they'll make `cargo fmt` work in macros

svix-jplatte: some day..

svix-cli/src/main.rs
```diff
@@ -37,6 +37,13 @@ const BIN_NAME: &str = env!("CARGO_BIN_NAME");
 struct Cli {
     #[command(flatten)]
     color: Color,
+    #[arg(
+        short,
+        long,
+        action = clap::ArgAction::Count,
+        help="Log more. This option may be repeated up to 3 times"
```
svix-jplatte:
```suggestion
        help = "Log more. This option may be repeated up to 3 times"
```

svix-jplatte (approved):

PR #2115, opened 2025-11-06
feat: adding configurable deployment.name in otel

vanhtuan0409:
## Motivation

Adding configuration to tag otel trace and metrics with `deployment.environment` attribute. Supporting signoz environment
https://signoz.io/docs/faqs/general/#q-how-to-setup-signoz-across-different-environments

## Solution

As above

### DIFF
--- a/server/svix-server/src/cfg.rs
+++ b/server/svix-server/src/cfg.rs
@@ -149,6 +149,8 @@ pub struct ConfigurationInner {
     pub opentelemetry_sample_ratio: Option<f64>,
     /// The service name to use for OpenTelemetry. If not provided, it defaults to "svix_server".
     pub opentelemetry_service_name: String,
+    /// The deployment environment to use for OpenTelemetry. It will be translated to "deployment.environment" resource if provided
+    pub opentelemetry_deployment_env: Option<String>,
     /// Whether to enable the logging of the databases at the configured log level. This may be
     /// useful for analyzing their response times.
     pub db_tracing: bool,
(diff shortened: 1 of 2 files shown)


svix-jbrown (commented):
This doesn't seem unreasonable, but a couple of questions to think about...

Question 1: I see that signoz recommends `deployment.environment`, but the [official opentelemetry docs](https://opentelemetry.io/docs/specs/semconv/resource/deployment-environment/) recommend `deployment.environment.name`; do you have any ideas as to which key is more commonly-used? Does signoz also support the "official" one?

Question 2: Do we want to special case `deployment.environment`, or add something like `opentelemetry_resource_attributes: HashMap<String, String>` to let people set other tags? Most otel applications just read from `$OTEL_RESOURCE_ATTRIBUTES` (although ours does not), which would be another option.

svix-james (approved):

svix-james (changes requested):
Ok, apologies, I didn't read @svix-jbrown's previous message closely. We should definitely not support a non-standard resource name here.

svix-james:
Honestly at this point I would just argue for using the original proposal (#2110), i.e., passing by environment variable, and being done with it. The reason we wanted to avoid this is to prevent inadvertently sharing OTEL parameters between Svix and user-specific OTEL streams downstream, but I think we can avoid that if we're careful. 

@vanhtuan0409 Have you confirmed that setting the OTEL env var actually works for your non-standard attribute name? If so, please update _this_ PR (no need to close/open another one) with the changes from #2110, and we can get this in.

PR #2117, opened 2025-11-17
Remove EndpointPatch#secret

svix-jplatte:
It is being deprecated in the cloud API which we use for generating the SDKs. It never did anything and was added by mistake (the corresponding `PUT` route doesn't have it).


svix-mman:
Is this removing `version` or `secret`?

PR #2120, opened 2025-11-21
Update spec and rebuilds SDKs / CLI

svix-jplatte:


rust/src/model_ext.rs
```diff
@@ -44,6 +44,22 @@ impl MessageIn {
...
+    type Err = ConnectorProductFromStrError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "Dispatch" => Ok(Self::Dispatch),
+            "Stream" => Ok(Self::Stream),
+            _ => Err(ConnectorProductFromStrError),
+        }
+    }
+}
```
svix-jplatte, line 61: Would be nice if we didn't have to do this. Probably not too hard to incorporate in templates.

PR #2143, opened 2025-12-18
Update openapi spec, and regen libs

svix-mman:
Update to latest openapi spec


go/models/api_token_out.go
```diff
@@ -6,7 +6,7 @@ import "time"
 type ApiTokenOut struct {
 	CreatedAt time.Time  `json:"createdAt"`
 	ExpiresAt *time.Time `json:"expiresAt,omitempty"`
-	Id        string     `json:"id"` // The GlobalApplicationToken's ID.
```
svix-mman, line 9: Lol, no more ID comment

svix-jplatte: It wasn't particularily helpful anyways.

PR #2146, opened 2025-12-23
Add no-proxy config support

### LINKED ISSUE #2119: Add `no_proxy` support for selective proxy bypass
### Motivation
Currently, when `SVIX_PROXY_ADDR` is configured, **all** outbound webhook requests are routed through the proxy. There's no way to configure exceptions for specific domains or IP ranges that should bypass the proxy.

This feature would align Svix with standard HTTP client proxy behavior and is commonly supported by tools like curl, wget, and most HTTP libraries through `NO_PROXY` environment variable support.

#### Use Case
It is common to use internal connection between components.
We may need to use svix webhook for an internal service hosted on the same k8s cluster where svix-server is running.

### Proposal
Add support for a `no_proxy` configuration that allows specifying domains, IP addresses, or CIDR ranges that should bypass the proxy, similar to the standard `NO_PROXY` environment variable used by many HTTP clients.
**Environment variable:**
```bash
SVIX_PROXY_ADDR="http://proxy.example.com:8080"
SVIX_NO_PROXY="localhost,127.0.0.1,10.0.0.0/8,internal.company.com"
```
**Configuration file:**
```toml
[proxy_config]
proxy_addr = "http://proxy.example.com:8080"
no_proxy = ["localhost", "127.0.0.1", "10.0.0.0/8", "internal.company.com", "*.internal"]
```
#### Expected Behavior
- Requests to URLs matching `no_proxy` patterns should connect directly
- Requests to other URLs should continue using the configured proxy
- Support for:
  - Exact domain matches (`internal.company.com`)
  - Wildcard domain matches (`*.internal`)
  - IP addresses (`127.0.0.1`)
  - CIDR ranges (`10.0.0.0/8`, `192.168.0.0/16`)
  - Port-specific matches (`localhost:8080`)
### Alternatives
So far to by pass the proxy we need to use non standard way that may introduce infrastructure complexity:
- Configure proxy server itself to handle no_proxy rules
- Use network-level routing
- Run multiple Svix instances (one with proxy, one without)

jaymell:
Allow passing a list of IP addresses, CIDR ranges, or hostnames that will bypass the defined 
proxy server. Hostnames can be complete or partial (i.e., `.svix.com` and `svix.com` are 
equivalent and will match both `svix.com` and all its subdomains).

Note that we do not try to resolve hostnames prior to determining whether to route to
the proxy. So in effect, IP addresses and CIDR ranges are only relevant for endpoints that 
actually use an IP address for the hostname.

Fixes #2119

### DIFF
--- a/server/svix-server/config.default.toml
+++ b/server/svix-server/config.default.toml
@@ -123,3 +123,11 @@ worker_max_tasks = 500
 
 # Maximum seconds of queue long-poll
 queue_max_poll_secs = 20
+
+# Optional address of socks5 or http/https proxy server.
+# proxy_addr = socks5://localhost:1080
+
+# Optional comma-separated list of IPV4/IPV6 addresses, CIDR blocks, and URLs that should bypass proxy server. Hostnames
+# may or may not contain a leading "." and include all subdomains. Note that IP addresses are not resolved prior
+# to executing the request, so IP-based exceptions will only apply if endpoint URLs are actually defined as IP addresses.
+# noproxy = "10.0.0.0/8, .example.com, 0ec2:1652:6021:693b:f928:565d:5a0e:de9f"
--- a/server/svix-server/src/cfg.rs
+++ b/server/svix-server/src/cfg.rs
@@ -251,6 +251,9 @@ pub struct ConfigurationInner {
 }
 
 #[derive(Clone, Debug, Deserialize)]
+pub struct ProxyBypassCfg(pub String);
+
+#[derive(Clone, Debug, Deserialize)]
 pub struct ProxyConfig {
     /// Proxy address.
     ///
@@ -261,6 +264,9 @@ pub struct ProxyConfig {
     ///   both HTTP and HTTPS targets are supported
     #[serde(rename = "proxy_addr")]
     pub addr: ProxyAddr,
+
+    #[serde(default)]
+    pub noproxy: Option<ProxyBypassCfg>,
 }
 
 #[derive(Clone, Debug)]
--- a/server/svix-server/src/core/webhook_http_client.rs
+++ b/server/svix-server/src/core/webhook_http_client.rs
@@ -5,7 +5,7 @@ use std::{
     pin::Pin,
     str::FromStr,
     sync::Arc,
-    task::Poll,
+    task::{ready, Poll},
     time::{Duration, Instant},
 };
 
@@ -19,13 +19,16 @@ use http_body_util::Full;
 use hyper::{body::Incoming, ext::HeaderCaseMap, Uri};
 use hyper_openssl::client::legacy::{HttpsConnector, MaybeHttpsStream};
 use hyper_util::{
-    client::legacy::{
-        connect::{
-            dns::Name,
-            proxy::{SocksV5, Tunnel},
-            HttpConnector,
+    client::{
+        legacy::{
+            connect::{
+                dns::Name,
+                proxy::{SocksV5, Tunnel},
+                HttpConnector,
+            },
+            Client,
         },
-        Client,
+        proxy::matcher::Matcher,
     },
     rt::{TokioExecutor, TokioIo},
 };
@@ -36,7 +39,7 @@ use thiserror::Error;
 use tokio::{net::TcpStream, sync::Mutex};
 use tower::Service;
 
-use crate::cfg::{ProxyAddr, ProxyConfig};
+use crate::cfg::{ProxyAddr, ProxyBypassCfg, ProxyConfig};
 
 pub type CaseSensitiveHeaderMap = HashMap<String, HeaderValue>;
 
@@ -84,6 +87,16 @@ pub struct WebhookClient {
     whitelist_nets: Arc<Vec<IpNet>>,
 }
 
+fn ssl_builder(disable_tls_verification: bool) -> SslConnectorBuilder {
+    // Openssl is required here -- in practice, rustls does not support many
+    // ciphers that we encounter on a regular basis:
+    let mut ssl = SslConnector::builder(SslMethod::tls()).expect("SslConnector build failed");
+    if disable_tls_verification {
+        ssl.set_verify(SslVerifyMode::NONE);
+    }
+    ssl
+}
+
 impl WebhookClient {
     pub fn new(
         whitelist_nets: Option<Arc<Vec<IpNet>>>,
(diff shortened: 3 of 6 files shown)


server/svix-server/src/core/webhook_http_client.rs
```diff
@@ -446,33 +454,71 @@ impl RequestBuilder {
...
     ) -> Result<Self, Box<dyn std::error::Error>> {
+        let https =
+            HttpsConnector::with_connector(inner.clone(), ssl_builder(disable_tls_verification))?;
+
         match proxy_cfg {
             Some(proxy_cfg) => match proxy_cfg.addr.clone() {
                 ProxyAddr::Socks5(proxy_addr) => {
+                    let mut matcher = Matcher::builder()
+                        .http(proxy_addr.to_string())
+                        .https(proxy_addr.to_string());
```
svix-jplatte: Since we enforce that the scheme is either `http` or `https`, could also use `all`, right?

server/svix-server/src/core/webhook_http_client.rs
```diff
@@ -488,16 +534,56 @@ impl Service<Uri> for SvixHttpsConnector {
...
+                match proxy.poll_ready(cx) {
+                    Poll::Ready(Ok(_)) => {}
+                    err @ Poll::Ready(Err(_)) => return err,
+                    Poll::Pending => return Poll::Pending,
+                }
+                match bypass.poll_ready(cx) {
+                    Poll::Ready(Ok(_)) => {}
+                    err @ Poll::Ready(Err(_)) => return err,
+                    Poll::Pending => return Poll::Pending,
+                }
```
svix-jplatte: Stupid readiness thing. I wonder if both should always be polled conceptually, but in practice I'm almost certain it won't matter. Anyways, can be simplified using `std::task::ready` + `?`:

```suggestion
                ready!(proxy.poll_ready(cx)?);
                ready!(bypass.poll_ready(cx)?);
```

server/svix-server/src/core/webhook_http_client.rs
```diff
@@ -488,16 +534,56 @@ impl Service<Uri> for SvixHttpsConnector {
...
```
svix-jplatte: Same here.

server/svix-server/tests/it/e2e_proxy.rs
```diff
@@ -70,3 +80,138 @@ async fn run_proxy_test(client: &TestClient) {
...
+    }
+}
+
+#[tokio::test]
+async fn test_http_proxy_exceptions() {
+    let listener = MockProxyServer::new();
+    test_proxy_exceptions(format!("http://{}", listener.addr), listener).await
+}
+
+#[ignore] // This doesn't currently work for socks5 servers
```
svix-jplatte: The proxy exclusion feature doesn't work for socks, or just the test?

svix-james: Just the test -- since parsing the socks5 request is something I'm reluctant to try. Will update the comment.

svix-jplatte (approved):

jaymell:
## OSS Review

Note that `socks5-proto` is added only to support testing.

svix-jplatte (dismissed):
Other than the license concern (comment above), LGTM.

tasn (changes requested):
GPL: it's fine if it's just used in tests, though I'm concerned about eager scanners by our customers/their companies flagging our repo because of it. Can we get rid of the dep/find an alternative one?

Also, why didn't our scanners catch it? Don't we use a dependency license checker?

jaymell:
Previously we didn't really test proxies (there are some ignored tests that assert messages succeed using a proxy). The added functionality here is very simple but at least asserts that a mocked socks5/http proxy server receives the expected requests for the expected addresses.

svix-jplatte (dismissed):

server/svix-server/tests/it/e2e_proxy.rs
```diff
@@ -70,3 +83,247 @@ async fn run_proxy_test(client: &TestClient) {
...
+
+    create_test_endpoint(&client, &app_id, "https://10.0.0.1")
+        .await
+        .unwrap();
+
+    create_test_endpoint(
+        &client,
+        &app_id,
+        "https://[0ec2:1652:6021:693b:f928:565d:5a0e:de9f]",
+    )
```
svix-jbrown, line 289: maybe add a test for the various v4/v6 encodings and at least document how they're handled. Ideally a `NO_PROXY=10.0.0.0/8` would also trigger for `::ffff:a01:203`, `::a01:203`, and `64:ff9b::a01:203`.

server/svix-server/src/cfg.rs
```diff
@@ -261,6 +264,9 @@ pub struct ProxyConfig {
     ///   both HTTP and HTTPS targets are supported
     #[serde(rename = "proxy_addr")]
     pub addr: ProxyAddr,
+
+    #[serde(default)]
+    pub noproxy: Option<ProxyBypassCfg>,
```
svix-jbrown, line 269: maybe a comment explaining the syntax (or linking to a source doc) for future reference? I assume https://docs.rs/hyper-util/latest/src/hyper_util/client/proxy/matcher.rs.html#280 is the relevant definition?

svix-james: Will add comment in followup.

server/svix-server/config.default.toml
```diff
@@ -123,3 +123,11 @@ worker_max_tasks = 500
...
 # Maximum seconds of queue long-poll
 queue_max_poll_secs = 20
+
+# Optional address of socks5 or http/https proxy server. 
+# proxy_addr = socks5://localhost:1080
+
+# Optional comma-separated list of IPV4/IPV6 addresses, CIDR blocks, and URLs that should bypass proxy server. Hostnames
+# may or may not contain a leading "." and include all subdomains. Note that IP addresses are not resolved prior 
+# to executing the request, so IP-based exceptions will only apply if endpoint URLs are actually defined as IP addresses.
+# noproxy = "10.0.0.0/8, .example.com, 0ec2:1652:6021:693b:f928:565d:5a0e:de9f"
```
svix-jbrown: nit: missing trailing newline

server/svix-server/tests/it/e2e_proxy.rs
jaymell: Added an assertion for endpoint `::ffff:a01:203` as well.

jaymell: Oh wait. That doesn't work.

jaymell: I'm not going to sweat this much for now. The IP functionality is not that useful, anyway, since we don't use it post-DNS resolution.

tasn:
> EDIT: Note that if this becomes an issue, we can pretty easily manually implement the bit of parsing needed to support the tests. It just wasn't something that seemed like merited time spent right now, hence the additional crate.

I'm concerned that it'll silently be an issue, without us knowing.

svix-jbrown (approved):
bummed that this matcher doesn't support ipv6 correctly, but c'est la vie

PR #2147, opened 2025-12-29
Use `standard-webhooks` as the JS and Python webhook verification impl

svix-mman:
Use the javascript and python `standardwebhooks` packages for webhook verification,


python/svix/webhooks.py
```diff
@@ -1,81 +1,23 @@
...
-                continue
-            sig_bytes = base64.b64decode(signature)
-            if hmac.compare_digest(expected_sig, sig_bytes):
-                return json.loads(data)
+        for h in ["id", "signature", "timestamp"]:
+            if (
+                headers.get(f"svix-{h}") is not None
+                and headers.get(f"webhook-{h}") is None
+            ):
+                headers[f"webhook-{h}"] = headers[f"svix-{h}"]
```
tasn: It's more lines and more opaque than just doing what we did in the code you removed.
(it's less line with your current implementation, but you did something different. If you do the exact same thing, the original implementation is 3 lines).
```python
headers["webhook-id"] = headers.get("svix-id", headers.get("webhook-id"))
...
```
For example.

python/svix/webhooks.py
```diff
@@ -1,81 +1,23 @@
...
 import typing as t
-from datetime import datetime, timedelta, timezone
-from math import floor
-
-
-def hmac_data(key: bytes, data: bytes) -> bytes:
-    return hmac.new(key, data, hashlib.sha256).digest()
-
-
-class WebhookVerificationError(Exception):
```
tasn, line 14: This needs to be re-exported, otherwise it can't be used.

svix-mman: I guess I need to re-export it from the same location right?
(and not just from the `__init__.py`)

python/svix/webhooks.py
svix-mman: ya, I thought of this. but mypy kept yelling at me :shrug:

python/svix/webhooks.py
tasn: Yup. I don't even know how, maybe something like this is enough?
```python
WebhookVerificationError = StdWebhookVerificationError
```

python/svix/webhooks.py
tasn: What is it saying? complaining about null? If so, you can just do exactly what we had before, which didn't get mypy anger. Or alternatively just pass `""` like this:
```python
headers["webhook-id"] = headers.get("svix-id", headers.get("webhook-id", ""))
...

File: python/svix/webhooks.py (hunk 2 above)
Comment (svix-mman) [reply python/svix/webhooks.py:14]: oh no, we simply import `WebhookVerificationError` and add it to `__all__` that it :+1:

File: python/svix/webhooks.py (hunk 1 above)
Comment (svix-mman) [reply python/svix/webhooks.py]: this work :+1:

File: javascript/src/webhook.ts [hunk 3]
```diff
@@ -75,72 +36,20 @@ export class Webhook {
...
-    if (!msgSignature || !msgId || !msgTimestamp) {
-      msgId = headers["webhook-id"];
-      msgSignature = headers["webhook-signature"];
-      msgTimestamp = headers["webhook-timestamp"];
-
-      if (!msgSignature || !msgId || !msgTimestamp) {
-        throw new WebhookVerificationError("Missing required headers");
+    for (const h of ["id", "signature", "timestamp"]) {
+      if (headers[`svix-${h}`] !== undefined && headers[`webhook-${h}`] === undefined) {
+        headers[`webhook-${h}`] = headers[`svix-${h}`];
```
tasn: Already commented on this with Python. Not good.

javascript/src/webhook.ts
```diff
@@ -75,72 +36,20 @@ export class Webhook {
...
-    let msgTimestamp = headers["svix-timestamp"];
-
-    if (!msgSignature || !msgId || !msgTimestamp) {
-      msgId = headers["webhook-id"];
-      msgSignature = headers["webhook-signature"];
-      msgTimestamp = headers["webhook-timestamp"];
-
-      if (!msgSignature || !msgId || !msgTimestamp) {
-        throw new WebhookVerificationError("Missing required headers");
+    for (const h of ["id", "signature", "timestamp"]) {
```
tasn: Btw, you can probably change both Python and here to iterate on tuples like `("webhook-id", "svix-id")` and then do this. Will be slightly less repetitive and also clear.

It's just that the dynamically generating strings is bad.

javascript/src/webhook.ts
svix-mman: fixed :+1:

javascript/src/webhook.ts
svix-mman: meh, I updated to do this 
```js
headers["webhook-id"] = headers["svix-id"] ?? headers["webhook-id"] ?? "";
```

I don't have strong feelings about using a tuple, so I will just stick to this :+1:

javascript/src/webhook.ts
```diff
@@ -75,72 +36,20 @@ export class Webhook {
...
   public sign(msgId: string, timestamp: Date, payload: string | Buffer): string {
-    if (typeof payload === "string") {
-      // Do nothing, already a string
-    } else if (payload.constructor.name === "Buffer") {
-      payload = payload.toString();
-    } else {
-      throw new Error(
-        "Expected payload to be of type string or Buffer. Please refer to https://docs.svix.com/receiving/verifying-payloads/how for more information."
-      );
-    }
```
svix-lucho, line 122: This isn't the same logic, right? We still need to check if type is Buffer before calling toString

svix-mman: We delegate to the standard-webhooks sign logic. (which does this verification)

PR #2148, opened 2025-12-29
Release v1.84.0

svix-mman:
## Version 1.84.0
* Libs/(JavaScript Python): Updated webhook verification logic to use official standard webhooks packages
* Cli: support disabling TLS verification when relaying requests (`svix listen`)


tasn:
Mind waiting for https://github.com/svix/svix-webhooks/pull/2149 ? 🙏🏻

PR #2149, opened 2025-12-29
Svix CLI: support disabling TLS verification when relaying requests.

tasn:
Makes local debugging easier when you're already using self-signed TLS certs.

### DIFF
--- a/svix-cli/src/relay/mod.rs
+++ b/svix-cli/src/relay/mod.rs
@@ -215,6 +215,7 @@ pub async fn listen(
     relay_token: String,
     relay_debug_url: Option<&str>,
     relay_disable_security: bool,
+    disable_tls_verification: bool,
 ) -> Result<()> {
     let scheme = if relay_disable_security { "ws" } else { "wss" };
     let api_host = relay_debug_url.unwrap_or(DEFAULT_API_HOST);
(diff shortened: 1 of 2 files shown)


svix-cli/src/relay/mod.rs
```diff
@@ -215,18 +215,23 @@ pub async fn listen(
...
     let scheme = if relay_disable_security { "ws" } else { "wss" };
     let api_host = relay_debug_url.unwrap_or(DEFAULT_API_HOST);
     let token = format!("c_{relay_token}");
 
     let websocket_url = format!("{scheme}://{api_host}/{API_PREFIX}/listen/").parse()?;
 
+    let http_client = HttpClient::builder()
+        .danger_accept_invalid_certs(disable_tls_verification)
+        .build()?;
+
```
svix-mman, line 229: We should check that `relay_debug_url` was used and not `DEFAULT_API_HOST` (if we want to only disable the tls verification if a user specified the `relay_debug_url`) right?

tasn: It disables it for listen_url, not relay_debug_url.

svix-mman: okie :+1:

PR #2152, opened 2025-12-30
regen_openapi: Limit max threads to cpu_count, and update codegen image

svix-mman:
Bump the codegen image, and limit concurrency to cpu_count


svix-jplatte (approved):
Maybe make the PR title / commit a little clearer by prefixing with `regen_openapi: ` or something. "workers" is usually something different in this repo :smile:

PR #2160, opened 2026-01-22
Add some documentation to codegen/

svix-jplatte:
I had to think a bit how to test a new version of `openapi-codegen` before pushing it. It's also been a while that we've change anything about the overall regeneration process, so I think it's overdue we document that process.

### DIFF
--- a/codegen/README.md
+++ b/codegen/README.md
@@ -1,0 +1,30 @@
+# Client SDK code generation
+
+This directory contains code to generate the SDKs (+ Svix CLI).
+Here is a quick overview for how to use it:
+
+## Updating the SDKs
+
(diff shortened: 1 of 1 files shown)


codegen/README.md
```diff
@@ -0,0 +1,30 @@
+# Client SDK code generation
+
+This directory contains code to generate the SDKs (+ Svix CLI).
+Here is a quick overview for how to use it:
+
+## Updating the SDKs
+
+To regenerate the code after updating `lib-openapi.json`, run `regen_openapi.py` from the repository root.
```
svix-jbrown, line 8: maybe mention that you need to be on a unix-like system with python3.11+?

PR #2177, opened 2026-02-09
server: add case-preserving header map

svix-jbrown:
Currently, the OSS version of the server uses a simple `HashMap<String, String>` for headers. This results in confusing behavior if the same header is inserted twice with different casings (for example, if a custom override header sets `User-Agent`, and our code inserts `user-agent`).

This PR modifies the server to use a case *preserving* map for headers, which will still correctly remember the difference between `user-agent` and `User-Agent`, but will not allow duplicates.

This is mostly migrated from svix/svix-webhooks-private#5916 in the private fork, with modifications to work with the OSS layout.

Part of svix/monorepo-private#12554

### DIFF
--- a/server/svix-server/src/core/types/case_preserving_header_map.rs
+++ b/server/svix-server/src/core/types/case_preserving_header_map.rs
@@ -10,11 +10,11 @@ use bytes::Bytes;
 use http::{header::InvalidHeaderName, HeaderMap, HeaderName, HeaderValue};
 
 /// A map from String -> T that preserves insertion case,
-/// but doesn't allow duplicates (that is to say, inserting
-/// "foo" -> "bar", and "Foo" -> "baz" will only result in one key:
-/// "Foo" -> "baz"
+/// but doesn't allow duplicates.
 ///
-/// Intended use case is really for storing headers (hence the name)
+/// That is to say, inserting "foo" -> "bar", and "Foo" -> "baz" will
+/// only result in one key: "Foo" -> "baz".
+/// Intended use case is really for storing headers, hence the name.
 #[derive(Debug, Clone)]
 pub struct CasePreservingHeaderMap<T = HeaderValue> {
     /// We implement this as a map from normalized key to (real key, value).
@@ -118,8 +118,8 @@ impl<T> CasePreservingHeaderMap<T> {
 
     /// Iterate over all key-value pairs. The key is returned in
     /// non-normalized (original) form
-    pub fn iter(&self) -> impl Iterator<Item = &(String, T)> {
-        self.inner.values()
+    pub fn iter(&self) -> impl Iterator<Item = (&str, &T)> {
+        self.inner.values().map(|(k, v)| (k.as_str(), v))
     }
 
     /// Convert into two maps: one from HeaderName -> T, and one from HeaderName -> Raw Name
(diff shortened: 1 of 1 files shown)


server/svix-server/src/core/types/case_preserving_header_map.rs
```diff
@@ -0,0 +1,325 @@
...
+
+use bytes::Bytes;
+use http::{header::InvalidHeaderName, HeaderMap, HeaderName, HeaderValue};
+
+/// A map from String -> T that preserves insertion case,
+/// but doesn't allow duplicates (that is to say, inserting
+/// "foo" -> "bar", and "Foo" -> "baz" will only result in one key:
+/// "Foo" -> "baz"
+///
+/// Intended use case is really for storing headers (hence the name)
```
svix-jplatte:
```suggestion
/// but doesn't allow duplicates.
///
/// That is to say, inserting "foo" -> "bar", and "Foo" -> "baz"
/// will only result in one key: "Foo" -> "baz".
/// Intended use case is really for storing headers (hence the name).
```

server/svix-server/src/core/types/case_preserving_header_map.rs
```diff
@@ -0,0 +1,325 @@
...
+            return false;
+        };
+        self.inner.contains_key(key.as_ref())
+    }
+
+    /// Iterate over all key-value pairs. The key is returned in
+    /// non-normalized (original) form
+    pub fn iter(&self) -> impl Iterator<Item = &(String, T)> {
+        self.inner.values()
+    }
```
svix-jplatte:
```suggestion
    pub fn iter(&self) -> impl Iterator<Item = (&str, &T)> {
        self.inner.values().map(|(k, v)| (k, v))
    }
```

server/svix-server/src/core/types/case_preserving_header_map.rs
```diff
@@ -0,0 +1,325 @@
...
+    fn into_iter(self) -> Self::IntoIter {
+        self.inner.into_values()
+    }
+}
+
+impl<T: Clone + PartialEq> PartialEq<Self> for CasePreservingHeaderMap<T> {
+    // this is terrible but it's only used for tests
+    fn eq(&self, other: &Self) -> bool {
+        self.iter().cloned().collect::<HashMap<String, T>>()
+            == other.iter().cloned().collect::<HashMap<String, T>>()
```
svix-jplatte:
```suggestion
        self.iter().collect::<HashMap<_, _>>()
            == other.iter().collect::<HashMap<_, _>>()
```

server/svix-server/src/core/types/case_preserving_header_map.rs
```diff
@@ -0,0 +1,325 @@
...
+}
+
+impl<V> Extend<(HeaderName, V)> for CasePreservingHeaderMap<V> {
+    #[inline]
+    fn extend<T: IntoIterator<Item = (HeaderName, V)>>(&mut self, iter: T) {
+        let iter = iter.into_iter();
+        self.inner.reserve(iter.size_hint().0);
+        iter.for_each(move |(k, v)| {
+            self.insert(k, v);
+        });
```
svix-jplatte:
```suggestion
        for (k, v) in iter {
            self.insert(k, v);
        }
```

PR #2179, opened 2026-02-11
ruby: Return nil if payload does is empty

donovanlopez:
## Motivation

I'm using the Svix gem in Ruby to verify a `GET` request using the `verify` method. The payload of this request is empty so when calling `verify`, I pass in an empty string as the `payload`. Tracing the code, it looks like it's able to verify the request correctly but an error gets throw when returning the empty payload: `return JSON.parse(payload, symbolize_names: true)`

## Solution

The problem is that calling `JSON.parse` on an empty string throws an error: `unexpected end of input at line 1 column 1 (JSON::ParserError)`. The solution is to just return `nil` if the `payload` is empty.

### DIFF
--- a/ruby/lib/svix/webhook.rb
+++ b/ruby/lib/svix/webhook.rb
@@ -40,7 +40,7 @@ module Svix
         end
 
         if ::Svix::secure_compare(signature, expectedSignature)
-          return {} if payload.empty?
+          return nil if payload.empty?
           return JSON.parse(payload, symbolize_names: true)
         end
       end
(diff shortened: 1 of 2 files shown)


donovanlopez:
> Thanks for the PR! Any reason to return an empty hash, rather than nil? That seems like it would more closely match the actual payload.

@svix-jplatte Since this method is expected to return a hash, I went with an empty hash but I can go either way. I'll change it to `nil`.

svix-jplatte:
I've started the release process over in #2180. Should be available later today.

PR #2203, opened 2026-02-28
Respect Retry-After on error responses

### LINKED ISSUE #2200: Respect Retry-After on 429/503 Responses
hey svix team been digging through the repo and found something interesting I want to fix—just want to make sure I'm on the right track before I open a PR.

Right now when a destination server sends back a 429 or 503 with a Retry-After header, Svix ignores it and retries on its own schedule. The server is saying give me 120 seconds but we might hit it again in 30 — which wastes retries and doesn't help it recover.

My proposal is to use whichever is larger — the server's requested delay or our internal backoff. This would only ever increase the delay, never cause earlier retries. We'd also need a cap since a server could send Retry-After: 86400 and we don't want to blindly respect that — tom mentioned something like retry_schedule_delay * 2 + 2hrs which sounds reasonable to me, but curious what the team thinks.

Before I touch anything — should the cap be relative to our backoff stage or a fixed ceiling? Should we support both seconds and HTTP-date formats or just seconds? And should this apply to 429/503 only or other 5xx too?

vinay0826:
This adds Retry-After handling as discussed in #2200 

If a destination responds with 429 or 503 and includes a Retry-After header, we now incorporate it into retry scheduling. The effective delay becomes max(internal_backoff, retry_after) and is capped at internal_backoff + 2h (using saturating_add) to avoid very large values causing excessive sleep.

Retry-After is honored for both 429 and 503 responses.

Existing overload penalty and jitter behavior remain unchanged. Invalid Retry-After headers are logged and ignored, falling back to the normal backoff logic. Unit tests are included for both seconds and HTTP-date formats.

Based on the Cloud implementation shared by @svix-jplatte 

Closes #2200

### DIFF
--- a/server/svix-server/src/worker.rs
+++ b/server/svix-server/src/worker.rs
@@ -10,11 +10,11 @@ use std::{
 };
 
 use base64::{Engine, engine::general_purpose::STANDARD};
-use chrono::Utc;
+use chrono::{DateTime, Utc};
 use futures::future;
 use http::{
-    HeaderName, HeaderValue, StatusCode, Version,
-    header::{CONTENT_TYPE, USER_AGENT},
+    HeaderMap, HeaderName, HeaderValue, StatusCode, Version,
+    header::{CONTENT_TYPE, RETRY_AFTER, USER_AGENT},
 };
 use http_body_util::BodyExt as _;
 use itertools::Itertools;
@@ -256,7 +256,8 @@ struct WorkerContext<'a> {
     webhook_client: &'a WebhookClient,
 }
 
-struct FailedDispatch(messageattempt::ActiveModel, Error);
+struct FailedDispatch(messageattempt::ActiveModel, Error, Option<DateTime<Utc>>);
+
 struct SuccessfulDispatch(messageattempt::ActiveModel);
 
 #[allow(clippy::large_enum_variant)]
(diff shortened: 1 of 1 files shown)


server/svix-server/src/worker.rs
```diff
@@ -412,6 +422,7 @@ async fn make_http_call(
                 Some(err) => Ok(CompletedDispatch::Failed(FailedDispatch(
                     attempt,
                     Error::generic(err),
+                    retry_after,
                 ))),
```
svix-jplatte: Can you move the `let retry_after = ` to this branch? No need to check the status code then really, we can honor `retry-after` for any failed request. As in

```rust
Some(err) => {
    let retry_after = ...;
    Ok(CompletedDispatch::Failed(...))
}
```

server/svix-server/src/worker.rs
```diff
@@ -487,14 +499,50 @@ async fn handle_successful_dispatch(
...
         std::cmp::max(duration, Duration::from_secs(OVERLOAD_PENALTY_SECS))
     } else {
         duration
     };
+
+    let duration = if let Some(retry_after) = retry_after {
+        let retry_after = (retry_after - Utc::now())
+            .to_std()
+            .unwrap_or(Duration::from_secs(0));
+        limit_retry_delay(duration, retry_after)
```
svix-jplatte, line 535: I see I failed to share the implementation of this function. It's actually just a helper function because it's used in multiple places in cloud. Let's use the same thing here, which is clamping the retry-after value to (duration, 2 * duration):

```suggestion
        retry_after.clamp(duration, duration * 2)
```

svix-jplatte (commented):
Thanks for the PR!

svix-jplatte (approved):

PR #2270, opened 2026-04-20
fix postman CI: install openapi-to-postmanv2 before running script

tasn:
It worked locally before because I haven't wiped my `node_modules` so I had this installed without realizing.


svix-jbrown:
isn't that what `npx` is supposed to do? ephemerally install a package?

svix-pluiz:
> isn't that what `npx` is supposed to do? ephemerally install a package?

I think `npx` only executes the binary/package, `update_postman.js` needs to use the library

svix-jbrown (approved):
ok

PR #2274, opened 2026-04-21
Release CI: ask for access to environment at the start of the run.

tasn:
At the moment it asks it at the end, asking at the start is nicer as it means we don't have to watch it and wait until the end to approve it.

We do it with a simple gate that just asks for the approval at the start.


.github/workflows/release.yml
svix-jbrown, line 109: should this depend on `gate` to be explicit?

svix-jbrown (approved):

PR #2277, opened 2026-04-21
Use 'export type' for TypeScript type-only re-exports

svix-pluiz:
## Motivation

...So we can upgrade packages that relies on TS libraries that relies on `export type` to resolve imports in newer versions.

### component_type_summary.ts.jinja

Check if `type.kind` is struct or struct_enum, if so, prepend `type`

### summary.ts.jinja

Add `export type` to types being re-exported


codegen/templates/javascript/component_type_summary.ts.jinja
```diff
@@ -1,4 +1,7 @@
 // this file is @generated
-{% for type in api.types -%}
-export { {{ type | to_upper_camel_case }} } from "./{{ type | to_lower_camel_case }}";
-{% endfor %}
+{% for name, type in api.types | items -%}
+export {
+    {% if type.kind == "struct" or type.kind == "struct_enum" %}type {% endif -%}
+    {{- name | to_upper_camel_case -}}
+} from "./{{ name | to_lower_camel_case }}";
+{% endfor %}
```
svix-jplatte: Can you (re-) add the final newline to this file?

svix-jplatte (approved):

PR #2278, opened 2026-04-21
feat(cli): show command examples on --help

svix-pluiz:
closes https://github.com/svix/monorepo-private/issues/12825

## Motivation

Add real world examples for the CLI, making it easier to use

## Solution

- Add a map with semi-realistic fake tokens `svix-cli/api_resource.rs.jinja`. 
- Update the CLI to show command examples when `--help` is passed

### svix ingest endpoint set-transformation --help
```
Set or unset the transformation code associated with this ingest endpoint

Usage: svix ingest endpoint set-transformation [OPTIONS] <SOURCE_ID> <ENDPOINT_ID> [INGEST_ENDPOINT_TRANSFORMATION_PATCH]

Example: svix ingest endpoint set-transformation src_abc000000000000000000 ep_abc000000000000000000000000 {...}


Example body:
{
  "code": "...",
  "enabled": true
}

Arguments:
  <SOURCE_ID>                             
  <ENDPOINT_ID>                           
  [INGEST_ENDPOINT_TRANSFORMATION_PATCH]  

Options:
      --color <WHEN>  Controls when to use color [default: auto] [possible values: auto, always, never]
  -h, --help          Print help
```

### svix ingest endpoint get-headers  --help

```
Get the additional headers to be sent with the ingest

Usage: svix ingest endpoint get-headers [OPTIONS] <SOURCE_ID> <ENDPOINT_ID>

Example: svix ingest endpoint get-headers src_abc000000000000000000 ep_abc000000000000000000000000


Example response:
{
  "headers": {
    "X-Example": "123",
    "X-Foobar": "Bar"
  },
  "sensitive": ["Authorization"]
}

Arguments:
  <SOURCE_ID>    
  <ENDPOINT_ID>  

Options:
      --color <WHEN>  Controls when to use color [default: auto] [possible values: auto, always, never]
  -h, --help          Print help
```

### DIFF
--- a/codegen/templates/svix-cli/api_resource.rs.jinja
+++ b/codegen/templates/svix-cli/api_resource.rs.jinja
@@ -24,7 +24,7 @@ use svix::api::*;
 use super::{{ sub.name | to_snake_case }}::{{ sub.name | to_upper_camel_case }}Args;
 {% endfor %}
 
-{%- macro get_type_example(field, api) -%}
+{%- macro get_type_example(field, api, indent) -%}
     {%- if field.type.is_string_const() -%}
         \"{{ field.type.string_const_val() }}\"
     {%- elif field.type.is_unix_timestamp_ms() -%}
@@ -40,9 +40,9 @@ use super::{{ sub.name | to_snake_case }}::{{ sub.name | to_upper_camel_case }}A
     {%- elif field.type.is_int_or_uint() -%}
         123
     {%- elif field.type.is_list() or field.type.is_set() -%}
-        [{{ get_type_example({"type": field.type.inner_type()}, api) }}]
+        [{{ get_type_example({"type": field.type.inner_type()}, api, indent) }}]
     {%- elif field.type.is_map() -%}
-        {\"key\": {{ get_type_example({"type": field.type.value_type()}, api) }}}
+        {\"key\": {{ get_type_example({"type": field.type.value_type()}, api, indent) }}}
     {%- elif field.type.is_json_object() -%}
         {\"key\": \"...\"}
     {%- elif field.type.is_schema_ref() -%}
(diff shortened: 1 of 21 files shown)


svix-jplatte:
One thing I noticed in diom: it's only shown with `--help`, not with `-h`. Is there a way to add it to both?

svix-pluiz:
@svix-jplatte 
> One thing I noticed in diom: it's only shown with `--help`, not with `-h`. Is there a way to add it to both?


I was researching about this and I noticed that Clap has two different 'help' modes, one for short help and other for long help. `-h` stands for short and `--help` for long, respectively.

Now I'm thinking about how we could take advantage of that. Because I think it makes sense to have long and short versions of help messages, however, I don't believe that users will know that beforehand

codegen/templates/svix-cli/api_resource.rs.jinja
```diff
@@ -12,6 +24,72 @@ use svix::api::*;
...
+{%- macro render_inline_object(schema, api, indent) -%}
+{
+{%- for f in schema.fields %}
+{{ indent }}  \"{{ f.name }}\": {% if f.example is defined %}{{ render_example_value(f.example, indent + "  ") }}{% else %}{{ get_type_example(f, api, indent + "  ") }}{% endif %}{{ "," if not loop.last else "" }}
+{%- endfor %}
+{{ indent }}}
+{%- endmacro -%}
+
+{%- macro render_schema_example(title, schema, api) -%}
+\x1b[1;4m{{ title }}:\x1b[0m
```
svix-jbrown: I'm not sure we want to use ANSI escape codes by default; they're going to screw up some terminals (and pretty much all pagers). I took a peek and didn't see anything in clap to conditionally render escape codes if stdout is a tty, so I think we should just forgo ANSI altogether

svix-pluiz: > they're going to screw up some terminals (and pretty much all pagers)

Nice catch, I'll update this PR and then open one for Diom to change there as well

svix-pluiz: https://github.com/svix/diom/pull/1169

svix-cli/src/cmds/api/connector.rs
svix-jplatte, line 1: Doesn't have to be in this PR, but we should mark these files as generated via `.gitattributes`.

svix-jplatte (approved):

svix-cli/src/cmds/api/connector.rs
svix-pluiz: I'll do this asap

PR #2285, opened 2026-04-23
fix(javascript): upgrade uuid to ^14.0.0 and TypeScript to ^5.0

th3nolo:
## Motivation

`uuid@14` fixes [GHSA-w5hq-g745-h8pq](https://github.com/advisories/GHSA-w5hq-g745-h8pq) (CWE-122, buffer bounds check in uuid v3/v5/v6). The existing Dependabot PR #2284 bumped uuid but CI failed with:

```
node_modules/uuid/dist/index.d.ts(5,1): error TS1383: Only named exports may use 'export type'.
```

`uuid@14`'s type declarations use `export type *` syntax which requires TypeScript 5.0+.

## Solution

Bump `uuid` from `^10.0.0` to `^14.0.0` and `typescript` from `^4.0` to `^5.0` together. svix only uses `uuid.v4()` (idempotency keys in `request.ts`), no v3/v5/v6 usage, so no API risk.

- [x] `npm install` completes without errors
- [x] `npm run build` passes
- [x] `npm test` passes (44 pass, 1 skipped, 0 fail)
- [x] `npm audit` shows 0 critical vulnerabilities for uuid

### DIFF
--- a/javascript/package-lock.json
+++ b/javascript/package-lock.json
@@ -10,14 +10,14 @@
       "license": "MIT",
       "dependencies": {
         "standardwebhooks": "1.0.0",
-        "uuid": "^10.0.0"
+        "uuid": "^14.0.0"
       },
       "devDependencies": {
         "@biomejs/biome": "2.2.4",
         "@stablelib/utf8": "^2.0.0",
         "@types/uuid": "^10.0.0",
         "mockttp": "^3.15.5",
-        "typescript": "^4.0"
+        "typescript": "^5.0"
       }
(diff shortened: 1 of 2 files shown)


svix-jplatte:
Thanks for the PR! I think we'll end up merging https://github.com/svix/svix-webhooks/pull/2286 instead though.
Either way, I'll make sure we publish a new release right after.

svix-jplatte:
Decided to merge #2286 instead. Thanks again though :)

PR #2286, opened 2026-04-23
js: Replace dependency `uuid` with native `crypto.randomUUID()`

ulrichstark:
## Motivation

Same motivation as #2285 to fix the reported security vulnerability, but instead of upgrading `uuid`, I propose to replace it with the native `crypto.randomUUID()`. This also reduces dependencies by 2.

I'm not sure what the baseline requirements for this project are, but `crypto.randomUUID` is [well established](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID#browser_compatibility).

## Solution

Replace dependency `uuid` with native `crypto.randomUUID()`

### DIFF
--- a/javascript/package-lock.json
+++ b/javascript/package-lock.json
@@ -9,13 +9,11 @@
       "version": "1.91.1",
       "license": "MIT",
       "dependencies": {
-        "standardwebhooks": "1.0.0",
-        "uuid": "^10.0.0"
+        "standardwebhooks": "1.0.0"
       },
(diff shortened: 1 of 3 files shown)


svix-jplatte (approved):
Thanks for the PR!
This makes sense to me and I prefer it over the other solution, though I'll wait to hear from others on the team before merging.
As another piece of data: [caniuse.com page](https://caniuse.com/mdn-api_crypto_randomuuid)

PR #2296, opened 2026-04-27
Add background sql migrations

jaymell:
This adds support for running certain migrations as background tasks 
within the application, which prevents situations where long-running 
migrations, e.g., adding indexes to existing tables, cannot complete
without blocking application startup for an excessive amount of time.

The approach is straightforward. The changes add a background worker, 
similiar to our other background workers, which, if enabled, will 
invoke the long-running migrations in a worker thread that doesn't 
block the rest of the application. If the migration fails, or if 
the application  is restarted before it completes, an optional cleanup 
script will run before the migration is attempted again. The migrations 
can be reverted with a separate `revert-background-migration` command 
if needed. 

There is an optional flag on the migration struct that, if true, will
prevent application startup if the migration does not exist. This 
allows for mandatory migrations to first be added with the flag set
to false, then set to true in a later version of the application that
will actually leverage the changes made by the migration.

This changeset also adds the migration needed for #2183.

### DIFF
--- a/server/svix-server/src/cfg.rs
+++ b/server/svix-server/src/cfg.rs
@@ -246,6 +246,9 @@ pub struct ConfigurationInner {
     #[serde(default = "default_redis_pending_duration_secs")]
     pub redis_pending_duration_secs: u64,
 
+    /// Should this instance run background migrations
+    pub background_migrations_enabled: bool,
+
     #[serde(flatten)]
     pub internal: InternalConfig,
 }
--- a/server/svix-server/src/db/background_migrations.rs
+++ b/server/svix-server/src/db/background_migrations.rs
@@ -41,10 +41,13 @@ async fn advisory_lock_acquire(
 }
 
 async fn advisory_lock_release(conn: &mut sqlx::PgConnection, key: i64) {
-    let _ = sqlx::query("SELECT pg_advisory_unlock($1)")
+    if let Err(e) = sqlx::query("SELECT pg_advisory_unlock($1)")
         .bind(key)
         .execute(conn)
-        .await;
+        .await
+    {
+        tracing::warn!(error = %e, "Failed to release advisory lock");
+    }
 }
 
 // Modeled after sqlx implementation
@@ -184,7 +187,7 @@ pub async fn run(cfg: &Configuration) {
 
 pub async fn run_with_pool(pool: &PgPool, migrations: &[BackgroundMigration]) {
     if let Err(e) = ensure_table(pool).await {
-        tracing::error!("Failed to create _svix_background_migrations table: {e}");
+        tracing::error!(error = %e, "Failed to create _svix_background_migrations table");
         return;
     }
 
@@ -231,8 +234,8 @@ pub async fn try_revert(
     migration: &BackgroundMigration,
 ) -> Result<bool, sqlx::Error> {
     if migration.revert_sql.is_none() {
-        tracing::info!("No revert SQl for migration {}", migration.id);
-    };
+        tracing::warn!(id = migration.id, "No revert SQl for migration");
+    }
 
     let key = lock_key(migration.id);
     let mut conn = pool.acquire().await?;
--- a/server/svix-server/src/db/background_migrations.rs
+++ b/server/svix-server/src/db/background_migrations.rs
@@ -17,7 +17,6 @@ pub static MIGRATIONS: &[BackgroundMigration] = &[BackgroundMigration {
     "#],
     cleanup_sql: None,
     revert_sql: Some(&["DROP INDEX CONCURRENTLY IF EXISTS messageattempt_per_endp_no_status"]),
-    required_for_startup: false,
 }];
 
 pub struct BackgroundMigration {
@@ -29,12 +28,6 @@ pub struct BackgroundMigration {
     pub cleanup_sql: Option<&'static [&'static str]>,
     /// Optional Sql statements to run for reverting the migration.
     pub revert_sql: Option<&'static [&'static str]>,
-    /// If true, the application will fail to start if this migration is missing.
-    ///
-    /// Ideally you would first add the migration with this field set to false, and only
-    /// set it to true in a subsequent application version after verifying the migration has
-    /// successfully run.
-    pub required_for_startup: bool,
 }
 
 async fn advisory_lock_acquire(
@@ -187,45 +180,6 @@ async fn apply_inner(pool: &PgPool, migration: &BackgroundMigration) -> Result<b
 pub async fn run(cfg: &Configuration) {
     let pool = super::connect(&cfg.db_dsn, cfg.db_pool_max_size).await;
     run_with_pool(&pool, MIGRATIONS).await;
-}
-
-pub async fn check_required_for_startup(cfg: &Configuration) {
-    let pool = super::connect(&cfg.db_dsn, cfg.db_pool_max_size).await;
-    if let Err(e) = check_required_for_startup_with_pool(&pool, MIGRATIONS).await {
-        panic!("{e}");
-    }
-}
-
-pub async fn check_required_for_startup_with_pool(
-    pool: &PgPool,
-    migrations: &[BackgroundMigration],
-) -> Result<(), String> {
-    if let Err(e) = ensure_table(pool).await {
-        return Err(format!("Failed to check required migrations: {e}"));
-    }
-    for migration in migrations.iter().filter(|m| m.required_for_startup) {
-        let success: Result<Option<bool>, sqlx::Error> =
-            sqlx::query_scalar("SELECT success FROM _svix_background_migrations WHERE id = $1")
-                .bind(migration.id)
-                .fetch_optional(pool)
-                .await;
-        match success {
-            Ok(Some(true)) => {}
-            Ok(_) => {
-                return Err(format!(
-                    "Required background migration '{}' has not completed.",
-                    migration.id
-                ));
-            }
-            Err(e) => {
-                return Err(format!(
-                    "Failed to check required migration '{}': {e}",
-                    migration.id
-                ));
-            }
-        }
-    }
-    Ok(())
 }
 
 pub async fn run_with_pool(pool: &PgPool, migrations: &[BackgroundMigration]) {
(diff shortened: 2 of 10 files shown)


jaymell:
## OSS Review

Added `crc`, which is already a transitive dependency.

server/svix-server/src/db/background_migrations.rs
```diff
@@ -0,0 +1,335 @@
...
+    pub apply_sql: &'static [&'static str],
+    /// Optional Sql statements that should be run before re-applying if the migration fails.
+    pub cleanup_sql: Option<&'static [&'static str]>,
+    /// Optional Sql statements to run for reverting the migration.
+    pub revert_sql: Option<&'static [&'static str]>,
+    /// If true, the application will fail to start if this migration is missing.
+    ///
+    /// Ideally you would first add the migration with this field set to false, and only
+    /// set it to true in a subsequent application version after verifying the migration has
+    /// successfully run.
```
svix-jplatte: We don't control OSS deployments and this value is defined in code, so what's the idea with this comment?

server/svix-server/src/cfg.rs
```diff
@@ -246,6 +246,9 @@ pub struct ConfigurationInner {
     #[serde(default = "default_redis_pending_duration_secs")]
     pub redis_pending_duration_secs: u64,
 
+    /// Should this instance run background migrations
+    pub background_migrations_enabled: bool,
```
svix-jplatte, line 250: So this is a new _required_ config field? I think usually we introduce stuff as optional and tie the effectively default value to some other setting (maybe `api_enabled` in this case).

server/svix-server/src/db/background_migrations.rs
jaymell: Sorry, maybe the wording is not clear. These are intended to be instructions to us.

If we need to introduce a long-running migration that should be required at some point, we would first add it as NOT required, then in a subsequent version of the application, we would make it required. This will require documentation for upgrading users, and obviously it's not even used right now, but that was the idea.

server/svix-server/src/cfg.rs
jaymell: Well, it has a default of `false` in config.default.toml. There can also be a serde-derived default, though.

server/svix-server/src/db/background_migrations.rs
jaymell: I can remove this idea of required migrations entirely for now. It was suggested elsewhere as a possibility when we first discussed this solution.

svix-jplatte: Okay, I get it now. Maybe easier to not have it at all for now since we've never had releases with mandatory update requirements before and I don't have the impression that we want to start that anytime soon.

But I'm fine with keeping this if you prefer.

server/svix-server/src/cfg.rs
svix-jplatte: Ah, right. But that default means that the bg migrations won't ever run for people if they don't explicitly configure this, right? Is that what we want?

jaymell: My thought was to only enable this for workers. But since we don't really expose a multi-node setup as a default configuration in OSS, I will set the default to true.

server/svix-server/src/db/background_migrations.rs
jaymell: This could be something we need downstream/later. But will remove.

server/svix-server/src/db/background_migrations.rs
```diff
@@ -0,0 +1,289 @@
...
+        .bind(key)
+        .fetch_one(conn)
+        .await
+}
+
+async fn advisory_lock_release(conn: &mut sqlx::PgConnection, key: i64) {
+    let _ = sqlx::query("SELECT pg_advisory_unlock($1)")
+        .bind(key)
+        .execute(conn)
+        .await;
```
svix-jplatte: Shouldn't errors from this at least be logged?

server/svix-server/src/db/background_migrations.rs
```diff
@@ -0,0 +1,289 @@
...
+            .await?;
+
+    match existing {
+        Some(row) if row.get::<bool, _>("success") => return Ok(true),
+        Some(row) => {
+            let attempts: i32 = row.get("attempts");
+            tracing::warn!(
+                id = migration.id,
+                attempts,
+                "Background migration previously failed; running cleanup before retry"
```
svix-jplatte:
```suggestion
                "Background migration previously failed"
```

The 'running cleanup' bit should only be printed if there is some cleanup SQL (and it already is in a separate `tracing::info!`).

server/svix-server/src/db/background_migrations.rs
```diff
@@ -0,0 +1,289 @@
...
+pub async fn ensure_table(pool: &PgPool) -> Result<(), sqlx::Error> {
+    sqlx::query(
+        "CREATE TABLE IF NOT EXISTS _svix_background_migrations (
+            id          TEXT        PRIMARY KEY,
+            started_at  TIMESTAMPTZ NOT NULL DEFAULT now(),
+            finished_at TIMESTAMPTZ,
+            success     BOOLEAN     NOT NULL DEFAULT FALSE,
+            attempts    INTEGER     NOT NULL DEFAULT 0,
+            last_error  TEXT
+        )",
```
svix-jplatte, line 68: Could / should this table be created by a separate normal schema migration instead? I imagine the background migrations would require schemas to be up-to-date anyways, right?

server/svix-server/src/db/background_migrations.rs
```diff
@@ -0,0 +1,289 @@
...
+    if let Err(e) = ensure_table(pool).await {
+        tracing::error!("Failed to create _svix_background_migrations table: {e}");
+        return;
+    }
+
+    let shutdown = crate::shutting_down_token();
+    let mut backoff = Duration::from_secs(5);
+
+    loop {
+        let mut pending = false;
```
svix-jplatte, line 198: Why track this? If we just stopped the background-migration worker once all migrations are either done, or in progress within another process, would that not make more sense? The only reason this would happen is if another process that is also set up to run bg migrations is currently running, right?

server/svix-server/src/db/background_migrations.rs
```diff
@@ -0,0 +1,289 @@
...
+    }
+}
+
+pub async fn try_revert(
+    pool: &PgPool,
+    migration: &BackgroundMigration,
+) -> Result<bool, sqlx::Error> {
+    if migration.revert_sql.is_none() {
+        tracing::info!("No revert SQl for migration {}", migration.id);
+    };
```
svix-jplatte:
```suggestion
    }
```

server/svix-server/src/db/background_migrations.rs
```diff
@@ -0,0 +1,289 @@
...
+        return Ok(false);
+    }
+
+    let result = revert_inner(pool, migration, migration.revert_sql).await;
+    advisory_lock_release(&mut conn, key).await;
+    result
+}
+
+// If there's no actual revert script, only the migrations table row
+// will be deleted.
```
svix-jplatte, line 253: When would it be useful to delete the row for a migration that can't be reverted? I'd think it makes more sense to log an error and stop.

server/svix-server/src/db/background_migrations.rs
jaymell: It could, I don't see a reason to, though. It makes more sense to me to keep all this in one place.

svix-jplatte: Okay

server/svix-server/src/db/background_migrations.rs
jaymell: Possibly. My thinking was that in the event of a failed migration that required some manual database intervention, it might be helpful to be able to remove the row and allow things to reapply. This is a scenario I run into often locally, anyway.

svix-jplatte: Okay, then maybe keep the logic as-is but add a `tracing::warn!` for when there isn't a down migration?

server/svix-server/src/db/background_migrations.rs
jaymell: Well, the worker does stop when all the migrations are done.

My thought was that it's safer to have all processes that were tasked with running migrations to keep checking on them until they're all done. In practice, it's probably not necessary, but it seems worse to me to have some edge case where all the migration workers end up stopping before the migrations are actually run.

svix-jplatte: > seems worse to me to have some edge case where all the migration workers end up stopping before the migrations are actually run

I don't see why that would happen, but okay.

server/svix-server/src/db/background_migrations.rs
```diff
@@ -0,0 +1,292 @@
...
+            .bind(migration.id)
+            .fetch_optional(pool)
+            .await?;
+
+    if success != Some(true) {
+        return Ok(false);
+    }
+
+    tracing::info!(id = migration.id, "Reverting background migration");
+    if let Some(statements) = statements {
```
svix-jplatte, line 270:
```suggestion
    if let Some(statements) = statements {
        tracing::info!(id = migration.id, "Reverting background migration");
```

server/svix-server/src/db/background_migrations.rs
```diff
@@ -0,0 +1,292 @@
...
+                    "UPDATE _svix_background_migrations SET last_error = $1 WHERE id = $2",
+                )
+                .bind(e.to_string())
+                .bind(migration.id)
+                .execute(pool)
+                .await;
+                return Err(e);
+            }
+        }
+    }
```
svix-jplatte, line 283:
```suggestion
    } else {
        tracing::warn!("There is no revert SQL, dropping the migration row only");
    }
```

svix-jplatte (approved):
Overall looks good to me, though would be nice to add that warning.

server/svix-server/src/db/background_migrations.rs
jaymell: Oh, there's already a warn previously, at line 237, which I think is sufficient.

svix-jplatte: Oh okay, missed that.

PR #2297, opened 2026-04-29
Ensure ordered background migrations

jaymell:
Ensure that background migrations can't be run out of order by using a global instead of per-migration lock. Also, don't allow reverting a migration if it's not the last-applied migration.

This also simplifies the workers by simply exiting if there's already a migration lock acquired.

### DIFF
--- a/server/svix-server/src/db/background_migrations.rs
+++ b/server/svix-server/src/db/background_migrations.rs
@@ -105,7 +105,7 @@ pub async fn run_migrations(
     Ok(true)
 }
 
-async fn apply(pool: &PgPool, migration: &BackgroundMigration) -> Result<bool, sqlx::Error> {
+async fn apply(pool: &PgPool, migration: &BackgroundMigration) -> Result<(), sqlx::Error> {
     let existing =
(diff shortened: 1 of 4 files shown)


server/svix-server/src/db/background_migrations.rs
```diff
@@ -72,29 +82,30 @@ pub async fn ensure_table(pool: &PgPool) -> Result<(), sqlx::Error> {
...
+    if !advisory_lock_acquire(&mut conn).await? {
+        return Ok(false);
     }
 
-    let result = apply_inner(pool, migration).await;
-    advisory_lock_release(&mut conn, key).await;
-    result
+    let mut apply_result = Ok(());
+    for migration in migrations {
+        if let Err(e) = apply(pool, migration).await {
```
svix-jbrown, line 97: what if `apply` returns `Ok(false)`?

jaymell: Nothing. The bool was vestigial. Removed.

svix-jbrown (approved):

PR #2302, opened 2026-05-04
Rust: Add AutoConfig module

svix-lucho:
Uses internal APIs for now.

`AutoConfig` decodes the autoconfig token, saves its params and uses them to call `subscribe()` or `verify()`. 

I'll do the other langs once we like the approach with this one. 

Note that I had to do some tweaks to the templates to make the `models` import work for both public and internal models inside the api resource files.


rust/src/internal/api/endpoint.rs
```diff
@@ -0,0 +1,37 @@
+// this file is @generated
+use super::EndpointAutoConfig;
+use crate::{error::Result, models::*, Configuration};
```
svix-lucho, line 3: this is wrong because `SubscribeIn` is in `internal::models`

codegen/templates/rust/api_resource.rs.jinja
```diff
@@ -5,6 +5,7 @@ use crate::{
 {%- if resource.operations | length > 0 %}
     error::Result,
     models::*,
+    internal::models::*,
 {%- endif %}
     Configuration,
 };
```
svix-jplatte, line 10:
```suggestion
use crate::{
    {%- if resource.operations | length > 0 %}
    error::Result,
    {%- endif %}
    Configuration,
};
use super::models::*;
```

svix-jplatte (commented):
Got a simple solution for you ;)

codegen/templates/rust/api_resource.rs.jinja
svix-lucho: Hmm I tried this and it didn't work.

svix-lucho: oh I see the fix, one sec

codegen/templates/rust/api_resource.rs.jinja
```diff
@@ -2,12 +2,12 @@
...
 use crate::{
-{%- if resource.operations | length > 0 %}
+    {%- if resource.operations | length > 0 %}
     error::Result,
-    models::*,
-{%- endif %}
+    {%- endif %}
     Configuration,
 };
+use super::models::*;
```
svix-jplatte: Sorry for the fail, actually has to be `super::super::models::*`. Ugly, but works. Maybe also move it in the existing `use super::{}` block below?

svix-lucho: I couldn't move it in the existing one because I have to include it conditionally, otherwise I get clippy errors for the unused import

rust/src/autoconfig.rs
```diff
@@ -0,0 +1,136 @@
...
+    endpoint_id: String,
+    endpoint: EndpointIn,
+    webhook: Webhook,
+    svix: Svix,
+}
+
+const AUTOCONFIG_TOKEN_PREFIX_V1: &str = "auto_v1_";
+
+#[derive(thiserror::Error, Debug)]
+pub enum AutoConfigError {
```
svix-jplatte, line 38:
```suggestion
#[non_exhaustive]
pub enum AutoConfigError {
```

rust/src/autoconfig.rs
```diff
@@ -0,0 +1,136 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn decode_autoconfig_token_v1_parses_payload() {
+        let json = r#"{"aid":"app_1","eid":"ep_2","surl":"https://api.example.test","esec":"whsec_Zm9v","tok":"sk_test_xyz"}"#;
+        let token = format!("{}{}", AUTOCONFIG_TOKEN_PREFIX_V1, base64::encode(json));
```
svix-jplatte:
```suggestion
        let token = format!("{AUTOCONFIG_TOKEN_PREFIX_V1}{}", base64::encode(json));
```

rust/src/autoconfig.rs
```diff
@@ -0,0 +1,136 @@
...
+            Err(AutoConfigError::InvalidToken)
+        ));
+    }
+
+    #[test]
+    fn decode_autoconfig_token_v1_rejects_invalid_json() {
+        let token = format!(
+            "{}{}",
+            AUTOCONFIG_TOKEN_PREFIX_V1,
+            base64::encode("not json")
```
svix-jplatte:
```suggestion
            "{AUTOCONFIG_TOKEN_PREFIX_V1}{}",
```

svix-jplatte (dismissed):

svix-jplatte (approved):

PR #2304, opened 2026-05-04
Tests: Add some cleanup

jaymell:
Cover a couple spots where cleanup might be needed before test runs.


tasn (approved):
Unclear commit message. :|

PR #2308, opened 2026-05-05
python: autoconfig sdk

svix-lucho:
Same as https://github.com/svix/svix-webhooks/pull/2302 - for Python, using an internal API

### DIFF
--- a/python/svix/api_internal/endpoint.py
+++ b/python/svix/api_internal/endpoint.py
@@ -1,0 +1,61 @@
+# This file is @generated
+from deprecated import deprecated
+
+from ..models import EndpointTransformationIn
+from .common import ApiBase
+from .endpoint_auto_config import (
+    EndpointAutoConfig,
+    EndpointAutoConfigAsync,
+)
+
+
+class EndpointAsync(ApiBase):
+    @property
+    def auto_config(self) -> EndpointAutoConfigAsync:
+        return EndpointAutoConfigAsync(self._client)
+
+    @deprecated
+    async def transformation_partial_update(
+        self,
+        app_id: str,
+        endpoint_id: str,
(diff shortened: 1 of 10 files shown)


python/svix/api_internal/endpoint.py
```diff
@@ -0,0 +1,61 @@
...
+)
+
+
+class EndpointAsync(ApiBase):
+    @property
+    def auto_config(self) -> EndpointAutoConfigAsync:
+        return EndpointAutoConfigAsync(self._client)
+
+    @deprecated
+    async def transformation_partial_update(
```
svix-jplatte, line 18: Uhhh, why is this here 😄
Has to be a bug in openapi-codegen, right?

python/svix/autoconfig.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+class AutoConfigError(Exception):
+    """Raised when an autoconfig token cannot be decoded or validated."""
+
+
+class _AutoConfigTokenContentV1(pydantic.BaseModel):
+    """Payload embedded in a v1 autoconfig token (short JSON keys)."""
+
+    model_config = pydantic.ConfigDict(populate_by_name=True)
```
svix-jplatte: Why? This is deprecated and allows both field name and alias to be used in deserialization. Alias is used by default.

https://pydantic.dev/docs/validation/latest/api/pydantic/config/#pydantic.config.ConfigDict.populate_by_name

python/svix/api_internal/endpoint.py
svix-lucho: hmm I think so? 😆 
happened with the other languages too, I don't know exactly where it's coming from.

svix-jplatte (approved):

PR #2312, opened 2026-05-06
Java: fix java 1.8 support in the Webhooks verification too

tasn:


svix-jbrown (commented):
gross

can we refactor this into a `WebhookBase` class that has all the constants and whatnot in it, and then subclass from that in both the Java8 and Java11 implementations?

svix-jbrown (approved):

ktham:
Hi, this PR had a slightly breaking change for users on Java 11 and higher.

For `com.svix.Webhook`, the method signature for `verify()` changed the `headers` parameter type from `HttpHeaders` to `Map<String, List<String>>`.

It's not a big deal, but it would be nice if this was in the Java SDK changelog for `1.93.0` since this is a breaking change for those on Java 11 and higher.

In 1.92.2:
https://github.com/svix/svix-webhooks/blob/7c8ce9ac1ae8f4eb3993302e5965bc05a04a8863/java/lib/src/main/java/com/svix/Webhook.java#L43

In 1.93.0:
https://github.com/svix/svix-webhooks/blob/81bf9e97b381c2b005ed50a4184c8aebb67d3059/java/lib/src/main/java/com/svix/WebhookBase.java#L48

Or perhaps consider keeping an overloaded version `public void verify(final String payload, final HttpHeaders headers)` to avoid breaking those on Java 11 and higher (edit: actually, I realized you do have this overloaded verify method in the java11 version of `Webhook.java`, hmm....)

---
For reference, we are seeing this error when attempting to upgrading the Svix Java SDK from `1.92.2` to `1.93.0` (note: we are using this library from Scala 2.13.18 and JDK17)

> [error]  found   : java.net.http.HttpHeaders
[error]  required: java.util.Map[String,java.util.List[String]]
[error]       webhook.verify(payload, javaHeaders)
[error]                               ^
[error] one error found
[error] (Compile / compileIncremental) Compilation failed

tasn:
Whoops, this was not intended, thank you for letting me know. Fix incoming!

PR #2323, opened 2026-05-12
php: autoconfig sdk

svix-lucho:
Same as https://github.com/svix/svix-webhooks/pull/2302 - for PHP, using an internal API

Had to add the same weird perl script we have for Go to rewrite the internal package names.

### DIFF
--- a/php/src/ApiInternal/Endpoint.php
+++ b/php/src/ApiInternal/Endpoint.php
@@ -1,0 +1,31 @@
+<?php
+
+// this file is @generated
+declare(strict_types=1);
+
+namespace Svix\ApiInternal;
+
+use Svix\Models\EndpointTransformationIn;
+use Svix\Request\SvixHttpClient;
+
+class Endpoint
+{
+    public EndpointAutoConfig $autoConfig;
+
+    public function __construct(
+        private readonly SvixHttpClient $client,
+    ) {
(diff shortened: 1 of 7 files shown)


php/src/ApiInternal/Endpoint.php
```diff
@@ -0,0 +1,31 @@
...
+{
+    public EndpointAutoConfig $autoConfig;
+
+    public function __construct(
+        private readonly SvixHttpClient $client,
+    ) {
+        $this->autoConfig = new EndpointAutoConfig($client);
+    }
+
+    /** This operation was renamed to `set-transformation`. */
```
svix-james, line 21: Can/should we do this? https://www.php.net/manual/en/class.deprecated.php

svix-lucho: I looked into this and apparently it's a PHP 8.4 feature, and our minimum supported version is 8.1. There may be other ways to add the deprecated flag, but it requires other changes to the codegen. I'll create a ticket for this, since it's not strictly related to my change, and I need Jonas to know how to do it properly.

svix-lucho: https://github.com/svix/monorepo-private/issues/13056

svix-james (approved):

PR #2327, opened 2026-05-13
[PHP] Add missing throws phpdoc

VincentLanglet:
## Motivation

Hi, the actual class methods are missing `@throws` phpdoc to explains which exception are thrown.
This is useful
- To know quickly what should be caught
- To have an alert from PHPStorm for uncaught exception
- To have an alert from Static analysis tool like PHPStan about uncaught exception

Currently when I try to `try/catch` some of svix code, PHPStan report that no exception are thrown because of the wrong PHPDoc.

## Solution

I update the phpdoc to show the expected result, but I also so some files are auto-generated, so I dunno where/what should be changed instead.

Could you help me on this @svix-lucho @svix-jbrown @svix-jplatte (I pinged some recent commiters, feel free to redirect me to the right person).

Thanks

### DIFF
--- a/codegen/templates/php/api_extra/application_create.php
+++ b/codegen/templates/php/api_extra/application_create.php
@@ -1,4 +1,8 @@
-/** Get or create a new application */
+/**
+ * Get or create a new application
+ *
+ * @throws ApiException
+ */
 public function getOrCreate(
     ApplicationIn $applicationIn,
     ?ApplicationCreateOptions $options = null,
--- a/codegen/templates/php/api_resource.php.jinja
+++ b/codegen/templates/php/api_resource.php.jinja
@@ -5,6 +5,7 @@ declare(strict_types=1);
 
 namespace Svix\Api;
 
+use Svix\Exception\ApiException;
 use Svix\Request\SvixHttpClient;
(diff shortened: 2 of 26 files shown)


svix-jbrown:
So, everything under `php/src/Api` is generated, which means this change will actually have to go in the template in `codegen/templates/php/`, and then you'll need to rerun the codegen script as described [in the README](https://github.com/svix/svix-webhooks/blob/main/codegen/README.md).

VincentLanglet:
> So, everything under `php/src/Api` is generated, which means this change will actually have to go in the template in `codegen/templates/php/`, and then you'll need to rerun the codegen script as described [in the README](https://github.com/svix/svix-webhooks/blob/main/codegen/README.md?rgh-link-date=2026-05-13T16%3A42%3A03Z).

Thanks for your quick answer, I updated my PR using the templates jinja then.
`python3 regen_openapi.py` now produces no changes.

svix-jbrown (approved):

svix-jbrown:
Thank you for your contribution!

VincentLanglet:
Thanks ! When is planned the next release ?

PR #2332, opened 2026-05-14
Java webhook verification: remove final from sign and verify functions.

tasn:
This was a mistake, and was not intended.

Reported in https://github.com/svix/svix-webhooks/pull/2312\#issuecomment-4454450060.

Thanks @ktham!

### DIFF
--- a/java/lib/src/main/java/com/svix/WebhookBase.java
+++ b/java/lib/src/main/java/com/svix/WebhookBase.java
@@ -45,7 +45,7 @@ abstract class WebhookBase {
         this.key = secret;
     }
 
-    public final void verify(final String payload, final Map<String, List<String>> headers)
+    public void verify(final String payload, final Map<String, List<String>> headers)
             throws WebhookVerificationException {
         String msgId = firstHeader(headers, SVIX_MSG_ID_KEY);
         String msgSignature = firstHeader(headers, SVIX_MSG_SIGNATURE_KEY);
@@ -101,7 +101,7 @@ abstract class WebhookBase {
         return null;
     }
 
-    public final String sign(final String msgId, final long timestamp, final String payload)
+    public String sign(final String msgId, final long timestamp, final String payload)
             throws WebhookSigningException {
         try {
             String toSign = String.format("%s.%s.%s", msgId, timestamp, payload);


svix-jbrown (approved):

svix-pluiz:
> Reported in [https://github.com/svix/svix-webhooks/pull/2312\#issuecomment-4454450060](https://github.com/svix/svix-webhooks/pull/2312%5C#issuecomment-4454450060).

Issue link is broken, correct is https://github.com/svix/svix-webhooks/pull/2312#issuecomment-4454450060

ktham:
Thank you for the quick follow up @tasn !!

Quick note though, the issue for us isn't that newly added `void verify(final String payload, final Map<String, List<String>> headers)` method has a `final` modifier, but rather the compiler no longer sees the existence of a `verify(final String payload, final HttpHeaders headers)` method.

Given that, I suspect that removing the `final` modifier might not help here.

Here's the error we hit against 1.93.0 (JDK 17, Scala 2.13.18):

```
[error] found : java.net.http.HttpHeaders
[error] required: java.util.Map[String,java.util.List[String]]
[error] webhook.verify(payload, javaHeaders)
[error] ^
[error] one error found
[error] (Compile / compileIncremental) Compilation failed
```

I do see that you added a version of `Webhook.java` under `src/main/java11` that has the existing (now overloaded) `verify(final String payload, final HttpHeaders headers)` method (which I initially overlooked and missed when looking at #2312): https://github.com/svix/svix-webhooks/blob/81bf9e97b381c2b005ed50a4184c8aebb67d3059/java/lib/src/main/java11/com/svix/Webhook.java#L20-L23

But for some reason, the java/scala compiler isn't able to find it from the svix `1.93.0` jar 🤔 You may want to try to test and confirm on your end as well.

I personally don't have experience with Java's [Multi-Release JAR (MRJAR) Files](https://openjdk.org/jeps/238) which #2312 seems to be attempting to use, however, I see this in JEP 238 (bolded for emphasis):

> An MRJAR is essentially one unit of release, so it has just one release version (which is no different from a normal JAR, distributed say via Maven Central), even though internally it contains multiple versions of a library implementation for use on different Java platform releases. **Every version of the library should offer the same API**

This leads me to suspect that MRJAR can only be used to swap out _implementations_ of a specific API rather than add new APIs (i.e. additional overloaded methods). And if that's the case, you may want to consider whether you want to keep `src/main/java11/com/svix/Webhook.java` around.

Again, this isn't a big deal, we can adjust our call sites to pass `Map<String, List<String>>` instead of `HttpHeaders`, so this isn't blocking us.

But it might be good to update the release notes or changelog for https://github.com/svix/svix-webhooks/releases/tag/v1.93.0 if possible to make note of the fact that users of on JDK11 or higher using the `verify(final String payload, final HttpHeaders headers)` method will need to make changes when upgrading to `1.93.0` to pass in `Map<String, List<String>>` instead of `HttpHeaders`.

tasn:
Thanks @ktham for the feedback, and sorry for misunderstanding you. I thought the issue was inheritance!

Anyhow, we have tests to test verify with HttpHeaders and they seem to work: https://github.com/svix/svix-webhooks/blob/main/java/lib/src/test/com/svix/test/WebhookTest.java

Maybe it's because the "linking" is done differently with tests? I'm not sure, but it looks like you're right.

I'm checking what we can do. :|

ktham:
No worries, thank you for looking into it!

> Anyhow, we have tests to test verify with HttpHeaders and they seem to work: https://github.com/svix/svix-webhooks/blob/main/java/lib/src/test/com/svix/test/WebhookTest.java

Hmm, it looks like #2312 updated that test such that `HttpHeaders` is no longer being passed in to `verify(...)` anymore.

I see there are 3 calls where `testPayload.headers()` is passed in:

https://github.com/svix/svix-webhooks/blob/81bf9e97b381c2b005ed50a4184c8aebb67d3059/java/lib/src/test/com/svix/test/WebhookTest.java#L34

https://github.com/svix/svix-webhooks/blob/81bf9e97b381c2b005ed50a4184c8aebb67d3059/java/lib/src/test/com/svix/test/WebhookTest.java#L49

https://github.com/svix/svix-webhooks/blob/81bf9e97b381c2b005ed50a4184c8aebb67d3059/java/lib/src/test/com/svix/test/WebhookTest.java#L68

However, this value is not a `HttpHeaders` type anymore but a `Map<String, List<String>>` type: https://github.com/svix/svix-webhooks/blob/81bf9e97b381c2b005ed50a4184c8aebb67d3059/java/lib/src/test/com/svix/test/WebhookTest.java#L214

ktham:
Hi @tasn! Just wanted to check in if you plan to make any changes here?

Should we update our call sites for `webhook.verify(payload, javaHeaders)` to pass `javaHeaders: Map<String, List<String>>` instead of `javaHeaders: HttpHeaders` in order to upgrade the Svix SDK from `1.92.2` to a version higher than or equal to `1.93.0`?

I'd like to get confirmation that if we do this refactor, that the interface for `verify(...)` won't be changing again (i.e. potentially reverting changes from #2312, which would cause the interface for `verify(...)` to change back to `javaHeaders: HttpHeaders`)

Again I would suggest updating the [changelog for 1.93](https://github.com/svix/svix-webhooks/releases/tag/v1.93.0) (if possible) to make note of this breaking change.

tasn:
Let me check with the team, I don't remember what the consensus was for this. I also asked other people for feedback but haven't heard back.

tasn:
@svix-jplatte is on it.

svix-jplatte:
Hey @ktham 

> Should we update our call sites for `webhook.verify(payload, javaHeaders)` to pass `javaHeaders: Map<String, List<String>>` instead of `javaHeaders: HttpHeaders` in order to upgrade the Svix SDK from `1.92.2` to a version higher than or equal to `1.93.0`?

Yes. Since we can apparently not have the `HttpHeaders` overload without sacrificing Java 8 compatibility, we will not add it back for the foreseeable future (in any way other than the mrjar thing implemented here, which appears to not work for you, or at all - I do want to get back to that though).

> I'd like to get confirmation that if we do this refactor, that the interface for `verify(...)` won't be changing again (i.e. potentially reverting changes from #2312, which would cause the interface for `verify(...)` to change back to `javaHeaders: HttpHeaders`)

Of course, we never intended to break things in the first place, and just had to change it to restore compatibility with Java 8 that we had accidentally broken. I have a PR up to test Java 8 in CI so that never happens again (https://github.com/svix/svix-webhooks/pull/2382).

> Again I would suggest updating the [changelog for 1.93](https://github.com/svix/svix-webhooks/releases/tag/v1.93.0) (if possible) to make note of this breaking change.

I want to better understand the issues with mrjar before doing this, so it might take a bit, but we'll update the changelog once we know the exact details.

ktham:
Hi @svix-jplatte ! 😃 

Thank you so much for following up on this and for confirming. What you said makes sense (it would also be another breaking change to revert these changes 😅 ) We'll proceed with updating our code to use the new version of the `verify` method when we upgrade to `1.93.0+`

> Of course, we never intended to break things in the first place, and just had to change it to restore compatibility with Java 8 that we had accidentally broken. I have a PR up to test Java 8 in CI such that never happens again (https://github.com/svix/svix-webhooks/pull/2382).

Hmm, I'm not 100% sure as I haven't tried building under Java 8, but I suspect there never was compatibility with Java 8.

I see that the `verify(final String payload, final HttpHeaders headers)` method  had always used `java.net.http.HttpHeaders` since the initial release 5 years ago (until `1.93.0`, when it changed to use a Java Map): https://github.com/svix/svix-webhooks/blob/9ec53f5a76a3a9e8d4d89b938f62e614313a172b/java/lib/src/main/java/com/svix/Webhook.java#L26

So it seemed like what happened here was that we _added_ rather than _restored_ compatibility for Java 8.

PR #2345, opened 2026-05-26
Avoid deprecation about null values

VincentLanglet:
## Motivation

Hi @svix-jbrown ; I got recently the error
```
Deprecated: DateTimeImmutable::__construct(): Passing null to parameter #1 ($datetime) of type string is deprecated
```
Seems like the method `deserializeDt` doesn't handle null correctly

## Solution

Early null check

### DIFF
--- a/php/src/Utils.php
+++ b/php/src/Utils.php
@@ -53,10 +53,14 @@ class Utils
         return $val;
     }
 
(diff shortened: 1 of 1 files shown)


svix-pluiz:
LGTM

svix-pluiz (approved):

PR #2357, opened 2026-05-27
py: make tests pass ruff and ty, and work better

svix-jbrown:
I'm optimistic that this will be the last Python client library change for a little while.

Changes:

 - Make all tests pass `ruff` and `ty`
 - Make tests that don't need the docker server not depend on it, so you can run `pytest tests/test_webhooks.py` quickly
 - Make client tests create a distinct event name and object UIDs for every test so that they work even if not running the wiper
 - Bump pgbouncer image to one that has an arm64 build for me
 - Use host.docker.internal instead of hardcoding an address from some specific version of docker when running locally (fixes tests on orbstack and probably podman)
 - Add an editorconfig for the python directory
 - Add some missing items to `.dockerignore`

### DIFF
--- a/python/scripts/lint.sh
+++ b/python/scripts/lint.sh
@@ -3,6 +3,9 @@
 set -ex
 
 mypy svix
-ty check svix
-ruff check svix
-ruff format --check svix
+ty check svix/
+ty check tests/
+ruff check svix/
+ruff check tests/
+ruff format --check svix/
+ruff format --check tests/
--- a/.github/workflows/python-tests.yml
+++ b/.github/workflows/python-tests.yml
@@ -22,12 +22,8 @@ jobs:
         working-directory: ./python
         run: uv sync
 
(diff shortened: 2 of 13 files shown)


python/scripts/lint.sh
```diff
@@ -3,6 +3,9 @@
 set -ex
 
 mypy svix
-ty check svix
-ruff check svix
-ruff format --check svix
+ty check svix/
+ty check tests/
+ruff check svix/
+ruff check tests/
```
svix-jplatte, line 9: Just `ruff check` will cover both, no? Will it cover anything we don't want? Same for ty probably.

svix-jbrown: it'll also get the `example.py` (which doesn't pass because it references pretend modules) and, depending on your ignore settings, can get the virtualenv

svix-jplatte (approved):

PR #2365, opened 2026-06-01
py: include the python version and platform in the user-agent header

svix-jbrown:
This is some *extremely minor*, non-PII-related telemetry so we can try to get a feel for what versions of Python are in use by cloud customers.

### DIFF
--- a/codegen/templates/python/summary.py.jinja
+++ b/codegen/templates/python/summary.py.jinja
@@ -2,6 +2,7 @@
 # this file is @generated
 
 import typing as t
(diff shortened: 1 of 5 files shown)


svix-jplatte (dismissed):

svix-jplatte (commented):
Oh, it's a generated file. We should probably split the `__init__` logic (and `SvixOptions` decl) out into a function in a different file that's not generated..

svix-jplatte (approved):

PR #2370, opened 2026-06-01
js: add node version and platform to user-agent string

svix-jbrown:
similar to #2365

### DIFF
--- a/javascript/src/request.ts
+++ b/javascript/src/request.ts
@@ -2,7 +2,21 @@ import { ApiException, type XOR } from "./util";
 import type { HttpErrorOut, HTTPValidationError } from "./HttpErrors";
 
 export const LIB_VERSION = "1.95.1";
-const USER_AGENT = `svix-libs/${LIB_VERSION}/javascript`;
+
+function getUserAgent() {
+  var fields = [`svix-libs/${LIB_VERSION}/javascript`];
+  if (process !== undefined) {
+    if (process.version !== undefined) {
+      fields.push(`node/${process.version}`);
+    }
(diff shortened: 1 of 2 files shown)


javascript/src/request.ts
```diff
@@ -2,7 +2,21 @@ import { ApiException, type XOR } from "./util";
 import type { HttpErrorOut, HTTPValidationError } from "./HttpErrors";
 
 export const LIB_VERSION = "1.95.1";
-const USER_AGENT = `svix-libs/${LIB_VERSION}/javascript`;
+
+function getUserAgent() {
+  var fields = [`svix-libs/${LIB_VERSION}/javascript`];
+  if (process !== undefined) {
+    if (process.version !== undefined) {
+      fields.push(`node/${process.version}`);
```
svix-jplatte, line 10: What about other runtimes? `process.version` doesn't sound very node-specific.

svix-jbrown: bun doesn't have it (they do expose a compatibility module at `node:process` which  I guess I could handle). Our code doesn't work in Bun at all, though. Can't even import the Svix module...

```
-------------------------------
1 | import { Svix } from "..";
2 | import type { SvixRequestContext } from "../request";
3 | 
4 | export class SvixInternal extends Svix {
                                      ^
ReferenceError: Cannot access 'Svix' before initialization.
      at /Users/jbrown/pg/svix-webhooks/javascript/src/api_internal/index.ts:4:35
```

I filed #2379 for that and will try to fix it

svix-jplatte (approved):

PR #2372, opened 2026-06-01
rust: add rustc version and platform to the user-agent string

svix-jbrown:
Basically the same as #2365.

I'm sort of the least interested in this one, because the Rust client has the least behavioral dependence on the platform, but it seems worth including for completeness.

<details>

<summary>dependency review</summary>

 - adds [`version_check`](https://crates.io/crates/version_check) as a direct dependency, which was already an indirect dependency. Apache-2.0/MIT license, actively maintained
 - adds [`nix`](https://crates.io/crates/nix) as a dependency, which every other rust program on the planet has. MIT license, actively maintained.

</details>

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -43,6 +43,7 @@ hyper-tls = { version = "0.6.0", optional = true }
 hyper-util = { version = "0.1.16", features = ["client", "client-legacy", "tokio"] }
 itertools = "0.14.0"
 js_option = "0.1.1"
+nix = "0.31"
 percent-encoding = "2.3.1"
 rand = "0.9.0"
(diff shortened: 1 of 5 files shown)


rust/Cargo.toml
```diff
@@ -43,6 +43,7 @@ hyper-tls = { version = "0.6.0", optional = true }
 hyper-util = { version = "0.1.16", features = ["client", "client-legacy", "tokio"] }
 itertools = "0.14.0"
 js_option = "0.1.1"
+nix = "0.31"
```
svix-jplatte: Surely you meant to only include the target-specific one below?

svix-jplatte (approved):

PR #2380, opened 2026-06-02
python: Handle with_content client-side

svix-jplatte:
Part of https://github.com/svix/monorepo-private/issues/13162.

### DIFF
--- a/codegen/templates/python/api_resource.py.jinja
+++ b/codegen/templates/python/api_resource.py.jinja
@@ -50,7 +50,12 @@ class {{ resource_type_name }}{{ op.name | to_upper_camel_case }}Options(BaseOpt
     def _query_params(self) -> t.Dict[str, str]:
         return serialize_params({
         {%- for p in op.query_params %}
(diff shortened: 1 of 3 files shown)


codegen/templates/python/api_resource.py.jinja
```diff
@@ -149,7 +154,12 @@ class {{ resource.name | to_upper_camel_case }}{% if is_async %}Async{% endif %}
                 {% if op.request_body_schema_name is defined -%}
                 ,json_body={{ op.request_body_schema_name | to_snake_case }}.model_dump_json(exclude_unset=True, by_alias=True)
                 {%- endif -%})
-            {% if op.response_body_schema_name is defined -%}
+            {% if op.id == "v1.message.create" -%}
+            message_out = MessageOut.model_validate(response.json())
+            if options.with_content is None or options.with_content:
```
svix-jbrown: tiny nit, but I would spell this as `if options.with_content is not False:`

svix-jbrown (approved):

PR #2387, opened 2026-06-05
Autoconfig for polling endpoints

svix-lucho:
Add an AutoConfig SDK for polling endpoints, similar to the one we have for regular webhook endpoints (https://docs.svix.com/receiving/webhooks-autoconfig). I decided to make this a separate class/type because the methods provided by the autoconfig polling class are completely different from the webhooks autoconfig. There are no `receive` and `commit` methods in webhooks autoconfig, and there is no signature verification in polling autoconfig.

The changes here involve updating the codegen to add the `pollerv2.consumer-poll` and `pollerv2.consumer-commit` methods, and adding a new `AutoConfigConsumer` class in each language.

The API looks roughly like this:

```typescript
import { AutoConfigConsumer } from "svix";

const consumer = new AutoConfigConsumer(AUTO_CONFIG_TOKEN, {
  filterTypes: ["invoice.paid", "user.created"],
  url: "https://api.us.example.com/webhooks/acme",
});

// Creates the polling endpoint - run once per deploy or app start
await consumer.subscribe();

// Inside some worker loop
const msgs = consumer.receive(consumer_id);
// ... process msgs
consumer.commit(consumer_id, offset);

```


codegen/templates/python/api_resource.py.jinja
```diff
@@ -16,6 +16,10 @@ from ..models import {{ c | to_upper_camel_case }}
 {% if resource.name == "authentication" -%}
 from ..models import DashboardAccessOut
 {% endif -%}
+{# HACK: api_internal/message.py generates `message_in_raw` which requires `MessageIn` -#}
+{% if resource.name == "message" -%}
+from ..models import MessageIn
+{% endif -%}
```
svix-lucho: bad hack

svix-jplatte: How about modifying the existing template hack that generates `message_in_raw` so it doesn't trigger in `api_internal`, instead?

svix-lucho: yesss I was in the middle of doing exactly that

svix-lucho: fixed

codegen/templates/python/api_resource.py.jinja
```diff
@@ -162,6 +166,7 @@ class {{ resource.name | to_upper_camel_case }}{% if is_async %}Async{% endif %}
             {% elif op.response_body_schema_name is defined -%}
             return {{ op.response_body_schema_name | to_upper_camel_case }}.model_validate(response.json())
             {% endif %}
+            {% if not (context is defined and context.api_internal) -%}
```
svix-jplatte: What is this needed for? The `api_operation_extra` file is only going to exist in the first place if the same operation is included in both the public and internal API, no?

csharp/Svix/ApiInternal/Message.cs
svix-jplatte, line 1: Looks like some of the new files aren't listed in the `.gitattributes` globs, mind fixing that?

svix-lucho: Fixed. This is much easier to review now 😅

.gitattributes
```diff
@@ -20,8 +20,8 @@
 /php/src/Models/** linguist-generated
 /php/src/Svix.php linguist-generated
 /python/svix/__init__.py linguist-generated
+/python/svix/api/** linguist-generated
```
svix-jplatte, line 23: We should probably move `api/client.py` somewhere else (up one level), because it's not generated.

svix-lucho: apparently `-linguist-generated` is a thing, we can use that https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github

codegen/templates/python/api_resource.py.jinja
svix-lucho: Ahh ok. Yeah, I just tried it, you're right. No changes in the generation after removing this.

.gitattributes
svix-jplatte: nice 👍

csharp/Svix/AutoConfigConsumer.cs
```diff
@@ -0,0 +1,114 @@
...
+using Svix.Models;
+
+namespace Svix
+{
+    public class AutoConfigConsumer
+    {
+        private readonly string _appId;
+        private readonly string _sinkId;
+        private readonly SinkInCommon _sinkIn;
+        private readonly SvixClient _client;
```
svix-jplatte: Why the underscore prefixes? I see we do this in some other places for C# as well, but not consistently.

go/autoconfig.go
```diff
@@ -34,7 +34,8 @@ type AutoConfig struct {
 	svix       *internalapi.InternalSvix
 }
 
-func decodeAutoConfigTokenV1(token string) (*autoConfigTokenContentV1, error) {
+// DecodeAutoConfigTokenV1 parses and validates a v1 auto-config token.
+func DecodeAutoConfigTokenV1(token string) (*autoConfigTokenContentV1, error) {
```
svix-jplatte: I don't think this should be public? Can you not access it from the other file which is part of the same package without making it public? If so, we should probably move it to the internal (not `internalapi`) package.

svix-jplatte: Or, as another option, move `AutoConfigConsumer` to this file.

java/lib/src/main/java/com/svix/AutoConfig.java
```diff
@@ -8,164 +8,160 @@
...
 import com.svix.models.SubscribeIn;
-
 import java.io.IOException;
 import java.util.Base64;
 import java.util.List;
 import java.util.Map;
 
 public final class AutoConfig {
-    public static final String AUTOCONFIG_TOKEN_PREFIX_V1 = "auto_v1_";
+  public static final String AUTOCONFIG_TOKEN_PREFIX_V1 = "auto_v1_";
```
svix-jplatte: Unrelated to his PR I guess, but why is this public?

javascript/src/mockttp.test.ts
```diff
@@ -5,7 +5,6 @@ import * as mockttp from "mockttp";
 import { ApiException } from "./util";
 import { Ordering } from "./models/ordering";
 import type { ValidationError, HttpErrorOut } from "./HttpErrors";
-import { LIB_VERSION } from "./request";
```
svix-jplatte, line 8: Huh? Drive-by improvement?

kotlin/lib/src/main/kotlin/AutoConfig.kt
```diff
@@ -64,7 +64,7 @@ constructor(token: String, endpoint: EndpointIn) {
     }
 
     @Serializable
-    private data class AutoConfigTokenContentV1(
+    data class AutoConfigTokenContentV1(
```
svix-jplatte: Instead of making it public, does `internal` work?

php/src/AutoConfig.php
```diff
@@ -66,7 +66,7 @@ public function verify(string $payload, array $headers): mixed
     /**
      * @return array{app_id: string, endpoint_id: string, server_url: string, endpoint_secret: string, token_plaintext: string}
      */
-    private static function decodeToken(string $token): array
+    public static function decodeToken(string $token): array
```
svix-jplatte, line 84: Same situation here, doesn't seem great to make it public

python/svix/autoconfig.py
```diff
@@ -29,7 +29,7 @@ class _AutoConfigTokenContentV1(pydantic.BaseModel):
     token_plaintext: str = pydantic.Field(alias="tok")
 
 
-def _decode_autoconfig_token_v1(token: str) -> _AutoConfigTokenContentV1:
+def decode_autoconfig_token_v1(token: str) -> _AutoConfigTokenContentV1:
```
svix-jplatte: Move it to some internal module?

svix-jplatte: Or move the consumer class to this file. Probably easier.

csharp/Svix/AutoConfigConsumer.cs
svix-lucho: Underscore prefix is the recommended standard in C# (https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/identifier-names) - I think the other places are wrong in that case

javascript/src/mockttp.test.ts
svix-lucho: yup, I saw this was unused, dunno why the linter didn't pick it up

java/lib/src/main/java/com/svix/AutoConfig.java
svix-lucho: fixed

go/autoconfig.go
svix-lucho: yeah you''re right. Made it private

php/src/AutoConfig.php
svix-lucho: I couldn't have it be fully private while being able to reuse it in AutoConfigConsumer. According to my brief investigation and limited knowledge of PHP, the recommended way to do this is by adding `@internal` as doc comment since PHP has no proper access modifiers

python/svix/autoconfig.py
svix-lucho: done

svix-gabriel (approved):

PR #2392, opened 2026-06-10
cli: support using svix listen with an existing token

svix-lucho:
Allow using `svix listen --token <token> <local-url>` to connect to an existing Play session. Remove the config variable, since it makes no sense now

### DIFF
--- a/svix-cli/src/relay/mod.rs
+++ b/svix-cli/src/relay/mod.rs
@@ -256,7 +256,7 @@ pub async fn listen(
             eprintln!("Failed to connect to Webhook Relay: {e:#}");
             if e.downcast_ref::<TokenInUse>().is_some() && allow_token_rotation {
                 eprintln!("Generating a new token for this session.");
-                client.token = generate_token()?;
+                client.token = format!("c_{}", generate_token()?);
             }
         } else {
             eprintln!("Failed to connect to Webhook Relay");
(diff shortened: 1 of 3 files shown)


svix-cli/src/relay/mod.rs
```diff
@@ -254,12 +254,9 @@ pub async fn listen(
 
         if let Err(e) = client.connect(show_welcome_message).await {
             eprintln!("Failed to connect to Webhook Relay: {e:#}");
-            if e.downcast_ref::<TokenInUse>().is_some() {
+            if e.downcast_ref::<TokenInUse>().is_some() && allow_token_rotation {
                 eprintln!("Generating a new token for this session.");
-                client.token = {
-                    let relay_token = generate_token()?;
-                    format!("c_{relay_token}")
```
svix-jbrown, line 261: we don't need `c_` any more?

svix-lucho: well, actually, we can do this only after deploying https://github.com/svix/monorepo-private/pull/13299

I'll undo this for now, so I don't create interleaved dependencies

svix-cli/src/relay/mod.rs
```diff
@@ -254,12 +254,9 @@ pub async fn listen(
...
-                };
+                client.token = generate_token()?;
```
svix-jbrown: this one needs `c_` too, right?

svix-jbrown (approved):

PR #2399, opened 2026-06-18
Ignore RUSTSEC-2026-0173 until sea_orm drops the dep

svix-gabriel:
See https://rustsec.org/advisories/RUSTSEC-2026-0173

We have the offending dep indirectly through sea_orm, so we can't drop the dep until sea_orm does. :'(


svix-gabriel:
Ooops, this is a dupe

PR #2400, opened 2026-06-19
Python: AutoConfigConsumer fixes

svix-pluiz:
I was trying to set up AutoConfigConsumer for the `svix-hermes` plugin and hit this problem. I let Claude take a look while I was working on other stuff

---

Two bugs prevented the polling `AutoConfigConsumer` from working in the Python SDK:

1. **`subscribe()` crashed** with `TypeError: 'AutoConfigSinkType' object does not support item assignment`. The struct-enum wrap-validator did `data["config"] = {}` assuming a dict, but pydantic hands it a model instance when re-validating a nested struct-enum (exactly what `SubscribeIn(sink=…)` triggers). Guarded with `if not isinstance(data, dict): return handler(data)`.

2. **Tokens failed to decode** with "Incorrect padding". The server emits tokens as unpadded base64 (`STANDARD_NO_PAD`), but the decoder used strict `base64.b64decode`. Other SDKs work via padding-tolerant primitives (e.g. Node's `Buffer.from`). Re-add padding before decoding.

### DIFF
--- a/codegen/templates/python/types/struct_enum.py.jinja
+++ b/codegen/templates/python/types/struct_enum.py.jinja
@@ -28,6 +28,12 @@ class {{ type_name }}(BaseModel):
     def validate_model(
         cls, data: t.Any, handler: ModelWrapValidatorHandler[Self]
     ) -> Self:
+        {# `data` is a mapping on construction, but an already-built model
+           instance when pydantic re-validates this nested model (e.g. when it
+           is a field of another model). An instance is already correctly
+           typed, so skip the mapping-only content-field massaging below. -#}
+        if not isinstance(data, dict):
+            return handler(data)
         {# sadly this is needed because of pydantic -#}
         if "{{ type.content_field }}" not in data:
             data["{{ type.content_field }}"] = {}
--- a/codegen/templates/python/types/struct_enum.py.jinja
+++ b/codegen/templates/python/types/struct_enum.py.jinja
@@ -32,7 +32,7 @@ class {{ type_name }}(BaseModel):
            instance when pydantic re-validates this nested model (e.g. when it
            is a field of another model). An instance is already correctly
            typed, so skip the mapping-only content-field massaging below. -#}
-        if not isinstance(data, dict):
+        if isinstance(data, cls):
             return handler(data)
         {# sadly this is needed because of pydantic -#}
         if "{{ type.content_field }}" not in data:
--- a/python/svix/autoconfig.py
+++ b/python/svix/autoconfig.py
@@ -47,6 +47,11 @@ def _decode_autoconfig_token_v1(token: str) -> _AutoConfigTokenContentV1:
     if not token.startswith(_AUTOCONFIG_TOKEN_PREFIX_V1):
         raise AutoConfigError("invalid token")
     b64 = token[len(_AUTOCONFIG_TOKEN_PREFIX_V1) :]
+    # The server emits the token with standard-but-unpadded base64
+    # (Rust `STANDARD_NO_PAD`), so re-add the padding that Python's strict
+    # `b64decode` requires. (Other SDKs decode via padding-tolerant
+    # primitives, e.g. Node's `Buffer.from(b64, "base64")`.)
+    b64 += "=" * (-len(b64) % 4)
     try:
         decoded = base64.b64decode(b64)
     except Exception as exc:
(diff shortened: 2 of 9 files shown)


codegen/templates/python/types/struct_enum.py.jinja
```diff
@@ -28,6 +28,12 @@ class {{ type_name }}(BaseModel):
     def validate_model(
         cls, data: t.Any, handler: ModelWrapValidatorHandler[Self]
     ) -> Self:
+        {# `data` is a mapping on construction, but an already-built model
+           instance when pydantic re-validates this nested model (e.g. when it
+           is a field of another model). An instance is already correctly
+           typed, so skip the mapping-only content-field massaging below. -#}
+        if not isinstance(data, dict):
```
svix-jbrown: shouldn't this just be `if isinstance(data, Self)`?

python/tests/test_autoconfig.py
```diff
@@ -3,8 +3,24 @@
...
+def _make_token(**overrides) -> str:
+    payload = {
+        "aid": "app_1",
+        "eid": "ep_2",
+        "surl": "https://api.example.test",
+        "esec": "whsec_Zm9v",
+        "tok": "sk_test_xyz",
+    }
+    payload.update(overrides)
+    raw = json.dumps(payload).encode()
```
svix-jbrown: `encode()` without an argument is bad in python because it depends on the system locale; always specify what you want encoded as (typically `'utf-8'`).

python/svix/autoconfig.py
```diff
@@ -47,6 +47,11 @@ def _decode_autoconfig_token_v1(token: str) -> _AutoConfigTokenContentV1:
     if not token.startswith(_AUTOCONFIG_TOKEN_PREFIX_V1):
         raise AutoConfigError("invalid token")
     b64 = token[len(_AUTOCONFIG_TOKEN_PREFIX_V1) :]
+    # The server emits the token with standard-but-unpadded base64
+    # (Rust `STANDARD_NO_PAD`), so re-add the padding that Python's strict
+    # `b64decode` requires. (Other SDKs decode via padding-tolerant
+    # primitives, e.g. Node's `Buffer.from(b64, "base64")`.)
+    b64 += "=" * (-len(b64) % 4)
```
svix-jbrown: I don't think using negative values with modulus is very reasonable (off the top of my head, I couldn't tell you what it does in Python, since it's different in every language). I believe `-len(b64) % 4` is equivalent in python to `4 - (len(b64) % 4)`; if so, please use that

python/svix/autoconfig.py
```diff
@@ -47,6 +47,11 @@ def _decode_autoconfig_token_v1(token: str) -> _AutoConfigTokenContentV1:
     if not token.startswith(_AUTOCONFIG_TOKEN_PREFIX_V1):
         raise AutoConfigError("invalid token")
     b64 = token[len(_AUTOCONFIG_TOKEN_PREFIX_V1) :]
+    # The server emits the token with standard-but-unpadded base64
+    # (Rust `STANDARD_NO_PAD`), so re-add the padding that Python's strict
+    # `b64decode` requires. (Other SDKs decode via padding-tolerant
+    # primitives, e.g. Node's `Buffer.from(b64, "base64")`.)
```
svix-jbrown: this comment is unnecessarily verbose
```suggestion
    # Ensure b64 has padding, which is required by the python base64 module
```

python/tests/test_autoconfig.py
```diff
@@ -3,8 +3,24 @@
...
+    return f"{_AUTOCONFIG_TOKEN_PREFIX_V1}{base64.b64encode(raw).decode('ascii')}"
```
svix-jbrown: presumably this should emulate the server
```suggestion
    return f"{_AUTOCONFIG_TOKEN_PREFIX_V1}{base64.b64encode(raw).decode('ascii').replace('=', '')}"
```

codegen/templates/python/types/struct_enum.py.jinja
svix-pluiz: I tried using `Self` and ran into this problem...

```
  svix/models/auto_config_sink_type.py:21: in validate_model
      if isinstance(data, Self):
    def __instancecheck__(self, obj):
  >     raise TypeError(f"{self} cannot be used with isinstance()")
  E     TypeError: typing.Self cannot be used with isinstance()
```

`not isinstance` does the trick but it looks indeed strange, I changed to check the `cls` param as it's basically Self but valid in runtime

svix-lucho:
I missed this PR last week, but just came across the same issue with the base64 padding. I think we should just fix it in the backend https://github.com/svix/svix-webhooks-private/pull/6602

svix-jbrown (approved):

PR #2401, opened 2026-06-23
test(js): Cover idempotency key reuse across POST retries

bharathkumar39293:
## Summary

Adds regression coverage to ensure automatically generated idempotency keys are reused across POST retries.

## Motivation

The JavaScript SDK automatically generates an idempotency key for POST requests when one is not provided by the caller.

Retries currently reuse the same request headers, meaning the same auto-generated idempotency key is preserved across retry attempts. This behavior is important for safe retry semantics and helps avoid duplicate side effects when a request is retried after a transient failure.

This test documents and protects that existing behavior.

## Changes

- Added a test covering POST retries with automatic idempotency key generation.
- Verifies the same idempotency key is reused across all retry attempts.
- No production code changes.

## Test plan

```bash
cd javascript
npm test

### DIFF
--- a/javascript/src/mockttp.test.ts
+++ b/javascript/src/mockttp.test.ts
@@ -470,6 +470,27 @@ test("mockttp tests", async (t) => {
     assert(idempotencyKey.startsWith("auto_"));
   });
 
+  await t.test("POST retry reuses the same auto idempotency key", async () => {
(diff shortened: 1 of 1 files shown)


svix-jplatte (approved):
Thanks!

PR #2403, opened 2026-06-24
Rust SDK: Fix visibility of MessagePollerv2Consumer*Options

svix-gabriel:
While trying to use the AutoConfigConsumer `receive` and `commit` methods (https://docs.rs/svix/1.96.0/svix/autoconfig_consumer/struct.AutoConfigConsumer.html), I was unable to use anything other than `None` for `options` field, because `MessagePollerv2ConsumerPollOptions` and `MessagePollerv2ConsumerCommitOptions` are in the `api_internal` module.

My guess is this wasn't intentional, so I tried to fix the issue and add a lint so issues like this would get caught in the future. (There were some places where the privacy was clearly intention, which I explicitly whitelisted).

I just did this for Rust, not the other languages. I don't know if the same issue impacts other SDKs.

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -77,6 +77,10 @@ allowed = [
 [lints.rust]
 # below clippy allow rule is for a lint that's not on stable yet 😒
 unknown_lints = "allow"
+# Catch public APIs that reference types which cannot be named outside the
+# crate (e.g. `pub` types living in a private module). The default
+# `private_interfaces` lint misses these because the types are nominally `pub`.
+unnameable_types = "deny"
 
 [lints.clippy]
 all = { priority = -1, level = "warn" }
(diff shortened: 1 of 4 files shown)


rust/Cargo.toml
```diff
@@ -77,6 +77,10 @@ allowed = [
 [lints.rust]
 # below clippy allow rule is for a lint that's not on stable yet 😒
 unknown_lints = "allow"
+# Catch public APIs that reference types which cannot be named outside the
+# crate (e.g. `pub` types living in a private module). The default
+# `private_interfaces` lint misses these because the types are nominally `pub`.
+unnameable_types = "deny"
```
svix-jplatte:
```suggestion
unnameable_types = "warn"
```

CI will fail when warnings are emitted, no need to have it shown as an error locally.

svix-jplatte (approved):

Issue #1435, opened 2024-09-13
Improve error messages when server is misconfigured


tasn:
E.g. try not setting `jwt_secret` and see how confusing the errors are.

svix-jplatte:
With https://github.com/svix/svix-webhooks/pull/1520 merged, I think what's left here is:

- retry schedule
- database config

which produce bad error messages due to `#[serde(untagged)]`.

svix-jplatte:
Yes, nothing has happened with regard to this since the last update I posted in November.

PR #2418, opened 2026-06-27
bridge: Improve config parsing error messages

Soroushsrd:
## Summary
  - Replaced #[serde(untagged)] on EitherReceiver, SenderInputOpts, and ReceiverOutputOpts with custom Deserialize impls that route on the type field
  - No breaking changes made. YAML format is identical, only error messages improved

## Context
While i was trying to find the remaining items from #1435 (retry schedule and database config), I found that:
  - The retry schedule issue appears to already be resolved
  - The database config types in the server don't use #[serde(untagged)], so it's unclear what the original item referred to

but, the bridge config had the same class of #[serde(untagged)] problem, with an explicit FIXME in the test suite explaining poor error diagnostics. This PR addresses that.

  ### Before
  The error msg before the fix read as:
  `Failed to parse config: receivers: data did not match any variant of untagged enum EitherReceiver at line 3 column 7`

  ### After
  The new parses outputs an error (in case there is one) in the following format:
  `unknown receiver input type svix-events, expected one of: webhook, svix-webhook, svix-message-poller`

  ## Test plan
  - [X] Existing tests pass

  Relates to #1435


bridge/svix-bridge/src/config/mod.rs
```diff
@@ -157,14 +187,56 @@ pub struct WebhookSenderConfig {
...
+        let type_val = value
+            .as_mapping()
+            .and_then(|m| m.get("type"))
+            .and_then(serde_yaml::Value::as_str)
+            .ok_or_else(|| {
+                let valid = if cfg!(feature = "kafka") {
+                    "kafka, rabbitmq, redis, sqs, gcp-pubsub"
+                } else {
+                    "rabbitmq, redis, sqs, gcp-pubsub"
+                };
```
svix-jplatte: I think it makes more sense to always treat kafka as valid, but print a more specific error message for kafka if the feature isn't enabled.

bridge/svix-bridge/src/config/mod.rs
```diff
@@ -17,13 +17,43 @@ use tracing::Level;
...
+    where
+        D: Deserializer<'de>,
+    {
+        let value = serde_yaml::Value::deserialize(deserializer)?;
+        let input_type = value
+            .as_mapping()
+            .and_then(|m| m.get("input"))
+            .and_then(|v| v.as_mapping())
+            .and_then(|m| m.get("type"))
+            .and_then(serde_yaml::Value::as_str)
```
svix-jplatte:
```suggestion
            .and_then(|m| m.get("input")?.as_mapping()?.get("type"))
```

bridge/svix-bridge/src/config/mod.rs
```diff
@@ -157,14 +187,56 @@ pub struct WebhookSenderConfig {
...
+impl<'de> Deserialize<'de> for SenderInputOpts {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let value = serde_yaml::Value::deserialize(deserializer)?;
+        let type_val = value
+            .as_mapping()
+            .and_then(|m| m.get("type"))
+            .and_then(serde_yaml::Value::as_str)
```
svix-jplatte:
```suggestion
            .and_then(|m| m.get("type")?.as_str())
```

bridge/svix-bridge/src/config/mod.rs
```diff
@@ -214,16 +286,61 @@ pub struct WebhookReceiverConfig {
...
+impl<'de> Deserialize<'de> for ReceiverOutputOpts {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let value = serde_yaml::Value::deserialize(deserializer)?;
+        let value_type = value
+            .as_mapping()
+            .and_then(|m| m.get("type"))
+            .and_then(serde_yaml::Value::as_str)
```
svix-jplatte:
```suggestion
            .and_then(|m| m.get("type")?.as_str())
```

svix-jplatte (commented):
Thanks for the contribution, and sorry for the delay in reviewing. Left a couple comments below.

Soroushsrd:
Thank you for the comments. i'll resolve them today.

bridge/svix-bridge/src/config/mod.rs
Soroushsrd: done

bridge/svix-bridge/src/config/mod.rs
Soroushsrd: done

bridge/svix-bridge/src/config/mod.rs
Soroushsrd: i moved the kafka validation into a #[cfg(not(feature = "kafka"))] branch. i think it made the whole thing much simpler

svix-jplatte (approved):
Looks good, thanks!

PR #2419, opened 2026-06-28
test(js): Cover message.create withContent false payload behavior

bharathkumar39293:
## Summary

Adds regression coverage for `message.create(..., { withContent: false })`.

## Motivation

`message.create` requests `with_content=false` from the API and normally restores the caller's original payload locally.

When `withContent` is explicitly set to `false`, the SDK intentionally skips that restoration and preserves the payload returned by the server.

This test documents and protects that existing behavior.

## Changes

- Verifies the caller's payload is not restored when `withContent` is `false`.
- Verifies the request uses `with_content=false`.
- No production code changes.

## Test plan

```bash
cd javascript
npm test

Result:

- 50 passed
- 1 skipped


bharathkumar39293:
Hi @svix-jplatte, just a friendly follow-up on this when you have a chance. This PR only adds regression coverage for the existing withContent: false behavior (no production code changes). I'd appreciate a review whenever you have time. Thanks!

svix-jplatte (approved):
Thanks!

PR #2427, opened 2026-06-29
rust: make http 0.2 dependency optional

svix-jbrown:
This changes the Rust library to make the http 0.2 dependency optional.

I recognize that this is a (very minor) breaking API change because it's in one `pub` field of one struct, so someone might be depending on the exact type (as opposed to just, I dunno, that it's `Debug` or calling `.is_success()`). I think that's such a minor edge case that it's not worth worrying about compared to making every consumer compile two independent copies of the `http` crate.

### DIFF
--- a/.github/workflows/rust-lint.yml
+++ b/.github/workflows/rust-lint.yml
@@ -68,3 +68,12 @@ jobs:
 
       - name: Run tests
         run: cargo nextest run
+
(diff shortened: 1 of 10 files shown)


svix-jplatte:
The http crate isn't very big so the impact of having the old dependency should be minimal. Let's just wait for SDK v2?

PR #2429, opened 2026-06-30
bridge/kafka: Include partition in idempotency key

baxori:
## Motivation

The current Kafka sender idempotency key uses `group_id`, `topic`, and `offset`.

Kafka offsets are unique only within a partition, not within a topic. Because of that, two distinct Kafka records from different partitions can produce the same idempotency key.

This can cause Svix to deduplicate non-duplicate messages.

This changes the idempotency key format. Replayed records that cross an upgrade boundary may not deduplicate against keys produced by older bridge versions. This is limited to replay and reprocessing scenarios.

## Solution

- include Kafka `partition` in the sender idempotency key
- add an integration test that asserts the `idempotency-key` header sent by the Kafka bridge consumer

## Test plan

- `cargo fmt --package svix-bridge-plugin-kafka`
- `cargo test -p svix-bridge-plugin-kafka`

If you'd prefer a migration-oriented rollout, I'm happy to adjust the approach.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 * Server: Fix warning when opentelemetry is disabled
+* Bridge: Include the Kafka partition in consumer idempotency keys
 
 ## Version 1.98.0
 * CLI, Server, Bridge: Set OCI metadata on Docker images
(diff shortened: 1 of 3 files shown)


svix-jplatte:
Thanks for the PR! I'm currently pretty busy, but I'll get to reviewing this soon.

svix-jplatte (commented):
Looks good to me! Could you please add a changelog entry?

baxori:
Thanks! I've added a changelog entry.

svix-jplatte (approved):

PR #2438, opened 2026-07-08
Update cloud OpenAPI spec and regenerate SDKs

svix-jplatte:
Due to a change in how we generate the spec, a lot of things got reordered. This results in the declared order of some operations in the SDKs changing, but nothing else.


svix-pluiz (approved):
lgtm

Issue #2395, opened 2026-06-17
request-id support in Ruby gem


paul-palmer:
## Feature Request

Please add support in the Ruby gem to be able to associate a request-id with the requests to the server such that it will be included in the server's traces and log messages.

### Motivation

We wish to be able to support end-to-end traceability of requests in our system. The X-Request-Id header is the most commonly supported mechanism for passing a request identifier along with HTTP requests through all layers of the distributed system so that logs and traces that include that identifier from all services involved can be correlated. The SVIX server already supports the incorpration of a request identifier from an X-Request-Id header in its logs and traces.

### Proposal

This might be implemented with a new "request_id" option on all requests that would add a X-Request-Id header to the HTTP requests to the server much in the same way that the "idempotency-key" option is handled today.

### Alternatives

Everything else seems to be unnecessarily complicated.

bharathkumar39293:
@svix-jplatte  I traced the current Ruby SDK request path while looking into this.

The existing pattern for forwarding request-specific headers is `idempotency-key`, which generated API methods pass through to `SvixHttpClient#execute_request`.

One scope question before I open a PR:

The issue mentions support for **all requests**, but some generated Ruby methods, for example `get`, `delete`, and `update`, currently do not accept an `options = {}` argument.

That makes full `request_id` support look like a generator-level API shape change rather than just adding another header alongside `idempotency-key` in a few methods.

Would you prefer:

1. supporting `request_id` only on methods that already accept `options`, or
2. updating the Ruby generator/API so every request can carry a `request_id`?

Happy to work on whichever direction you prefer.

paul-palmer:
I think option 1 is  non-starter as even now many of the requests that need to support an idempotency-key via options do not support one. That is, the idempotency-key is supported inconsistently. For example, application.create supports it, but application.patch, application.update, and application.delete do not.

I think every request potentially needs to support a request_id to get a useful level of traceability.

PR #2441, opened 2026-07-09
feat(ruby): support request_id on API requests

bharathkumar39293:
## Summary

Adds Ruby SDK support for a caller-provided `request_id` option, sent as the `X-Request-Id` header for Svix API requests.

## Motivation

Svix server traces already support `x-request-id` and fall back to the SDK-generated `svix-req-id`. The Ruby SDK currently sends the internal `svix-req-id`, but callers cannot attach their own request id for end-to-end request correlation.

This addresses #2395.

## Changes

- Updates the Ruby API generator so generated API methods accept optional `options = {}`.
- Sends `options["request_id"]` as the `X-Request-Id` header on generated Ruby API requests.
- Uses `.compact` on generated header hashes so omitted `request_id` values are not sent.
- Updates Ruby extra API templates that define custom request helpers.
- Adds WebMock coverage for:
  - a method that already accepted options (`message.create`)
  - a method that previously did not accept options (`application.delete`)
  - omitted `request_id` not sending `X-Request-Id`

## Test plan

```bash
git diff --check

### DIFF
--- a/codegen/templates/ruby/api_resource.rb.jinja
+++ b/codegen/templates/ruby/api_resource.rb.jinja
@@ -31,16 +31,12 @@ module Svix
       {%- if op.request_body_schema_name is defined -%}
       {{ op.request_body_schema_name | to_snake_case }},
       {%- endif -%}
-      {# header/query params -#}
-      {%- if op | has_query_or_header_params -%}
+      {# request options -#}
       options = {}
-      {%- endif -%}
     {% endset -%}
 
     def {{ op.name | to_snake_case }}({{ func_args | strip_trailing_comma }})
-      {% if op | has_query_or_header_params -%}
       options = options.transform_keys(&:to_s)
-      {% endif -%}
       {% if op.response_body_schema_name is defined -%}res = {% endif -%}
       @client.execute_request(
         "{{ op.method | upper }}",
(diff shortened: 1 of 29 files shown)

### CODE REVIEW

File: codegen/templates/ruby/api_resource.rb.jinja [hunk 1]
```diff
@@ -52,13 +48,12 @@ module Svix
...
         {% endif -%}
-        {% if op.header_params | length >0 -%}
         headers:{
+          "X-Request-Id" => options["request_id"],
           {% for p in op.header_params -%}
             "{{ p.name }}" => options["{{ p.name }}"],
           {% endfor -%}
-        },
-        {% endif -%}
+        }.compact{% if op.request_body_schema_name is defined %},{% endif %}
```
svix-jplatte, line 56: Why add this extra conditional around the comma?

ruby/lib/svix/api/application.rb
```diff
@@ -21,7 +21,10 @@ def list(options = {})
           "limit" => options["limit"],
           "iterator" => options["iterator"],
           "order" => options["order"]
-        }
+        },
+        headers: {
+          "X-Request-Id" => options["request_id"]
```
svix-jplatte:
```suggestion
          "x-request-id" => options["request_id"]
```

svix-jplatte (commented):
Overall this makes sense to me. I'm not sure if I'll want to merge this before the SDK-v2 PR though. Ideally we'd do this for all SDKs at the same time (doesn't seem like any other SDK supports custom request-IDs either).

PR #2445, opened 2026-07-09
rust: Default-related cleanup

svix-jplatte:
A bunch of `Default` implementations that did not yield meaningful "default" values will be removed with SDK-v2. Drop usage of them ahead of the big SDK-v2 PR.

### DIFF
--- a/rust/tests/it/kitchen_sink.rs
+++ b/rust/tests/it/kitchen_sink.rs
@@ -30,13 +30,7 @@ async fn test_endpoint_crud() {
 
     let app = client
         .application()
-        .create(
-            ApplicationIn {
-                name: "app".to_string(),
-                ..Default::default()
-            },
-            None,
-        )
+        .create(ApplicationIn::new("app".to_owned()), None)
         .await
         .unwrap();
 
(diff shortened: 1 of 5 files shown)


rust/tests/it/kitchen_sink.rs
```diff
@@ -30,23 +30,23 @@ async fn test_endpoint_crud() {
...
+                name: "event.started".to_owned(),
+                description: "Something started".to_owned(),
+                archived: None,
+                deprecated: None,
+                #[expect(deprecated)]
+                feature_flag: None,
+                feature_flags: None,
+                group_name: None,
+                schemas: None,
             },
```
svix-jbrown, line 50: this is so incredibly verbose, though; are we also going to add `new` impls that take the most common options so users don't have to write code like this test?

or, better yet, builders for every In struct...

```rust
EventTypeIn::builder_with_name("event.ended").description("Something ended").build()
```

svix-jplatte: Yeah, I plan on having at least `EventTypeIn::new(name, description)`, the annoying thing is right now we have `EventTypeIn::new(description, name)`. I figured I'd do the easy thing here for now.

svix-jbrown (approved):

PR #2446, opened 2026-07-09
rust: Upgrade private dependencies

svix-jplatte:


rust/Cargo.toml
```diff
@@ -41,10 +41,10 @@ hyper = "1.7.0"
 hyper-rustls = { version = "0.27.7", optional = true }
 hyper-tls = { version = "0.6.0", optional = true }
 hyper-util = { version = "0.1.16", features = ["client", "client-legacy", "tokio"] }
-itertools = "0.14.0"
+itertools = "0.15.0"
 js_option = "0.1.1"
 percent-encoding = "2.3.1"
-rand = "0.9.0"
+rand = "0.10.2"
```
svix-jbrown, line 47: I thought we were staying on 0.9 for now because a bunch of dependencies are on 0.9 and we didn't want to bring in two copies of rand?

svix-jbrown (commented):
I'm not sure I'd call these "private" but whatevs

rust/Cargo.toml
svix-jplatte: Pretty sure rand 0.10 is sufficiently old to have made its way into all our dependency trees by now. It doesn't seem to be new in the two lockfiles in this repo.

svix-jplatte:
> I'm not sure I'd call these "private" but whatevs

What do you mean? Both of them are not part of our public API.

rust/Cargo.toml
svix-jbrown: It is new for svix-cli...?

svix-jbrown:
> > I'm not sure I'd call these "private" but whatevs
> 
> What do you mean? Both of them are not part of our public API.

I mean, I would say that our public client library is part of our public API. Doesn't really matter, though, just weird nomenclature.

rust/Cargo.toml
svix-jplatte: After `cargo update` there, quinn is upgraded to 0.10 and only tungstenite remains on rand 0.9 (upstream PR: https://github.com/snapview/tungstenite-rs/pull/529).

svix-jplatte:
By "our public API" I meant that of the client library. It is not a breaking change (in terms of Rust API compatibility) to upgrade these dependencies, because they are not a part of that. (unlike e.g. `http 0.2`, `js_option`)

svix-jbrown (approved):

PR #2449, opened 2026-07-10
 fix: replace Mutex<DnsState> with OnceCell in NonLocalDnsResolver

Rohan-Singla:
# Motivation

`NonLocalDnsResolver` used `Arc<Mutex<DnsState>>`  for lazy initialization of the DNS resolver. During the Init state, the mutex lock was held across an `.await` boundary specifically across the `new_resolver().await` call which performs I/O (reads system DNS config and builds the resolver).

This meant that under concurrent webhook dispatch, all DNS calls arriving before initialization completed would block waiting to acquire the mutex, serializing them behind a single I/O operation. On a high-throughput instance processing thousands of webhooks per second, this is a contention point at startup.

# Solution

Replaced `Arc<Mutex<DnsState>>` with `Arc<tokio::sync::OnceCell<TokioResolver>>`. The `get_or_try_init` method handles the lazy initialization correctly only one caller runs the initializer, concurrent callers wait on a lightweight internal notifier (not a mutex), and all subsequent calls after initialization is complete are lock-free atomic loads.

The `DnsState` enum and the Arc wrapper around `TokioResolver` are both removed as they are no longer needed. `new_resolver()` now returns `TokioResolver` directly. No behavior change the DNS resolution logic and IP filtering are untouched.

`tokio::sync::OnceCell` requires no new dependency as tokio is already pulled in with `features = ["full"]`.

### DIFF
--- a/server/svix-server/src/core/webhook_http_client.rs
+++ b/server/svix-server/src/core/webhook_http_client.rs
@@ -704,7 +704,10 @@ mod tests {
         net::{IpAddr, TcpListener},
         path::PathBuf,
         str::FromStr,
-        sync::Arc,
+        sync::{
+            Arc,
+            atomic::{AtomicUsize, Ordering},
+        },
     };
 
     use axum::{Router, routing};
@@ -712,7 +715,7 @@ mod tests {
     use http::{HeaderValue, Method, Version, header::AUTHORIZATION};
     use ipnet::IpNet;
 
-    use super::{RequestBuilder, WebhookClient, is_allowed};
+    use super::{NonLocalDnsResolver, RequestBuilder, WebhookClient, is_allowed, new_resolver};
     use crate::core::types::CasePreservingHeaderMap;
 
     #[test]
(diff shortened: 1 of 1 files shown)


server/svix-server/src/core/webhook_http_client.rs
```diff
@@ -592,24 +586,15 @@ impl Service<Name> for NonLocalDnsResolver {
...
-                    resolver
-                }
-
-                DnsState::Ready(resolver) => resolver.clone(),
-            };
-
-            drop(lock);
+            let resolver = this
+                .resolver
+                .get_or_try_init(|| async { new_resolver().await })
```
svix-jplatte:
```suggestion
                .get_or_try_init(new_resolver)
```

this should work as well, no?

svix-jplatte (commented):
Thanks for the PR! I'm wondering why we ever made `new_resolver` async given that it does no *async* I/O. Would you mind also changing that function to do all its actual work in a blocking tokio task?

server/svix-server/src/core/webhook_http_client.rs
```diff
@@ -942,4 +930,40 @@ mod tests {
         let whc_without_validation = WebhookClient::new(Some(whitelist), None, true, None);
         assert!(whc_without_validation.execute(request).await.is_ok());
     }
+
+    #[tokio::test]
+    async fn test_dns_resolver_initializes_once_under_concurrency() {
```
svix-jplatte: I'm not really convinced about the value of this test. Really all it tests is that tokio's `OnceCell` behaves as advertised, no?

Rohan-Singla: Ah makes sense..  will remove it doesn't makes sense . thanks jplatte !

server/svix-server/src/core/webhook_http_client.rs
Rohan-Singla: Done, thanks for pointing out the closure wrapper was redundant. 

new_resolver now runs its work inside tokio::task::spawn_blocking.

server/svix-server/src/core/webhook_http_client.rs
```diff
@@ -653,10 +635,15 @@ impl Iterator for SocketAddrs {
...
-    Ok(Arc::new(builder.build()?))
+async fn new_resolver() -> Result<TokioResolver, NetError> {
+    tokio::task::spawn_blocking(|| {
+        let mut builder = Resolver::builder_tokio()?;
+        builder.options_mut().ip_strategy =
+            hickory_resolver::config::LookupIpStrategy::Ipv4thenIpv6;
+        builder.build()
+    })
+    .await
+    .unwrap_or_else(|e| std::panic::resume_unwind(e.into_panic()))
```
svix-jplatte, line 646: Nice! I hadn't encountered this pattern before, but it seems more sensible than a plain `unwrap`.

svix-jplatte (approved):
Thanks again!

server/svix-server/src/core/webhook_http_client.rs
Rohan-Singla: Ahan cool!

Rohan-Singla:
> Thanks again!

Welcome looking forward to contribute more!

Issue #2379, opened 2026-06-02
JavaScript client library doesn't work under Bun


svix-jbrown:
The JS client library can't be imported from Bun, and the tests don't work in bun:

```
-------------------------------
1 | import { Svix } from "..";
2 | import type { SvixRequestContext } from "../request";
3 | 
4 | export class SvixInternal extends Svix {
                                      ^
ReferenceError: Cannot access 'Svix' before initialization.
      at svix-webhooks/javascript/src/api_internal/index.ts:4:35
```

The test suite doesn't work because bun requires that you have a `describe()` block and doesn't supported nested `test` calls.

Ideally at the end of this ticket, the library would work equally well under Node and Bun, and the test suite would run automatically under Bun.

ashuydv:
hey, i would like to contribute to svix and i've been contributing to codebase of cloudshot.io

svix-jplatte:
@ashuydv if you want to help, you could check if the latest pre-release ([2.0.0-rc.2](https://www.npmjs.com/package/svix/v/2.0.0-rc.2)) fixes this issue.

ashuydv:
sure

PR #2460, opened 2026-07-15
js: Bundle with tsdown, drop commonJS support

svix-lucho:
Use [tsdown](https://tsdown.dev/) for bundling, and drop explicit CJS support (CJS-style requires are supported in recent Node versions)

I had to make a small refactor in the generated code to avoid a circular dependency error that didn't pop up before due to how the files were used in runtime.

Also made some updates to the tsconfig to make it more strict

Removing support for older syntax significantly reduces the resulting bundle size 

Screenshot 2026-07-15 at 2 01 51 PM

### DIFF
--- a/.github/workflows/javascript-lint.yml
+++ b/.github/workflows/javascript-lint.yml
@@ -11,8 +11,8 @@ jobs:
 
       - uses: actions/setup-node@v4
         with:
-          node-version: 'latest'
-          registry-url: 'https://registry.npmjs.org'
+          node-version: "latest"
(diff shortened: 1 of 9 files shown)


svix-jbrown:
does this fix #2379?

svix-lucho:
> does this fix #2379?

I'll verify that, I think it probably will. 

I'm not going to merge this PR for now though, I think we'll include the change in Jonas' SDK v2 branch. Just posting it so he can cherry-pick it

svix-jplatte:
Merged into #2481, thanks Lucho!

PR #2471, opened 2026-07-17
docker: build CLI with alpine instead of debian

svix-jbrown:
This is an experiment, maybe people like it, maybe not.

It switches the Docker image for the CLI from debian to alpine. This drops the image size from 121MB to 22MB and the CVE count from 161 to 3.

I also embedded an SPDX for the CLI while I was at it.

The downside is probably somewhat reduced performance, but the CLI doesn't seem performance-sensitive to me, especially not when run through Docker where the startup time dwarfs everything else.

### DIFF
--- a/bridge/scripts/check-deps.sh
+++ b/bridge/scripts/check-deps.sh
@@ -1,8 +1,10 @@
-#!/bin/bash
+#!/bin/sh
+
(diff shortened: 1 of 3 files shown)


svix-pluiz (approved):

svix-jbrown:
closed in favor of #2473

PR #2472, opened 2026-07-17
server: switch to fork of sea-bae to remove proc-macro-error2 from dependencies

svix-jbrown:
`proc-macro-error2` has a very annoying warning printed on every build under rust 1.97+. This removes our last user of proc-macro-error2 by switchign sea-bae to use https://github.com/SeaQL/sea-bae/pull/3.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -15,9 +15,6 @@ unused-ignored-advisory = "allow"
 ignore = [
     # TODO: Update dependencies that use rsa crate
     "RUSTSEC-2023-0071",
-    # TODO: Wait for dependencies to upgrade off of proc-macro-error (validator)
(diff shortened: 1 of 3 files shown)


deny.toml
```diff
@@ -81,6 +78,7 @@ allow-git = [
     "https://github.com/svix/omniqueue-rs.git",
     "https://github.com/svix/validator.git",
```
svix-jplatte: Not used anymore, right?

svix-jplatte (dismissed):

deny.toml
svix-jbrown: indeed!

svix-jplatte (approved):

PR #2473, opened 2026-07-17
docker: build CLI with wolfi instead of debian

svix-jbrown:
This is an alternative to #2471 that @svix-jplatte  suggested; rather than using alpine, it uses Chainguard's `wolfi`.

The image is a bit bigger than alpine (28MB instead of 22MB; entirely due to the fact that glibc is 6MB larger than musl libc), but it should be a lot faster due to using "real" glibc.

Also, 0 CVEs. :-)

### DIFF
--- a/bridge/scripts/check-deps.sh
+++ b/bridge/scripts/check-deps.sh
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-# this has to run under both bash on debian and busybox sh on alpine
+# this has to run under both bash on debian and busybox sh on alpine/wolfi
 
 BINPATH="$1"
 
--- a/bridge/scripts/check-deps.sh
+++ b/bridge/scripts/check-deps.sh
@@ -1,8 +1,10 @@
-#!/bin/bash
+#!/bin/sh
+
+# this has to run under both bash on debian and busybox sh on alpine
 
 BINPATH="$1"
 
-if [[ -z "$BINPATH" ]]; then
+if [ -z "$BINPATH" ]; then
     echo >&2 "Usage $0 path/to/binary"
     exit 2
 fi
@@ -13,6 +15,6 @@ MISSING="$(ldd "$BINPATH" | grep 'not found')"
 # There'll be 1 linebreak for the empty case. Anything else and it means we
 # caught some unexpected output from ldd.
 if [ "1" != "$(echo "$MISSING" | wc -l)" ]; then
-    echo -e "Binary $BINPATH missing dependencies:\n$MISSING"
+    printf "Binary %s missing dependencies:\n%s" "$BINPATH" "$MISSING"
     exit 1
 fi
--- a/svix-cli/Dockerfile
+++ b/svix-cli/Dockerfile
@@ -3,22 +3,18 @@
 # NOTE: this should be built from the root of `svix-webhooks`
 # with `docker build -f svix-cli/Dockerfile .`# Base build
 
-FROM docker.io/rust:1.96-slim-trixie AS build
-
-SHELL ["/bin/bash", "-eux", "-o", "pipefail", "-c"]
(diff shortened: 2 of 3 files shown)


bridge/scripts/check-deps.sh
```diff
@@ -1,8 +1,10 @@
-#!/bin/bash
+#!/bin/sh
+
+# this has to run under both bash on debian and busybox sh on alpine
```
svix-jplatte: No alpine in this PR ;)

svix-cli/Dockerfile
```diff
@@ -3,22 +3,18 @@
...
-    apt-get install -y \
-        build-essential=12.* \
-        checkinstall=1.* \
-        curl=8.* \
-        libssl-dev=* \
-        pkg-config=1.8.* \
-        zlib1g-dev=1:* \
-        cmake=3.* \
-        --no-install-recommends
+FROM cgr.dev/chainguard/wolfi-base@sha256:02dab76bd852a70556b5b2002195c8a5fdab77d323c433bf6642aab080489795 AS build
```
svix-jplatte, line 6: Why a sha, and are we going to have dependabot take care of updating this?

svix-cli/Dockerfile
```diff
@@ -30,48 +26,42 @@ EOF
 
 WORKDIR /app/svix-cli
 
-# Hack to enable docker caching
 COPY rust /app/rust
-COPY svix-cli/Cargo.toml .
-COPY svix-cli/Cargo.lock .
+COPY svix-cli /app/svix-cli
```
svix-jplatte, line 30: No Docker caching anymore? I guess maybe it's overkill for CLI.

svix-cli/Dockerfile
```diff
@@ -30,48 +26,42 @@ EOF
...
-RUN touch src/main.rs
-RUN cargo build --release --frozen
+# ensure that the check-deps.sh script never ends up in the real image
+FROM alpine:3.24 AS checker
+RUN mkdir /scripts
+COPY bridge/scripts/check-deps.sh /scripts/check-deps.sh
 
 # Production
-FROM docker.io/debian:trixie-20260713-slim AS prod
+FROM cgr.dev/chainguard/wolfi-base@sha256:02dab76bd852a70556b5b2002195c8a5fdab77d323c433bf6642aab080489795 AS prod
```
svix-jplatte, line 42: I thought the main "thing" with Wolfi was that you'd build your container a bit differently with the package manager not being contained in the final container. Looking at docs, I guess they support both with this being the "you can use the exact same workflow as before, but get a slightly larger container" approach. Seems like a good start but I do find the alternative intriguing - maybe as a next step :)

svix-jplatte (dismissed):

svix-cli/Dockerfile
svix-jbrown: wolfi (like the other chainguard images) is a rolling distribution, so if we don't pin by sha we may get random breakages (particularly for pinned packages).

My plan is to use their tool ([digestabot](https://github.com/chainguard-dev/digestabot)) once Tom makes the appropriate permission changes to let it run.

svix-cli/Dockerfile
svix-jbrown: It was spending more time on caching infrastructure than on actually building, so I decided I'd just strip it out

svix-cli/Dockerfile
svix-jbrown: we could use apko to build a "distroless" runtime for production that doesn't have the package manager in it, but I don't know how to do that for a multiarch build without bringing in a ton of complexity, because then we'd need multiple phases of building artifacts and base images across multiple architectures and then merging them

svix-jplatte (approved):

svix-cli/Dockerfile
svix-jplatte: That's unfortunate that apko "conflicts" w/ multiarch like that. Probably not a big deal though.

PR #2476, opened 2026-07-21
Backwards-compatibly rename filterTypes => eventTypes

svix-jplatte:
Same as we did in the cloud server, except here we already update the OpenAPI spec to only contain the new name (since we have nothing depending on it – for Cloud the same thing is part of the SDK-v2 OpenAPI changes).

### DIFF
--- a/server/svix-server/src/v1/endpoints/endpoint/mod.rs
+++ b/server/svix-server/src/v1/endpoints/endpoint/mod.rs
@@ -56,8 +56,8 @@ use crate::{
 pub fn validate_event_types_ids(event_types_ids: &EventTypeNameSet) -> Result<(), ValidationError> {
     if event_types_ids.0.is_empty() {
         Err(validation_error(
-            Some("filterTypes"),
-            Some("filterTypes can't be empty, it must have at least one item."),
+            Some("eventTypes"),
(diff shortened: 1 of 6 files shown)


svix-pluiz (approved):

server/svix-server/src/v1/endpoints/endpoint/mod.rs
```diff
@@ -1041,22 +1047,22 @@ mod tests {
...
-             "url": URL_VALID,
-             "rateLimit": RATE_LIMIT_VALID,
-             "uid": ENDPOINT_ID_VALID,
-             "filterTypes": EVENT_TYPES_VALID,
-             "channels": EVENT_CHANNELS_VALID
+            "url": URL_VALID,
+            "rateLimit": RATE_LIMIT_VALID,
+            "uid": ENDPOINT_ID_VALID,
+            "filterTypes": EVENT_TYPES_VALID,
+            "channels": EVENT_CHANNELS_VALID
```
svix-pluiz, line 1065: There's no test for a valid input that has `eventTypes` as a property so we're only testing if the backwards compatible client works

svix-pluiz: not a big problem imo because it ~obviously works but we can add this in a new pr

Issue #1813, opened 2025-03-10
Missing Endpoint Events related classes


aherkowiak:
### Version

` >= java svix-1.60.0

### Platform
Various (MacOS, Linux x64 mostly)

### Description

We listen on event notifications that svix sends when endpoint is created/deleted/disabled/enabled/updated, after version upgrade to 1.60.0 corresponding classes don't exist anymore:

- com.svix.models.EndpointCreatedEvent.java
- com.svix.models.EndpointCreatedEventData.java
- com.svix.models.EndpointDeletedEvent.java
- com.svix.models.EndpointDeletedEventData.java
- com.svix.models.EndpointDisabledEvent.java
- com.svix.models.EndpointDisabledEventData.java
- com.svix.models.EndpointEnabledEvent.java
- com.svix.models.EndpointEnabledEventData.java
- com.svix.models.EndpointUpdatedEvent.java
- com.svix.models.EndpointUpdatedEventData.java

It's an important part of our solution, I can't find release notes about changes in this area or replacement I could use. Would that be possible to bring it back?

svix-mman:
@aherkowiak I am discussing this internally, I will get back to you later today!

aherkowiak:
@svix-mman  thank you for information and for you support 👍

svix-mman:
@aherkowiak v.1.61.4 was released with a fix for this bug

svix-jplatte:
@aherkowiak-trackunit hey, I'm looking to remove these models again as part of v2.0 of the SDKs, for a couple of different reasons:

- one of the models (for background-task.finished) is pretty crappy (we can improve that of course, but up until I found this issue I thought nobody was using these models anyways)
- this seems like bloat for almost all users
- we don't provide a public stand-alone API for (de)serialization of schemas separately from API calls for most SDKs
  - for Java in particular, we never documented which libraries serialization annotations we add to our models (and it has changed once already IIRC)

I think a good way forward here would be for us to provide a _separate_ package for Java that only contains the op-webhook model definitions, using jackson annotations just like the SDK *but actually committing to this* (or providing public parse / stringify methods). Would this work for you?

PR #2478, opened 2026-07-21
SDK v2: Update openapi-codegen to 2026-07-04

svix-jplatte:
… to remove webhook models from the SDKs. They were never referenced, and of unclear value. This should also reduce times for SDKs that have a build (compile) step.

Part 1/x of the SDK-v2 changes. Splitting things up for easier review, and to figure out what's wrong with the Python tests.


svix-jplatte:
Looks like there is actually at least one user with a use case for the op-webhook models: https://github.com/svix/svix-webhooks/issues/1813
I've left a comment on that issue about how I think we could proceed. Either way I don't think that needs to hold up v2.0(-rc.1) of the SDKs. We can always add the models back in v2.1.

svix-gabriel (approved):

svix-jbrown (approved):

PR #2481, opened 2026-07-21
SDK v2: Language-specific changes

### LINKED ISSUE #1483: Adding svix for webhook validation adds almost 1 MB to the JS bundle

### LINKED ISSUE #2396: Upgrade OkHttp to 5.x
### Motivation

OkHttp 4.x is not native ready, 5.x is

### Proposal

Please upgrade OkHttp deps to 5.x

### Alternatives

native-stubs, annoying to maintain

svix-jplatte:
Part 3/x of the SDK-v2 changes.

- Closes #1483
- Closes #2396


svix-jplatte:
Note to self: There is actually one thing missing here, removing JSON parsing from Python SDK's `Webhook.verify`. Blocked on standard-webhooks Python SDK release. Maybe we'll just ship rc.1 of the Svix SDKs without it.

svix-gabriel (dismissed):

java/build.gradle
```diff
@@ -36,7 +36,7 @@ dependencies {
     implementation "com.fasterxml.jackson.datatype:jackson-datatype-jsr310:2.21.4"
     implementation "com.fasterxml.jackson.core:jackson-annotations:2.21"
     implementation "com.fasterxml.jackson.core:jackson-databind:2.21.4"
-    implementation "com.squareup.okhttp3:okhttp:4.12.0"
+    implementation "com.squareup.okhttp3:okhttp:5.4.0"
     implementation "com.google.code.findbugs:jsr305:3.0.2"
```
svix-jbrown, line 41: the minimum-java-version bump didn't make it into the SDKv2 list, hm?

svix-jbrown (dismissed):

java/build.gradle
svix-jplatte: It's not very long ago that we had somebody ask about that webhook.verify signature thing which was preventing us from supporting Java 8.

svix-jbrown: I know, but a man can dream

svix-jbrown (approved):

PR #2487, opened 2026-07-21
Try to fix CI....

svix-jplatte:
Having a hard time understanding the CI failure in https://github.com/svix/svix-webhooks/pull/2486, and can't reproduce locally (something about the docker setup doesn't work for me and the error messages are no help in finding out what it might be).


svix-jplatte:
Issue has been found out, will be fixed separately.

PR #2491, opened 2026-07-22
bridge/kafka: Add transformation envelope

baxori:
## Motivation

Kafka JSON transformations currently receive only the record payload. This prevents transformations from routing or enriching messages using Kafka headers, keys, or record metadata.

## Solution

Add an opt-in `kafka_transformation_input: envelope` mode for Kafka JSON transformations.

When enabled, transformations receive the record headers, key, topic, partition, offset, and parsed payload. The default remains `payload`, preserving the existing transformation input contract.

Headers are represented as an ordered array to preserve duplicate keys. UTF-8 values are passed as strings; binary values are represented as base64.

Added integration coverage for envelope metadata and a configuration test that verifies the default remains `Payload`.

### DIFF
--- a/bridge/svix-bridge-plugin-kafka/src/input.rs
+++ b/bridge/svix-bridge-plugin-kafka/src/input.rs
@@ -3,10 +3,12 @@ use std::{
     time::{Duration, Instant},
 };
 
+use base64::{Engine as _, engine::general_purpose::STANDARD};
 use rdkafka::{
     Message as _,
     consumer::{CommitMode, Consumer as _},
     error::KafkaError,
+    message::Headers as _,
 };
 use svix_bridge_types::{
     CreateMessageRequest, JsObject, SenderInput, SenderOutputOpts, TransformationConfig,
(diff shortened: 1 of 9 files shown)


bridge/svix-bridge-plugin-kafka/src/input.rs
```diff
@@ -46,6 +51,41 @@ impl KafkaConsumer {
...
     }
 
+    fn kafka_bytes_as_json(value: Option<&[u8]>) -> serde_json::Value {
+        value.map_or(serde_json::Value::Null, |value| {
+            str::from_utf8(value).map_or_else(
+                |_| serde_json::json!({ "base64": STANDARD.encode(value) }),
+                |value| serde_json::Value::String(value.to_owned()),
+            )
+        })
+    }
```
svix-jbrown, line 64: I'm a little skeptical of doing this; I think having a key that's either a string or a JSON blob containing a base64-encoded binary blob is kind of weird.

For keys, I'd probably be inclined to just do `String::from_utf8_lossy` and expect that anyone doing non-unicode-encoded headers in Kafka can fix their application.

For values, I'd say that we should probably just base64 every time for consistency.

In general, the more consistent the better

baxori: Agreed, updated accordingly.

svix-jplatte:
Note for @svix-jbrown: Please let's not merge this before https://github.com/svix/svix-webhooks/pull/2499, it shouldn't conflict with anything from that PR stack but the new stacked PR feature still seems a bit brittle so I'd like to be extra careful ^^

edit: Nevermind, looks like we have another thing we should merge before the SDK v2 changes anyways...

svix-jbrown (approved):

PR #2494, opened 2026-07-22
bridge/kafka: Add configurable auto.offset.reset

baxori:
## Motivation

Kafka consumers currently rely on librdkafka's default `auto.offset.reset` behavior. Bridge users cannot configure a consumer to replay records when creating a new consumer group or when stored offsets are no longer valid.

## Solution

Add an optional `kafka_auto_offset_reset` setting for Kafka inputs.

The setting defaults to `latest`, preserving existing behavior. Set it to `earliest` to consume existing records for a new consumer group.

Added configuration tests for the default and an explicit `earliest` value, plus Kafka integration coverage for consuming a record published before the consumer starts.

### DIFF
--- a/bridge/svix-bridge-plugin-kafka/src/config.rs
+++ b/bridge/svix-bridge-plugin-kafka/src/config.rs
@@ -2,6 +2,7 @@ use rdkafka::{
     ClientConfig, consumer::StreamConsumer, error::KafkaResult, producer::FutureProducer,
 };
 use serde::Deserialize;
+use strum::IntoStaticStr;
 use svix_bridge_types::{ReceiverOutput, SenderInput, SenderOutputOpts, TransformationConfig};
 
 use crate::{KafkaProducer, Result, input::KafkaConsumer};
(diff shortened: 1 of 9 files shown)


bridge/svix-bridge-plugin-kafka/src/config.rs
```diff
@@ -27,6 +27,13 @@ pub enum KafkaInputOpts {
         #[serde(rename = "kafka_topic")]
         topic: String,
 
+        /// The value for 'auto.offset.reset' in the kafka config.
+        #[serde(
+            rename = "kafka_auto_offset_reset",
+            default = "default_auto_offset_reset"
+        )]
+        auto_offset_reset: String,
```
svix-jbrown: I think we're all big fans of making unrepresentable states impossible, so what do you think about something like

```rust
#[derive(Clone, Deserialize, Default, strum::IntoStaticStr)]
#[serde(rename_all="snake_case")]
#[strum(serialize_all="snake_case")]
enum KafkaAutoOffsetReset {
  #[default]
  Latest,
  Earliest
}
```

and then changing the type of this value from String to `KafkaAutoOffsetReset`

svix-jbrown (dismissed):

svix-jbrown:
Please rebase on top of the other change

svix-jplatte (approved):
Thanks for your contribution!

PR #2502, opened 2026-07-24
bridge/kafka: Add idempotency namespace

baxori:
## Motivation

The current Kafka sender idempotency key uses `group_id`, `topic`, `partition`, and `offset`.

`group_id` is scoped to a Kafka cluster and is not globally unique across independent Kafka sources. When multiple Kafka clusters deliver to the same Svix instance, distinct records can produce the same idempotency key.

This can cause Svix to deduplicate non-duplicate messages.

When configured, the idempotency namespace changes the key format. Replayed records that cross an upgrade boundary may not deduplicate against keys produced without the namespace. This is limited to replay and reprocessing scenarios.

## Solution

- add an optional `kafka_idempotency_namespace` Kafka input setting
- include the namespace in the sender idempotency key when configured
- preserve the existing idempotency key format when the setting is omitted
- add config and integration coverage for namespaced keys

The namespace represents a stable source identity and can identify an environment, Kafka cluster, region, or backfill source.

## Test plan

- `cargo fmt --check`
- `cargo test -p svix-bridge-plugin-kafka`

### DIFF
--- a/bridge/ChangeLog.md
+++ b/bridge/ChangeLog.md
@@ -3,6 +3,7 @@
 ## Unreleased
 * Add an opt-in Kafka record envelope for JSON transformations
 * Allow configuring Kafka consumer `auto.offset.reset`
+* Add an optional Kafka idempotency namespace for independent sources
 
 ## Version 1.99.0
 * Include the Kafka partition in consumer idempotency keys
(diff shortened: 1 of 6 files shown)


svix-jbrown (approved):

svix-jplatte:
Thanks for the contribution!

PR #2510, opened 2026-07-28
rust: Make Configuration private

svix-jplatte:
It's only public for historical reasons.
Our own use of svix.cfg() only uses it to get the access token, which is now covered by svix.token().

This allows us more refactors down the line.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -6,9 +6,10 @@ The Svix Server changelog has moved to [server/ChangeLog.md](./server/ChangeLog.
 The Svix Bridge changelog has moved to [bridge/ChangeLog.md](./bridge/ChangeLog.md).
 
 ## Unreleased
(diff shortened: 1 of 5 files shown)


svix-pluiz:
> This allows us more refactors down the line.

What kind of refactor we cannot do as a consequence of this being public?

svix-jplatte:
For example: Representing the retry schedule in a more sensible way where we don't have two different fields, one of which is only used if the other is `None`.

svix-pluiz:
> For example: Representing the retry schedule in a more sensible way where we don't have two different fields, one of which is only used if the other is None.

Oh, I see. It was not uncommon in the JS Land to have exposed-but-not-expected-to-be-used properties due to the previous lack of private fields... So this kind of property would be named like `__INTERNAL_DO_NOT_USE_FOO` 😁

svix-pluiz (approved):

PR #2515, opened 2026-07-29
php: Add support for guzzle 8

VincentLanglet:
## Motivation

Hi @svix-jbrown @svix-pluiz, guzzle released the v8 recently so I tried to add the compatibility.

### DIFF
--- a/.github/workflows/php-ci.yml
+++ b/.github/workflows/php-ci.yml
@@ -8,6 +8,7 @@ jobs:
       fail-fast: false
       matrix:
         php-versions: ['8.3', '8.4', '8.5']
+        dependencies: ['highest', 'lowest']
     runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v6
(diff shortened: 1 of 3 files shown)


svix-jplatte:
Hey! Any reasons why we would support both versions at once, rather than just upgrading to version 8?

VincentLanglet:
> Hey! Any reasons why we would support both versions at once, rather than just upgrading to version 8?

Sure.

Guzzle is a library used by looooooot of other libraries. And guzzle 8 is currently not supported by a lot of them.

If you restrict your package to guzzle 8 it means lot of users will have incompatibility between guzzle versions for a long time and will be stuck with guzzle 7 and an older version of svix lib.
Since the support of both 7 and 8 does not seems hard it will offer more flexibility to developers.

svix-jbrown (commented):
We should probably have a test matrix that builds with both guzzle 7 and guzzle 8 if we're going to support both. Do you feel comfortable making that kind of change to the github actions workflow for PHP?

VincentLanglet:
> We should probably have a test matrix that builds with both guzzle 7 and guzzle 8 if we're going to support both. Do you feel comfortable making that kind of change to the github actions workflow for PHP?

Easiest way to do this is to run highest and lowest build.
This also ensure that lowest supported version are really supported.

I updated the workflow
See
- guzzle 8: https://github.com/svix/svix-webhooks/actions/runs/30482777490/job/90680847906?pr=2515
- guzzle 7: https://github.com/svix/svix-webhooks/actions/runs/30482777490/job/90680847812?pr=2515

svix-jplatte (approved):
Ordinarily I'd be a bit worried about ever more combinations of CI runs, but it looks like these are super fast 👍 

Thanks for this contribution!

VincentLanglet:
> Ordinarily I'd be a bit worried about ever more combinations of CI runs, but it looks like these are super fast 👍 
> 
> 
> 
> Thanks for this contribution!

Yeah I tried like this but if preferred it is possible to have only one run dedicated to lowest versions

PR #2520, opened 2026-07-30
ci: Update server-release workflow to get version from Cargo

svix-jplatte:
Now that we no longer use the same version for SDKs and server, bridge. This should also allow us to create the server release tag _after_ successful publish to ghcr, Dockerhub.

### DIFF
--- a/.github/workflows/server-release.yml
+++ b/.github/workflows/server-release.yml
@@ -46,9 +46,12 @@ jobs:
           filter: tree:0
       - id: get-version-number
         run: |
-          echo "tag=$(git describe --tags --abbrev=0)" >> "$GITHUB_OUTPUT"
+          version=$(cargo metadata --format-version 1 \
+            | jq -r '.packages[] | select(.name == "svix-server") | .version')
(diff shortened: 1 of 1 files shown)


.github/workflows/server-release.yml
```diff
@@ -46,9 +46,12 @@ jobs:
           filter: tree:0
       - id: get-version-number
         run: |
-          echo "tag=$(git describe --tags --abbrev=0)" >> "$GITHUB_OUTPUT"
+          version=$(cargo metadata --format-version 1 \
+            | jq -r '.packages[] | select(.name == "svix-server") | .version')
```
svix-jbrown, line 50: nit: if this fails for some reason, you might get nonsense (like two rows separated by a newline, or an empty string)

I would maybe add some sanity checking here
```suggestion
          version=$(cargo metadata --format-version 1 \
            | jq -r '.packages[] | select(.name == "svix-server") | .version')
          if ! [[ "$version" =~ ^[0-9.]+$ ]] ; then
             echo >&2 "Invalid version $version"
             exit 1
          fi
```

svix-jbrown (approved):

PR #2521, opened 2026-07-30
server: Send log events to the configured OpenTelemetry collector

clinejj:
## Motivation

`opentelemetry_address` currently only enables OTLP export for traces and metrics — log events (`tracing::info!`, `warn!`, etc.) only ever reach stdout, with WARN/ERROR additionally forwarded to Sentry. This means anyone routing observability data through an OTel collector (e.g. to Loki, DataDog, or another OTLP-compatible backend) gets traces and metrics but has to fall back to scraping stdout/container logs separately to get logs, which is inconsistent and easy to miss in an otherwise fully OTel-based setup.

We're deploying the OSS server on render, which unfortunately makes it difficult to ship logs from stdout to somewhere else (other than DataDog).

## Solution

Adds a new `opentelemetry_logs_enabled` config flag (default `false`). When set to `true` *and* `opentelemetry_address` is configured, `setup_tracing` builds an OTLP `LogExporter` + `SdkLoggerProvider` (reusing the same collector address and `opentelemetry_service_name` as the existing trace/metric exporters) and registers an `OpenTelemetryTracingBridge` (`opentelemetry-appender-tracing`) layer alongside the existing stdout `fmt` layer, Sentry layer, and trace layer. stdout logging is unaffected either way — this only adds an additional export destination.

Log export is opt-in rather than automatically following `opentelemetry_address`, per review feedback — enabling OTLP traces/metrics no longer implicitly turns on OTLP log export too.

Implementation notes:
- Uses `opentelemetry_sdk::logs::BatchLogProcessor`, which runs its own dedicated background thread rather than requiring an async-runtime feature flag
- The logger provider is shut down gracefully alongside the tracer provider on process exit.
- Verified end-to-end locally against a real `otel/opentelemetry-collector-contrib` instance with a `debug` exporter — confirmed log records arrive with correct severity, body, and `service.name` resource attribute.
- Updated `OpenTelemetry.md`, `README.md`, and `config.default.toml` docs to reflect the new flag and that `opentelemetry_address` now covers logs (when enabled) in addition to traces/metrics.

Would also be open to allowing a separate log collector address/service name from the traces/metrics config, wouldn't be that hard to add in.


svix-james:
Hi, thanks for the PR!

I think it would be best if log export can be selectively enabled rather than be automatically on if an opentelemetry endpoint is defined.

clinejj:
> Hi, thanks for the PR!
> 
> I think it would be best if log export can be selectively enabled rather than be automatically on if an opentelemetry endpoint is defined.

Sounds good! Added a new `opentelemetry_logs_enabled` flag that controls whether the logs are sent to the configured otel endpoint

svix-james:
Hi, apologies for the delay in reviewing further. The code looks great, but I want to spend some additional time testing this, as I haven't previously used the logging features provided by the otel crates.

svix-james:
Ok, tested locally a bit and things seem good here. I have one last concern I haven't looked into deeply, which is the possibility for a log feedback loop, as suggested by this issue: https://github.com/open-telemetry/opentelemetry-rust/issues/2877. I wasn't able to create one in my very brief testing, but should we consider disabling events from certain crates in order to prevent this possibility?

clinejj:
@svix-james good catch, that is a concern and was able to reproduce locally using `RUST_LOG=debug`. just added a [patch to filter the events](https://github.com/clinejj/svix-webhooks/blob/4d6ddf9ec4bfeddcc629487ee6ae0591580f4b10/server/svix-server/src/lib.rs#L287-L305), but details on the issue here (with help from claude):

____
**Why it happens**

Exporting a log record emits log records of its own (OTel SDK internals, plus the `hyper`/`h2`/`tonic`/`tower` transport stack underneath). Those get queued and exported on the next tick, which emits more. The SDK does guard against this — `SdkLogger::emit` checks `is_current_telemetry_suppressed()` and `BatchLogProcessor` enters a suppressed scope on its worker thread — but the suppression rides on the OpenTelemetry `Context`, which hyper/tonic don't propagate across the thread boundaries used for the actual transport. That's exactly the gap in #2877.

Two things that make it less alarming than it first looks:

- **It's rate-limited, not exponential.** `BatchLogProcessor`'s 1s `scheduled_delay` caps it at a constant trickle rather than a spiral. It just never goes quiet.
- **It's unreachable at default config.** Our `EnvFilter` only has target-specific directives (`svix_server`, `tower_http`, optionally `sqlx`), and unlisted targets are fully off — I verified an explicit `error!(target: "opentelemetry_sdk", …)` doesn't pass. So OTel internals never reach the bridge unless someone sets a broad `RUST_LOG`. That's probably why it didn't show up in your testing — but a broad `RUST_LOG` is exactly what you'd set while debugging, which is the worst time for a log flood.

**How I confirmed it**

Ran a probe against a real `otel/opentelemetry-collector-contrib` with a debug exporter: emit **one** application log line, then sit completely idle for 20s, with `RUST_LOG=debug` and `opentelemetry_logs_enabled = true`.

- **Before:** collector received **207 records** at a steady ~10/sec that never stopped, from a single app log line and zero traffic. (With the collector unreachable instead, same picture — a flat ~9 lines/sec of failed-export churn.)
- **After the fix:** **2 records** — precisely the two real application logs, nothing else.

**The fix**

A per-layer target filter on the OTLP logs bridge only, excluding `opentelemetry*`, `hyper*`, `h2`, `tonic`, and `tower`. Deliberately does *not* match `tower_http`, which is our own HTTP layer. stdout is untouched — those events are still fully visible for debugging, they're just not fed back into the exporter.

One tradeoff worth naming: `hyper`/`h2` are also used by outbound webhook dispatch, so those events won't reach the collector either — they can't be distinguished from export-path events by target alone. They remain on stdout.

I've added a `TODO` to drop the whole workaround once #2877 is fixed and we've upgraded to a release containing it — the SDK's suppression mechanism is the right fix, it just doesn't survive the transport's thread hops today.

svix-james:
Hi, sorry again for delays reviewing. In testing this a bit further, I notice that none of the span info we include in console logs has made it into the Grafana stack I'm using for local testing. I only see the raw message, e.g., "Webhook success.", and links to the relevant trace (which is great!), but none of the additional span info has made it into the logs, which limits utility when searching.

e.g., JSON output of a typical event.
```
{
  "timestamp": "2026-09-09T19:26:52.491760Z",
  "level": "INFO",
  "message": "Webhook success.",
  "target": "svix_server::worker",
  "span": {
    "response_code": 204,
    "name": "handle_successful_dispatch"
  },
  "spans": [
    {
      "app_id": "app_3HjvcaCQmtTsnUpxGdsxjhsdRjc",
      "msg_id": "msg_3J6WUKFLe2Hs1pyJYIjvwXOnIOY",
      "org_id": "org_23rb8YdGqMT0qIzpgGwdXfHirMu",
      "task_type": "MessageBatch",
      "name": "process_queue_task"
    },
    {
      "endp_id": "ep_3HjvhgvuzwqGSfAX7EceWHmz7fD",
      "name": "dispatch_message_task"
    },
    {
      "response_code": 204,
      "name": "handle_successful_dispatch"
    }
  ]
}
```

clinejj:
@svix-james no worries! pushed an update that includes the span attributes, using `experimental_span_attributes` on `opentelemetry-appender-tracing`. The example you have now exports as:

```
Body: Str(Webhook success.)
Attributes:
     -> msg_id: Str(msg_…)
     -> app_id: Str(app_…)
     -> org_id: Str(org_…)
     -> task_type: Str(MessageBatch)
     -> endp_id: Str(ep_…)
     -> response_code: Int(204)
```

Worth noting for completeness:
- Span _names_ aren't included. This is dependent on https://github.com/open-telemetry/opentelemetry-rust/pull/3466, and also only adds the leaf span name not the full array of spans. The names are still recording on the span itself, so you can look up via ID
- That being said, if the span isn't included due to sampling that gets lost. The sampled flag still informs if the span exists, but you lose the ability to look up logs by name if you need

server/svix-server/src/lib.rs
```diff
@@ -349,6 +372,44 @@ pub fn setup_tracing(
 
     let (otel_layer, otel_tracer_provider) = mapped.unzip();
 
+    let mapped_logs = cfg
+        .opentelemetry_address
+        .as_ref()
+        .filter(|_| cfg.opentelemetry_logs_enabled)
```
svix-james, line 392: One more nit: It's probably worth logging a warning if `opentelemetry_logs_enabled` is true but `opentelemetry_address` is `None`.

clinejj: Updated at https://github.com/svix/svix-webhooks/pull/2521/commits/b2448e749c15115ba431b77856b713eb31a14c7a

svix-james (approved):
Approved. Thanks for the contribution!

server/svix-server/src/lib.rs
clinejj: There were a couple options (using `eprintln!` directly, or logging through `tracing::dispatcher` since we have access in the method), but chose to follow the existing pattern for the redis timeout warning

PR #2523, opened 2026-07-31
fix(ruby): serialize false patch values

clinejj:
## Motivation

Ruby patch models used truthiness when serializing non-nullable boolean fields, so explicit `false` values were omitted. This prevented re-enabling disabled endpoints and toggling other boolean patch fields back off.

## Solution

- Serialize non-nullable booleans unless their value is `nil`, preserving explicit `false` values.
- Update the Ruby code-generation template and generated models for endpoint, endpoint transformation, ingest endpoint transformation, event type, and stream event type patches.
- Add regression coverage for `false` patch values and confirm an endpoint description can be cleared with an empty string.

### DIFF
--- a/codegen/templates/ruby/types/macros.rb.jinja
+++ b/codegen/templates/ruby/types/macros.rb.jinja
@@ -31,6 +31,8 @@
     {% endif -%}
     {% if type.name is endingwith "Patch" and field.nullable -%}
     out['{{ field.name }}'] = {{ val_expr }} if @__{{ field.name | to_snake_case }}_is_defined
+    {% elif type.name is endingwith "Patch" and field.type.is_bool() -%}
+    out['{{ field.name }}'] = {{ val_expr }} unless @{{ field.name | to_snake_case }}.nil?
     {% else -%}
     out['{{ field.name }}'] = {{ val_expr }} if @{{ field.name | to_snake_case }}
     {% endif -%}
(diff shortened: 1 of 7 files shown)


svix-jbrown:
that's an awfully silly bug, thanks for finding it!

svix-jbrown (approved):

codegen/templates/ruby/types/macros.rb.jinja
```diff
@@ -31,6 +31,8 @@
     {% endif -%}
     {% if type.name is endingwith "Patch" and field.nullable -%}
     out['{{ field.name }}'] = {{ val_expr }} if @__{{ field.name | to_snake_case }}_is_defined
+    {% elif type.name is endingwith "Patch" and field.type.is_bool() -%}
+    out['{{ field.name }}'] = {{ val_expr }} unless @{{ field.name | to_snake_case }}.nil?
     {% else -%}
     out['{{ field.name }}'] = {{ val_expr }} if @{{ field.name | to_snake_case }}
```
svix-jplatte, line 37: Why the separate condition? Can't we use `unless @{{ field.name | to_snake_case }}.nil?` regardless of whether it's a `Patch` type?

clinejj: Sure! I did this mostly from being conservative for `Patch` endpoints, but can open a PR to remove this condition

clinejj: https://github.com/svix/svix-webhooks/pull/2542

Issue #2524, opened 2026-08-01
svix-bridge fails to build with --no-default-features


Soroushsrd:
### Version
```
svix-bridge v1.99.0
svix-bridge-plugin-kafka v1.99.0
svix-bridge-plugin-queue v1.99.0
svix-bridge-types v1.99.0
svix v2.0.0-rc.2
svix-ksuid v0.10.0
```

### Platform
```
Linux archlinux 7.1.5-arch1-1 #1 SMP PREEMPT_DYNAMIC Sun, 26 Jul 2026 00:49:06 +0000 x86_64 GNU/Linux
```

### Description

`svix-bridge` fails to compile with `--no-default-features` because the two feature-gated definitions of `get_allocator_stats` in `allocator.rs` have divergent signatures

I tried this:

```
cargo build -p svix-bridge --no-default-features
```

It should have compiled successfully but it failed with this msg:

```
error[E0308]: mismatched types
   --> svix-bridge/src/main.rs:291:88
    |
291 | ...ted, resident))) = get_allocator_stats(true, &mibs) {
    |                       -------------------       ^^^^^ expected `Arc<AllocatorStatMibs>`, found `&Arc<AllocatorStatMibs>`
    |                       |
    |                       arguments to this function are incorrect
    |
    = note: expected struct `std::sync::Arc<_>`
            found reference `&std::sync::Arc<_>`
note: function defined here
   --> svix-bridge/src/allocator.rs:60:12
    |
 60 |     pub fn get_allocator_stats(
    |            ^^^^^^^^^^^^^^^^^^^
 61 |         _bust_cache: bool,
 62 |         _mibs: Arc<AllocatorStatMibs>,
    |         -----------------------------
```

Reproduces on unmodified `main`, independent of any PR.

PR #2527, opened 2026-08-02
gha: run bridge check in CI with various combinations of features

svix-jbrown:
follow-up to #2524

### DIFF
--- a/.github/workflows/bridge-ci.yml
+++ b/.github/workflows/bridge-ci.yml
@@ -57,7 +57,7 @@ jobs:
       matrix:
         rust: ["1.91.1", stable, beta]
     steps:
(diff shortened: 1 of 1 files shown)


svix-jplatte (commented):
Do we really care to do a full build for these different feature combinations? We'd catch almost all the issues this can find with just `cargo check`.

svix-jbrown:
> Do we really care to do a full build for these different feature combinations? We'd catch almost all the issues this can find with just `cargo check`.

I'm fine with a check for now, although I have seen issues now and again that only crop up in build (e.g., problems linking with C libraries).

svix-jplatte (approved):

PR #2532, opened 2026-08-03
python, go: Strip trailing slashes from user-provided server URLs

conorbronsdon:
## Motivation

Part of #1154, and a follow-up to #2526, which fixed this for Rust and JavaScript. `https://api.svix.com/` and `https://api.svix.com` should behave the same as a server URL. Today a trailing slash produces a double slash once the request path is appended, since every path already begins with `/`.

#2526's description flagged Python and Go as having the same bug. This is that fix.

**Python:** `_make_httpx_client` constructs `httpx.Client(mounts=..., cookies=...)` with no `base_url` argument, so httpx's own normalization never applies. `base_url` merging only affects relative request URLs, and `common.py` builds an absolute one:

```python
```

**Go:** `svix_http_client.go` concatenated `client.BaseURL + replacePathKeys(...)`.

## Solution

**Python:** `Client.base_url` gets an attrs converter in `python/svix/api/client.py` that strips trailing slashes when the value is stored.

**Go:** `ExecuteRequest` now builds its URL through a `requestURL` helper in `go/internal/svix_http_client.go` that trims trailing slashes off the base URL first.

The request path is untouched in both. It always starts with a slash.

### Where this differs from #2526, and why

**Go normalizes per request; the other three normalize once at construction.** Rust, JavaScript and Python strip when the client is built. Go strips inside `requestURL`, called from `ExecuteRequest`. That is forced rather than preferred: `internal.DefaultSvixHttpClient` only ever receives the built-in default, and a user-supplied URL is assigned straight to the field in generated `go/svix.go:36`, so the request path is the only hand-written code that ever sees a user-supplied base URL.

On #1969 you asked for the normalization to run only when the user actually supplied a server URL:

> I think it's cleaner to only run the `replace` if the user specified a server URL. The fallback values all don't end in `/` and I wouldn't want us to become sloppy with things like that internally.
>
> Same for all the other SDKs.

I could not do that from hand-written code in either of these two SDKs. In both, the place that knows whether the URL came from the user is generated:

- `python/svix/api/svix.py:84` does `host = options.server_url or regional_url or DEFAULT_SERVER_URL`, then `AuthenticatedClient(base_url=host, ...)`. Provenance is gone by the time any hand-written code sees it.
- `go/svix.go:36` does `svixHttpClient.BaseURL = options.ServerUrl.String()`, writing the field directly. The hand-written constructor `internal.DefaultSvixHttpClient()` only ever receives the default URL, so it never sees the user's.

So this patch normalizes at the storage boundary instead, which means the default and regional URLs pass through it too. They don't end in a slash, so it's a no-op for them today, but it is not the guard you asked for: if a built-in URL ever gained a trailing slash, this would hide it rather than surface it.

Doing it the way you described is a one-line change in each template:

- In `codegen/templates/go/summary.go.jinja:36`, wrap `options.ServerUrl.String()` in `strings.TrimRight(..., "/")`. `strings` is already imported there.
- In `codegen/templates/python/summary.py.jinja:72`, strip only when `options.server_url` is set.

I left those out because it means regenerating `go/svix.go` and `python/svix/api/svix.py`, and I didn't want to put generated output in a PR you hadn't asked for it in. Say the word and I'll switch this to the template version.

## How it was verified

```
python/   uv run pytest -q          30 passed, 14 skipped (baseline 24 passed, 14 skipped)
          uv run ./scripts/lint.sh  clean (mypy, ty, ruff check, ruff format --check)
go/       go test ./go/...          44 test funcs passed, 3 skipped (baseline 40 / 3, four added here)
          golangci-lint run         2 govet findings, both pre-existing and identical on main
          go vet, gofmt             clean
```

One pytest error appears in both the baseline and the patched run: `tests/test_openapi.py`, which needs Docker Compose and doesn't have it on this machine. The 14 skips are the Docker-backed client tests. Neither is affected by this change.

Tests go through the real constructors. Python builds a `Svix` / `SvixAsync` and reads the resulting `base_url`. Go has a table test on `requestURL` plus two end-to-end tests: one at the `internal` layer, and one in `go/svix_http_client_test.go` that goes through `svix.New` with a trailing-slash `SvixOptions.ServerUrl` and asserts on the path the server receives.

The tests were also checked by reverting the source change and keeping them. Python goes to 3 failed, 3 passed:

```
FAILED tests/test_server_url.py::test_server_url_trailing_slashes_are_stripped
FAILED tests/test_server_url.py::test_async_client_normalizes_server_url
FAILED tests/test_server_url.py::test_request_url_has_no_double_slash
  AssertionError: assert 'https://api.example.test//api/v1/app' == 'https://api.example.test/api/v1/app'
```

Go goes to 3 failed:

```
--- FAIL: TestServerUrlTrailingSlashIsStripped (0.00s)
    svix_http_client_test.go:339: Expected path `/api/v1/app` got `//api/v1/app`
--- FAIL: Test_requestURL (0.00s)
    --- FAIL: Test_requestURL/trailing_slash_is_stripped (0.00s)
        svix_http_client_test.go:224: expected "https://api.example.com/api/v1/app", got "https://api.example.com//api/v1/app"
    --- FAIL: Test_requestURL/multiple_trailing_slashes_are_stripped (0.00s)
        svix_http_client_test.go:224: expected "https://api.example.com/api/v1/app", got "https://api.example.com////api/v1/app"
    --- FAIL: Test_requestURL/a_path_prefix_is_kept (0.00s)
        svix_http_client_test.go:224: expected "https://api.example.com/prefix/api/v1/app", got "https://api.example.com/prefix//api/v1/app"
    --- FAIL: Test_requestURL/path_params_are_still_substituted (0.00s)
        svix_http_client_test.go:224: expected "https://api.example.com/api/v1/app/app_123/msg", got "https://api.example.com//api/v1/app/app_123/msg"
--- FAIL: Test_executeRequestStripsTrailingSlashFromBaseURL (0.00s)
    svix_http_client_test.go:262: expected path "/api/v1/app/app_123/msg", got "//api/v1/app/app_123/msg"
```

The cases that should pass without the fix — a base URL with no trailing slash, and the default and regional URLs — still pass in both languages. Go had no assertion on the default or regional URL before this branch; that gap mattered because the built-in URLs are exactly where this change departs from the preference stated on #1969.

The new Go test is mutation-checked rather than assumed. Pointing the default at another host fails it, and collapsing the regional URL into the default fails it. Adding a trailing slash to a built-in does *not* fail it — the fix normalizes that away before the assertion runs. So it guards host and regional routing, not built-in sloppiness; nothing hand-written can guard the latter, for the same reason described above.

## Notes

No generated files are touched. All six changed files are absent from `codegen/generated_files.json`:

```
ChangeLog.md                          not in manifest
go/internal/svix_http_client.go       not in manifest
go/internal/svix_http_client_test.go  not in manifest
go/svix_http_client_test.go           not in manifest
python/svix/api/client.py             not in manifest
python/tests/test_server_url.py       not in manifest
```

`.gitattributes` already carries `/python/svix/api/client.py -linguist-generated`, the same negation `rust/src/api/client.rs` has.

The ChangeLog bullet #2526 added is amended in place rather than duplicated, since both are in the same unreleased block.

This leaves #1154 open for Ruby, C#, Java, Kotlin, and PHP.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/python/tests/test_server_url.py
+++ b/python/tests/test_server_url.py
@@ -42,7 +42,9 @@ def request_url(httpx_mock, server_url: str) -> str:
         },
     )
     svx.application.list()
-    return str(httpx_mock.get_request().url)
+    # `get_requests`, not `get_request`: the latter asserts at most one request
+    # was made, so it would break the moment a test calls this helper twice.
+    return str(httpx_mock.get_requests()[-1].url)
 
 
 def test_request_url_has_no_double_slash(httpx_mock) -> None:
--- a/python/tests/test_server_url.py
+++ b/python/tests/test_server_url.py
@@ -30,8 +30,9 @@ def test_async_client_normalizes_server_url() -> None:
     assert svx._client.base_url == "https://api.example.com"
 
 
-def test_request_url_has_no_double_slash(httpx_mock) -> None:
-    svx = Svix("token", SvixOptions(server_url="https://api.example.test/"))
+def request_url(httpx_mock, server_url: str) -> str:
+    """Send a request through a client built with `server_url`, return the URL hit."""
+    svx = Svix("token", SvixOptions(server_url=server_url))
     httpx_mock.add_response(
         json={
             "data": [],
@@ -41,4 +42,18 @@ def test_request_url_has_no_double_slash(httpx_mock) -> None:
         },
     )
     svx.application.list()
-    assert str(httpx_mock.get_request().url) == "https://api.example.test/api/v1/app"
+    return str(httpx_mock.get_request().url)
+
+
+def test_request_url_has_no_double_slash(httpx_mock) -> None:
+    assert (
+        request_url(httpx_mock, "https://api.example.test/")
+        == "https://api.example.test/api/v1/app"
+    )
+
+
+def test_request_url_keeps_path_prefix(httpx_mock) -> None:
+    assert (
+        request_url(httpx_mock, "https://api.example.test/prefix/")
+        == "https://api.example.test/prefix/api/v1/app"
+    )
--- a/python/tests/test_server_url.py
+++ b/python/tests/test_server_url.py
@@ -1,0 +1,44 @@
+from svix.api import Svix, SvixAsync, SvixOptions
+
+
+def base_url(server_url: str) -> str:
+    return Svix("token", SvixOptions(server_url=server_url))._client.base_url
+
+
+def test_server_url_trailing_slashes_are_stripped() -> None:
+    assert base_url("https://api.example.com/") == "https://api.example.com"
+    assert base_url("https://api.example.com///") == "https://api.example.com"
+    assert (
+        base_url("https://api.example.com/prefix/") == "https://api.example.com/prefix"
+    )
+
+
+def test_server_url_without_trailing_slash_is_unchanged() -> None:
+    assert base_url("https://api.example.com") == "https://api.example.com"
+
+
+def test_default_server_url() -> None:
+    assert Svix("token")._client.base_url == "https://api.svix.com"
+
+
+def test_regional_server_url() -> None:
+    assert Svix("testsk.eu")._client.base_url == "https://api.eu.svix.com"
+
+
+def test_async_client_normalizes_server_url() -> None:
+    svx = SvixAsync("token", SvixOptions(server_url="https://api.example.com/"))
+    assert svx._client.base_url == "https://api.example.com"
+
+
+def test_request_url_has_no_double_slash(httpx_mock) -> None:
+    svx = Svix("token", SvixOptions(server_url="https://api.example.test/"))
+    httpx_mock.add_response(
+        json={
+            "data": [],
+            "done": True,
+            "iterator": None,
+            "prevIterator": None,
+        },
+    )
+    svx.application.list()
+    assert str(httpx_mock.get_request().url) == "https://api.example.test/api/v1/app"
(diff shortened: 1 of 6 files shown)


python/tests/test_server_url.py
```diff
@@ -0,0 +1,44 @@
...
+    httpx_mock.add_response(
+        json={
+            "data": [],
+            "done": True,
+            "iterator": None,
+            "prevIterator": None,
+        },
+    )
+    svx.application.list()
+    assert str(httpx_mock.get_request().url) == "https://api.example.test/api/v1/app"
```
svix-jbrown: for funsies, can you add an equivalent test where the `server_url` has a prefix (i.e., `https://api.example.test/prefix/`) so we catch it if url-joining is ever done with concatenation?

conorbronsdon: For sure, will do

conorbronsdon: Added in 7986fe7 as `test_request_url_keeps_path_prefix` — it drives a real request through `application.list()` with `server_url="https://api.example.test/prefix/"` and asserts the outgoing URL is `https://api.example.test/prefix/api/v1/app`. The existing wire-level test and this one now share a `request_url` helper.

One thing worth flagging, because it inverts the reason for the request: concatenation is what *preserves* the prefix here. `common.py` does `f"{base_url}{path}"` today, and since the request path is absolute, a real join is what drops the prefix — `urljoin(".../prefix", "/api/v1/app")` is `.../api/v1/app`. So the test guards against a future move *away* from concatenation. 

Confirmed by mutation rather than assumed — swapping the concatenation for `urljoin(base_url, path)`:

```
test_request_url_keeps_path_prefix   -> FAIL
    got  https://api.example.test/api/v1/app
    want https://api.example.test/prefix/api/v1/app
test_request_url_has_no_double_slash -> still passes
```

The new test is the only thing in the suite that catches it. Go already covers the same case in `Test_requestURL` ("a path prefix is kept"), added earlier in this PR.

While checking: Rust and JS concatenate too (`format!("{}{}", conf.base_path, path)` and `ctx.baseUrl + this.path`), and after #2526 both assert only the stored base URL, never a wire URL. No bug, just the same missing test. Happy to add it there in a follow-up if you want the parity.

svix-jbrown (approved):

PR #2548, opened 2026-08-04
ruby, php: Make the webhook timestamp tolerance configurable

conorbronsdon:
## Motivation

Per the review discussion below: #93 wants the timestamp tolerance to be configurable, the separate-method approach from the earlier version of this PR is out.

## Solution

Both SDKs now take an optional `tolerance` constructor argument (seconds, default 5 minutes). It must be a non-negative integer:
- NaN/null/floats are rejected at construction
- NaN silently disables both window checks and PHP coerces null to a zero-second window

One edge case: a tolerance reaching past the epoch still rejects non-positive timestamps, matching Rust ("we won't honor timestamps before 1970") and keeping PHP's `sign()` from throwing on attacker-controlled input.

Tests cover accept/reject in both directions vs the default, signature failure under a widened window, and the constructor validation. rspec: 56 examples, 0 failures; phpunit: 52 tests, OK.

_PR assisted by Claude_

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -13,9 +13,9 @@ The Svix Bridge changelog has moved to [bridge/ChangeLog.md](./bridge/ChangeLog.
 * Libs/Rust, Libs/JavaScript, Libs/Python, Libs/Go: Strip trailing slashes from a
   user-provided server URL, so `https://api.svix.com/` and `https://api.svix.com`
   behave the same
-* Libs/Ruby, Libs/PHP: Add a webhook verification method that skips the timestamp
-  tolerance check (`verify_ignoring_timestamp` / `verifyIgnoringTimestamp`),
-  matching the Go and Rust SDKs
+* Libs/Ruby, Libs/PHP: The timestamp tolerance used by webhook verification can
+  now be customized through a `tolerance` constructor argument (in seconds;
+  default five minutes)
 
 ## Version 2.0.0-rc.2
 * Libs/JS: Fix a publishing issue
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -13,6 +13,9 @@ The Svix Bridge changelog has moved to [bridge/ChangeLog.md](./bridge/ChangeLog.
 * Libs/Rust, Libs/JavaScript, Libs/Python, Libs/Go: Strip trailing slashes from a
   user-provided server URL, so `https://api.svix.com/` and `https://api.svix.com`
   behave the same
+* Libs/Ruby, Libs/PHP: Add a webhook verification method that skips the timestamp
+  tolerance check (`verify_ignoring_timestamp` / `verifyIgnoringTimestamp`),
+  matching the Go and Rust SDKs
 
 ## Version 2.0.0-rc.2
 * Libs/JS: Fix a publishing issue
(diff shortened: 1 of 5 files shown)


svix-jplatte:
Hey, thanks for the PR. I think we made the wrong call previously though, and we should instead just allow customizing the tolerance. I'll bring it up with the team.

svix-jplatte:
Team consensus is that our previous decision was wrong (sorry @ogenodisho for closing #2055 back then!).

Would you mind updating the PR to allow customizing the tolerance to a user-provided value?

conorbronsdon:
sure thing

svix-jplatte:
Also, could you please write the PR descriptions yourself going forward? The overall length and note at the bottom has me under the impression that the whole thing is generated by Claude.

conorbronsdon:
Will do - I have claude generate the first draft of the description and then edit, but agree has been too dense etc

Will adjust

conorbronsdon:
Done: swapped to tolerance constructor, PR description updated

php/src/Webhook.php
```diff
@@ -7,13 +7,24 @@ class Webhook
     public const SECRET_PREFIX = "whsec_";
     public const TOLERANCE = 5 * 60;
```
svix-jplatte: While you're at it, can you make this private and rename to `DEFAULT_TOLERANCE`? I didn't realize we had this public constant, and this is a great opportunity to get rid of it since we haven't released the final v2 of the SDKs yet. Same for `SECRET_PREFIX` actually which should also become private because we'll likely add support for asymmetric secrets at some point (and those use a different prefix), but that doesn't really belong in this PR.

ruby/lib/svix/webhook.rb
svix-jplatte: Can you rename to `DEFAULT_TOLERANCE`?

ChangeLog.md
```diff
@@ -13,6 +13,9 @@ The Svix Bridge changelog has moved to [bridge/ChangeLog.md](./bridge/ChangeLog.
 * Libs/Rust, Libs/JavaScript, Libs/Python, Libs/Go: Strip trailing slashes from a
   user-provided server URL, so `https://api.svix.com/` and `https://api.svix.com`
   behave the same
+* Libs/Ruby, Libs/PHP: The timestamp tolerance used by webhook verification can
+  now be customized through a `tolerance` constructor argument (in seconds;
+  default five minutes)
```
svix-jplatte:
```suggestion
  defaults to the previously hardcoded value of five minutes)
```

svix-jplatte (approved):
Thanks!

conorbronsdon:
Looks like a CI sissue

ogenodisho:
> Team consensus is that our previous decision was wrong (sorry @ogenodisho for closing #2055 back then!).
> 
> Would you mind updating the PR to allow customizing the tolerance to a user-provided value?

No worries @svix-jplatte , thank you for addressing this. Shall I reopen #2055 in that case? Because it currently does the same thing as this PR but for the Python version (configurable tolerance which defaults to 5 minutes)

svix-jplatte:
Sure, if you're still interested! For Python though, we're now calling into standardwebhooks there so I think that lib will need a PR first.

PR #2561, opened 2026-08-13
Update OpenAPI spec yet again

svix-jplatte:
We identified a couple more changes that we wanted to make before the proper release of v2 of the SDKs.


svix-pluiz (approved):
Do we already have the setup for extraordinary patches for the v1 SDK? We should test those

svix-jplatte:
We can start a branch off of the last release. The workflow files themselves have no issue running from a separate branch, the only thing I'm not sure about (but can't do anything about myself) is trusted publishing. I think we can figure that out once we get to the point of needing to release another v1.x

PR #2570, opened 2026-08-20
ruby, php: Accept webhook headers of any casing as a fallback

### LINKED ISSUE #1013: Support header names of all casing (as a fallback), not just all lowercase
Some of the libraries (e.g. PHP!) assume the headers are all lower-case. And while they should be lowercase (that's how Svix sends them).

Doing this for every webhook is probably not that efficient, so we should probably only fallback to this if headers are not found.

dylanpulver:
## Motivation

The PHP and Ruby verifiers look the signature headers up by their exact lowercase names, so a request whose headers arrive as `Svix-Id` rather than `svix-id` is rejected with "Missing required headers" even though the signature itself is perfectly good. HTTP header names are case-insensitive, so nothing on the sending side is wrong here. It is the receiving side that decides the casing, and the casing it picks is often not lowercase. PHP's `getallheaders()` commonly hands back `Svix-Id`, and Rack presents the same shape to Ruby apps, so an otherwise correct integration fails purely on how the server chose to spell the header names.

The other libraries already handle this. `python/svix/webhooks.py` lowercases every incoming key before it looks anything up, and `javascript/src/webhook.ts` does the same at the top of `Webhook.verify`. Go never had the problem, since `http.Header.Get` is case-insensitive by spec. PHP and Ruby are the two that were left behind.

## Solution

Both verifiers now try the exact lowercase names first, exactly as they did before, and build a lowercased view of the headers only when that first lookup comes up empty. The path for headers spelled the way Svix actually sends them is unchanged and does no extra work, which is what the issue asks for. The lowercased retry preserves the existing preference for the branded `svix-` names over the unbranded `webhook-` ones. When both sets are present at once, an exact lowercase match still wins over a differently cased one, which is the cost of keeping the common path free of any allocation.

Tests were added in `php/tests/WebhookTest.php` and `ruby/spec/webhook_spec.rb`. Each language covers a title-cased header set and a deliberately mixed-case one using `SVIX-ID` alongside `sViX-SiGnAtUrE`. The unbranded `Webhook-` names are covered in title case as well. Each language also has a case where one of the title-cased headers is genuinely absent and verification still raises "Missing required headers", so the fallback cannot quietly paper over an incomplete request. The exact lowercase path was already covered by the existing tests and still passes.

I ran `vendor/bin/phpunit --configuration phpunit.xml` and `bundle exec rspec spec/webhook_spec.rb` locally, on PHP 8.5.9 and Ruby 3.3.0. Reverting the two source files while keeping the new tests makes the casing tests fail, so they do exercise the change. There is a ChangeLog entry under Unreleased.

Fixes #1013

### DIFF
--- a/php/src/Webhook.php
+++ b/php/src/Webhook.php
@@ -40,26 +40,19 @@ class Webhook
      */
     public function verify($payload, $headers)
     {
-        if (
-            isset($headers['svix-id'])
-            && isset($headers['svix-timestamp'])
-            && isset($headers['svix-signature'])
-        ) {
-            $msgId = $headers['svix-id'];
-            $msgTimestamp = $headers['svix-timestamp'];
-            $msgSignature = $headers['svix-signature'];
-        } elseif (
-            isset($headers['webhook-id'])
-            && isset($headers['webhook-timestamp'])
-            && isset($headers['webhook-signature'])
-        ) {
-            $msgId = $headers['webhook-id'];
-            $msgTimestamp = $headers['webhook-timestamp'];
-            $msgSignature = $headers['webhook-signature'];
-        } else {
+        $signedHeaders = self::findSignedHeaders($headers);
+        if ($signedHeaders === null && is_array($headers)) {
+            // Svix sends the headers in lowercase, so the lookup above is the
+            // common case. Some servers hand them back with a different casing
+            // (`getallheaders()` typically returns `Svix-Id`), so retry once
+            // against lowercased keys rather than paying for that every time.
+            $signedHeaders = self::findSignedHeaders(array_change_key_case($headers, CASE_LOWER));
+        }
+        if ($signedHeaders === null) {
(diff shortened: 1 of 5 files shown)


svix-lucho:
Thanks for the contribution! And sorry for the late review.

php/src/Webhook.php
svix-lucho, line 52: I feel like this would be more readable if we just constructed a normalized dictionary with the lowercase versions, like we do in the javascript version.

```
  public verify(
    payload: string | Buffer,
    headers_:
      | WebhookRequiredHeaders
      | WebhookUnbrandedRequiredHeaders
      | Record<string, string>
  ): undefined {
    const headers: Record<string, string> = {};
    for (const key of Object.keys(headers_)) {
      headers[key.toLowerCase()] = (headers_ as Record<string, string>)[key];
    }

    [...]
  }
```

what do you think?

svix-lucho: same for ruby

PR #2575, opened 2026-08-21
server: implement bulk-expunge in OSS

svix-jbrown:
This matches the cloud implementation (svix/svix-webhooks-private#6996).

Part of svix/monorepo-private#14037

### DIFF
--- a/server/svix-server/src/lib.rs
+++ b/server/svix-server/src/lib.rs
@@ -432,7 +432,7 @@ pub fn setup_tracing_for_tests() {
                 // Output is only printed for failing tests, but still we shouldn't overload
                 // the output with unnecessary info. When debugging a specific test, it's easy
                 // to override this default by setting the `RUST_LOG` environment variable.
-                "svix_server=debug".into()
+                "it=debug,svix_server=debug,info".into()
             }),
         )
         .with(tracing_subscriber::fmt::layer().with_test_writer())
--- a/server/svix-server/tests/it/e2e_message.rs
+++ b/server/svix-server/tests/it/e2e_message.rs
@@ -1,6 +1,8 @@
 // SPDX-FileCopyrightText: © 2022 Svix Authors
 // SPDX-License-Identifier: MIT
 
+use std::collections::BTreeMap;
+
 use chrono::{Duration, Utc};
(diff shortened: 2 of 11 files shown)


server/svix-server/src/lib.rs
```diff
@@ -432,7 +432,7 @@ pub fn setup_tracing_for_tests() {
                 // Output is only printed for failing tests, but still we shouldn't overload
                 // the output with unnecessary info. When debugging a specific test, it's easy
                 // to override this default by setting the `RUST_LOG` environment variable.
-                "svix_server=debug".into()
+                "it=debug,svix_server=debug,info".into()
```
svix-jplatte, line 435: Oh. Yeah this makes much more sense 😄

server/svix-server/tests/it/e2e_message.rs
```diff
@@ -520,6 +526,166 @@ async fn test_expunge_message_payload() {
...
+    let app_id = create_test_app(&client, "V1MessageBulkExpungeApp")
+        .await?
+        .id;
+    let receiver = TestReceiver::start(StatusCode::OK);
+    create_test_endpoint(&client, &app_id, &receiver.endpoint).await?;
+
+    let message_1: MessageOut = client
+        .post(
+            &format!("api/v1/app/{app_id}/msg/"),
+            message_in("event-1", json!({ "test": "value" }))?,
```
svix-jplatte, line 542: Why not use a `json!` literal for the entire second arg to `post`, like we've been doing for new tests in `-private`?

svix-jplatte (approved):

server/svix-server/tests/it/e2e_message.rs
svix-jbrown: all the expunge tests in -private also use `message_in` and `message_in_with_uid` (in fact, this test is just copied from private), but if you'd prefer me to use a `json!` literal I obviously can in a follow-up

svix-jplatte: Alright. If you care, I wouldn't mind if you joined me in the test-cleanup marathon but I can also do it myself.

PR #2581, opened 2026-08-25
server: Install a rustls CryptoProvider at startup

### LINKED ISSUE #2579: Startup panic since v1.87.0: could not determine rustls CryptoProvider
### Version

- Broken: `svix/svix-server:v1.100.0` and `:latest` (same digest, `sha256:ff30f0e665cd9449646785ac77098a1b258ea110cf6c0e6b766528ed8c63642a`), and every image from `v1.87.0` onward that we tested (`v1.87.0`, `v1.98.0`, `v1.99.0`, `v1.100`).
- Last working: `svix/svix-server:v1.86.0`.
- Relevant transitive dependency, from `server/Cargo.lock`: `rustls 0.23.43` (0.23.37 on the v1.8x/v1.9x images), with **both** `aws-lc-rs` and `ring` enabled.

### Platform

- Image: `linux/arm64` (the reproduction below was run on an arm64 host; the cause is feature unification in the lockfile, so we do not expect it to be arch-specific).
- Production: Kubernetes on Amazon Linux 2023, kernel `6.12.100-125.179.amzn2023.aarch64`.
- Postgres: Aurora with TLS required; we connect using `sslmode=verify-ca` and the AWS RDS CA bundle.

### Description

`svix-server` panics during startup and exits 101 before logging anything, whenever `SVIX_DB_DSN` uses `sslmode=verify-ca` with an explicit `sslrootcert`. The process never reaches the point of serving traffic, so the pod crash-loops.

I tried this:

```bash
# Postgres with TLS enabled, using a leaf certificate signed by a local test CA
docker run -d --name pg --network t --network-alias pg \
  -e POSTGRES_PASSWORD=p -e POSTGRES_DB=svix my-pg-with-certs \
  -c ssl=on -c ssl_cert_file=/certs/server.crt -c ssl_key_file=/certs/server.key

docker run --rm --network t -v "$PWD/ca.crt:/ca.crt:ro" \
  -e SVIX_DB_DSN="postgresql://postgres:p@pg:5432/svix?sslmode=verify-ca&sslrootcert=/ca.crt" \
  -e SVIX_JWT_SECRET=x -e SVIX_CACHE_TYPE=memory -e SVIX_QUEUE_TYPE=memory \
  svix/svix-server:latest svix-server --run-migrations
```

I expected to see this happen: the migrations run and the server starts, as they do on `v1.86.0`.

Instead, this happened:

```
thread 'main' (1) panicked at /usr/local/cargo/registry/src/index.crates.io-.../rustls-0.23.43/src/crypto/mod.rs:249:14:
Could not automatically determine the process-level CryptoProvider from Rustls crate features.
Call CryptoProvider::install_default() before this point to select a provider manually,
or make sure exactly one of the 'aws-lc-rs' and 'ring' features is enabled.
See the documentation of the CryptoProvider type for more information.
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

One container run per image, same harness:

| image                            | outcome                                       |
|----------------------------------|-----------------------------------------------|
| `v1.86.0`                        | migrations complete, server starts and serves |
| `v1.87.0`                        | CryptoProvider panic, exit 101                |
| `v1.98.0`                        | CryptoProvider panic, exit 101                |
| `v1.99.0`                        | CryptoProvider panic, exit 101                |
| `v1.100` / `v1.100.0` / `latest` | CryptoProvider panic, exit 101                |

#### What appears to be going on

`server/Cargo.lock` enables both rustls crypto backends, so `rustls` cannot resolve a default provider and the first code path requesting one panics. The `rustls` node lists them directly:

| tag                                                | rustls          | providers in the `rustls` lock node |
|----------------------------------------------------|-----------------|-------------------------------------|
| `v1.86.0`                                          | 0.22.4, 0.23.23 | `ring`                              |
| `v1.87.0`                                          | 0.23.37         | **`aws-lc-rs`, `ring`**             |
| `v1.98.0`                                          | 0.23.37         | `aws-lc-rs`, `ring`                 |
| `v1.99.0`                                          | 0.23.37         | `aws-lc-rs`, `ring`                 |
| `server-v1.100.0`                                  | 0.23.43         | `aws-lc-rs`, `ring`                 |
| `v2.0.0` (monorepo tag; no server image published) | 0.23.43         | `aws-lc-rs`, `ring`                 |

```bash
for t in v1.86.0 v1.87.0 server-v1.100.0 v2.0.0; do
  echo "== $t"
  curl -s "https://raw.githubusercontent.com/svix/svix-webhooks/$t/server/Cargo.lock" \
    | rg -A14 '^name = "rustls"$' | rg 'version =|"aws-lc-rs"|"ring"'
done
```

`aws-lc-rs` reaches the graph through `quinn-proto` (via `reqwest`) while `ring` stays enabled from elsewhere, and Cargo unifies features across the graph, so both end up on.

`svix-server` never calls `CryptoProvider::install_default()`. A repo-wide search finds exactly one such call, in `svix-cli/src/main.rs` — added for this same symptom in v1.81.0 ("Fix `svix listen` failing with 'Could not automatically determine the process-level CryptoProvider'"). The server binary presumably never needed it until the lockfile changed in v1.87.0.

Since the last `v2.0.0` row above is the monorepo tag and no `v2.0.0` server image exists, that row is source inspection rather than something we ran — but it suggests current source is still affected.

#### Why this may not have shown up in testing

Plain TLS is not enough to trigger it; a custom root store is. Measured on `v1.100`:

| `SVIX_DB_DSN`                                  | Result                                                       |
|------------------------------------------------|--------------------------------------------------------------|
| `…?sslmode=require`                            | starts fine, migrations run (`pg_stat_ssl` confirms TLSv1.3) |
| `…?sslmode=verify-ca&sslrootcert=/path/ca.crt` | CryptoProvider panic, exit 101                               |

A TLS-enabled-Postgres smoke test therefore passes on the broken builds. `verify-ca` with an explicit CA bundle is a common configuration for RDS/Aurora deployments.

#### Suggested fix

Call `CryptoProvider::install_default()` early in the server's startup, the way `svix-cli` does, or pin the crate features so that exactly one backend is enabled. Happy to open a PR if useful.

gergely-kurucz-konghq:
## Motivation

Since the dependency update in v1.87.0, `svix-server` panics and exits 101 before logging anything whenever rustls has to build a custom root store — most visibly when connecting to Postgres with `sslmode=verify-ca` and an explicit `sslrootcert`, a common RDS/Aurora setup:

```
Could not automatically determine the process-level CryptoProvider from Rustls crate features.
```

Both rustls backends are enabled in the server's graph, so rustls cannot pick one:

- `aws-lc-rs` — `reqwest`'s `rustls` feature (also `jsonschema`, the `svix` SDK, `lapin`)
- `ring` — `sqlx` and `sea-orm` via `runtime-tokio-rustls`

`sslmode=require` builds no custom root store and still works, which is why a TLS-enabled Postgres smoke test passes on the affected builds.

Fixes #2579

## Solution

Install the `aws-lc-rs` provider as the first statement of `main()`, mirroring what `svix-cli/src/main.rs` already does for the same symptom. Requires `rustls` as a direct dependency of `svix-server`.

Pinning crate features so exactly one backend is enabled would also work, but `sea-orm` exposes no `aws-lc-rs` TLS feature of its own, and any future dependency re-enabling `rustls/ring` would silently bring the panic back.

### Verification

Docker image built from this branch, against Postgres with a leaf cert signed by a local CA:

| build | `sslmode=verify-ca&sslrootcert=…` |
|---|---|
| `svix/svix-server:latest` | CryptoProvider panic, exit 101 |
| this branch | migrations complete, server starts and serves |

`pg_stat_ssl` confirms `TLSv1.3` on the connection; `sslmode=require` still starts clean. `cargo clippy --all-targets --all-features -- -D warnings` is clean, and the `queue=memory,cache=none` configuration of `run-tests.sh` passes (161 tests).

No test is included: the failure is process-global rustls state at startup, which the integration suite (running in-process against a non-TLS Postgres) cannot observe.

### Disclaimer

Created with the assistance of Claude Opus 5, but with human oversight.

### DIFF
--- a/server/Cargo.lock
+++ b/server/Cargo.lock
@@ -5312,6 +5312,7 @@ dependencies = [
  "regex",
  "reqwest",
  "rust-executable-metadata",
+ "rustls",
  "schemars",
  "sea-orm",
  "sentry",
--- a/server/ChangeLog.md
+++ b/server/ChangeLog.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+* Fix startup panic ("Could not automatically determine the process-level CryptoProvider") when connecting to Postgres with `sslmode=verify-ca`
 
 ## Version 1.100.0
 * Upgrade dependencies
--- a/server/svix-server/Cargo.toml
+++ b/server/svix-server/Cargo.toml
@@ -72,6 +72,7 @@ rand = "0.10"
 redis = { version = "0.32.5", features = ["tokio-comp", "tokio-native-tls-comp", "streams", "cluster-async", "tcp_nodelay", "connection-manager", "sentinel"] }
 regex = "1.5.5"
 reqwest = { version = "0.13.4", default-features = false, features = ["json", "rustls", "hickory-dns"] }
+rustls = "0.23.43"
 schemars = { version = "0.8.11", features = ["chrono", "url", "preserve_order"] }
 sea-orm = { version = "1.1.1", features = ["sqlx-postgres", "runtime-tokio-rustls", "macros", "with-chrono", "with-json"], default-features = false }
 sentry = { version = "0.48.2", default-features = false, features = ["backtrace", "contexts", "debug-images", "panic", "tracing", "transport"] }
--- a/server/svix-server/src/main.rs
+++ b/server/svix-server/src/main.rs
@@ -146,6 +146,12 @@ fn org_id_parser(s: &str) -> Result<OrganizationId, String> {
 
 #[tokio::main]
 async fn main() -> anyhow::Result<()> {
+    // rustls requires a crypto backend ("provider") choice to be made explicitly when more than
+    // one is compiled in, which is the case here: reqwest enables aws-lc-rs while sqlx and sea-orm
+    // enable ring. Without this, the first code path that asks rustls for the process-level default
+    // provider panics (for example when connecting to Postgres with `sslmode=verify-ca`).
+    _ = rustls::crypto::aws_lc_rs::default_provider().install_default();
+
     dotenv().ok();
 
     let args = Args::parse();


svix-jplatte (approved):
Thanks!

PR #2583, opened 2026-08-25
Fix missing queue prefixes in redis queue migration code 

### LINKED ISSUE #2580: Legacy queue drain ignores SVIX_QUEUE_PREFIX on destination keys
### Version

- Observed on `svix/svix-server:v1.86.0` and `:v1.100`.
- The relevant code is unchanged as of the `v2.0.0` monorepo tag, so we believe current source is affected as well.
- Relevant history: `queue_prefix` became a user-facing config option in **v1.82.0** (#2107); it is absent from `cfg.rs` at v1.81.0.

### Platform

- Image: `linux/arm64`.
- Production: Kubernetes on Amazon Linux 2023, kernel `6.12.100-125.179.amzn2023.aarch64`.
- Redis: ElastiCache/Valkey over `rediss://`, with an ACL scoped by key prefix. `SVIX_QUEUE_TYPE=redis`, `SVIX_CACHE_TYPE=memory`, `SVIX_QUEUE_PREFIX` set.

### Description

The legacy queue drain does not account for `SVIX_QUEUE_PREFIX` on its **destination** keys, and it has no way to be disabled. Two distinct problems follow: possible silent data stranding for deployments that do have legacy data, and permanent error noise for those that do not.

To be clear about what is *not* the problem: the legacy **source** keys are read unprefixed, and that looks correct — no server before v1.82.0 could set a prefix, so any v1/v2-era keys must be unprefixed.

#### Problem 1: migrations write to unprefixed destinations

In `server/svix-server/src/queue/redis.rs`, the live queue names get the prefix in `new_pair_inner`:

```rust
let main_queue_name    = format!("{queue_prefix}{main_queue_name}");
let delayed_queue_name = format!("{queue_prefix}{delayed_queue_name}");
```

The migrations, however, use the bare constants on both sides:

```rust
async fn migrate_v2_to_v3_queues(conn: &mut RedisConnection<'_>) -> Result<()> {
    migrate_list_to_stream(conn, LEGACY_V2_MAIN, MAIN).await?;       // MAIN is unprefixed
    migrate_list_to_stream(conn, LEGACY_V2_PROCESSING, MAIN).await?;
    Ok(())
}

async fn migrate_v1_to_v2_queues(conn: &mut RedisConnection<'_>) -> Result<()> {
    migrate_list(conn, LEGACY_V1_MAIN, LEGACY_V2_MAIN).await?;
    migrate_list(conn, LEGACY_V1_PROCESSING, LEGACY_V2_PROCESSING).await?;
    migrate_sset(conn, LEGACY_V1_DELAYED, DELAYED).await?;           // DELAYED is unprefixed
    Ok(())
}
```

`migrate_list_to_stream` then does `pipe.xadd(queue, ...)` with that unprefixed name.

I expected to see this happen, on a deployment with `SVIX_QUEUE_PREFIX` set and legacy data present: the drain moves the messages into the queues the running server actually consumes, i.e. `<prefix>{queue}_svix_v3_main`.

Instead, what the code does: it writes them to `{queue}_svix_v3_main` and `{queue}_svix_delayed` — live-shaped keys without the prefix, which the running server never reads. The messages would be silently stranded, with the legacy keys already drained.

This needs a deployment that both still holds v1/v2-era keys and has moved to >= v1.82.0 with a prefix configured, so it is narrow — but the failure mode is silent message loss, which is why we are reporting it. In our case it never fires, because our ACL denies the unprefixed keys and the drain aborts before writing anything (see problem 2). The permission error is effectively protecting the data.

#### Problem 2: on a prefix-scoped ACL the drain can only ever fail, and cannot be turned off

Our Redis user's ACL permits only our prefix (`on ~om:* +@all -@dangerous`). The drain reads unprefixed keys, so every attempt fails:

```
ERROR svix_server::queue::redis: Error migrating queue: NOPERM: this user has no permissions to access one of the keys used as arguments
```

`run_migration_schedule` retries on its delay ladder (8 attempts, ~11 min to 24 h) for the lifetime of every pod, so a correctly configured prefixed deployment logs errors indefinitely. There is no config gate — the drain is an unconditional `tokio::spawn` — and we cannot widen the ACL, because our Redis cluster is multi-tenant and partitioned by key prefix per team, so granting unprefixed keys is not an option.

We would rather not filter the message out on our side either: any filter narrow enough to catch this would also hide a genuine Redis misconfiguration later.

#### Suggested fix

1. Apply `queue_prefix` to the migration **destinations** (`MAIN`, `DELAYED`, and the `LEGACY_V2_*` intermediates if they are to stay self-consistent), leaving the legacy sources unprefixed as they are. That closes the stranding case.
2. Add a config flag to skip the legacy drain, so deployments that never had v1/v2 queues can turn it off. That resolves the noise for everyone in our position, and is useful regardless of the ACL question.

svix-jplatte:
… and allow disabling the migration task entirely.

Fixes #2580.

### DIFF
--- a/server/svix-server/src/cfg.rs
+++ b/server/svix-server/src/cfg.rs
@@ -246,8 +246,20 @@ pub struct ConfigurationInner {
     /// Should this instance run background migrations
     pub background_migrations_enabled: bool,
 
+    /// Enable migration task for old redis queue tasks.
+    ///
+    /// Defaults to `true`. Can be safely set to `false` if redis contains no
+    /// (more) pre-v0.64.0 queue tasks.
+    /// Has no effect if the queue backend is not redis.
+    #[serde(default = "default_true")]
+    pub enable_redis_v1_v2_queue_migration: bool,
+
     #[serde(flatten)]
     pub internal: InternalConfig,
+}
+
+fn default_true() -> bool {
+    true
 }
 
 #[derive(Clone, Debug, Deserialize)]
--- a/server/svix-server/src/queue/redis.rs
+++ b/server/svix-server/src/queue/redis.rs
@@ -181,34 +181,36 @@ async fn new_pair_inner(
         .try_into()
         .expect("Pending duration out of bounds");
 
-    // Migrate v1 queues to v2 and v2 queues to v3 on a loop with exponential backoff.
-    tokio::spawn({
-        let pool = pool.clone();
-        let prefix = queue_prefix.to_owned();
+    if cfg.enable_redis_v1_v2_queue_migration {
+        // Migrate v1 queues to v2 and v2 queues to v3 on a loop with exponential backoff.
+        tokio::spawn({
+            let pool = pool.clone();
+            let prefix = queue_prefix.to_owned();
 
-        async move {
-            let delays = [
-                // 11.25 min
-                Duration::from_secs(60 * 11 + 15),
-                // 22.5 min
-                Duration::from_secs(60 * 22 + 30),
-                // 45 min
-                Duration::from_secs(60 * 45),
-                // 1.5 hours
-                Duration::from_secs(60 * 30 * 3),
-                // 3 hours
-                Duration::from_secs(60 * 60 * 3),
-                // 6 hours
-                Duration::from_secs(60 * 60 * 6),
-                // 12 hours
-                Duration::from_secs(60 * 60 * 12),
-                // 24 hours
-                Duration::from_secs(60 * 60 * 24),
-            ];
+            async move {
+                let delays = [
+                    // 11.25 min
+                    Duration::from_secs(60 * 11 + 15),
+                    // 22.5 min
+                    Duration::from_secs(60 * 22 + 30),
+                    // 45 min
+                    Duration::from_secs(60 * 45),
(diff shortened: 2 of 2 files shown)


server/svix-server/src/cfg.rs
```diff
@@ -246,10 +246,22 @@ pub struct ConfigurationInner {
     /// Should this instance run background migrations
     pub background_migrations_enabled: bool,
 
+    /// Enable migration task for old redis queue tasks.
+    ///
+    /// Defaults to `true`. Can be safely set to `false` if redis contains no
+    /// (more) pre-v0.64.0 queue tasks.
+    /// Has no effect if the queue backend is not redis.
+    #[serde(default = "default_true")]
+    pub enable_redis_v1_v2_queue_migration: bool,
```
svix-james, line 255: I wonder if this should even just be set to false at this point. This is 4 years old at this point, correct?

svix-jplatte: Yeah, I know but then might as well delete it? I'm not sure, we've been very conservative and this isn't exactly complex to keep around.

svix-james (approved):

PR #2604, opened 2026-08-31
libs/ruby: fix scalar query parameter encoding

### LINKED ISSUE #2595: Ruby client rejects scalar query parameter values
## Summary

The generated Ruby client passes non-array, non-Time query values directly to `CGI.escape`. Ruby's CGI API expects a string, so ordinary integer and boolean query values raise `TypeError` before a request is sent.

## Reproduction

On current `main` and the 2.1.0 release, invoking the query encoder with `{ "limit" => 5, "with_content" => true }` raises `TypeError: no implicit conversion of Integer into String`.

## Proposed correction

Convert the scalar value with `v.to_s` before escaping. Arrays and Time values retain their existing handling.

## Verification

An isolated candidate passes the full 56-example Ruby suite and focused scalar-query models on Ruby 3.2.11 and 4.0.6. No generated API shape changes are involved.

anishfyi:
## Motivation

Fixes #2595.

The Ruby HTTP client passes ordinary query values directly to `CGI.escape`, which only accepts strings. Integer and boolean query parameters therefore raise `TypeError` before a request is sent.

## Solution

Convert ordinary scalar query values to strings before escaping them. Array and `Time` handling remain unchanged.

Added regression coverage for integer and boolean query parameters.

Validation:

- `bundle exec rake build`

### DIFF
--- a/ruby/lib/svix/svix_http_client.rb
+++ b/ruby/lib/svix/svix_http_client.rb
@@ -113,7 +113,7 @@ module Svix
           elsif v.is_a?(Time)
             encoded_query_pairs.append("#{k}=#{CGI.escape(v.utc.to_datetime.rfc3339)}")
           else
-            encoded_query_pairs.append("#{k}=#{CGI.escape(v)}")
+            encoded_query_pairs.append("#{k}=#{CGI.escape(v.to_s)}")
           end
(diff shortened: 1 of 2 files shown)


svix-jbrown (approved):
seems fine

PR #2605, opened 2026-08-31
Accept symbol keys in Ruby struct enums

### LINKED ISSUE #2596: Ruby struct enums reject symbol-keyed discriminator content
## Summary

Generated Ruby struct-enum initializers accept either string or symbol keys when validating fields, but compare the discriminator-content key only to a string. A valid symbol-keyed hash therefore sets the model value without setting its discriminator and then raises that the required field was not set.

## Reproduction

Constructing a generated struct enum such as `Svix::AutoConfigSinkType.new(config: Svix::AutoConfigSinkTypeConfig::Http.new(...))` accepts `:config` in `ALL_FIELD.include?(k.to_s)` but misses `k == "config"`, ending with `ArgumentError: Required config field was not set`.

## Proposed correction

Compare `k.to_s` with the generated content-field name in the struct-enum template and regenerate the six affected models.

## Verification

An isolated generator-first candidate passes the full 56-example Ruby suite and focused string/symbol-key models on Ruby 3.2.11 and 4.0.6.

anishfyi:
## Motivation

Fixes #2596.

Generated Ruby struct enums accept symbol keys during field validation, but compare the discriminator content key only as a string. A valid symbol-keyed `config:` value is then left without its generated type discriminator.

## Solution

Normalize the content key with `to_s` in the Ruby struct-enum template and regenerate all affected models.

Added regression coverage that constructs and serializes a struct enum with a symbol-keyed `config` field.

Validation:

- `./regen_openapi.py`
- `bundle exec rake build`

### DIFF
--- a/codegen/templates/ruby/types/struct_enum.rb.jinja
+++ b/codegen/templates/ruby/types/struct_enum.rb.jinja
@@ -69,7 +69,7 @@ module Svix
         unless ALL_FIELD.include?(k.to_s)
           fail(ArgumentError, "The field #{k} is not part of Svix::{{ class_ty }}")
         end
-        if k == "{{type.content_field | to_snake_case }}"
+        if k.to_s == "{{type.content_field | to_snake_case }}"
           unless TYPE_TO_NAME.key?(v.class)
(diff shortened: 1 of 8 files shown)


svix-jbrown (approved):
seems fine

PR #2558, opened 2026-08-12
gha: run ruby e2e tests in github

svix-jbrown:
This is very similar to how the Python library runs its kitchen sink tests.

I thought about using the [docker-compose gem](https://rubygems.org/gems/docker-compose), but it doesn't support Ruby 4.x yet

PR #2606, opened 2026-08-31
Align Ruby IRB dependency with lockfile

### LINKED ISSUE #2599: Ruby release packaging rewrites inconsistent irb lock metadata
## Summary

`ruby/Gemfile.lock` records `irb (~> 1.18)`, but `ruby/Gemfile` declares unpinned `gem "irb"`. The Ruby release workflow runs `bundle lock --update`, so release packaging rewrites the lockfile before building and the packaged lock metadata no longer matches the tagged source.

## Reproduction

On the 2.1.0 tag/current Ruby source, a clean Bundler lock refresh removes the `~> 1.18` dependency constraint. The official gem's `Gemfile.lock` reflects that release-time rewrite rather than the repository lockfile.

## Proposed correction

Declare `gem "irb", "~> 1.18"` so the manifest and checked-in lock agree and release packaging is reproducible without an incidental metadata change.

## Verification

The isolated candidate is lock-stable and passes the Ruby package build on Ruby 3.2.11 and 4.0.6.

anishfyi:
## Motivation

Fixes #2599.

The Ruby `Gemfile` leaves `irb` unconstrained while `Gemfile.lock` records `irb (~> 1.18)`. Bundler consequently rewrites the lockfile during release packaging.

## Solution

Declare the existing `~> 1.18` IRB constraint in the `Gemfile` so the dependency declaration and lock metadata agree.

Validation:

- `bundle lock --update irb` leaves `Gemfile.lock` unchanged
- `bundle exec rake build`
- `bundle exec rspec --exclude-pattern spec/kitchen_sink_spec.rb` (56 examples)

### DIFF
--- a/ruby/Gemfile
+++ b/ruby/Gemfile
@@ -3,6 +3,6 @@ source "https://rubygems.org"
 gem 'base64'
 gem 'logger'
 gem 'ostruct'
-gem "irb"
+gem "irb", "~> 1.18"
 
 gemspec


svix-jbrown:
I actually think we should just remove the IRB dependency; I didn't really mean to check it in in #2558 and was just using it for some debugging.

svix-jbrown:
closed in favor of #2609

PR #2607, opened 2026-08-31
libs/ruby: support nested model JSON generation

### LINKED ISSUE #2598: Generated Ruby models do not implement the JSON generator protocol
## Summary

Generated Ruby models define zero-argument `to_json` methods. Ruby's JSON generator calls nested values with a state argument, so placing a Svix model inside an Array or Hash raises `ArgumentError` instead of producing JSON.

## Reproduction

`JSON.generate([Svix::ApplicationIn.new(name: "example")])` raises `ArgumentError: wrong number of arguments (given 1, expected 0)` on current `main` and 2.1.0. Direct `model.to_json` happens to work.

## Proposed correction

Generate `def to_json(*args); serialize.to_json(*args); end` in both struct templates and regenerate the affected models. This preserves direct output while honoring JSON's state/options protocol.

## Verification

A generator-first candidate updates 219 generated models and passes the full 56-example Ruby suite plus focused direct/nested/options models on Ruby 3.2.11 and 4.0.6.

anishfyi:
## Motivation

Fixes #2598.

Generated Ruby models define `to_json` without accepting JSON generator state. `JSON.generate` passes this state when a model is nested in an array or object, causing an argument-count error.

## Solution

Update both Ruby model templates to accept `to_json(*args)` and forward the JSON generator state through the serialized hash. Regenerated all affected Ruby models.

Added regression coverage for direct model serialization and models nested in arrays and objects.

Validation:

- `./regen_openapi.py`
- `bundle exec rake build`

### DIFF
--- a/codegen/templates/ruby/types/struct.rb.jinja
+++ b/codegen/templates/ruby/types/struct.rb.jinja
@@ -54,8 +54,8 @@ module Svix
 
     # Serializes the object to a json string
     # @return String
-    def to_json
-      JSON.dump(serialize)
+    def to_json(*args)
+      serialize.to_json(*args)
     end
   end
 end
--- a/codegen/templates/ruby/types/struct_enum.rb.jinja
+++ b/codegen/templates/ruby/types/struct_enum.rb.jinja
@@ -22,8 +22,8 @@ module Svix
       end
       # Serializes the object to a json string
       # @return String
-      def to_json
-        JSON.dump(serialize)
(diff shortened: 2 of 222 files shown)


svix-jbrown:
~~Please rebase~~

edit: oh yeah, I can do that myself

svix-jbrown (approved):

svix-owenf (approved):

PR #2621, opened 2026-09-04
libs: add support for AutoConfig v2 token

svix-lucho:
In the AutoConfig and AutoConfigConsumer classes, add support for parsing the v2 autoconfig token (prefix `auto_v2_`). 

The external API for these classes is unchanged, so existing users using the v1 token will continue to work without any breaking changes in the updated version of the SDK. When the new SDK receives a v2 token, it will decode different fields from it and call different APIs. 

The change here is relatively straightforward, the only annoying thing is that it had to be done manually for every language.

We don't have anything setup in CI to test this e2e, but I did try it locally and I validated it works for all langs

Can be reviewed commit by commit to see each language separately


php/src/AutoConfigConsumer.php
```diff
@@ -90,12 +106,46 @@ public function commit(
...
             $consumerId,
             PollerV2CommitIn::create($offset),
-            $options,
+            $options ?? new MessagePollerv2\MessagePollerv2ConsumerCommitOptions(),
         );
     }
+
+    private static function destinationOutFromV1Endpoint(EndpointOut $endpoint): DestinationOut
+    {
+        return DestinationOut::fromMixed([
```
svix-james, line 122: Are we sure this works? It looks like DestinationIn/Out are just empty files in the php lib

java/lib/src/main/java/com/svix/AutoConfigConsumer.java
```diff
@@ -63,4 +88,37 @@ public void commit(final String consumerId, final long offset,
   public void commit(final String consumerId, final long offset) throws IOException, ApiException {
     commit(consumerId, offset, new MessagePollerv2ConsumerCommitOptions());
   }
+
+  private static DestinationOut destinationOutFromV1Endpoint(final EndpointOut endpoint) {
+    DestinationOut destination = new DestinationOut();
```
svix-james, line 93: It looks like we may be at risk of NPE here if all the fields on DestinationOut aren't explicitly set.

svix-lucho: added the missing fields

svix-pluiz:
Noticed that we're changing the return type from EndpointOut to DestinationOut for all SDKs and this is gonna be a breaking change, that's expected, right?

php/src/AutoConfigConsumer.php
svix-lucho: Ah yeah I noticed this - it's an issue with the PHP codegen, that was also not working for the Stream API. I think the best choice here is to remove `AutoConfigConsumer` from the PHP SDK and implement it later. I'll ping Jonas about this.

svix-lucho:
> Noticed that we're changing the return type from EndpointOut to DestinationOut for all SDKs and this is gonna be a breaking change, that's expected, right?

Yes, that is expected since polling endpoints are a `Destination` in the `/destination` API. AutoConfigConsumer should be consistent with that. I updated the changelog.

svix-pluiz:
Looks good to me, I'd just ensure the test coverage is the same for all SDKs (TS tests has a mock server to simulate http requests).
We should also report in the changelog that 74c9a3ff9386c9729cc6bb5d068b74ffed6a2b65 removes the AutoConfigConsumer class for PHP

svix-pluiz (approved):

PR #2625, opened 2026-09-08
csharp: Strip trailing slashes from server URLs

conorbronsdon:
﻿## Motivation

A C# client configured with `https://api.svix.com/` appends resource paths directly to that URL, producing requests such as `https://api.svix.com//api/v1/app/app1`. This follows up on #1154 for the C# SDK.

## Solution

Trim trailing `/` characters once when constructing `SvixHttpClient`. This preserves reverse-proxy path prefixes and internal slashes.

Add seven mock-server cases covering URLs with no trailing slash, one or several trailing slashes, and path prefixes. Five cases fail against the original code; the two URLs without trailing slashes pass as controls. The updated C# suite passes 53 tests and skips one credential-dependent integration test. CSharpier 1.0.1 passes on both changed files.

Related to #1154. Other SDKs remain outside this patch, so the issue should stay open.

Prepared with Codex assistance.

### DIFF
--- a/csharp/Svix.Tests/WiremockTests.cs
+++ b/csharp/Svix.Tests/WiremockTests.cs
@@ -71,6 +71,27 @@ namespace Svix.Tests
             );
         }
 
+        [Theory]
+        [InlineData("", "/api/v1/app/app1")]
+        [InlineData("/", "/api/v1/app/app1")]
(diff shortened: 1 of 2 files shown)


svix-jplatte (approved):
Thanks!

PR #2627, opened 2026-09-08
Only write executable metadata in Docker builds

svix-jplatte:
Allow local dev with linkers that don't fully support linker scripts (like wild).

### DIFF
--- a/bridge/Dockerfile
+++ b/bridge/Dockerfile
@@ -57,6 +57,9 @@ RUN <<EOF
 EOF
 
 COPY bridge /app/bridge
+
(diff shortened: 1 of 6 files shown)


svix-jbrown (approved):

svix-jbrown:
I don't really love making this docker-only, because I would assume many users aren't using docker and are just building regular unix binaries, but if you really want the default to support non-GNU linkers, so be it...

svix-jplatte:
The docker images are basically the only thing we officially support. If people want to build their own binaries they can do it, and use the Dockerfile as a reference.

I'm happy to change the defaults once wild supports the necessary features :)

PR #2633, opened 2026-09-09
server: treat concurrent CREATE TABLE as success for background migrations

### LINKED ISSUE #2589: ensure_table for background migrations races when replicas start together (CREATE TABLE IF NOT EXISTS is not atomic)
### Version

- Observed on `svix/svix-server:v1.101.0` (image digest `sha256:7ba594c766c882dea9c868cd38957755b1d5c8bc5020f81cf5521ba5ab5adc11`).
- Code referenced below is from the `server-v1.101.0` tag.

### Platform

- Image: `linux/arm64`.
- Kubernetes on Amazon Linux 2023, kernel `6.12.100-125.179.amzn2023.aarch64`; 3 replicas per region, rolled out together.
- Postgres: Aurora (PostgreSQL 15).

### Description

When several `svix-server` replicas start at the same time against a database where `_svix_background_migrations` does not exist yet, one of them logs an error and then skips background migrations for the rest of its lifetime. `CREATE TABLE IF NOT EXISTS` is not atomic against a concurrent `CREATE TABLE` in PostgreSQL, so the losing session fails while inserting the new table's row type into `pg_type`.

This is cosmetic for us — the work is globally serialized behind an advisory lock anyway, so it still happens exactly once — but the error log is alarming, and the early return looks unintended.

I tried this: deploy v1.101.0 with 3 replicas into a region whose database had no `_svix_background_migrations` table yet, so all three pods reached `run_with_pool` at roughly the same moment.

I expected to see this happen: each replica either creates the table or observes that it already exists, then contends for the advisory lock, with a single winner running the migrations.

Instead, this happened. One pod logged, at `11:39:30.968Z`:

```
ERROR svix_server::db::background_migrations: Failed to create _svix_background_migrations table
  error: error returned from database: duplicate key value violates unique constraint "pg_type_typname_nsp_index"
```

A different pod in the same region logged `Starting background migration` and `Background migration completed` one second later, at `11:39:31Z` — so the migration itself was fine. It has not recurred since, as expected: the table now exists, and `IF NOT EXISTS` is a no-op from here on.

#### What appears to be going on

`ensure_table` (`server/svix-server/src/db/background_migrations.rs:69`) issues a bare statement:

```rust
pub async fn ensure_table(pool: &PgPool) -> Result<(), sqlx::Error> {
    sqlx::query(
        "CREATE TABLE IF NOT EXISTS _svix_background_migrations (
        ...
```

PostgreSQL documents `IF NOT EXISTS` as subject to exactly this race: the existence check and the creation are not performed atomically, so two concurrent sessions can both pass the check, and the loser fails on the unique index over `pg_type`.

The call site then returns early (`background_migrations.rs:201-205`):

```rust
pub async fn run_with_pool(pool: &PgPool, migrations: &[BackgroundMigration]) {
    if let Err(e) = ensure_table(pool).await {
        tracing::error!(error = %e, "Failed to create _svix_background_migrations table");
        return;
    }
```

So the losing replica does not go on to contend for the lock. In practice that costs nothing — `run_migrations` takes `pg_try_advisory_lock` (`background_migrations.rs:44`, called at `:91` and `:240`), so only one instance was ever going to do the work — but the replica also will not pick up any later migration attempt for the rest of its lifetime.

#### Suggested fix

Either would do:

1. Wrap `ensure_table` in the same advisory lock the migrations already use, so table creation is serialized with everything else; or
2. Treat the duplicate-key error as success — a `23505` on `pg_type_typname_nsp_index` here means another instance created the table concurrently, which is the desired end state.

Option 2 is the smaller change and keeps startup lock-free.

Tyagiquamar:
## Problem

When several `svix-server` replicas start together against a database that does not yet have `_svix_background_migrations`, PostgreSQL can reject `CREATE TABLE IF NOT EXISTS` with `23505` (`pg_type_typname_nsp_index`) or `42P07`. `ensure_table` then returns an error, and `run_with_pool` returns early, so the losing replica skips background migrations for the rest of its lifetime.

This was observed on `svix-server` v1.101.0 (see #2589). The migration itself still ran on the replica that won the race; the losing replica logged an error and never contended for the advisory lock.

## Reproduction

1. Drop `_svix_background_migrations` (or start from an empty database).
2. Start several server processes at once so they all reach `ensure_table` together.
3. One process logs `Failed to create _svix_background_migrations table` with a duplicate-key error on `pg_type_typname_nsp_index`.

Locally the new test races `CREATE TABLE IF NOT EXISTS` in an isolated schema:

```text
cargo test -p svix-server --test it test_ensure_table_concurrent_create_from_absent
```

## Root cause

`CREATE TABLE IF NOT EXISTS` is not atomic against a concurrent `CREATE TABLE`. Two sessions can both pass the existence check. The loser fails while inserting the new row type into `pg_type`.

## Fix

Treat PostgreSQL `23505` and `42P07` from that `CREATE TABLE IF NOT EXISTS` as success: another replica created the table, which is the desired end state. `run_with_pool` then continues to the existing advisory lock.

## Regression test

`test_ensure_table_concurrent_create_from_absent` issues 16 concurrent `CREATE TABLE IF NOT EXISTS` statements against a fresh schema and asserts every caller succeeds and the table exists.

## Verification

- `git diff --check` (no whitespace errors)
- `rustfmt` on the two touched files
- Targeted cargo test was not run on this Windows host (no local Postgres for the svix-server integration suite). CI should run the existing background-migration tests plus the new race test.

## Scope

No change to migration SQL, advisory locking, or retry backoff.

Fixes #2589

### DIFF
--- a/server/svix-server/src/db/background_migrations.rs
+++ b/server/svix-server/src/db/background_migrations.rs
@@ -66,8 +66,30 @@ fn lock_key(id: &str) -> i64 {
     0x7a9f2c51 * (CRC_IEEE.checksum(id.as_bytes()) as i64)
 }
 
+/// PostgreSQL `CREATE TABLE IF NOT EXISTS` is not atomic against a concurrent
+/// `CREATE TABLE`. Two sessions can both pass the existence check; the loser
+/// then fails while inserting the new type into `pg_type` (`23505`) or with
+/// `duplicate_table` (`42P07`). Either error means the table now exists.
+fn is_concurrent_create_table_error(err: &sqlx::Error) -> bool {
+    match err {
+        sqlx::Error::Database(db) => matches!(db.code().as_deref(), Some("23505") | Some("42P07")),
+        _ => false,
+    }
+}
+
+/// Run `CREATE TABLE IF NOT EXISTS`, treating a concurrent-create race as success.
+#[doc(hidden)]
+pub async fn create_table_if_not_exists(pool: &PgPool, sql: &str) -> Result<(), sqlx::Error> {
+    match sqlx::query(sql).execute(pool).await {
+        Ok(_) => Ok(()),
+        Err(e) if is_concurrent_create_table_error(&e) => Ok(()),
+        Err(e) => Err(e),
+    }
+}
+
 pub async fn ensure_table(pool: &PgPool) -> Result<(), sqlx::Error> {
-    sqlx::query(
+    create_table_if_not_exists(
+        pool,
         "CREATE TABLE IF NOT EXISTS _svix_background_migrations (
             id          TEXT        PRIMARY KEY,
             started_at  TIMESTAMPTZ NOT NULL DEFAULT now(),
@@ -77,9 +99,7 @@ pub async fn ensure_table(pool: &PgPool) -> Result<(), sqlx::Error> {
             last_error  TEXT
         )",
     )
-    .execute(pool)
-    .await?;
-    Ok(())
+    .await
 }
 
 pub async fn run_migrations(
--- a/server/svix-server/tests/it/background_migrations.rs
+++ b/server/svix-server/tests/it/background_migrations.rs
@@ -3,7 +3,8 @@
 use futures::future::join_all;
 use sqlx::PgPool;
 use svix_server::db::background_migrations::{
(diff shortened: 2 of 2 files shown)


svix-jbrown (approved):
seems ok to me

PR #2640, opened 2026-09-11
Libs/PHP: Add struct enums support

svix-lucho:
Adds a jinja template for struct enums, and regenerates the code based on that.

API looks like this to use them

```php
  $svx->streaming->sink->create(
      'stream_1srOrx2ZWZBpBUvZwXKQmoEYga2',
      StreamSinkIn::create(StreamSinkInConfig::poller())
  );

  $svx->streaming->sink->create(
      'stream_1srOrx2ZWZBpBUvZwXKQmoEYga2',
      StreamSinkIn::create(
          StreamSinkInConfig::amazonS3(S3ConfigIn::create('my-bucket')->withRegion('us-east-1'))
      )
  );
```

For `Patch` models, it uses the same `$setFields = []` pattern we already have in the PHP lib.

This also required a change in `composer.json`. PHP consumers load classes by path. `Svix\Foo\Bar` must live at `php/src/Foo/Bar.php`. Struct-enum variants are classes like `Svix\Models\DestinationInConfig\Http`, but they live in `DestinationInConfig.php`, not `DestinationInConfig/Http.php`. The composer change helps the package manager find those classes. It doesn't require any change from package consumers, it should just work.

I validated this by using the PHP lib as a consumer would, in addition to the mock tests.

Closes https://github.com/svix/monorepo-private/issues/14282


codegen/templates/php/types/struct_enum.php.jinja
```diff
@@ -0,0 +1,270 @@
+{% from "types/macros.php.jinja" import field_from_json, field_to_json -%}
+{% set type_name = type.name | to_upper_camel_case -%}
+{% set has_fields = type.fields | length > 0 -%}
+{% set enum_type_name = type_name ~ "Config" if has_fields else type_name -%}
```
svix-jplatte: We don't have different branches for when a struct enum has fields vs. when it doesn't in any other languages - why here? Do we really want to make adding a field to a previously "fieldless" (outside variants) struct enum a breaking change?

codegen/templates/php/types/struct_enum.php.jinja
```diff
@@ -0,0 +1,270 @@
...
+     * @param {{ field.type.to_phpdoc() }}{{ optional_prefix }} ${{ field.name | to_lower_camel_case }} {{ field.description | to_doc_comment(style="php_field") }}
+    {% endfor -%}
+     */
+    private function __construct(
+        public readonly {{ enum_type_name }} $config,
+        {% for field in type.fields -%}
+            {% if field.required -%}
+        public readonly {{ field.type.to_php() }} ${{ field.name | to_lower_camel_case }},
+            {% endif -%}
+        {% endfor -%}
```
svix-jplatte:
```suggestion
        {% for field in type.fields | selectattr("required") -%}
        public readonly {{ field.type.to_php() }} ${{ field.name | to_lower_camel_case }},
        {% endfor -%}
```

codegen/templates/php/types/struct_enum.php.jinja
```diff
@@ -0,0 +1,270 @@
...
+        {% for field in type.fields -%}
+            {% if not field.required -%}
+        public readonly ?{{ field.type.to_php() }} ${{ field.name | to_lower_camel_case }} = null,
+            {% endif -%}
+        {% endfor -%}
```
svix-jplatte:
```suggestion
        {% for field in type.fields | selectattr("required", "false") -%}
        public readonly ?{{ field.type.to_php() }} ${{ field.name | to_lower_camel_case }} = null,
        {% endfor -%}
```

codegen/templates/php/types/struct_enum.php.jinja
```diff
@@ -0,0 +1,270 @@
...
+        $this->setFields = $setFields;
+    }
+
+    /**
+     * Create an instance of {{ type_name }} with required fields
+     */
+    public static function create(
+        {{ enum_type_name }} $config,
+        {% for field in type.fields -%}
+            {% if field.required -%}
```
svix-jplatte: Same here

codegen/templates/php/types/struct_enum.php.jinja
```diff
@@ -0,0 +1,270 @@
...
+                'config' => true,
+                {%- for field in type.fields | selectattr("required") -%}
+                '{{ field.name | to_lower_camel_case }}' => true,
+                {%- endfor -%}
+            ]
+        );
+    }
+
+    {% for field in type.fields -%}
+        {% if not field.required -%}
```
svix-jplatte: … and here

codegen/templates/php/types/struct_enum.php.jinja
```diff
@@ -0,0 +1,270 @@
...
+    /**
+     * Create an instance from a json string
+     */
+    public static function fromJson(string $json): self
+    {
+        $data = json_decode(json: $json, associative: true, depth: 512, flags: JSON_THROW_ON_ERROR);
+        return self::fromMixed($data);
+    }
+}
+{% do generate_extra_file(output_dir ~ "/" ~ enum_type_name ~ ".php", config_file()) %}
```
svix-jplatte: Do we have to put stuff in separate files in PHP? If we can, I'd prefer putting related stuff in one file. Currently we only use this weird helper for Java, where the language forbids us from putting two (non-nested) classes in the same file.

codegen/templates/php/types/struct_enum.php.jinja
svix-lucho: right, no reason for this. I'll do the same as the other langs

codegen/templates/php/types/struct_enum.php.jinja
svix-lucho: yeah, it's possible in PHP. I'll change it

codegen/templates/php/types/struct_enum.php.jinja
svix-lucho: fixed this and the others

svix-lucho:
@svix-jplatte this is ready for another look

svix-jplatte (approved):

PR #2642, opened 2026-09-12
fix(server): report validation locations without the __all__ placeholder

### LINKED ISSUE #920: Query validation shows an incorrect location in some cases
E.g. when you "list messages" with `event_types=1,` (, is not allowed in event type names), you get this:
```rust
Error: Varidation error Some(HttpValidationError { detail: [ValidationError { loc: ["query", "__all__"], msg: "String must match the following pattern: [a-zA-Z0-9\\-_.].", _type: "value_error" }] })
```

Location should be `query, event_types` or even better: `query, event_types, 0` as it's the first item.

Tyagiquamar:
Fixes #920

## Problem

Wrapper-type validators in `core::types` (e.g. `EventTypeName` via `validate_limited_str`, `BaseUid`) key their errors under the validator crate's `__all__` placeholder. When such a type is validated at the top level of a request, the placeholder leaks into the user-facing location. Listing messages with `?event_types=1,` currently returns:

```json
{"loc": ["query", "__all__"], "msg": "String must match the following pattern: [a-zA-Z0-9\-_.]."}
```

## Fix

- `validation_errors` now skips the `__all__` key when flattening the error tree, so wrapper-type errors render at their parent location. This also cleans up other paths that leaked the placeholder, e.g. an invalid endpoint `uid` previously reported `["body", "uid", "__all__"]` and now reports `["body", "uid"]`.
- The `event_types` query extractor now names the parameter, so the location is `["query", "event_types"]` as the issue requests.

Reported per-element locations like `["query", "event_types", "0"]` are not meaningful here because the parameter is parsed into an unordered `HashSet`, so the parameter-level location is used instead.

## Testing

- New unit test `test_validation_errors_wrapper_type_location` asserts the reported location and message.
- `cargo fmt --check`, `cargo clippy --all-targets --all-features -- -D warnings`, the `v1::utils` unit tests, and `./generate-openapi.sh --check` all pass on Rust 1.88 (matching the CI matrix).

### DIFF
--- a/server/svix-server/src/core/types/mod.rs
+++ b/server/svix-server/src/core/types/mod.rs
@@ -25,7 +25,7 @@ pub use case_preserving_header_map::CasePreservingHeaderMap;
 
 use super::cryptography::{AsymmetricKey, Encryption};
 
-const ALL_ERROR: &str = "__all__";
+pub(crate) const ALL_ERROR: &str = "__all__";
 
 macro_rules! enum_wrapper {
     ($name_id:ty) => {
--- a/server/svix-server/src/v1/utils/mod.rs
+++ b/server/svix-server/src/v1/utils/mod.rs
@@ -31,7 +31,7 @@ use validator::{Validate, ValidationError};
 
 use crate::{
     core::types::{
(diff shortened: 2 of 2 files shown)


Tyagiquamar:

PR #2644, opened 2026-09-12
feat(server): add expired_message_cleaner_enabled config option

### LINKED ISSUE #1111: Make the deletion of events from postgres configurable
### Discussed in https://github.com/svix/svix-webhooks/discussions/1108

<div type='discussions-op-text'>

<sup>Originally posted by **cheleswer** October 19, 2023</sup>
Even after deleting the subscription, I see that events are still persisting in the Postgres DB.
Is there any certain time duration after which all these events will be deleted?  Is it configurable?</div>

Tyagiquamar:
Fixes #1111

## Problem

The expired message cleaner, which nulls payloads and deletes `messagecontent` rows once a message's expiration has passed, always ran with the worker. From the discussion linked in the issue: message payloads persist only until expiration, and operators who want to keep them around (e.g. for their own retention tooling) had no way to opt out of the deletion.

## Fix

Add an `expired_message_cleaner_enabled` setting (`SVIX_EXPIRED_MESSAGE_CLEANER_ENABLED`), defaulting to `true` to preserve current behavior. When disabled, the cleaner loop is not started.

Message rows themselves are kept regardless (pagination iterators depend on them); full deletion of old messages remains available via the `svix-server prune` command.

The option is documented in `config.default.toml` and covered by environment parsing tests in `tests/config.rs`.

## Testing

- `cargo test --test config` passes (defaults to enabled, `SVIX_EXPIRED_MESSAGE_CLEANER_ENABLED=false` disables).
- `cargo fmt --check` and `cargo clippy --all-targets --all-features -- -D warnings` pass on Rust 1.88 (matching the CI matrix).

### DIFF
--- a/server/svix-server/config.default.toml
+++ b/server/svix-server/config.default.toml
@@ -109,6 +109,11 @@ api_enabled = true
 # Should this instance run the message worker
 worker_enabled = true
 
+# Should this instance run the expired message cleaner, which nulls payloads and deletes
+# message contents for messages past their expiration. Message rows themselves are kept
+# regardless, to preserve pagination, and can be deleted with the `prune` command.
+expired_message_cleaner_enabled = true
+
 # Should this instance run background migrations
 background_migrations_enabled = true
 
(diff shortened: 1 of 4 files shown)


Tyagiquamar:

svix-jbrown (approved):
seems fine to me

